PR #475, opened 2023-07-14
Ensure L1 txns built by commitment task don't exceed size limit

jbearer:
When working on zkEVM preconfirmations, with HotShot much faster than the L1, I have been running into a transaction size limit that does not exist in EVM, but does exist in Geth's RPC implementation, and therefore is of practical importance.

This limit affects the HotShot commitment task when it batches many blocks together; the resulting L1 contract call can exceed 131072 bytes. This change consists of 2 fixes:
* Instead of sending the whole QC to the contract for each block in the chain, only send 3 dummy words. This more closely matches our proposed contract design, where most of the QC fields are encapsulated by a succinct commitment. This drastically reduces the size of the contract call transactions.
* Lower the MAX_BLOCKS constant in HotShot.sol from 1000 to 500, so that transactions with this many blocks are still smaller than the Geth size limit.

I also modified the test to build a transaction with the max number of blocks, check that it is not too large, and then submit it.


nomaxg (approved):
LGTM

PR #483, opened 2023-07-18
Remove nix/just requirement for docker demo

sveitser:
Depends on 
- https://github.com/EspressoSystems/espresso-sequencer/pull/479


nomaxg (approved):
Thanks!

PR #489, opened 2023-07-19
Only hold state wirte guard to execute block

sveitser:
This should reduce API response times.

### DIFF
--- a/example-l2/src/executor.rs
+++ b/example-l2/src/executor.rs
@@ -98,12 +98,11 @@ pub async fn run_executor(opt: &ExecutorOptions, state: Arc<RwLock<State>>) {
 
         // Execute new blocks, generating proofs.
         let mut proofs = vec![];
-        let mut state = state.write().await;
         tracing::info!(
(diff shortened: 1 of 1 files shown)


example-l2/src/executor.rs
```diff
@@ -137,12 +136,18 @@ pub async fn run_executor(opt: &ExecutorOptions, state: Arc<RwLock<State>>) {
...
+                    .await
+                    .execute_block(nmt_root, namespace_proof)
+                    .await,
+            );
         }
 
         // Compute an aggregate proof.
         let proof = BatchProof::generate(&proofs).expect("Error generating batch proof");
-        let state_comm = commitment_to_u256(state.commit());
+        let state_comm = commitment_to_u256(state.read().await.commit());
```
sveitser, line 150: If something else was modifying `state` then I think the commitment and proof could go out of sync. However I think wouldn't be an issue for the current example rollup.

nomaxg (approved):

PR #490, opened 2023-07-19
Migrate from CDN -> Orchestrator + Web servers 

### LINKED ISSUE #485: Migrate sequencer from HotShot cdn_server to web_server
Work off of https://github.com/EspressoSystems/HotShot/tree/paper_benchmarking for now

nomaxg:
Refactor sequencer network to use web-servers/orchestrator instead of a cdn-server:
1. The orchestrator takes the place of the cdn-server, providing the network config to sequencer nodes and providing the start signal
2. Two web servers are necessary, one for DA, and one for consensus

closes #485


docker/web-server.Dockerfile
```diff
@@ -1,26 +1,26 @@
...
+COPY target/release/web-server /bin/web-server
+RUN chmod +x /bin/web-server
 
-ENV ESPRESSO_CDN_SERVER_PORT=50000
+ENV ESPRESSO_WEB_SERVER_PORT=50000
 
 # Set up view timing for optimal performance in high volume conditions. We set a fairly long minimum
 # propose time to wait to build up a large block before proposing.
-ENV ESPRESSO_CDN_SERVER_MIN_PROPOSE_TIME=10s
+ENV ESPRESSO_WEB_SERVER_MIN_PROPOSE_TIME=10s
```
nomaxg, line 14: change to orchestrator

scripts/build-docker-images
```diff
@@ -33,7 +33,8 @@ cleanup(){
 }
 
 mkdir -p ${WORKDIR}/target/release
-cp -v target/${TARGET}/release/cdn-server ${WORKDIR}/target/release
+cp -v target/${TARGET}/release/web-server ${WORKDIR}/target/release
```
sveitser, line 36: @nomaxg I suggest merging `main`. I think the new script has a higher chance of working.

.env
```diff
@@ -3,20 +3,25 @@
...
+ESPRESSO_SEQUENCER3_API_PORT=50003
+ESPRESSO_SEQUENCER4_API_PORT=50004
 ESPRESSO_SEQUENCER_URL=http://sequencer0:$ESPRESSO_SEQUENCER_API_PORT
 ESPRESSO_SEQUENCER_STORAGE_PATH=/store/sequencer
 ESPRESSO_SEQUENCER_L1_PORT=8545
 ESPRESSO_SEQUENCER_L1_PROVIDER=http://demo-l1-network:$ESPRESSO_SEQUENCER_L1_PORT
 ESPRESSO_SEQUENCER_HOTSHOT_ADDRESS=0x5fbdb2315678afecb367f032d93f642f64180aa3
 ESPRESSO_SEQUENCER_ETH_MNEMONIC="test test test test test test test test test test test junk"
-
+t
```
sveitser, line 24: There is an extra `t` on this line.

README.md
```diff
@@ -35,27 +35,39 @@ Docker images and the [docker-compose-demo.yaml](docker-compose-demo.yaml) file
...
-Once you have started the CDN server, you must connect `$N` sequencer nodes to it, after which the network will start up
-automatically. To start one node, run `target/release/sequencer --cdn-url tcp://localhost:$PORT`. A useful Bash snippet
-for running `$N` nodes simultaneously in the background of your shell is:
+Once you have started the orchestrator and the web servers, you must connect `$N` sequencer nodes to them, after which the network will start up automatically. To start one node, run
+```bash
+target/release/sequencer \
+    --orchestrator-url http://localhost:$PORT \
+    --da-server-url http://localhost:$DA_PORT \
+    --consensus-server-url http://localhost:$CONSENSUS_PORT \
+    -- api --port 8083 --storage-path storage
```
jbearer, line 56: The module syntax has changed, it should be 
```suggestion
    -- http --port 8083 -- query --storage-path storage -- submit
```

justfile
```diff
@@ -29,14 +29,24 @@ test:
...
+dev-consensus-server:
+    target/release/web-server -p 8082
 
 dev-sequencer:
-    target/release/sequencer --cdn-url tcp://127.0.0.1:8080 -- api --port 8081 --storage-path storage
+    target/release/sequencer \
+    --orchestrator-url http://localhost:8080 \
+    --da-server-url http://localhost:8081 \
+    --consensus-server-url http://localhost:8082 \
+    -- api --port 8083 --storage-path storage
```
jbearer, line 46: This will also need to change (maybe after merging with main?)

sequencer/src/lib.rs
```diff
@@ -304,18 +289,32 @@ async fn init_hotshot<N: network::Type>(
...
+    _metrics: Box<dyn Metrics>,
+) -> (HotShotHandle<SeqTypes, Node<network::Web>>, u64) {
+    // Orchestrator client
+    let validator_args = ValidatorArgs {
+        host: network_params.orchestrator_url.host().unwrap().to_string(),
+        port: network_params.orchestrator_url.port().unwrap(),
+        public_ip: None,
+    };
+    let orchestrator_client = OrchestratorClient::connect_to_orchestrator(validator_args).await;
+    let public_ip = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
jbearer, line 310: In what sense is this IP "public", or how does it identify the node? If all nodes are using 127.0.0.1?

sequencer/src/lib.rs
```diff
@@ -326,16 +325,31 @@ pub async fn init_node(
...
-            connected,
-            config.config.total_nodes,
-        );
-        sleep(Duration::from_secs(1)).await;
-    }
+    // // Wait for other nodes to connect.
+    orchestrator_client
+        .wait_for_all_nodes_ready(node_index.into())
+        .await;
+    let wait_time = Duration::from_secs(1);
```
jbearer, line 332: What is wait_time used for?

sequencer/src/lib.rs
nomaxg: I believe this is only used for the libp2p setup, which we aren't using here so this can pretty much be anything, I can add a comment explaining that

sequencer/src/lib.rs
nomaxg: The wait time in between polls (replaces the previous `sleep` statement)

jbearer: So to be clear, this sleep is only used for connecting to the server, not during consensus? I'm just thinking 1 second seems a little slow for consensus messages but totally fine for startup-related stuff

nomaxg: Ah actually this is the wait time in between sequencer -> web server polls, so I guess 1 second would be too slow. How about 100 ms?

jbearer: That sounds more reasonable, let's double check with the consensus team

jbearer: Yes sounds like 100 ms is good, long term we should consider making this configurable

jbearer (approved):

PR #495, opened 2023-07-20
rust: 1.70 -> 1.71

sveitser:
- Make CI job names unique.
- Fix clippy.

### DIFF
--- a/sequencer/src/lib.rs
+++ b/sequencer/src/lib.rs
@@ -509,7 +509,7 @@ mod test {
         handle: SystemContextHandle<SeqTypes, I>,
         txn: &ApplicationTransaction,
     ) -> Transaction {
-        let tx = Transaction::new(TestVm::default().id(), bincode::serialize(txn).unwrap());
+        let tx = Transaction::new(TestVm {}.id(), bincode::serialize(txn).unwrap());
 
         handle
             .submit_transaction(tx.clone())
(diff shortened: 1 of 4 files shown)


sequencer/src/lib.rs
```diff
@@ -509,7 +509,7 @@ mod test {
         handle: SystemContextHandle<SeqTypes, I>,
         txn: &ApplicationTransaction,
     ) -> Transaction {
-        let tx = Transaction::new(TestVm::default().id(), bincode::serialize(txn).unwrap());
+        let tx = Transaction::new(TestVm {}.id(), bincode::serialize(txn).unwrap());
```
jbearer, line 512: Curious why Clippy doesn't like `default()` here? It seems like that would actually be less likely to break in the future (suppose we add a field but give it a default value)

jbearer (approved):

jbearer:
Re-running to see if they pass

sequencer/src/lib.rs
jbearer: https://rust-lang.github.io/rust-clippy/master/index.html#/default_constructed_unit_structs

Their rationale is just "unnecessary function call". I think this is a dumb lint and I would maybe even be in favor of `#[ignore]`ing it. `default()` semantically says something abstract about intent, whereas a unit struct literal says something very specific about the representation. They're not interchangable.

nomaxg (approved):

PR #497, opened 2023-07-20
Run HTTP server and use healthcheck by default

jbearer:


nomaxg (approved):

docker/sequencer.Dockerfile
```diff
@@ -9,4 +9,6 @@ RUN apt-get update \
 COPY target/$TARGETARCH/release/sequencer /bin/sequencer
 RUN chmod +x /bin/sequencer
 
-CMD [ "/bin/sequencer"]
+CMD ["/bin/sequencer", "--", "http"]
+HEALTHCHECK CMD curl --fail http://localhost:$ESPRESSO_SEQUENCER_API_PORT/healthcheck  || exit 1
```
sveitser, line 13: Does it need to be specified in both the Dockerfile and the docker compose file?

jbearer: I don't think so. I can remove it from the compose file

sveitser: Good idea to add them to the dockerfiles instead, I think.

PR #510, opened 2023-07-26
Use correct node index

nomaxg:
We were using the node_index from the config returned by the orchestrator (which appears to always be 0) instead of the node_index we receive when identifying with the orchestrator, making all validators replicas.


philippecamacho (approved):
LGTM

PR #514, opened 2023-07-28
Explicitly set contract address in event filters.

jbearer:
This works around what appears to be a bug or deficiency in Ethers: https://github.com/gakonst/ethers-rs/issues/2528


philippecamacho (approved):
LGTM

PR #509, opened 2023-07-26
Update hotshot after run_view_refactor

jbearer:
Depends on https://github.com/EspressoSystems/hotshot-query-service/pull/151

PR #520, opened 2023-08-01
Improve error handling in `contract_send`

jbearer:
I noticed this while working on [the HotShot upgrade](#509), and this is cherry-picked from that branch, but I think this could affect `main` as well, so we may not want to wait for that other PR in order to merge this.

While stress testing, I observed that this function would occasionally hang, presumably in an ethers.rs call, stalling the commitment task and in turn all updates to the L2. I'm not sure exactly where it was hanging, as the Future impl for `PendingTransaction` is quite complicated and doesn't have useful logging. But this rewrite handles all error cases and at the very least logs at DEBUG level every time it loops. It seems to fix the problem, and if we ever run into similar problems in the future, it will be more transparent where exactly things are getting stuck.

### DIFF
--- a/utils/src/lib.rs
+++ b/utils/src/lib.rs
@@ -242,19 +242,46 @@ pub async fn contract_send<T: Detokenize>(
             return None;
         }
     };
-    let receipt = match pending.await {
+
+    let hash = pending.tx_hash();
+    let provider = pending.provider();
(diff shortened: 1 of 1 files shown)


philippecamacho (approved):
LGTM

PR #524, opened 2023-08-04
Add a script to count transactions sequenced by Espresso

jbearer:
This should really be a query service endpoint, but I thought we might want metrics for Doppio without having to restart the query service.

Results from just before the public launch:
For blocks 0-23636:
Total transactions: 59889
Unique: 19970


sveitser (approved):
LGTM.

I don't think it matters much but according the logs it looks like the last block isn't fetched, but the final output indicates that it's included. 

```  2023-08-15T10:55:30.699822Z  INFO count_transactions: task 0 processing block 101966/101967
    at sequencer/src/bin/count-transactions.rs:66

For blocks 101955-101967:
Total transactions: 0
Unique: 0
```

jbearer:
Oh yeah. That's an off-by-one between block indexes and block counts, but the final count should still be accurate

PR #528, opened 2023-08-07
Fix race condition in rollup tests

jbearer:
There was a race condition in the executor tests, where we read the state from the executor and the smart contract and asserted that they were the same, but nothing guaranteed that the smart contract had been updated by then, or that the smart contract hadn't been updated with a new block since we read the executor's state. I have seen a failure here once in CI, so I think this is a race that can really happen.


example-l2/src/executor.rs
```diff
@@ -273,9 +293,69 @@ mod test {
...
+        ///
+        /// At each intermediate state, including the terminal state, this function checks that the
+        /// state reported by the executor matches the state in the contract. This implies that, if
+        /// this function returns, `predicate` holds on both the executor state and some state that
+        /// has been verified by the smart contract.
+        pub async fn wait_for_effect(&self, predicate: impl Fn(State) -> bool) {
+            let vm_id: VmId = self.vm.into();
+            let mut exec_stream = self.subscribe_executor().await;
+            let mut l1_stream = self.subscribe_contract().await;
+            loop {
```
nomaxg, line 323: Should this timeout?

nomaxg (approved):
LGTM

PR #530, opened 2023-08-08
Set MIN_PROPOSE_TIME

jbearer:
We are setting MAX_PROPOSE_TIME short, which is somewhat pointless when MIN_PROPOSE_TIME defaults to 10s. These parameters should always be set together.


sveitser:
LGTM

sveitser (approved):

PR #537, opened 2023-08-09
Add timestamp and L1 block info to sequencer blocks

### LINKED ISSUE #536: Add L1 block number and timestamp to `next_block`

jbearer:
Note: I think a better design for this would be to have some shared memory (ie a `Mutex` or channel) where we store the latest known L1 block, and have an async task to update this state each time it receives a new block from the L1 provider. Then we could move L1 queries, and the complexities that occur when they are slow or fail, from the critical path of proposal; proposal would only need to read from the `Mutex`. Unfortunately, since this design is stateful, it can't really be implemented with the current HotShot traits. Consider this change a short term workaround.

The timestamp is needed because we work in terms of block commitments,
not leaf commitments, so clients cannot authenticate the timestamp
from the leaf. Perhaps we should change this.
    
I had to do a pretty ugly hack to set up an L1 provider and get the
latest L1 block when proposing a new block. There are some (I think
simple) changes to HotShot APIs that would make this much cleaner.
    
Finally, ntoe that this is not secure. The proposer has full
discretion to choose this metadata, and safety invariants (like
monotonicity) as well as performance requirements (like using a
sufficiently recent L1 block) are not currently enforced. The plan
is to eventually hook validation code into HotShot replicas via
the metastate API.
    
Closes #536


jbearer:
Some weird issues with this change, debugging

jbearer:
Fixed (I think)

sequencer/src/state.rs
```diff
@@ -20,11 +26,79 @@ impl HotShotState for State {
 
     type Time = ViewNumber;
 
-    fn next_block(prev_state: Option<Self>) -> Self::BlockType {
-        match prev_state {
-            Some(_state) => unimplemented!("Using sequencing consensus, no previous state"),
-            None => Block::new(),
-        }
+    fn next_block(_prev_state: Option<Self>) -> Self::BlockType {
```
nomaxg, line 29: Ooof yeah these all seem like necessary improvements to the API

sequencer/src/state.rs
```diff
@@ -20,11 +26,79 @@ impl HotShotState for State {
...
+                        block.timestamp
+                    );
+                    break L1BlockInfo {
+                        number: number.as_u64(),
+                        timestamp: block.timestamp,
+                    };
+                }
+            })
+        } else {
+            // For unit testing, we may disable the L1 provider and use mock L1 blocks instead.
```
sveitser, line 97: Should we hide this behavior behind the `testing` feature flag?

sequencer/src/api.rs
```diff
@@ -295,17 +297,17 @@ impl<N: network::Type> Borrow<HotShotSequencingConsensusApi<SeqTypes, Node<N>>>
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct NamespaceProofQueryData {
-    proof: NamespaceProofType,
-    nmt_root: NMTRoot,
+    pub proof: NamespaceProofType,
+    pub header: Header,
```
sveitser, line 301: What's the reason for making these public if we also have getters?

sveitser (approved):
LGTM, some minor questions.

sequencer/src/api.rs
jbearer: So that you can move out of the fields

sequencer/src/state.rs
jbearer: Yes, good idea

PR #539, opened 2023-08-11
Prepare to port commitment scheme to Go

jbearer:
* Add reference objects with known commitments, which will be used to write test cases for the port
* Make the commitment scheme more consistent and easier to port to other languages


sequencer/src/block.rs
```diff
@@ -242,3 +256,120 @@ impl Block {
         &self.l1_block
     }
 }
+
+#[cfg(test)]
+mod reference {
```
sveitser, line 261: I think it would be nice for consumers if we commit json files with the test data to this repo.

sveitser (approved):
LGTM. One suggestion.

PR #555, opened 2023-08-18
Feat/timestamp endpoints

jbearer:
Closes https://github.com/EspressoSystems/espresso-sequencer/issues/545


sequencer/src/api.rs
```diff
@@ -179,7 +194,7 @@ impl Options {
             // Initialize availability and status APIs
             let mut options: AvailabilityOptions = Default::default();
             let namespace_extension =
-                toml::from_str::<toml::value::Value>(include_str!("namespace_api.toml"))
+                toml::from_str::<toml::value::Value>(include_str!("../api/availability.toml"))
```
sveitser, line 197: Change the variable name `namespace_extension`?

sequencer/api/availability.toml
```diff
@@ -0,0 +1,60 @@
...
+PATH = ["block/:height/namespace/:namespace"]
+":height" = "Integer"
+":namespace" = "Integer"
+DOC = "Get the namespace proof for a set of VM transations and the NMT root"
+
+[route.gettimestampwindow]
+PATH = [
+    "headers/window/:start/:end",
+    "headers/window/from/:height/:end",
+    "headers/window/from/hash/:hash/:end",
```
sveitser, line 11: What's the use case for this? It seems a bit strange to specify one edge of the interval with a hash and the other one with a timestamp.

sequencer/src/api.rs
```diff
@@ -285,8 +378,9 @@ pub type HandleFromMetrics<N> = Box<
 >;
 
 struct AppState<N: network::Type> {
-    pub submit_state: HotShotSequencingConsensusApi<SeqTypes, Node<N>>,
-    pub query_state: QueryData<SeqTypes, Node<N>, ()>,
+    submit_state: HotShotSequencingConsensusApi<SeqTypes, Node<N>>,
+    query_state: QueryData<SeqTypes, Node<N>, ()>,
+    blocks_by_time: BTreeMap<u64, Vec<u64>>,
```
sveitser, line 383: Not sure if worth the effort but internally we could use a dedicated type for the timestamps.

sveitser (approved):
LGTM. I think the indexing in the test is a bit difficult to follow but unfortunately I don't have a concrete suggestion on how to make it easier. The comments definitely help.

sequencer/api/availability.toml
jbearer: Use case for `from/hash` is the same as `from/:height`. It's if you've already fetched up to a certain block within the window, you can fetch the remaining blocks starting from there. I'm just trying to be consistent with the rest of the API, where whenever you specify a specific block, we accept either a block number or a hash

sequencer/src/api.rs
jbearer: Yeah I think this would be good, at least a type alias, but we can do it later.

PR #595, opened 2023-09-08
Fkrell/block scanner

fkrell:


jbearer (approved):
LGTM

sveitser:
@fkrell if you merge `main` into your branch the CI issues should be resolved.

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
+### Simple live Espresso block scanner
+### Scan all blocks looking for transaction on each VM. 
+### For each found transaction, display the VM block containing that transaction, including inclusion proofs for transactions in the Espresso block.
+
+import urllib.request
+import json;
```
sveitser, line 6: `;` aren't needed

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+import time;
+import datetime;
+import sys;
+host="localhost"
+port="50000"
+if len(sys.argv) > 1:
+	host=sys.argv[1]
+if len(sys.argv) > 2:
+	port=sys.argv[2]
+url="http://"+host+":"+port+"/availability/block/"
```
sveitser, line 16: You can write this as `f"http://{host}:{port}/availability/block/`"

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Keep track of transaction to avoid showing duplicates
+payloads = set()
+
+# Scan block by block
+block_number = 0
+while True:
```
sveitser, line 23: Recommend using `black` to format the code. It can be added via the flake for example  with

```diff
diff --git a/flake.nix b/flake.nix
index 443aa6d42..6c3422af7 100644
--- a/flake.nix
+++ b/flake.nix
@@ -58,7 +58,7 @@
             docker-compose # provides the `docker-compose` command

             # Python
-            (python3.withPackages (ps: with ps; [ ]))
+            (python3.withPackages (ps: with ps; [ black ]))
             jq
```

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+	try:
+		page = urllib.request.urlopen(url+str(block_number))
+	except urllib.error.HTTPError as e:
```
sveitser, line 26:
```suggestion
	except urllib.error.HTTPError:
```

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+				namespace_block_url=url+str(block_number)+"/namespace/"+str(vm)
+				print("GET "+namespace_block_url)
+				# find block info for Rollup
+				vm_page = urllib.request.urlopen(url+str(block_number)+"/namespace/"+str(vm))
+				vm_block_str = vm_page.read()
+				vm_block_struct = json.loads(vm_block_str)
+				
+				# Omit unnecessary data
+				del vm_block_struct['proof']['phantom']
+				if vm_block_struct['proof']['left_boundary_proof'] == None:
```
sveitser, line 72: `is None` (e. g. compare by identity and not value)

scripts/espresso_block_scanner.py
sveitser: It's probably better to just accept the base URL e. g. `http://localhost:50000` as argument otherwise this won't work if we have a https URL (for example when running behind the load balancer on AWS).

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+payloads = set()
+
+# Scan block by block
+block_number = 0
+while True:
+	try:
+		page = urllib.request.urlopen(url+str(block_number))
+	except urllib.error.HTTPError as e:
+		# "Block not produced yet. waiting for 1 second
+		time.sleep(1)
```
sveitser, line 28: Print something here so the script doesn't run forever without output if something is wrong? When I first tried it I kept getting 404's.

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+		struct = json.loads(block)
+
+		# throw exception if json is different from expected
+		try:
+			txns = struct['block']['transaction_nmt']
+			# we only need to show the block once per vm, so if several transactions are seen for same vm in same block, treat them as duplicates
+			seen_vms = set();
+			for txn in txns:
+				vm = txn['vm'] # extract Rollup ID
+				payload = payload_hex=bytearray(txn['payload']).hex()
```
sveitser, line 51: The `payload_hex` looks unused. 

If the bytes are not mutated `bytes` can be used in state of `bytearray`.

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,111 @@
...
+				print(json.dumps(vm_block_struct, indent=2))
+				print()
+						
+					
+					
+					
+			
+		except ValueError:
+			print("Error: invalid json structture")
+			exit(0)
```
sveitser, line 108: Since it's an error: `exit(1)`?

scripts/espresso_block_scanner.py
fkrell: I was thinking in the demo in which we show the rollup block. So if Ben wanted to explain it, we needed the block previously printed to stay still.

Ideas how to do better?

scripts/espresso_block_scanner.py
```diff
@@ -1,98 +1,129 @@
...
-	port=sys.argv[2]
-url="http://"+host+":"+port+"/availability/block/"
+import json
+import time
+import datetime
+import sys
+
+parser = argparse.ArgumentParser(
+    prog="Espresso Block Scanner",
+    description="Scans the espresso ledger looking for  transactions. FOr each transaction it  prints out the namespace  block containing it",
```
jbearer, line 14:
```suggestion
    description="Scans the espresso ledger looking for  transactions. For each transaction it prints out the namespace block containing it",
```

scripts/espresso_block_scanner.py
```diff
@@ -1,98 +1,129 @@
...
+    try:
+        page = urllib.request.urlopen(f"{url}{block_number}")
+    except urllib.error.HTTPError as e:
+        # "Block not produced yet. waiting for 1 second
+        time.sleep(1)
+        continue
+    except urllib.error.URLError as ex:
+        # Server is down, reset block number to 0
+
+        print("URLERROR:" + url + ". Ttrying again in 10 secs")
```
jbearer, line 39:
```suggestion
        print("URLERROR: " + url + ". Trying again in 10 secs")
```

scripts/espresso_block_scanner.py
```diff
@@ -1,98 +1,129 @@
...
+                if vm in seen_vms:
+                    continue
+                seen_vms.add(vm)
+
+                print(
+                    "### New transactions for vm "
+                    + str(vm)
+                    + " found on Espresso Block "
+                    + str(block_number)
+                    + " ###"
```
jbearer, line 73: Could use a format string here as well

jbearer (approved):
Few typos, but looks good

scripts/espresso_block_scanner.py
```diff
@@ -0,0 +1,129 @@
...
+                )
+                namespace_block_url = f"{url}{block_number}/namespace/{vm}"
```
sveitser, line 75:
```suggestion
                namespace_block_url = f"{url}-{block_number}/namespace/{vm}"
```

sveitser (approved):

scripts/espresso_block_scanner.py
sveitser: I think it's okay we can leave it as is.

PR #596, opened 2023-09-11
Deploy hotshot contract with foundry

### LINKED ISSUE #594: Add a stand-alone binary to deploy the HotShot contract
It should output as a JSON file all the information necessary to describe the deployment. This should include:
* chain ID of the chain we deployed on
* address of the deployed contract
* block number where the contract was deployed

sveitser:
Still need to test this with sepolia but the basics seem to be working.

Close #594


.gitignore
```diff
@@ -22,3 +22,7 @@ target/
 cache/
 
 node_modules/
+
```
sveitser, line 25: Actually, just saw the `.gitignore` in `contracts` already covers this.

.env
```diff
@@ -29,3 +29,8 @@ ESPRESSO_DEMO_L1_PROVIDER=$ESPRESSO_SEQUENCER_L1_PROVIDER
 ESPRESSO_DEMO_HOTSHOT_ADDRESS=$ESPRESSO_SEQUENCER_HOTSHOT_ADDRESS
 ESPRESSO_DEMO_ROLLUP_ADDRESS=0xe7f1725e7734ce288f8367e1bb143e90bb3f0512
 ESPRESSO_DEMO_ROLLUP_MNEMONIC="test test test test test test test test test test test junk"
+
+# Foundry
+MNEMONIC="test test test test test test test test test test test junk"
+SEPOLIA_RPC_URL=
```
nomaxg, line 35: Do we need these empty env variables?

nomaxg (approved):

.env
sveitser: I don't know a better place to put them. From experience the `.env` file is much more likely to be consulted than the README. Maybe we can add some comments to explain what they do.

PR #603, opened 2023-09-13
Update HotShot to tag 0.4.14

### LINKED ISSUE #598: Update to latest HotShot for Cortado deployment
HotShot has working catchup now, so we definitely want to pull that into Cortado if possible

nomaxg:
Depends on: https://github.com/EspressoSystems/hotshot-query-service/pull/221

Closes #598


.github/workflows/build_static.yml
```diff
@@ -12,6 +12,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  RUSTFLAGS: "--cfg async_executor_impl=\"async-std\" --cfg async_channel_impl=\"async-std\""
```
nomaxg, line 16: @elliedavidson @rob-maron would it be possible to have HotShot build with a default async runtime?

sequencer/src/lib.rs
```diff
@@ -224,7 +222,7 @@ pub type Storage = MemoryStorage<SeqTypes, Leaf>;
 pub type Event = hotshot::types::Event<SeqTypes, Leaf>;
 
 pub type SignatureSchemeType = BLSSignatureScheme;
-pub type SignatureKeyType = JfPubKey<SignatureSchemeType>;
+pub type SignatureKeyType = BN254Pub;
```
sveitser, line 225: Is `SignatureSchemeType` unused now?

.github/workflows/build_static.yml
rob-maron: We could probably change some of the conditional compilation attributes to use `async-std` if none are specified. Would that be an alright solution? Other than that, I don't think it's possible to set default `cfg` values without editing something like `.cargo/config.toml`.

nomaxg: yeah that sounds good!

sveitser (approved):

sequencer/src/lib.rs
nomaxg: yeah looks like it

.github/workflows/build_static.yml
elliedavidson: @DieracDelta Can you do this update when you update the profiles in our justfile? :)

PR #616, opened 2023-09-15
docker: use a network with a fixed name

### LINKED ISSUE #606: Script to start the sequencer docker compose env and deploy the HotShot contract
The compose file should define a named network that the rollup compose projects can join so that containers can talk to each other.

https://docs.docker.com/compose/networking/#use-a-pre-existing-network

sveitser:
- Improve logging output.
- Only wait 1 second before killing containers.
- Move the L1 service to the top.
- Use a 3 second block time for anvil.
- Use compose file version 3.

Close #606

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,20 @@
-version: "1"
+version: "3"
+
+networks:
+  default:
+    name: espresso-sequencer
+
 services:
+  demo-l1-network:
+    image: ghcr.io/foundry-rs/foundry:latest
+    ports:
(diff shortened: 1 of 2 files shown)


nomaxg (approved):
LGTM

docker-compose.yaml
```diff
@@ -1,5 +1,20 @@
-version: "1"
+version: "3"
+
+networks:
+  default:
+    name: espresso-sequencer
+
 services:
+  demo-l1-network:
```
nomaxg, line 8: Can we actually name this `l1` to be consistent with what we have in the OP repo?

PR #630, opened 2023-09-25
Use geth node in docker compose

nomaxg:
Anvil isn't compatible with our OP integration because of some issues with getProof


docker-compose.yaml
```diff
@@ -6,14 +6,13 @@ networks:
...
 services:
   demo-l1-network:
-    image: ghcr.io/foundry-rs/foundry:latest
+    image: ghcr.io/espressosystems/geth-l1:main
     ports:
       - $ESPRESSO_SEQUENCER_L1_PORT:8545
-    # The way the foundry image handles arguments is strange, the double quotes
-    # make it work here.
-    command: "'anvil --host 0.0.0.0 --chain-id 1337 --block-time $L1_BLOCK_TIME_SEC --accounts 20'"
-    healthcheck:
```
nomaxg, line 15: @sveitser @jbearer any ideas for a suitable healtcheck here? I don't really think we need one in the demo setup (haven't run into a situation where the L1 doesn't start up in time) but it would be safer to have

jbearer: Definitely important to have a healthcheck both for monitoring reasons and startup dependencies. I would suggest hitting the `chainid` endpoint, which tells you that both the HTTP server and the actual RPC server are up and running. Something like 
```
curl --fail http://localhost:8545 -X POST -H "Content-Type: application/json" \
  --data '{"method":"eth_chainId","params":[],"id":1,"jsonrpc":"2.0"}'
```

jbearer (approved):

docker-compose.yaml
nomaxg: sounds good, I can add this check

PR #643, opened 2023-10-02
Move L1 interactions off critical path, to asynchronous task

jbearer:
Instead of querying the L1 directly to find the latest head and finalized block each time we need to propose, we now have each sequencer node run a background task to update an in-memory snapshot of the L1 state each time there is a new L1 block, and the proposer logic just reads from this snapshot.

This has numerous benefits, including
* Drastically lower Infura usage. Instead of 2 RPC calls per HotShot block (one to fetch the head and one to fetch the finalized block) we have one RPC call per sequencer node per L1 block, plus one subscription stream per sequencer. This is far more optimized for the case where L1 blocks are much slower than HotShot blocks. For example, with a 1s HotShot block time, 12s L1 block time, and 9 HotShot nodes, the total RPC calls are reduced from 2/s to 0.75/s.
* Move L1 interactions off critical path. Before, failures or delays in Infura calls could cause a HotShot node to miss or delay a proposal. We suspect this may have contributed to irregular block production that was ultimately observed in the OP rollup. Now, Infura failures and delays can cause a sequencer node to propose with a slightly old L1 block number, but not to miss a proposal.
* Closer to the desired design. This is more similar to how the L1 interactions _should_ work once each sequencer is running its own L1 light client in the background. All we have to do to make this design production-ready, once the ValidatedState API is ready, is use ValidatedState to replace the process-global L1Client with an L1Client in the ValidatedState.

The only potential downside I see to deploying this right now is that it adds fairly complicated code that has not been thoroughly tested. However, I would argue that the failures in Cortado mean that all the local testing we did on the previous iteration was not sufficient to thoroughly test it either, and that therefore we might as well just deploy this code and stress test it in a realistic setting (Sepolia). That's what testnets are for, after all.

Edit: another potential downside is that this might increase the _per node_ cost of querying the L1, especially for very small stakers who are rarely leader. Previously, you only need to query the L1 at all when you became leader. Now, you need to be querying the L1 constantly, albeit only once every 12 seconds. IMO this is fine, since the previous design was, in my view, cheating a bit, by putting logic onto the consensus critical path that shouldn't have been there. Long term, we could optimize for small stakers by having the background L1 client only start updating when the node is about to become leader. Short term, with only 9 nodes, this is still much cheaper over all in terms of API usage than the previous design.

Local testing: I have run the sequencer unit tests. I have run the local devent with both Polygon and OP rollups deployed on the sequencer demo. End-to-end tests for both pass and the rollups run without crashing for at least 15-30 minutes. The op-e2e unit tests also work with the new sequencer images.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,5 +39,4 @@ jobs:
         run: cargo clippy --workspace --all-features --all-targets -- -D warnings
 
       - name: Audit
-        # See https://github.com/libp2p/rust-libp2p/issues/4327 for RUSTSEC-2022-0093
-        run: cargo audit --ignore RUSTSEC-2022-0093 --ignore RUSTSEC-2023-0052 --ignore RUSTSEC-2023-0053 --ignore RUSTSEC-2023-0063
+        run: cargo audit --ignore RUSTSEC-2023-0063 --ignore RUSTSEC-2023-0065
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -68,12 +68,14 @@ services:
       - ESPRESSO_SEQUENCER_CONSENSUS_SERVER_URL
       - ESPRESSO_SEQUENCER_API_PORT
       - ESPRESSO_SEQUENCER_STORAGE_PATH
-      - ESPRESSO_SEQUENCER_L1_PROVIDER
+      - ESPRESSO_SEQUENCER_L1_WS_PROVIDER
       - ESPRESSO_SEQUENCER_L1_USE_LATEST_BLOCK_TAG
       - RUST_LOG
       - RUST_LOG_FORMAT
     depends_on:
       orchestrator:
+        condition: service_healthy
+      demo-l1-network:
         condition: service_healthy
     stop_grace_period: 1s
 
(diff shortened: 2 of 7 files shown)


jbearer:
I am still seeing irregular block production locally, so while I think this is an architectural improvement, and at least it eliminates the variable of slow L1 RPC calls on the critical path, there is still some underlying problem in consensus I think

philippecamacho:
Regarding the number of queries to infura, do we consider at some point relying on [Amazon Managed Blockchain](https://aws.amazon.com/managed-blockchain/pricing/) service? It seems that if the number of nodes/queries increases significantly it might save money. WDYT  @Ancient123?

philippecamacho (approved):
LGTM

nomaxg (approved):
lgtm

Ancient123:
> Regarding the number of queries to infura, do we consider at some point relying on [Amazon Managed Blockchain](https://aws.amazon.com/managed-blockchain/pricing/) service? It seems that if the number of nodes/queries increases significantly it might save money. WDYT @Ancient123?

My quick napkin math says this is worth looking into. There are a few hangups that could be a problem, and I would have to do some significant spreadsheeting to see how this compares to our current setup with infura.

PR #687, opened 2023-10-05
Add reorg function to Anvil automation and example reorg test for L1 client

jbearer:
This also fixes a bug in the L1 client where the reconnection logic reconnects the WebSocket stream but not that actual Provider itself, which doesn't work.


philippecamacho (approved):
Tests are passing on my machine.

PR #692, opened 2023-10-05
Point back to main

nomaxg:


Cargo.toml
```diff
@@ -11,10 +11,3 @@ members = [
   "utils",
 ]
 
-[patch."https://github.com/EspressoSystems/hotshot"]
```
Ancient123, line 14: Was this supposed to go away?

nomaxg: Oh I see - we're still developing off of that branch, I assumed we had merged to HotShot main

PR #696, opened 2023-10-06
Configurable chain ID for Anvil utility

jbearer:
This is needed to use this Anvil automation with the zkEVM stack for the zkEVM reorg test.


philippecamacho (approved):
LGTM

Issue #704, opened 2023-10-11
Endpoint to expose most recent hotshot contract address


nomaxg:
Add an endpoint:
`hotshot_contract/get` 

The API should return the hex encoded string that represents the hotshot contract 0xED15E1FE0789c524398137a066ceb2EF9884E5D8.

Define the API [here](https://github.com/EspressoSystems/espresso-sequencer/tree/main/sequencer/api). The address should live in the sequencer API state, and can be passed in [here](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/main.rs#L60) at initialization:

PR #707, opened 2023-10-12
Add an endpoint to get the address of HotShot contract

ImJeremyHe:
#704

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ target/
 node_modules/
 
 **/*DS_Store
+
+# solidity files cache
+cache
--- a/sequencer/api/commitment_task.toml
+++ b/sequencer/api/commitment_task.toml
@@ -1,0 +1,3 @@
+[route.gethotshotcontract]
+PATH = ["/hotshot_contract"]
+DOC = "Get the address of HotShot contract on Layer1."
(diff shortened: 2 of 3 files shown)


sequencer/src/api.rs
```diff
@@ -1,4 +1,7 @@
-use crate::{network, transaction::Transaction, Header, Leaf, NamespaceProofType, Node, SeqTypes};
+use crate::{
```
jbearer, line 1: Ah, ok, I misunderstood something. You have discovered that only the commitment task deals with the HotShot contract, not the sequencers themselves, which is correct. However, the commitment task now runs as a completely separate service. So I don't think we should import commitment task stuff into the sequencer API. Instead, we should add this endpoint to the commitment task API .

Currently, the commitment task currently only has [a stub API](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/bin/commitment-task.rs#L128-L129), which gives us generated healthcheck and version endpoints and nothing else. You could pull that `start_http_server` out into another file and add a non-trivial endpoint like the one you defined here.

jbearer (changes requested):

jbearer (approved):
Couple minor comments, overall look good

PR #708, opened 2023-10-12
Point at HotShot main

jbearer:
Per the new methodology described by @elliedavidson, main will always be the latest stable HotShot release, while the new develop branch will be used for unstable development. Thus, we can point at main and not an (old) stable release tag.


philippecamacho (approved):
LGTM

PR #711, opened 2023-10-13
Non static dockers and build updates

Ancient123:
- Build AMD64 and ARM64 dockers as part of normal build
- Add concurrency option and remove old build cancelling tool
- Add option so dependabot doesn't try to push to cachix


Ancient123:
Please forgive the first runs of `build-arm` being slow. There is no rust cache for it to work off of.

nomaxg (approved):

jbearer (approved):

PR #712, opened 2023-10-13
Optimize docker builds

Ancient123:
Reduce build times by a minute by homogenizing installed packages. This means we only spend time fetching packages and installing on the first docker and the rest are fully cached.


Ancient123:
Before
Screenshot 2023-10-13 at 15 22 50

After
Screenshot 2023-10-13 at 15 22 27

nomaxg (approved):

PR #722, opened 2023-10-24
feat: add PlonkVerifier contracts and tests

### LINKED ISSUE #702: [contract] Move PlonkVerifier contract over
This issue aim to:

- [x] migrate `PlonkVerifier` contract code over
- [x] write unit test in solidity
- [x] write differential test against our jellyfish code in rust
  - [x] write Rust command line tool to using `clap` to help run jellyfish code
  - [x] write FFI code in Solidity to invoke rust command

alxiong:
Resolves #702


alxiong:
I will deal with some minor TODOs and fix CI failure tmr, but I think it's good to start reviewing this PR now. 

@sveitser @philippecamacho

.solhint.json
```diff
@@ -10,14 +10,12 @@
...
       , "func-visibility": ["error", { "ignoreConstructors": true }]
       , "imports-on-top": "error"
-      , "max-line-length": ["error", 100]
       , "modifier-name-mixedcase": "error"
       , "named-parameters-mapping": "error"
       , "no-console": "error"
       , "no-empty-blocks": "error"
       , "no-unused-vars": "error"
       , "no-global-import": "off"
-      , "private-vars-leading-underscore": ["error", { "strict": true }]
```
sveitser, line 20: @alxiong I thought the issue with this was fixed by updating solhint.

contracts/rust/Cargo.toml
```diff
@@ -8,27 +8,33 @@ edition = "2021"
...
-
-# This crate requires a newer version than 0.1.2-patch.1 that's used elsewhere
-# in this workspace. However not pinning any revision currently breaks
-# dependency resolution.
-jf-primitives = { git = "https://github.com/EspressoSystems/jellyfish", features = [
-    "std",
-] }
-jf-utils = { git = "https://github.com/EspressoSystems/jellyfish", rev = "483fb97324d42724720f9155f200973a3b83a7b4" }
+itertools = "0.10.3"
+jf-plonk = { git = "https://github.com/EspressoSystems/jellyfish", features = ["test-apis", "test-srs"], branch = "more-test-apis" }
```
sveitser, line 29: Should we stay pinned to this branch?

sveitser:
Currently I get 2 failures when running forge test

```
Ran 16 test suites: 16 tests passed, 2 failed, 0 skipped (18 total tests)

Failing tests:
Encountered 1 failing test in contracts/test/PlonkVerifier.t.sol:PlonkVerifier_computeChallenges_Test
[FAIL. Reason: Assertion failed. Counterexample: calldata=0x0x4c92422b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000035541b4d458e1cb2a3f47449bd72a5662dd2c0c21e5c6a6e3399a6986e8277e8e961b25350069538a884b453c367990c9b1c21416a4b0000000000000000000000, args=[0, [], 0x541b4d458e1cb2a3f47449bd72a5662dd2c0c21e5c6a6e3399a6986e8277e8e961b25350069538a884b453c367990c9b1c21416a4b]] testFuzz_computeChallenges_matches(uint64,uint256[],bytes) (runs: 0, μ: 0, ~: 0)

Encountered 1 failing test in contracts/test/Transcript.t.sol:Transcript_appendGroupElement_Test
[FAIL. Reason: EvmError: Revert Counterexample: calldata=0x0x459b2b260000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000608b60e34a2dfbd66cc298e855f381cefba459c88dcda34bcd9ddc727709000000e5b30b0d08de663ddd47c98d961948b26be9de3447d6704a96c5117e569e64df0000000000000000000000000000000000000000000000000000000000000000, args=[(0x, [0x8b60e34a2dfbd66cc298e855f381cefba459c88dcda34bcd9ddc727709000000, 0xe5b30b0d08de663ddd47c98d961948b26be9de3447d6704a96c5117e569e64df]), 0]] testFuzz_appendGroupElement_matches((bytes,bytes32[2]),uint256) (runs: 2, μ: 45148, ~: 45148)

Encountered a total of 2 failing tests, 16 tests succeeded
```

contracts/rust/Cargo.toml
alxiong: I plan to pin to a rev on the main branch once [this PR](https://github.com/EspressoSystems/jellyfish/pull/398) is merged.

.solhint.json
alxiong: the main problem is conflict requirements by forge linters and solhint linters.
when enabled, I would have:
```
contracts/src/libraries/PlonkVerifier.sol
   28:5  error  'COSET_K1' should start with _                           private-vars-leading-underscore
   30:5  error  'COSET_K2' should start with _                           private-vars-leading-underscore
   32:5  error  'COSET_K3' should start with _                           private-vars-leading-underscore
   34:5  error  'COSET_K4' should start with _                           private-vars-leading-underscore
```

but naming them `_COSET_K4` would upset `forge build`

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -0,0 +1,856 @@
...
+        BN254.validateG1Point(proof.wire2);
+        BN254.validateG1Point(proof.wire3);
+        BN254.validateG1Point(proof.wire4);
+        BN254.validateG1Point(proof.prodPerm);
+        BN254.validateG1Point(proof.split0);
+        BN254.validateG1Point(proof.split1);
+        BN254.validateG1Point(proof.split2);
+        BN254.validateG1Point(proof.split3);
+        BN254.validateG1Point(proof.split4);
+        BN254.validateG1Point(proof.zeta);
```
sveitser, line 137: Is zetaOmega missing?

alxiong: very good catch! G... I failed miserably by not catching this in tests, but you catch it using bare eyes. LOL irony.

alxiong: ah! I know why my test is missing it. 
1. I forget to test for `zetaOmega`
2. `vm.expectRevert()` should only be invoked once in each tests... I misuse it.

alxiong: fixed in [1d7c6c6](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/1d7c6c646e93c2077328e1a30d50e75f1327e04f)

contracts/rust/Cargo.toml
alxiong: done in [af7266c](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/af7266c6b8bca80d4be59103cbe40187092c798f)

alxiong:
This PR is properly ready for full review!

contracts/rust/src/bin/diff_test.rs
```diff
@@ -0,0 +1,1072 @@
...
+            println!("{}", (res,).encode_hex());
+        }
+        Action::PlonkPreparePcsInfo => {
+            let arg1 = cli.arg1.as_ref().expect("Should provide arg1=verifyingKey");
+            let arg2 = cli.arg2.as_ref().expect("Should provide arg2=publicInput");
+            let arg3 = cli.arg3.as_ref().expect("Should provide arg3=proof");
+            let arg4 = cli
+                .arg4
+                .as_ref()
+                .expect("Should provide arg3=extraTranscriptInitMsg");
```
sveitser, line 304:
```suggestion
                .expect("Should provide arg4=extraTranscriptInitMsg");
```

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,378 @@
...
+
+            // happy path
+            assert(V.batchVerify(verifyingKeys, publicInputs, proofs, extraTranscriptInitMsgs));
+
+            // unhappy path
+            // wrong vk
+            IPlonkVerifier.VerifyingKey[] memory badVks = verifyingKeys;
+            badVks[0].q1 = BN254.negate(verifyingKeys[0].q1);
+            badVks[0].qEcc = BN254.negate(verifyingKeys[0].qEcc);
+            assert(!V.batchVerify(badVks, publicInputs, proofs, extraTranscriptInitMsgs));
```
alxiong, line 189: I read "[data location and assignment behavior](https://docs.soliditylang.org/en/latest/types.html#data-location-and-assignment-behavior)" today and realize that this test logic is wrong.

cuz the new `badVks` mutation will also affect the underlying `verifyingKeys` since they are both marked as `memory`, thus the following bad tests are not isolating the bad input arguments.

I will work on this next working day.

alxiong: fixed in [bdf18ac](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/bdf18acddacaf8702c3c403bfb9346f1cd1c9056)

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+contract PlonkVerifier_batchVerify_Test is PlonkVerifierCommonTest {
+    /// forge-config: default.fuzz.runs = 30
+    /// @dev Test if some of the user inputs are invalid
+    function testFuzz_revertWhenInvalidArgs(
+        IPlonkVerifier.VerifyingKey[] memory verifyingKeys,
+        uint256[][] memory publicInputs,
+        IPlonkVerifier.PlonkProof[] memory proofs,
+        bytes[] memory extraTranscriptInitMsgs
+    ) external {
+        vm.assume(
```
philippecamacho, line 154: The tests still pass on my machine when I comment this line.

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+        IPlonkVerifier.PlonkProof memory proof = dummyProof(42);
+
+        // we are testing the `nthPoint` in the `proof`,
+        // only mutating a single field element (either x or y coordinate)
+        // There are 13 points in total.
+        nthPoint = bound(nthPoint, 0, 12);
+
+        assembly {
+            if testX {
+                // muteate the x coordinate
```
philippecamacho, line 288:
```suggestion
                // mutate the x coordinate
```

contracts/test/PlonkVerifier.t.sol
alxiong: this is perfectly normal and expected, `vm.assume` is just a rejection sampling, if the conditions are not met, then they will generate another input, 
in this case, we are trying to test if bad inputs will cause revert. and fuzz-generated inputs (that are not proper proofs and vk and public inputs) are bad most of the time!! so it's normal that all of them revert.

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+                mstore(mload(add(proof, mul(0x20, nthPoint))), 0x1234)
+            }
+            // else, mutate y coordinate
+            mstore(add(mload(add(proof, mul(0x20, nthPoint))), 0x20), 0x1234)
+        }
+
+        vm.expectRevert();
+        V._validateProof(proof);
```
philippecamacho, line 296: How do we know the proof validation failed because the group element is invalid and not because the proof is incorrect w.r.t. to the public inputs / verification key?

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+        a = bound(a, 0, BN254.R_MOD - 1);
+        BN254.validateScalarField(a);
+        return a;
+    }
+
+    /// @dev Sanitize all values in `a` to be valid scalar fields Bn254::Fr.
+    /// This is helpful to sanitize fuzzer-generated random `uint[]` values.
+    function sanitizeScalarFields(uint256[] memory a) public view returns (uint256[] memory) {
+        for (uint256 i = 0; i < a.length; i++) {
+            a[i] = bound(a[i], 0, BN254.R_MOD - 1);
```
sveitser, line 39: Wonder why this doesn't call the function above.

contracts/test/PlonkVerifier.t.sol
alxiong: wait... what do you mean?
we are testing `_validateProof` function, which only checks group & field elements validity, not doing anything else. (it doesn't even accept public inputs or vk)

contracts/test/PlonkVerifier.t.sol
alxiong: sure, I will use `sanitizeScalarField(a[i])` instead

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+
+        vm.expectRevert();
+        V._validateProof(proof);
+    }
+}
+
+contract PlonkVerifier_preparePcsInfo_Test is PlonkVerifierCommonTest {
+    /// forge-config: default.fuzz.runs = 5
+    /// @dev Test `preparePcsInfo` matches that of Jellyfish
+    function testFuzz_preparePcsInfo_matches(
```
sveitser, line 324:
```
Encountered 1 failing test in contracts/test/PlonkVerifier.t.sol:PlonkVerifier_preparePcsInfo_Test
[FAIL. Reason: InvalidPolyEvalArgs() Counterexample: calldata=0x0x99fdc07100000000000000000000000000000000000000000000000000000000000005fe000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000003b2, args=[1534, [], 0x00000000000000000000000000000000000000000000000000000000000003b2]] testFuzz_preparePcsInfo_matches(uint64,uint256[],bytes) (runs: 1, μ: 8001859, ~: 8001859)
```

foundry.toml
```diff
@@ -26,3 +27,6 @@ sepolia = { key = "${ETHERSCAN_API_KEY}" }
 line_length=100
 bracket_spacing=true
 wrap_comments=true
+
+[profile.ci]
+fuzz = { run = 50 }
```
sveitser, line 32:
```suggestion
fuzz = { runs = 50 }
```

Should we run a bit more?

I also noticed that the inline comments seems to take precedence over setting it via env var or in foundry.toml which seems weird. In that case should we not use the inline comments?

alxiong: I use inline ones previously because some tests run slower than others, but now the released built `diff-test` binary, all of these tests are fast, so I will remove any inline configs for fuzzer (just use 256 runs by default) and even remove this as well (256 for ci seems good enough)?

sveitser: Or we could add something like the following and run it once a day on the CI.

```
[profile.fuzz]
fuzz = { runs = 10000 }
```

flake.nix
```diff
@@ -196,6 +198,7 @@
                 # Prevent cargo aliases from using programs in `~/.cargo` to avoid conflicts
                 # with rustup installations.
                 export CARGO_HOME=$HOME/.cargo-nix
+                export PATH="./target/release:$PATH"
```
sveitser, line 201:
```suggestion
                export PATH="$PWD/target/release:$PATH"
```

contracts/test/PlonkVerifier.t.sol
alxiong: good catch, I will add test logic for empty `publicInput` array.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -0,0 +1,1072 @@
...
+use num_traits::Num;
+
+#[derive(Parser)]
+#[command(author, version, about, long_about=None)]
+struct Cli {
+    /// Identifier for the functions to invoke in Jellyfish
+    #[arg(value_enum)]
+    action: Action,
+    /// Optional 1st argument for the `action`
+    arg1: Option<String>,
```
nomaxg, line 41: How about a vector to allow arbitrary # of args:
```
    #[clap(value_parser, num_args = 1.., value_delimiter = ' ')]
    pub args: Vec<String>,
```

contracts/rust/src/bin/diff_test.rs
```diff
@@ -0,0 +1,1072 @@
...
+// ------- Helper functions and structs --------
+// ---------------------------------------------
+
+// constant in hex string copied from hardcoded constants from solidity contracts
+
+// TODO: (alex) change to simply using `MontFp!("0x..")` after
+// <https://github.com/arkworks-rs/algebra/pull/635> is on a tag release
+// Return cosets coefficients for circuits over BN254.
+fn coset_k() -> Vec<Fr> {
+    vec![
```
nomaxg, line 409: Not a big deal, but this would be more readable if you had a vec of hex strings that you mapped over to get the field elems

contracts/rust/src/bin/diff_test.rs
```diff
@@ -0,0 +1,1072 @@
...
+            .unwrap(),
+        ),
+    ]
+}
+
+/// Returns `OpenKeys` for KZG10 over BN254 curve from Aztec's SRS
+fn open_key() -> OpenKey<Bn254> {
+    let g = G1Affine::new_unchecked(MontFp!("1"), MontFp!("2"));
+    let h = G2Affine::new(
+        Fp2::new(
```
nomaxg, line 446: Same here

contracts/rust/src/bin/diff_test.rs
```diff
@@ -0,0 +1,1072 @@
...
+        .iter()
+        .map(|cs| cs.public_input().unwrap())
+        .collect();
+
+    Ok(izip!(proofs, ver_keys, public_inputs, extra_msgs, domain_sizes).collect())
+}
+
+// Different `m`s lead to different circuits.
+// Different `a0`s lead to different witness values.
+fn gen_circuit_for_test<F: PrimeField>(m: usize, a0: usize) -> Result<PlonkCircuit<F>> {
```
nomaxg, line 1040: If this is just for testing purposes, maybe you can just panic here instead of returning a result?

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+        // only mutating a single field element (either x or y coordinate)
+        // There are 13 points in total.
+        nthPoint = bound(nthPoint, 0, 12);
+
+        assembly {
+            if testX {
+                // muteate the x coordinate
+                mstore(mload(add(proof, mul(0x20, nthPoint))), 0x1234)
+            }
+            // else, mutate y coordinate
```
sveitser, line 291: The `y` coordinate is always mutated.

foundry.toml
sveitser: > I use inline ones previously because some tests run slower than others, but now the released built `diff-test` binary, all of these tests are fast, so I will remove any inline configs for fuzzer (just use 256 runs by default) and even remove this as well (256 for ci seems good enough)?

Yeah if that's not too slow that is fine I think.

Since it's almost free and only costs CPU time I think it's not a bad idea to run more iterations on the CI occasionally, but we don't have to do it as part of this PR.

contracts/test/PlonkVerifier.t.sol
philippecamacho: Ah sorry, I get it now. Thanks.

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+    /// forge-config: default.fuzz.runs = 15
+    /// @dev Randomly pick field in a proof mutate it to invalid value
+    /// test if our check will revert.
+    function testFuzz_RevertIfProofContainsInvalidField(uint256 nthField) external {
+        // a valid proof
+        IPlonkVerifier.PlonkProof memory proof = dummyProof(42);
+        uint256 invalidField = BN254.R_MOD;
+
+        // we are testing the `nthField` in the `proof`,
+        // There are 10 points in total (with 13 points in front)
```
philippecamacho, line 308:
```suggestion
        // There are 10 field elements in total (with 13 points in front)
```

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+        assertEq(chal.gamma, c.gamma);
+        assertEq(chal.zeta, c.zeta);
+        assertEq(chal.v, c.v);
+        assertEq(chal.u, c.u);
+    }
+}
+
+contract PlonkVerifier_prepareEvaluations_Test is PlonkVerifierCommonTest {
+    /// forge-config: default.fuzz.runs = 5
+    /// @dev Test if combinng the polynomial evaluations into a single evaluation is done correctly
```
philippecamacho, line 405:
```suggestion
    /// @dev Test if combining the polynomial evaluations into a single evaluation is done correctly
```

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+    function testFuzz_prepareEvaluations_matches(
+        uint64 seed,
+        uint256 linPolyConstant,
+        uint256[30] memory scalars
+    ) external {
+        IPlonkVerifier.PlonkProof memory proof = dummyProof(seed);
+        linPolyConstant = sanitizeScalarField(linPolyConstant);
+        uint256[] memory commScalars = sanitizeScalarFields(copyCommScalars(scalars));
+
+        string[] memory cmds = new string[](5);
```
philippecamacho, line 416: Would it be worth writing a function for preparing this command line and calling `vm.ffi` in order to reduce code duplication?

contracts/test/PlonkVerifier.t.sol
```diff
@@ -0,0 +1,432 @@
...
+            );
+
+            assert(V.batchVerify(verifyingKeys, publicInputs, proofs, extraTranscriptInitMsgs));
+        }
+    }
+
+    /// forge-config: default.fuzz.runs = 30
+    /// @dev Test when bad public inputs are supplied, the verification should fail
+    /// We know our `gen_circuit_for_test` in `diff_test.rs` has only 3 public inputs
+    function testFuzz_badPublicInputs_fails(uint256[3] calldata randPublicInput) external {
```
philippecamacho, line 188: Should we add a similar test with bad verifying keys?

contracts/rust/src/bin/diff_test.rs
alxiong: it is for testing only, I just find the syntax sugar `?` much easier to read than bunch of `unwrap()` when building the circuit. Consider:

```rust
    let c = cs.create_public_variable(
        (cs.witness(b[1])? + cs.witness(a[0])?) * (cs.witness(b[1])? - cs.witness(a[0])?),
    )?;

// v.s. 
    let c = cs.create_public_variable(
        (cs.witness(b[1]).unwrap() + cs.witness(a[0]).unwrap()) * (cs.witness(b[1]).unwrap() - cs.witness(a[0]).unwrap()),
    )?;
```

that said, I will turn more `?` into `expect("reason")` in the `fn gen_plonk_proof_for_test()` which is the wrapper for this `fn gen_circuit_for test` for faster panic with debuggable reasons.

contracts/test/PlonkVerifier.t.sol
alxiong: This is up for debate.
I used to have it, but if we consider the actual usage: VK is hardcoded contract that we ship our contract with, it's not a user-provided input, thus its integrity/correctness is not an issue.

OTOH, it's also fine to test the implementation of `batchVerify()` which should fail if there's an honest mistake in the process of hardcoding the VK contract. 🤔

contracts/test/PlonkVerifier.t.sol
alxiong: Not really imo. Few reasons:
- varying number of arguments: some commands offer more (this might be fine with Solidity's polymorphism support)
- varying argument types: some arguments are `bytes32` (if they are abi encoded), some are `uint256`, some are `uint64`
- varying return types: while the result of `vm.ffi() -> bytes memory` is always `bytes`, when you need concrete/test-specific info for abi decoder
- finally, I find it clearer to write this way and not add too many re-direction. (Optimism's code also does it explicitly without any helper function)

contracts/test/PlonkVerifier.t.sol
alxiong: done in [90bf8d1](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/90bf8d1c42a6613213ef9efe4711a57e3c803a89)

foundry.toml
alxiong: done in [90bf8d1](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/90bf8d1c42a6613213ef9efe4711a57e3c803a89)

contracts/rust/src/bin/diff_test.rs
alxiong: good suggestions! all of your comments addressed in [fc1484e](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/fc1484eccc33744db412645d721146bdf553c0de)

contracts/test/PlonkVerifier.t.sol
alxiong: fixed in [b914f77](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/b914f7761975e341ddd7f49eea9125f8f72efbd6)

contracts/test/PlonkVerifier.t.sol
alxiong: After some thoughts, I agree that we should add this bad path test, done in [b914f77](https://github.com/EspressoSystems/espresso-sequencer/pull/722/commits/b914f7761975e341ddd7f49eea9125f8f72efbd6)

alxiong:
Apologize for accidentally merged #743 by pushing my commits addressing PR comments to the wrong remote branch.

Please ignore file `interfaces/IStakeTable.sol` and `libraries/EdOnBn254.sol` in this PR. I'm addressing comments on them in a separate PR.

philippecamacho (approved):
Great work!

sveitser (approved):
LGTM

contracts/rust/src/bin/diff_test.rs
nomaxg: makes sense! I see what you're saying

PR #728, opened 2023-11-01
Add temporary route to return block header

nomaxg:
Necessary for our Arbitrum integration, which iteratively fetches block headers for sequencing purposes

### DIFF
--- a/sequencer/api/availability.toml
+++ b/sequencer/api/availability.toml
@@ -45,11 +45,15 @@ where a `Header` has the schema:
 
 ```json
 {
-    "timestamp": "integer",
-    "l1_block": {
-        "timestamp": "integer",
-        "number": "integer",
-    },
+    "metadata": {
+      "timestamp": "integer",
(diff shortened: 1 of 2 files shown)


sequencer/api/availability.toml
```diff
@@ -58,3 +58,22 @@ where a `Header` has the schema:
...
 """
+
+[route.getheader]
+PATH = ["header/:height"]
+":height" = "Integer"
+DOC = """
+If the header at the provided height is available, returns
+
+```json
+{
```
jbearer, line 69: I guess this is copied from above but this format is actually out of data. The header now looks like

```json
{
  "metadata": {
    "timestamp": "integer",
    "l1_head": "integer",
    "l1_finalized": { ... },
  },
  "transactions_root": { ... },
}
```

jbearer (approved):
Looks good, we just have some out-of-date documentation

sveitser (approved):

PR #739, opened 2023-11-08
Update bindings, fix CI

sveitser:
Third opinion for https://github.com/EspressoSystems/espresso-sequencer/pull/738


jbearer:
Cool, if this worked I guess the problem is probably my Nix install

jbearer (approved):

PR #743, opened 2023-11-13
feat: add stake table interfaces contract

### LINKED ISSUE #673: Define and review the interface of the stake table
interfaces need to support:

READ:
- stake table commitments (for the `Snapshot::LastEpochStart` and `Snapshot::EpochStart`)
- total stake and total number of participants/keys
- stake lookup (fine-grained results maybe through multiple getters, one returning actual owned amount, the other returning voting power with delegated included)
- registering and exiting queue info,

WRITE:
- register (batch?)
- deposit (you can deposit more once register)
- exit (signal your withdraw intent and get queued)
- withdraw
- delegate/approve
- disapprove

alxiong:
Closes #673 

_note: I'm currently targeting `plonk-verifier` branch for easier git diff for the reviewers, and will change to `main` once #722 is merged_

### DIFF
--- a/contracts/src/interfaces/IStakeTable.sol
+++ b/contracts/src/interfaces/IStakeTable.sol
@@ -1,0 +1,127 @@
+// SPDX-License-Identifier: Unlicensed
+
+pragma solidity ^0.8.0;
+
+import { BN254 } from "bn254/BN254.sol";
+import { EdOnBN254 } from "../libraries/EdOnBn254.sol";
+
+/// @title Interface for stake table that keep track of validators' external key and staked amount.
+/// @dev Stake delegation happens in a separate DelegationPool contract, and specific to
+/// instantiation thus not part of this interface
+interface IStakeTable {
+    /// @notice Supported stake type, either using native token or re-staked using ETH
+    enum StakeType {
+        Native,
+        Restake
+    }
+
+    /// @dev (sadly, Solidity doesn't support type alias on non-primitive types)
+    // We avoid declaring another struct even if the type info helps with readability,
+    // extra layer of struct introduces overhead and more gas cost.
+    //
+    /// @dev BLS verification key (used for Consensus voting in HotShot) is `BN254.G1Point`.
+    // `type BlsVerKey is BN254.G1Point;`
+    //
+    /// @dev Verification key of a Schnorr signature is just a `EdOnBN254.EdOnBN254Point`
+    // `type SchnorrVerKey is EdOnBN254.EdOnBN254Point;`
+
+    /// @notice Represents a HotShot validator node
+    /// In the dual-staking model, a HotShot validator could have multiple `Node` entries.
+    /// @param account The Ethereum account of the validator.
+    /// @param stakeType The type of token staked.
+    /// @param balance The amount of token staked.
+    /// @param registerEpoch The starting epoch for the validator.
+    /// @param exitEpoch The ending epoch for the validator.
+    struct Node {
+        address account;
+        StakeType stakeType;
+        uint64 balance;
+        uint64 registerEpoch;
+        uint64 exitEpoch;
+    }
+
+    // === Table State & Stats ===
+
+    /// @notice Get the commitment of the stake table used in current voting (i.e. snapshot at
+    /// the start of last epoch)
+    function votingStakeTableCommitment() external view returns (bytes32);
+    /// @notice Get the commitment of the stake table frozen for change (i.g. snapshot at the start
+    /// of the current epoch)
+    function frozenStakeTableCommitment() external view returns (bytes32);
+
+    /// @notice Get the total stake of the registered keys in the latest stake table (Head)
+    function totalStake() external view returns (uint256);
+    /// @notice Get the total number of stakers, uniquely identified by their `blsVK`
+    function totalKeys() external view returns (uint32);
+    /// @notice Get the total stake of the registered keys in the voting stake table
+    /// (LastEpochStart)
+    function totalVotingStake() external view returns (uint256);
+
+    /// @notice Look up the balance of `blsVK`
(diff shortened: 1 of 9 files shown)


contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// smaller than the current epoch)
+    ///
+    /// @return success status
+    ///
+    /// @dev No validity check on `schnorrVK`, as it's assumed to be sender's responsibility,
+    /// the contract only treat it as auxiliary info submitted by `blsVK`.
+    function register(
+        BN254.G1Point calldata blsVK,
+        EdOnBN254.EdOnBN254Point calldata schnorrVK,
+        uint64 amount,
```
chancharles92, line 95: Is `uint64` type sufficient for `amount`?

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,127 @@
...
+    // `type SchnorrVerKey is EdOnBN254.EdOnBN254Point;`
+
+    /// @notice Represents a HotShot validator node
+    /// In the dual-staking model, a HotShot validator could have multiple `Node` entries.
+    /// @param account The Ethereum account of the validator.
+    /// @param stakeType The type of token staked.
+    /// @param balance The amount of token staked.
+    /// @param registerEpoch The starting epoch for the validator.
+    /// @param exitEpoch The ending epoch for the validator.
+    struct Node {
```
philippecamacho, line 35: This struct should contain the BLS and Schnorr public keys right?

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    /// @notice Get the commitment of the stake table used in current voting (i.e. snapshot at
+    /// the start of last epoch)
+    function votingStakeTableCommitment() external view returns (bytes32);
+    /// @notice Get the commitment of the stake table frozen for change (i.g. snapshot at the start
+    /// of the current epoch)
+    function frozenStakeTableCommitment() external view returns (bytes32);
+
+    /// @notice Get the total stake of the registered keys in the latest stake table (Head)
+    function totalStake() external view returns (uint256);
```
philippecamacho, line 53: Should we differentiate between native and restake?

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,127 @@
...
+        bytes calldata blsSig,
+        uint64 maxWait
+    ) external returns (bool);
+
+    /// @notice Deposit more stakes to registered keys
+    ///
+    /// @param blsVK The BLS verification key
+    /// @param amount The amount to deposit
+    /// @return (newBalance, effectiveEpoch) the new balance effective at a future epoch
+    function depoist(BN254.G1Point calldata blsVK, uint64 amount)
```
philippecamacho, line 106:
```suggestion
    function deposit(BN254.G1Point calldata blsVK, uint64 amount)
```

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// withdraw past their `exitEpoch`.
+    ///
+    /// @param blsVK The BLS verification key to withdraw
+    /// @return The total amount withdrawn, equal to `Node.balance` associated with `blsVK`
+    function withdraw(BN254.G1Point calldata blsVK) external returns (uint64);
+
+    /// @notice Advance to the next epoch, the new `votingStakeTableCommitment` will be
+    /// `frozenStakeTableCommitment` and the new `frozenStakeTableCommitment` will be
+    /// `curEpochComm`,
+    function advanceEpoch(bytes32 curEpochComm) external returns (bool);
```
philippecamacho, line 126: Who calls this function and why?

alxiong:
I'm sorry I messed up and push to the wrong remote branch, somehow when I revert changes, Github recognize it as "merged" ...

contracts/src/interfaces/IStakeTable.sol
alxiong: I'm following our [doc here](https://docs.espressosys.com/sequencer/~/revisions/oijyFXG7VlZ4pxqo4Lgq/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table#data-structures) by @jbearer 

64 bits should be enough?

contracts/src/interfaces/IStakeTable.sol
alxiong: Half correct, should contains Schnorr key (and other auxiliary info)
there would be a mapping in storage keeping track of `mapping(BlsVerKey => Node)`, so no need to store BLS key again in the `Node` struct.

contracts/src/interfaces/IStakeTable.sol
alxiong: very interesting point! I see these options:
1. keep as it is, let instantiations specify its own conversion, the output is neither native nor restaked total stake, but an interpreted/exchange-rate-weighted total stake
2. returns `(uint256, uint256)`, first one for native, second one for restaked
3. returns `(StakeType[] memory, uint256[] memory)`

Option 3 is the most flexible and future-proof API, since it allows StakeType to be extended to more types (e.g. `EthRestaked, BtcRestaked, ...`, but Option 2 is most straightforward if the assumption that we only support ETH and native token is valid for the foreseeable future.

WDYT?

contracts/src/interfaces/IStakeTable.sol
alxiong: Think of this API as the "setter for stake commitments" (as we have getter as `votingStakeTableCommitment()` and `frozenStakeTableCommitment()`)

So light client prover will submit new states and call this function to update the new finalized stake table commitments.

PR #748, opened 2023-11-14
feat: add Stake Table contract API

### LINKED ISSUE #673: Define and review the interface of the stake table

alxiong:
Closes #673 

This is a replicate of #743 which is accidentally closed.


contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,129 @@
...
+    /// smaller than the current epoch)
+    ///
+    /// @return success status
+    ///
+    /// @dev No validity check on `schnorrVK`, as it's assumed to be sender's responsibility,
+    /// the contract only treat it as auxiliary info submitted by `blsVK`.
+    function register(
+        BN254.G1Point calldata blsVK,
+        EdOnBN254.EdOnBN254Point calldata schnorrVK,
+        uint64 amount,
```
alxiong, line 97: > [chancharles92](https://github.com/chancharles92)
Is uint64 type sufficient for amount?

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    /// @notice Represents a HotShot validator node
+    /// In the dual-staking model, a HotShot validator could have multiple `Node` entries.
+    /// @param account The Ethereum account of the validator.
+    /// @param stakeType The type of token staked.
+    /// @param balance The amount of token staked.
+    /// @param registerEpoch The starting epoch for the validator.
+    /// @param exitEpoch The ending epoch for the validator.
+    /// @param schnorrVK The Schnorr verification key associated.
+    struct Node {
```
alxiong, line 36: > [philippecamacho](https://github.com/philippecamacho)
This struct should contain the BLS and Schnorr public keys right?

> [alxiong](https://github.com/alxiong)
Half correct, should contains Schnorr key (and other auxiliary info)
there would be a mapping in storage keeping track of mapping(BlsVerKey => Node), so no need to store BLS key again in the Node struct.

Done in [47aa781](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/47aa7813c38e8b4a5b46fb20931924d3f836524e)

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    /// @notice Get the commitment of the stake table used in current voting (i.e. snapshot at
+    /// the start of last epoch)
+    function votingStakeTableCommitment() external view returns (bytes32);
+    /// @notice Get the commitment of the stake table frozen for change (i.g. snapshot at the start
+    /// of the current epoch)
+    function frozenStakeTableCommitment() external view returns (bytes32);
+
+    /// @notice Get the total stake of the registered keys in the latest stake table (Head)
+    function totalStake() external view returns (uint256);
```
alxiong, line 55: > [philippecamacho](https://github.com/philippecamacho)
Should we differentiate between native and restake?

alxiong: very interesting point! I see these options:

    keep as it is, let instantiations specify its own conversion, the output is neither native nor restaked total stake, but an interpreted/exchange-rate-weighted total stake
    returns (uint256, uint256), first one for native, second one for restaked
    returns (StakeType[] memory, uint256[] memory)

Option 3 is the most flexible and future-proof API, since it allows StakeType to be extended to more types (e.g. EthRestaked, BtcRestaked, ..., but Option 2 is most straightforward if the assumption that we only support ETH and native token is valid for the foreseeable future.

WDYT? @philippecamacho

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,129 @@
...
+    /// withdraw past their `exitEpoch`.
+    ///
+    /// @param blsVK The BLS verification key to withdraw
+    /// @return The total amount withdrawn, equal to `Node.balance` associated with `blsVK`
+    function withdraw(BN254.G1Point calldata blsVK) external returns (uint64);
+
+    /// @notice Advance to the next epoch, the new `votingStakeTableCommitment` will be
+    /// `frozenStakeTableCommitment` and the new `frozenStakeTableCommitment` will be
+    /// `curEpochComm`,
+    function advanceEpoch(bytes32 curEpochComm) external returns (bool);
```
alxiong, line 128: > [philippecamacho](https://github.com/philippecamacho)
Who calls this function and why?

alxiong: Think of this API as the "setter for stake commitments" (as we have getter as votingStakeTableCommitment() and frozenStakeTableCommitment())

So light client prover will submit new states and call this function to update the new finalized stake table commitments.
@philippecamacho

contracts/src/interfaces/IStakeTable.sol
alxiong: I'm following our [doc here](https://docs.espressosys.com/sequencer/~/revisions/oijyFXG7VlZ4pxqo4Lgq/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table#data-structures) by @jbearer
 
64 bits should be enough?

cc @chancharles92

alxiong (commented):
Moving previous review over

contracts/src/interfaces/IStakeTable.sol
philippecamacho: I think option 2 is the most reasonable as well.

contracts/src/interfaces/IStakeTable.sol
philippecamacho: I see. Are there some assumptions regarding when this function must be called? For example if nobody calls it during more than one epoch could we get into liveness issues?

alxiong: - We assume altruistic prover for now as we discussed during the retreat, so normally should be called periodically
- the logic of liveness stalling detection and response would be outside the scope of `IStakeTable.sol`, but probably in `LightClient.sol` or `HotShot.sol` directly.

contracts/src/interfaces/IStakeTable.sol
alxiong: done in [8e857eb](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/8e857eb9df7deb9d60b23eff43e51b06d325ebb4)

contracts/src/interfaces/IStakeTable.sol
jbearer: 64 bits is definitely enough *precision* for the contract. We may need to have some multiplier to convert from stake amounts to BEANs, since it's customary for ERC20s to have like 9 decimal places of precision, but we can definitely limit stake amounts to being a whole number or even a multiple of BEANs.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+/// @dev Stake delegation happens in a separate DelegationPool contract, and specific to
+/// instantiation thus not part of this interface
+interface IStakeTable {
+    /// @notice Supported stake type, either using native token or re-staked using ETH
+    enum StakeType {
+        Native,
+        Restake
+    }
+
+    /// @dev (sadly, Solidity doesn't support type alias on non-primitive types)
```
jbearer, line 18: New feature! https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint64 registerEpoch;
+        uint64 exitEpoch;
+        EdOnBN254.EdOnBN254Point schnorrVK;
+    }
+
+    // === Table State & Stats ===
+
+    /// @notice Get the commitment of the stake table used in current voting (i.e. snapshot at
+    /// the start of last epoch)
+    function votingStakeTableCommitment() external view returns (bytes32);
```
jbearer, line 49: I thought the plan was to have replicas read the entire stake table, compute the commitment offchain, and vote on it. Then the light client would update its stake table commitment based on a vote of the last stake table comitment.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+    /// @notice Get the commitment of the stake table frozen for change (i.g. snapshot at the start
+    /// of the current epoch)
+    function frozenStakeTableCommitment() external view returns (bytes32);
```
jbearer, line 52: The stake table contract would not maintain multiple snapshots. It would just have the bleeding edge state at any given time. It would then be the responsibility of the HotShot validators and light client contract to keep track of the different snapshots.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /// @notice Get the total stakes of the registered keys in the latest stake table (Head).
+    /// @return The total stake for native token and restaked token respectively.
+    function totalStake() external view returns (uint256, uint256);
+    /// @notice Get the total number of stakers, uniquely identified by their `blsVK`
+    function totalKeys() external view returns (uint32);
+    /// @notice Get the total stake of the registered keys in the voting stake table
+    /// (LastEpochStart)
+    function totalVotingStake() external view returns (uint256);
```
jbearer, line 61: Similarly, do we need all of these onchain?

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+    function lookup(BN254.G1Point calldata blsVK) external view returns (uint64);
+    /// @notice Look up the full `Node` state associated with `blsVK`
+    function fullLookup(BN254.G1Point calldata blsVK) external view returns (Node memory);
+
+    // === Queuing Stats ===
+
+    /// @notice Get the next available epoch for new registration
+    function nextRegisterationEpoch() external view returns (uint64);
+    /// @notice Get the number of pending registration requests in the waiting queue
+    function numPendingRegistrations() external view returns (uint64);
```
jbearer, line 73: Why do we need to expose this?

contracts/src/interfaces/IStakeTable.sol
jbearer: I don't think we need this function here. The stake table contract should have a reference to the light client contract, and the light client contract should have a `currentEpoch()` function which updates whenever the prover sends it a proof.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /// @notice Look up the balance of `blsVK`
+    function lookup(BN254.G1Point calldata blsVK) external view returns (uint64);
+    /// @notice Look up the full `Node` state associated with `blsVK`
+    function fullLookup(BN254.G1Point calldata blsVK) external view returns (Node memory);
+
+    // === Queuing Stats ===
+
+    /// @notice Get the next available epoch for new registration
+    function nextRegisterationEpoch() external view returns (uint64);
```
philippecamacho, line 71:
```suggestion
    function nextRegistrationEpoch() external view returns (uint64);
```

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+    /// @param stakeType The type of staking (native or restaking)
+    /// @param blsSig The BLS signature that authenticates the `blsVK` field
+    /// @param maxWait The maximum epoch the sender is waiting to wait to be included (cannot be
+    /// smaller than the current epoch)
+    ///
+    /// @return success status
+    ///
+    /// @dev No validity check on `schnorrVK`, as it's assumed to be sender's responsibility,
+    /// the contract only treat it as auxiliary info submitted by `blsVK`.
+    function register(
```
philippecamacho, line 95: There are some slight differences when dealing with stake delegation (see [PR #14](https://github.com/EspressoSystems/gitbook/pull/14)) . In all cases I think it is better to agree on this version of API first, and then make the changes to support delegation.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,130 @@
...
+        BN254.G1Point calldata blsVK,
+        EdOnBN254.EdOnBN254Point calldata schnorrVK,
+        uint64 amount,
+        StakeType stakeType,
+        bytes calldata blsSig,
```
philippecamacho, line 100: Verifying a BLS signature inside the contract is expensive. As we do not have to perform BLS signatures aggregation anymore, why do we need this signature? If someone registers a BLS public key without knowing the corresponding private key, it will act as a byzantine node unable to sign a QC/proposal, is that a problem?

contracts/src/interfaces/IStakeTable.sol
alxiong: > it's customary for ERC20s to have like 9 decimal places of precision

I'm not sure where "9" comes from, when reading openzepplin's [`IERC20` doc here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20-decimals--), it says "18" is the default.
btw, I'm not suggesting we should go for 18, I'm fine with whatever decimals we find most sensible.

 > we can definitely limit stake amounts to being a whole number or even a multiple of BEANs.

with 9 decimal places, just absorbing the whole precision, the largest amount is still `2**55~= 3.6*10^16`, should be more than enough.

In short, we should be fine with `uint64` here.

contracts/src/interfaces/IStakeTable.sol
alxiong: in my mind, it's the other way around, the light client contract would have a ref to the stake table contract, and call this function, as shown [here](https://github.com/EspressoSystems/espresso-sequencer/blob/light-client-pub-input/contracts/src/LightClient.sol#L123) (and the connection is established in its constructor [here](https://github.com/EspressoSystems/espresso-sequencer/blob/light-client-pub-input/contracts/src/LightClient.sol#L72))

contracts/src/interfaces/IStakeTable.sol
alxiong: Yes I am aware of this, it doesn't support **non-primitive/built-in underlying types tho**:

> A user-defined value type is defined using type C is V, where C is the name of the newly introduced type and V has to be a built-in value type (the “underlying type”).

contracts/src/interfaces/IStakeTable.sol
alxiong: yes, that's right. 

this is a getter for _other_ contracts in the ecosystem who might want to read on-chain.

contracts/src/interfaces/IStakeTable.sol
alxiong: correct, the contract will only keep the whole table for the "head", but it keeps two variable in storage -- the snapshoted table's commitments only, these values are used (& updated) during `advanceEpoch()`

contracts/src/interfaces/IStakeTable.sol
alxiong: when I add them, I try to think it from the perspective of "what normal users would expect a stake table API should offer" instead of "what's the minimum set of API our sequencer project need internally".

IMO, here's a conjured use case for the downstream app that might exist in our ecosystem:
a DAO that decide whether to stake more when its percentage drops below 10%. Then that contract would invoke these APIs to make its decision, right?

I think it'll be tough to require anyone who wants these basic stats to run the `espresso-sequencer` software and parse out the contract storage or history to read these values. WDYT?

contracts/src/interfaces/IStakeTable.sol
alxiong: in the same light of https://github.com/EspressoSystems/espresso-sequencer/pull/748#discussion_r1395134797,

Imagine a user (or contract) would want to make decision of whether to register based on the pending queue. Note that this is just a read operation, they don't want to try `register(maxWait = curBlock)` to trial-and-error to know the pending queue, as that costs gas.

contracts/src/interfaces/IStakeTable.sol
alxiong: sure! is there a TL;DR for the potential conflicts or major incompatibility you can foresee?
I can also try to address some concerns in this PR if desired.

contracts/src/interfaces/IStakeTable.sol
alxiong: > Verifying a BLS signature inside the contract is expensive.

this is not necessarily true. We can force every signature to be the same message, and precompute and hardcode the `HashToGroup(msg)` variable inside contract, then verifying the signature only takes 2 group validations and 2 pairings (via precompile)

> If someone registers a BLS public key without knowing the corresponding private key, it will act as a byzantine node unable to sign a QC/proposal, is that a problem?

Good point, I agree that we don't need to verify it anymore.
double checking with @jbearer

contracts/src/interfaces/IStakeTable.sol
jbearer: I suppose that could work. But then this function needs to be restricted so it can only be called from the light client.

contracts/src/interfaces/IStakeTable.sol
jbearer: Oh that's so annoying lol

contracts/src/interfaces/IStakeTable.sol
jbearer: But couldn't those contracts just read from the light client instead? It seems we're forcing ourselves to implement and maintain an expensive operation in this contract that we don't yet have any use case for

contracts/src/interfaces/IStakeTable.sol
jbearer: Ah ok, and then the light client contract reads the commitment from here 👍

contracts/src/interfaces/IStakeTable.sol
jbearer: Sure, seems reasonable

contracts/src/interfaces/IStakeTable.sol
jbearer: Wait, we still do BLS aggregation. Not for the light client proofs but in consensus itself, to shrink the QCs

philippecamacho: Oh that is right, we should keep the check then.

contracts/src/interfaces/IStakeTable.sol
philippecamacho: TL;DR
* We need to register pool members in addition to pool operators. Users do not need a BLS/Schnorr key because by they don't run a validator.
* We need to handle pool operator fees (when registering a pool and also allow the operator to modify the fee)
* Reward collection is performed by pool members for better UX.

But again I think it would be better to have the whole system working with delegation first. Adding delegation should not be too intrusive.

contracts/src/interfaces/IStakeTable.sol
alxiong: what? even if we do BLS aggregation, it has nothing to do with LightClient.sol as of now. 
We don't need to check them in contract, right? @jbearer

contracts/src/interfaces/IStakeTable.sol
alxiong: > But then this function needs to be restricted so it can only be called from the light client.

🤔 ah! this is a good point! 
I'm not sure how I can restrict it. because right now, I was expecting to deploy StakeTable.sol first to get its address and then pass it to LightClient.sol. But if we want `OnlyOwner`-ish control on this function, then we sort of need to deploy LightClient.sol first. 

one possible solution I can think of:
- Make `contract LightClient is StakeTable` -- let LightClient inherits StakeTable, and mark `advanceEpoch` to be `internal`


any idea? @jbearer @philippecamacho

contracts/src/interfaces/IStakeTable.sol
philippecamacho: We don't need to check the BLS signature in the contract but we need to be sure the registered participant knows the BLS private key to avoid key rogue attacks, unless this check is done somewhere else but it seems natural to do it here.

contracts/src/interfaces/IStakeTable.sol
jbearer: In the original design we had the dependency the other way around. So the light client contract has the authoritative state of HotShot including stake table commitments, and stake table only needs to read `currentEpoch()` from the light client. Nothing needs to be permissioned and the light client doesn't even know about the stake table contract.

jbearer: I think `LightClient is StakeTable` would also work but it may be more coupling than we want

alxiong: after discussion: we decide to keep `stakeTableCommitments` in the `LightClient.sol` (which is reasonable but a little counterintuitive), and move `advanceEpoch()` to `LightClient.sol`

contracts/src/interfaces/IStakeTable.sol
alxiong: Agree, after https://github.com/EspressoSystems/espresso-sequencer/pull/748#discussion_r1395976543, this function will be from LightClient instead.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,128 @@
...
+
+    // === Write APIs ===
+
+    /// @notice Register a validator in the stake table, transfer of tokens incurred!
+    ///
+    /// @param blsVK The BLS verification key
+    /// @param schnorrVK The Schnorr verification key (as the auxiliary info)
+    /// @param amount The amount to register
+    /// @param stakeType The type of staking (native or restaking)
+    /// @param maxWait The maximum epoch the sender is waiting to wait to be included (cannot be
```
sveitser, line 87: Typo?
```suggestion
    /// @param maxWait The maximum epoch the sender is waiting to be included (cannot be
```

sveitser: How about putting "epoch" into the name of the parameter?

contracts/src/interfaces/IStakeTable.sol
alxiong: Good point, in order to prevent Rogue key attack, we need to authenticate PoK on the secret key.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,128 @@
...
+    /// @param blsVK The BLS verification key to exit
+    /// @return success status
+    function exit(BN254.G1Point calldata blsVK) external returns (bool);
+
+    /// @notice Withdraw from the staking pool. Transfers occur! Only successfully exited keys can
+    /// withdraw past their `exitEpoch`.
+    ///
+    /// @param blsVK The BLS verification key to withdraw
+    /// @return The total amount withdrawn, equal to `Node.balance` associated with `blsVK`
+    function withdraw(BN254.G1Point calldata blsVK) external returns (uint64);
```
sveitser, line 122:
```suggestion
    function withdrawFunds(BN254.G1Point calldata blsVK) external returns (uint64);
```
Since `withdraw` can have multiple meanings.

contracts/src/interfaces/IStakeTable.sol
sveitser: > 🤔 ah! this is a good point! I'm not sure how I can restrict it. because right now, I was expecting to deploy StakeTable.sol first to get its address and then pass it to LightClient.sol. But if we want `OnlyOwner`-ish control on this function, then we sort of need to deploy LightClient.sol first.

We can deploy the contract in any order and set an owner after deployed (like with https://docs.openzeppelin.com/contracts/2.x/api/ownership#Ownable-transferOwnership-address- ) if needed.

flake.nix
```diff
@@ -18,20 +18,22 @@
   inputs.fenix.url = "github:nix-community/fenix";
```
sveitser, line 18: @alxiong I guess you still have another formatter running?

alxiong: do you mean another nix formatter? I don't think so, totally switch to `nil` since our last chat about it.
are you suggesting we remove this line?

sveitser: My comment just refers to the fact that the file has been reformatted, so I assumed it was formatted by a different formatter than `nixpkgs-fmt`.

contracts/src/interfaces/IStakeTable.sol
alxiong: `advanceEpoch()` removed in [65a87d0](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/65a87d0b981f90a204a7ba399e1741653d9f1295), will add to LightClient in #754

contracts/src/interfaces/IStakeTable.sol
alxiong: added back in [65a87d0](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/65a87d0b981f90a204a7ba399e1741653d9f1295)

contracts/src/interfaces/IStakeTable.sol
alxiong: done in [8ff8e8d](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/8ff8e8d77fb6eebd42fd313f36fab14add4b423d)

contracts/src/interfaces/IStakeTable.sol
alxiong: done [65a87d0](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/65a87d0b981f90a204a7ba399e1741653d9f1295)

flake.nix
```diff
@@ -142,7 +142,6 @@
                 types_or = [ "markdown" ];
                 pass_filenames = true;
               };
-              nixpkgs-fmt.enable = true;
```
sveitser, line 145: @alxiong this is an older version of the file. I guess your `main` was out of date?

alxiong: sorry about that, fixed

sveitser:
No idea what's wrong with the CI. I'm having a look.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -0,0 +1,121 @@
...
+    /// @notice Get the total number of stakers, uniquely identified by their `blsVK`
+    function totalKeys() external view returns (uint32);
+    /// @notice Get the total stake of the registered keys in the voting stake table
+    /// (LastEpochStart)
+    function totalVotingStake() external view returns (uint256);
+
+    /// @notice Look up the balance of `blsVK`
+    function lookup(BN254.G1Point calldata blsVK) external view returns (uint64);
+    /// @notice Look up the full `Node` state associated with `blsVK`
+    function fullLookup(BN254.G1Point calldata blsVK) external view returns (Node memory);
```
jbearer, line 61: Nit: I would prefer more descriptive names like `lookupStake` and `lookupNode`

jbearer (approved):
LGTM

Just one take-it-or-leave-it suggestion

contracts/src/interfaces/IStakeTable.sol
alxiong: I like it, done in [612a578](https://github.com/EspressoSystems/espresso-sequencer/pull/748/commits/612a5787aa1ab5d81c73765aa22de498c06a3f23)

sveitser:
@alxiong Can reproduce CI failure locally when using latest foundry so error is likely something related to anvil. I think we can ignore these tests until that's fixed.

PR #749, opened 2023-11-14
Make L1 client more resilient.

jbearer:
Previously, if we encountered an RPC error in `get_finalized_block`, we would retry indefinitely without resetting the WebSockets connection. This makes it likely that the failure will persist indefinitely.

This change improves resiliency in two ways:
* If anything goes wrong in `get_finalized_block`, we return an error and the update loop resets the WebSockets connection
* Before resetting the connection, the update loop will update the L1 head anyways, so that failure to fetch the finalized block does not impede our ability to update the latest block

### DIFF
--- a/sequencer/src/l1_client.rs
+++ b/sequencer/src/l1_client.rs
@@ -105,7 +105,7 @@ impl L1ClientOptions {
         // at least until the background task starts updating the snapshot.
         let snapshot = L1Snapshot {
             head: rpc.get_block_number().await?.as_u64(),
-            finalized: get_finalized_block(&rpc, self.finalized_block_tag, self.retry_delay).await,
+            finalized: get_finalized_block(&rpc, self.finalized_block_tag).await?,
         };
         tracing::info!("L1 client initialized with snapshot {snapshot:?}");
 
@@ -143,7 +143,7 @@ async fn update_loop(
     retry_delay: Duration,
     snapshot: Arc<RwLock<L1Snapshot>>,
 ) {
-    loop {
+    'reset_connection: loop {
         // Subscribe to new blocks. This task cannot fail; retry until we succeed.
         let rpc = match Provider::connect(url.clone()).await {
             Ok(rpc) => rpc,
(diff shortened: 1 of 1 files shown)


sequencer/src/l1_client.rs
```diff
@@ -195,15 +195,37 @@ async fn update_loop(
...
 
             // Drop our exclusive lock before logging anything.
             let snapshot = RwLockWriteGuard::downgrade(snapshot);
             tracing::info!("updated L1 snapshot to {:?}", *snapshot);
+
+            // If we encountered an error in `get_finalized_block`, reset the connection now that we
+            // have updated what we can.
+            if finalized.is_none() {
+                tracing::warn!("resetting connection due to error in get_finalized_block");
+                continue 'reset_connection;
```
nomaxg, line 227: We could just break here - or is your idea to use the loop tag for a more semantic break?

sequencer/src/l1_client.rs
```diff
@@ -213,40 +235,28 @@ async fn update_loop(
...
+        // with a null L1 block rather than wait for the L1 to finalize a block, which can take a
+        // long time.
+        tracing::warn!("no finalized block yet");
+        return Ok(None);
+    };
+
+    // The block number always exists unless the block is pending. The finalized block cannot be
+    // pending, unless there has been a catastrophic reorg of the finalized prefix of the L1
+    // chain, so it is OK to panic if this happens.
+    let number = block.number.expect("finalized block has no number");
```
nomaxg, line 251: Since this function returns a Result now, can we remove these panics for resiliency's sake?

nomaxg (approved):
a question and suggestion, overall LGTM

sequencer/src/l1_client.rs
jbearer: Sure, good idea

sequencer/src/l1_client.rs
jbearer: Idea was to avoid the redundant log after the inner loop but now that you mention it that seems like a bad reason to complicate the control flow. I'll change it to `break`

PR #754, opened 2023-11-15
feat: Add new LightClient contract skeleton

### LINKED ISSUE #705: [contract] Define public inputs to light client SNARK verification
This issue aims to define struct related to public inputs of the SNARKed hotshot state update.

Related to https://github.com/EspressoSystems/espresso-sequencer/issues/673 (at least the `StakeTableCommitment` as a global variable should come from the stake table interface.

- [x] Define `struct LightClientState` (corresponding to [same struct in `hotshot-state-prover`](https://github.com/EspressoSystems/HotShot/blob/78a41a306d84c27794952047d67aca06a8761d19/crates/types/src/traits/state.rs#L225-L236))

alxiong:
Closes #705 

This PR includes:
- Defining new `LightClient.sol` contract with main `newFinalizedState()` function and key struct like `struct LightClientState` that corresponds to the public inputs of the SNARK, as well as some key global states.


contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        // check plonk proof
+        // TODO: (alex) replace the vk with the correct one
+        IPlonkVerifier.VerifyingKey memory vk = VkTest.getVk();
+        uint256[] memory publicInput = preparePublicInput(newState);
+        PlonkVerifier.verify(vk, publicInput, proof, bytes(""));
+
+        // upon successful verification, update state
+        if (newState.blockHeight == epochEndingBlockHeight) {
+            bytes32 newStakeTableComm = keccak256(
```
philippecamacho, line 115: It might be redundant, but conceptually at least it seems reasonable to also hash the epoch number.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+                abi.encodePacked(
+                    newState.stakeTableBlsKeyComm,
+                    newState.stakeTableSchnorrKeyComm,
+                    newState.stakeTableAmountComm
+                )
+            );
+
+            stakeTable.advanceEpoch(newStakeTableComm);
+        }
+        if (finalizedState.blockHeight == epochEndingBlockHeight) {
```
philippecamacho, line 125:
```suggestion
        if (newState.blockHeight == epochEndingBlockHeight) {
```

Because `finalizedState` is updated afterwards only? Or you can merge with the if statement at line 114.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+    //
+    /// @notice Update the latest finalized light client state. It is expected to be updated
+    /// periodically, especially an update for the last block for every epoch has to be submitted
+    /// before any newer state can be accepted since the stake table commitments of that block
+    /// become the snapshots used for vote verifications later on.
+    function newFinalizedState(
+        LightClientState calldata newState,
+        IPlonkVerifier.PlonkProof calldata proof
+    ) external {
+        if (newState.viewNum <= finalizedState.viewNum) {
```
philippecamacho, line 86: Why do we need to perform check on views? Why not only compare block numbers like in line 97?

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+        // TODO: (alex) initialized stake table or at least store its contract address ref here
+    }
+
+    // === State Modifying APIs ===
+    //
+    /// @notice Update the latest finalized light client state. It is expected to be updated
+    /// periodically, especially an update for the last block for every epoch has to be submitted
+    /// before any newer state can be accepted since the stake table commitments of that block
+    /// become the snapshots used for vote verifications later on.
+    function newFinalizedState(
```
philippecamacho, line 82: Should not we also check that the new state threshold value is not too low ( I guess based on the total stake of the previous epoch)?

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    constructor(LightClientState memory genesis, address stakeTableAddr) {
+        if (genesis.viewNum != 0 || genesis.blockHeight != 0) {
+            revert InvalidArgs();
+        }
+
+        genesisState = genesis;
+        currentEpoch = 0;
+        stakeTable = IStakeTable(stakeTableAddr);
+        // TODO: (alex) initialized stake table or at least store its contract address ref here
```
philippecamacho, line 73: Add `finalizedState = genesisState` ?

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,152 @@
...
+        // format validity check
+        BN254.validateScalarField(newState.blockComm);
+        BN254.validateScalarField(newState.feeLedgerComm);
+        BN254.validateScalarField(newState.stakeTableBlsKeyComm);
+        BN254.validateScalarField(newState.stakeTableSchnorrKeyComm);
+        BN254.validateScalarField(newState.stakeTableAmountComm);
+
+        // check plonk proof
+        // TODO: (alex) replace the vk with the correct one
+        IPlonkVerifier.VerifyingKey memory vk = VkTest.getVk();
```
philippecamacho, line 109: When the block is the last one of the epoch, we need to check an extended QC instead of normal QC (see Section 2.4.2 of the HotShot paper). I wonder if we need a special circuit for that case. cc @chancharles92

chancharles92: Are we ever going to check QC given the new Schnorr signature flow?

contracts/src/LightClient.sol
alxiong: 🤔 actually I tend to disagree. 
Think from offchain version of the stake table, which doesn't care about "epoch" concept at all. That stake table will compute an overall commitment that is consistent with this value here. 

I do agree, tho, that there should be a bijective mapping between an epoch and a value (effectively binding them) -- this mapping exists in the contract history.

contracts/src/LightClient.sol
alxiong: > Are we ever going to check QC given the new Schnorr signature flow?

I believe not. The extended QC verification is part of HotShot node software, and honest node won't even sign on the schnorr signature for that special block if unsatisfied.

contracts/src/LightClient.sol
alxiong: no, (but you are right that my logic is also wrong):
Case 1: `newState` is in the middle of an epoch, --> just update `finalizedState`
Case 2: `newState` is the last block of an epoch --> `epoch` should _NOT_ increment, still just update `finalizedState`
Case 3: `newState` is in the next epoch --> `epoch` should increment, and `stakeTable` should advance with the last stored `finalizedState`'s commitment (cuz that was the state of the last block of the last epoch), only then update `finalizedState`

contracts/src/LightClient.sol
alxiong: Agree that under our honest majority assumption, these values are strictly increasing and checking one of them is sufficient.

I'm just checking both for peace of mind...

contracts/src/LightClient.sol
alxiong: yeah, again if we trust honest majority, then this threshold (as part of the SNARK public input) should be correct, or else they won't have enough signatures to produce the correct proof.

But I guess it doesn't hurt to check it here as well. (I personally like adding these checks for protection in defense, but maybe they are not useful, since if honest majority breaks, this is on the lower priority of concerns among other possible catastrophic outcomes)

contracts/src/LightClient.sol
philippecamacho: Mmm, regarding the bijective mapping there can be in theory  two equals commitments for different epochs if the stake table does not change.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+        if (finalizedState.blockHeight == epochEndingBlockHeight) {
+            bytes32 newStakeTableComm = keccak256(
+                abi.encodePacked(
+                    finalizedState.stakeTableBlsKeyComm,
+                    finalizedState.stakeTableSchnorrKeyComm,
+                    finalizedState.stakeTableAmountComm
+                )
+            );
+
+            stakeTable.advanceEpoch(newStakeTableComm);
```
philippecamacho, line 130: Why do we need this? Aren't epochs defined from HotShot honest nodes reaching consensus on the ethereum block that is read when the last block of some Hotshot epoch is generated? Otherwise depending on when `newFinalizedState` is called,  it seems that we may have different definitions of the stake table.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    // === State Modifying APIs ===
+    //
+    /// @notice Update the latest finalized light client state. It is expected to be updated
+    /// periodically, especially an update for the last block for every epoch has to be submitted
+    /// before any newer state can be accepted since the stake table commitments of that block
+    /// become the snapshots used for vote verifications later on.
+    function newFinalizedState(
+        LightClientState calldata newState,
```
sveitser, line 84:
```suggestion
        LightClientState memory newState,
```

IIRC it's cheaper in terms of gas if the data is loaded into memory in the function.

contracts/src/LightClient.sol
philippecamacho: Ok, no problem, let us keep it.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+        IPlonkVerifier.PlonkProof calldata proof
```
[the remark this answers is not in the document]
sveitser: I once checked in remix for a small test example. Not sure it's true here. Would be good to get to the bottom of it eventually.

contracts/src/LightClient.sol
philippecamacho: Ok I understand, this is just to store the new stake table commitment.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+// TODO: replace these, for now WIP and test only
+import { VkTest } from "../test/stubs/Transfer1In2Out24DepthVk.sol";
+
+/// @notice A light client for HotShot consensus. Keeping track of its finalized states in safe,
+/// authenticated ways.
+contract LightClient {
+    // === Constants ===
+    //
+    /// @notice System parameter: number of blocks per epoch
+    uint32 public constant BLOCKS_PER_EPOCH = 20_000;
```
jbearer, line 19: Maybe we should change this to `immutable` and set it in the constructor. We might want to play around with different epoch lengths and it would be nice to do this without changing the code of the contract.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+    uint64 public currentEpoch;
+    /// @notice address of the stake table contract
+    IStakeTable internal stakeTable;
+
+    // === Data Structure ===
+    //
+    /// @notice The finalized HotShot state (as the digest of the entire HotShot state)
+    /// @param viewNum The latest view number of the finalized HotShot chain
+    /// @param blockHeight The block height of the latest finalized block
+    /// @param blockComm The block commitment (type: BN254::ScalarField)
```
jbearer, line 37: What is this? Latest block commitment or merkle commitment of all blocks? And why is it a field element as opposed to just a hash output?

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+    /// @param feeLedgerComm The commitment to the fee ledger state (type: BN254::ScalarField)
```
jbearer, line 38: Same here, why field element vs `bytes32`?

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+            revert MissingLastBlockForCurrentEpoch(epochEndingBlockHeight);
+        }
+        // format validity check
+        BN254.validateScalarField(newState.blockComm);
+        BN254.validateScalarField(newState.feeLedgerComm);
+        BN254.validateScalarField(newState.stakeTableBlsKeyComm);
+        BN254.validateScalarField(newState.stakeTableSchnorrKeyComm);
+        BN254.validateScalarField(newState.stakeTableAmountComm);
+        // sanity check on the threshold
+        if (newState.threshold <= stakeTable.totalVotingStake() * 2 / 3) {
```
jbearer, line 107: I think we should remove this check:
* The stake table contract is no longer doing snapshots, so this would be the voting stake now, possibly mid-epoch
* If we are going to enforce that `threshold` is 2/3 + 1, why store threshold at all? Why not just compute it?
Instead, I think we can just use the previous threshold, which has been signed by a quorum of replicas from the previous epoch. And then update the threshold based on `newState` when the epoch changes.

contracts/src/LightClient.sol
jbearer: Why does it need to be bijective? As a client I definitely care about getting the stake table commitment for a given epoch, I don't know why I would ever need to get the epoch with a given stake table commitment

alxiong: you are right, it's not entirely bijection. 
I don't see why it's bad to have the same stake table commitments for two epochs if there's no change in stake table.

contracts/src/LightClient.sol
jbearer, line 130: We're getting rid of this, right? Instead we should store the stake table commitments in this contract.

contracts/src/LightClient.sol
jbearer: Ah, I see, because it's an input to the SNARK. However if in practice these commitments are generated from a hash function, they may not fit in a field element, we may need to hash to field.

contracts/src/LightClient.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    // === Pure or View-only APIs ===
+    /// @dev Transform a state into an array of field elements, prepared as public inputs of the
+    /// plonk proof verification
+    function preparePublicInput(LightClientState calldata state)
+        public
+        pure
+        returns (uint256[] memory)
+    {
+        uint256[] memory publicInput = new uint256[](8);
```
jbearer, line 146: Did we consider hashing these all together and unhashing in the circuit, so we just have a single public input? I don't know which would be cheaper, but I know Polygon zkEVM hashes everything before sending it into the SNARK

contracts/src/LightClient.sol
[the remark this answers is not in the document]
alxiong: Nope, try this:
```solidity
pragma solidity ^0.8.0;

contract Foo {
    struct Bar {
        uint a;
        uint b;
    }
    Bar bar;
    function memoryOption(Bar memory _bar) public {
        bar = _bar;
    }

    function calldataOption(Bar calldata _bar) public {
        bar = _bar;
    }
}
```

Then pass in `[1,2]` to both function on Remix:
`memoryOption`: 66486
`calldataOption`: 27304

contracts/src/LightClient.sol
philippecamacho: Yes, we can leave it as is for now, just that conceptually a commitment is a unique (in practice) representation of something, that is why I am inclined to include the epoch into the hash.

alxiong: >  just that conceptually a commitment is a unique (in practice) representation of something, that is why I am inclined to include the epoch into the hash.

true, but practically, in our context, if you receive a `stakeTableComm` value and checked that it was committing to `epoch: 5`, this info alone doesn't help or bind it. You still need to go to blockchain history and verify that at Epoch 5, whether the stake table commitment matches. Since you need to do that historical query anyway, you should find about the epoch number even if that field is not part of the preimage of `stakeTableComm`.

^^ that's why I didn't find "including epoch into the hash" that useful. (even tho I agree, conceptually it's nicer)

contracts/src/LightClient.sol
alxiong: good point, done in [6ef422c](https://github.com/EspressoSystems/espresso-sequencer/pull/754/commits/6ef422c33121d8566894d772413543c26142ef9a)

contracts/src/LightClient.sol
alxiong: This is an error [due a misunderstanding from the bad variable name](https://github.com/EspressoSystems/HotShot/pull/2055#pullrequestreview-1737485954). It should be `block_comm_root` the root of the merkle tree that accumulates all the historical block commitments. (which should be using Rescue-based commitment, thus guaranteed to be a field element) 

I will update its name

contracts/src/LightClient.sol
alxiong: Here is the `Field::from_le_bytes(actual_comm: bytes32)` result that is also used to generate the proof.

contracts/src/LightClient.sol
alxiong: good suggestion, done in [29392af](https://github.com/EspressoSystems/espresso-sequencer/pull/754/commits/29392afd5c14a3940ee1ea863c63ba3660e99380)

contracts/src/LightClient.sol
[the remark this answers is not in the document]
alxiong: after more tests by @sveitser, we realized that whenever we are doing copying from `calldata` to `memory`, then declaring these input arguments with `memory` to start with will be cheaper.

Thus updated in [3c2b4c4](https://github.com/EspressoSystems/espresso-sequencer/pull/754/commits/3c2b4c4250c3e03a052e8eb3f17f80f141c10f64)

philippecamacho (approved):
LGTM

contracts/src/LightClient.sol
jbearer: Ok. Sounds good to me, let's keep it as is

jbearer (approved):
LGTM

PR #752, opened 2023-11-15
Set CARGO_TARGET_DIR in nix dev shell

sveitser:
@ggutoski can you confirm if this helps before we merge?

Because I think it will prevent the `cargo-sweep` tool from finding the target directory when it runs recursively from outside this repo :-/

Close #751

PR #756, opened 2023-11-15
feat: build block payload bytes from tx iterator

### LINKED ISSUE #744: Build payload from tx iterator, single-namespace only
# New `VidBlockPayload` struct

Create new `VidBlockPayload` struct something like
```rust
struct VIDBlockPayload {
  payload: Vec<u8>,
  transactions: Vec<Range<usize>>,
  commitment: VIDCommitment,
}
```
Implement a constructor something like
```rust
fn build(transactions: impl IntoIterator<Item = Transaction>) -> Self;
```
where `Transaction` is something like `(namespace_id, tx_bytes)`.

# Tx table in payload

* 4 bytes: length of metadata section
* 4 bytes for each transaction: transaction offset relative to data section
* data section: all transaction bodies concatenated

# Support one namespace only

Implementation of the namespace table is out of scope for this issue.

ggutoski:
close #744 

- For now: new module `block2.rs`. Eventually it will replace the existing `block.rs`.
- Lots of wordy code comments on topics to discuss. We can reduce wordiness after decisions are made if desired.
- I'm not convinced that `BlockPayload` actually needs a `tx_ranges` field like we discussed. Even without this field it's still constant-time to find the start/end of a given tx. In any case it's not relevant for this PR---`tx_ranges` would only be used in future PRs anyway.
- all edits outside `block2.rs` are to fix build warnings
- `.gitignore` tweak is temporary. Will clean up in subsequent PRs after #752


sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
+// Question: jellyfish uses `ark_std` everywhere instead of `std`.
+// Is there a chance of badness if we (as a downstream user) use `std`?
```
jbearer, line 2: I don't think so, with the `std` feature enabled `ark_std` should just be a wrapper around `std`

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::mem::size_of;
+
+#[allow(dead_code)] // TODO temporary
+pub struct BlockPayload {
+    payload: Vec<u8>,
+    // tx_ranges: Vec<Range<u32>>, // not convinced we need this
```
jbearer, line 8: Yeah, I agree. We originally had the idea for this when we were planning on putting each transactions size with the transaction, so finding the transaction boundaries required a linear scan. Now with the transactions table this is no longer necessary.

jbearer: We should store the VID commitment in here, though, so that we can retrieve it in constant time.

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct Tx {
+    // namespace: u32, // TODO how many bytes for namespace id?
```
jbearer, line 13: We should be able to reuse the `Transaction` struct from `transaction.rs` here. It's basically exactly this. But we have a `VmId` newtype for the namespace.

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+        // so that the payload bytes of the ith tx is
+        // tx_bodies[tx_table[i-1]..tx_table[i]].
+        // edge case: tx_table[-1] is defined as 0.
+        //
+        // TODO clarification for future docs:
+        // perhaps the final entry of tx_table is redundant:
+        // it should always equal the length of the flattened tx bytes.
+        // But this might not always be the case.
+        // Example: If the flattened tx bytes are not a monolith at the end
+        // of the payload then the final tx_table entry is needed.
```
jbearer, line 30: We should know where the end of the payload or namespace is from the namespace table. I think it's fine to define the end of the last transaction as the end of that region. For an honest builder that should always be what you want anyways.

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut tx_table = Vec::new();
+
+        // concatenation of all tx payloads
+        let mut tx_bodies = Vec::new();
+
+        // build tx_table, tx_bodies
+        let mut end: u32 = 0;
+        for tx in txs.into_iter() {
+            // TODO idiomatic usize -> u32 conversion?
+            // Is panic acceptable? If not then we need to return `Result`.
```
jbearer, line 40: Yeah I think after @shenkeyao finishes the `BlockPayload` trait this function will return a `Result`

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+                .payload
+                .len()
+                .try_into()
+                .expect("tx byte length should fit into u32");
+
+            end = end
+                .checked_add(len)
+                .expect("total byte length of all tx bodies should fit into u32");
+
+            // TODO to_be_bytes or to_le_bytes?
```
jbearer, line 51: It doesn't really matter, as long as we're consistent. I usually use `to_le_bytes` since both x86 and ARM Mac are little endian, there's a miniscule and insignificant performance benefit to serializing in little endian

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+            tx_table.extend(end.to_be_bytes());
```
jbearer, line 52: We talked about leaving 3 unused bytes after every 7 table entires, so that we don't have entries spanning two 31-byte words. This reduces the chance of needing two evaluation proofs in a transaction proof. Still think this is worth doing?

sequencer/src/block2.rs
```diff
@@ -0,0 +1,144 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::{BlockPayload, Tx};
+    use std::mem::size_of;
+
+    #[test]
+    fn build_basic_correctness() {
+        // play with this
+        let test_cases = vec![
```
jbearer, line 84: These are good test cases. Just a note for the future, when we get around to functions that read the payload, like transaction getters, we'll need to test with some blocks that weren't built by `BlockPayload::build`, since that function only builds "nice" blocks. We'll need to maliciously build some blocks with overlapping and truncated and out-of-order transactions.

jbearer (approved):
Looks good so far!

sequencer/src/block2.rs
ggutoski: > We should store the VID commitment in here, though, so that we can retrieve it in constant time.

How would that work with `build()`? Seems like `build()` would need to compute the VID commitment, but I've been trying hard to avoid unnecessary computations of the VID commitment. (We could make the commitment an `Option` but I hope I don't need to argue against that.)

Note that the VID commitment is not needed to make namespace/tx proofs; it's needed only to verify such a proof.

Perhaps the VID commitment could instead be given to us as part of the metadata / header.

sequencer/src/block2.rs
ggutoski: I knew that. I was just testing.

We may wish to change it from `u64` to `u32` though. I don't anticipate ever having more than 4.2 billion namespaces.

sequencer/src/block2.rs
ggutoski: Ok. That means for `n` txs we'll have only `n-1` entries in the tx table. That's good because it saves space but now there's a pesky edge case to deal with. I'll post a new issue for it.

sequencer/src/block2.rs
ggutoski: I'm going to say no for now. Implementing this requires breaking abstraction with jellyfish. Let's get something working first and then revisit later.

We already talked offline about another example of abstraction violation: aligning namespaces with polynomial boundaries. You floated the idea of adding a `next_alignment_boundary(index) -> Option<usize>` method to jellyfish trait for this purpose.

That example is closely related to this problem. Whatever solution we use for namespace alignment should generalize to entry-word alignment. We may need to do it someday but today is not that day.

I'll post a new issue.

sequencer/src/block2.rs
jbearer: Currently in HotShot the `BlockPayload` trait has a `commit()` function, and it's understood that calling `build()` computes the commitment, and thereafter you keep the thing you built and it's cheap to call `commit()`. But maybe it makes more sense to have HotShot separately keep track of the commitment, so the `BlockPayload` implementation doesn't have to worry about it. After all, HotShot knows best when it needs to compute a commitment and when it needs to remember that commitment. @shenkeyao

sequencer/src/block2.rs
jbearer: Yeah I've been thinking about that. While there will never be more than 2^32 distinct namespaces, it is nice to be able to use a larger integer for the namespace ID. For example, for Vienna we use 0xc0ffee1 just for fun, which happens to fit in 4 bytes but you could imagine a similar idea that doesn't. Someone in Discord also suggested using the output of a hash function as a namespace ID, for example the hash of the rollup name (and indeed, Ethereum chain IDs are 32 bytes).

However, we can always just have a separate mapping from long namespace IDs to short ones, if that's desirable, so I think going with 4 bytes for now is fine.

sequencer/src/block2.rs
jbearer: Yeah. Unclear if the extra complexity is worth it to save 4 bytes per namespace. We can circle back to this in the future.

sequencer/src/block2.rs
jbearer: Sounds good, +1 for getting something working

sequencer/src/block2.rs
ggutoski: Roger that. At a bare minimum we'll need those tests for the implementation of `decode` / `from_bytes`.

sequencer/src/block2.rs
ggutoski: > But maybe it makes more sense to have HotShot separately keep track of the commitment, so the `BlockPayload` implementation doesn't have to worry about it. After all, HotShot knows best when it needs to compute a commitment and when it needs to remember that commitment.

Strongly agree. We're already taking great pains in hotshot to minimize where we compute the VID commitment. I'd hate to waste that effort by computing it again inside `build()`.

shenkeyao: This makes sense! Let me check the codebase to see if removing `commit()` is doable. Just to clarify, `BlockPayload` actually doesn't have a `commit()` function, but it has the `Committable` bound which has `commit()`, so we need to remove the bound.

shenkeyao: I think we still need a function to compute the commitment. If we are trying to avoid the computation in `build()`, I think we could add a `vid_commitment` function to `Payload` but unlike `commit()` which takes `&self` as input, it takes the encoded transactions. In this way, the payload build and the commitment computation are separated.

jbearer: Why do we need a trait function for this? The thing that's app-specific is `encode()`, but once you have a flat list of bytes you just need to call `disperse` or `commit_only`, and it should actually be HotSHot's responsibility to (a) know which one to call when and (b) set up the VID scheme with the proper parameters

shenkeyao: Agreed. I just created an [issue](https://github.com/EspressoSystems/HotShot/issues/2072) for this. Would like to do it in a separate PR.

sequencer/src/block2.rs
ggutoski: I put a comment about this in the code. I'll make it return `Result` after I have an upstream trait.

sequencer/src/block2.rs
jbearer: Yeah that makes sense

ggutoski:
Welp, I have no idea what caused this [CI failure](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/6884000530/job/18725669022) but I don't think it was me. I have an approval so I'll merge.

PR #762, opened 2023-11-16
Add nixpkgs-fmt pre-commit hook

sveitser:
Pre-commit hook that runs nixpkgs-fmt on the nix files.


alxiong (approved):
Love it! 
(btw since I use your config on in my `config.el` and the same `nil` lsp, I've been having no issues with diff format. adding this to precommit hook is also nice!)

PR #765, opened 2023-11-16
Add non-nix installation instructions

sveitser:
- Add instructions
- Run weekly CI job to execute installation instructions and run tests.

Room for improvement: currently the toolchain file is not read by nix so there is potentital for the toolchain file to go out of sync. It would be easy to do that however, in that case `nix flake update` won't update rust anymore. I read dependabot doesn't handle toolchain files so we'd have to manually update it. We could also use "stable" instead of a specific version tag.

### DIFF
--- a/.github/workflows/ubuntu-install-without-nix.yml
+++ b/.github/workflows/ubuntu-install-without-nix.yml
@@ -4,7 +4,6 @@ on:
   schedule:
     - cron: "0 0 * * MON"
   workflow_dispatch:
-  pull_request: # TODO remove
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


sveitser:
Since I removed the PR CI trigger, the latest run is [here](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/6896439272/job/18762566284).

.github/workflows/ubuntu-install-without-nix.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install and test
+        # Execute all lines that start with four spaces, and remove leading 'sudo '
+        # because we're in a container and already root.
+        run: |
+          grep '^    ' doc/ubuntu.md \
+            | sed 's/^    //' \
+            | sed 's/^sudo //' \
+            | bash -exo pipefail
```
jbearer, line 27: This is pretty nifty 👍

jbearer (approved):
LGTM

nyospe:
Thank you for this.

PR #766, opened 2023-11-17
CI: pin foundry version to working commit

sveitser:
The behaviour of anvil seems to have changed in the last 1 or 2 days in ways that break it for our tests. Until we figure out what exactly that is let's pin the CI to a working commit for minimal disruption.


jbearer (approved):
I saw this in the OP integration too. I think a recently nightly build is just buggy

PR #769, opened 2023-11-20
Implementation of the Stake table contract.

### LINKED ISSUE #737: Stake table contract implementation
Follow up of #673

TODO:

- [x]  `register` function 
- [x]  `deposit` function
- [x] `requestExit` function
- [x] `withdrawFunds` function
- [x] Queue (registration and exit) logic (https://github.com/EspressoSystems/espresso-sequencer/pull/803)

philippecamacho:
Closes #737.

DRAFT. Only have the code compile for now.


contracts/rust/src/bin/diff_test.rs
```diff
@@ -387,6 +393,26 @@ fn main() {
...
+            let pk_y_c0 = field_to_u256::<Fq>(vk_g2_proj.y.c0);
+            let pk_y_c1 = field_to_u256::<Fq>(vk_g2_proj.y.c1);
+
+            let sig: Signature = key_pair.sign(&msg, CS_ID_BLS_BN254);
+            let sig_affine_point = sig.sigma.into_affine();
+            let sig_x = field_to_u256::<Fq>(sig_affine_point.x);
+            let sig_y = field_to_u256::<Fq>(sig_affine_point.y);
+
+            let res = (sig_x, sig_y, pk_x_c0, pk_x_c1, pk_y_c0, pk_y_c1, msg);
+            println!("{}", res.encode_hex());
```
alxiong, line 414: We should define a `struct ParsedG2Point` (instead of parsing each field individually which is more manual and error prone), and implement two-way conversion, see `ParsedG1Point` for example.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,145 @@
...
+import { EdOnBN254 } from "../src/libraries/EdOnBn254.sol";
+import "../src/interfaces/IStakeTable.sol";
+
+// Token contract
+import { ExampleToken } from "../src/ExampleToken.sol";
+
+// Target contract
+import { StakeTable as S } from "../src/StakeTable.sol";
+
+contract StakeTable_keyRegister_Test is Test {
```
alxiong, line 20: nit: name should simply be `StakeTable_register_Test is Test` where the middle one matches exactly the corresponding target function. 

maybe skim through the[ guidance here](https://book.getfoundry.sh/tutorials/best-practices#general-test-guidance).

alxiong (commented):
General comment: 
all variables/struct/event, all functions (including those defined in `IStakeTable.sol`, except `constructor`) should add code doc. 
note that `@notice` is the general doc, `@dev` is dev-specific info, I noticed that sometimes we misuse the latter for the former. When in doubt, consult[ NatSpec](https://docs.soliditylang.org/en/latest/natspec-format.html).

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
...
+    /// @dev Computes a hash value of some G2 point
+    /// @param blsVK BLS verification key in G2
+    /// @return keccak256(blsVK)
+    function hashBlsKey(BN254.G2Point calldata blsVK) private pure returns (bytes32) {
+        uint256 x0 = blsVK.x0;
+        uint256 x1 = blsVK.x1;
+        uint256 y0 = blsVK.y0;
+        uint256 y1 = blsVK.y1;
+        bytes32 hash = keccak256(abi.encode(x0, x1, y0, y1));
+        return hash;
```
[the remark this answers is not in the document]
philippecamacho: Done in bd4e4f32db4ce6acc39ccfa4c22394cfe3be64f2.

contracts/src/interfaces/IStakeTable.sol
```diff
@@ -89,11 +91,11 @@ interface IStakeTable {
     /// the contract only treat it as auxiliary info submitted by `blsVK`.
     /// @dev `blsSig` field is necessary to prevent "rogue public-key attack".
     function register(
-        BN254.G1Point calldata blsVK,
+        BN254.G2Point calldata blsVK,
         EdOnBN254.EdOnBN254Point calldata schnorrVK,
         uint64 amount,
         StakeType stakeType,
-        bytes calldata blsSig,
+        BN254.G1Point calldata blsSig,
```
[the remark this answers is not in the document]
philippecamacho: Sure, done in 40a2034.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -27,6 +28,14 @@ use jf_plonk::{
...
 };
 use jf_primitives::pcs::prelude::{Commitment, UnivariateUniversalParams};
+
+use ark_ed_on_bn254::{EdwardsConfig as Param254, Fq as FqEd254};
+use jf_primitives::constants::CS_ID_BLS_BN254;
+use jf_primitives::signatures::bls_over_bn254::KeyPair as BLSKeyPair;
+use jf_primitives::signatures::bls_over_bn254::Signature;
+use jf_primitives::signatures::schnorr::KeyPair;
+use jf_primitives::signatures::schnorr::KeyPair as SchnorrKeyPair;
+
```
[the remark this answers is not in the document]
philippecamacho: Done in 80972684a0840d4252b7a1cdf1f16ee79d1a56f7 and 30f91ce49e32eb7062810de3dcc4b1e91fabe1c4.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -387,6 +398,53 @@ fn main() {
...
+                sig_x,
+                sig_y,
+                pk_x_c0,
+                pk_x_c1,
+                pk_y_c0,
+                pk_y_c1,
+                schnorr_pk_x,
+                schnorr_pk_y,
+                sender_address,
+            );
```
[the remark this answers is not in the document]
philippecamacho: Done in b0ba4b6.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -387,6 +398,53 @@ fn main() {
...
         }
+        Action::GenBLSSig => {
+            let mut rng = jf_utils::test_rng();
+
+            if cli.args.len() != 1 {
+                panic!("Should provide arg1=senderAddress");
+            }
+
+            let sender_address_string = cli.args[0].clone();
+            let sender_address: Address = AbiDecode::decode_hex(sender_address_string).unwrap();
```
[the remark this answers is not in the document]
alxiong: that's strange, have you tried `ethers::types::Address` instead? (see example [here](https://github.com/gakonst/ethers-rs/blob/c6efa65adc770f98475e5c4e7b644847c20d583e/examples/queries/examples/uniswapv2_pair.rs#L24))

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
...
+        bytes memory message = abi.encode(msg.sender);
+        BLSSig.verifyBlsSig(message, blsSig, blsVK);
+
+        // Find the earliest epoch at which this node can register. Usually, this will be
+        // currentEpoch() + 1 (the start of the next full epoch), but in periods of high churn the
+        // queue may fill up and it may be later. If the queue is so full that the wait time exceeds
+        // the caller's desired maximum wait, abort.
+        uint64 registerEpoch = this.nextRegistrationEpoch();
+        if (registerEpoch > validUntilEpoch) {
+            revert("Invalid next registration epoch.");
```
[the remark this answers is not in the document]
philippecamacho: Agreed. Done in fbf96f9.

contracts/test/StakeTable.t.sol
philippecamacho: Done in 8a12d7a.

contracts/rust/src/bin/diff_test.rs
philippecamacho: TODO created in b0ba4b6222fc9c02e9c65111c9663ebff4400d09

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,145 @@
...
+        assertEq(msg.sender, msgSenderAddress);
+        vm.prank(msgSenderAddress);
+
+        // Failed signature verification
+        BN254.G1Point memory badSig = BN254.P1();
+        vm.expectRevert(BLSSig.BLSSigVerificationFailed.selector);
+        stakeTable.register(
+            blsVk, schnorrVK, depositAmount, IStakeTable.StakeType.Native, badSig, validUntilEpoch
+        );
+
```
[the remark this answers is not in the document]
philippecamacho: In the tests that expect some call to `revert` we make sure the right error is raised, so I think the current test captures all the logic we want. 

I am ok to write a function for each error / the happy path but do you agree the test already covers all the execution paths? In particular the right signature is passed to the `register` function when checking for some other error to be raised.

contracts/rust/src/bin/diff_test.rs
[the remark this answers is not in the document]
alxiong: This [`FromStr for H160` impl](https://docs.rs/ethers/latest/ethers/core/abi/type.Address.html#impl-FromStr-for-H160) should makes the compiler happy, I don't know why it would err. let's chat.

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
...
+        uint64 validUntilEpoch
+    ) external returns (bool) {
+        bytes32 key = hashBlsKey(blsVK);
+        Node memory node = nodesTable[key];
+
+        // The node must not already be registered.
+        require(node.account == address(0x0), "The node has already been registered");
+
+        bytes memory message = abi.encode(msg.sender);
+        BLSSig.verifyBlsSig(message, blsSig, blsVK);
```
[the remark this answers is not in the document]
alxiong: this is not a problem. There's no benefit for registering someone else's signature. Squatting that pk doesn't give the attacker any advantage. afterall, from the HotShot system perspective, you, the attacker is the new staker now, good. you are expected to vote (oops.. you don't have the secret key, slashed)

plus when you register, you are paying yourself (w/ your own token)

plus, this so-called "rushing attack" (or I prefer "PK squatter attack") is orthogonal to what message we sign. As long as there's mempool, people could do that (even tho there's no point doing so)

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
...
+        // currentEpoch() + 1 (the start of the next full epoch), but in periods of high churn the
+        // queue may fill up and it may be later. If the queue is so full that the wait time exceeds
+        // the caller's desired maximum wait, abort.
+        uint64 registerEpoch = this.nextRegistrationEpoch();
+        if (registerEpoch > validUntilEpoch) {
+            revert("Invalid next registration epoch.");
+        }
+
+        // Create an entry for the node.
+        node.account = msg.sender;
```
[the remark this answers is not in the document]
alxiong: I ask a paymaster to help me register, but use my own wallet address for withdraw.

contracts/test/StakeTable.t.sol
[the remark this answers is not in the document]
alxiong: > In the tests that expect some call to revert we make sure the right error is raised, so I think the current test captures all the logic we want.

that's wrong.

`expectRevert` capture the next line's revert, then this function is finished, no further execution is carried out. (Line +2 is not executed because the EVM is already reverted!) 

This logic is similar to how panic handler works, there's main thread wrapping around your rust binary, then when it panic, it will give control flow to the panic handler, who will optionally log some trace if in debug mode, or directly panic if in production mode. You simply cannot go back to your rust binary again, i.e. you cannot panic twice!

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Yes, why not. WDYT @jbearer ?

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
...
+contract StakeTable is IStakeTable {
+    mapping(bytes32 keyHash => Node node) private nodesTable;
+    uint256[2] private totalStakeArr;
+    uint32 private totalNumberOfKeys;
+    uint256 private totalVotingStakeVal;
+    uint64 private numRegistrations;
+    uint64 private numPendingExits;
+    uint64 private constant BLOCKS_PER_EPOCH = 10; // TODO make an argument of the constructor?
+    uint256 private creationBlock;
+    address private tokenAddress;
```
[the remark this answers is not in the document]
philippecamacho: @philippecamacho TODO use an abstract contract instead of an interface and declare only global variables instead of getters.

contracts/test/StakeTable.t.sol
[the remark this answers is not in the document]
alxiong: I apologize, I did more tests and my understanding of forge is wrong. Sorry for the fuss.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Indeed the attack described above does not seem to do any harm, however allowing to sign the same message to  register a [rogue key](https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html ) as follows (For the ease of exposition we assume $\mathbb{G_1}=\mathbb{G_2 }=\mathbb{G}$ and $g$ generates $\mathbb{G}$):
1. An honest participant registers its public key $pk=g^{sk}$ where $sk$ is the signing key . To do so, he signs the unique message $m$ producing $\sigma=H(m)^{sk}$.
1. A malicious participant will now use $pk$ and $\sigma$ in order to register a rogue public key $pk^{\star}:=g^\beta pk^{-1}$, where $\beta$ is chosen by the attacker. To achieve this goal the attacker will need to compute a signature on the message $m$ that verifies against the rogue public key. He sets $\sigma^* := H(m)^\beta \sigma^{-1} = H(m)^\beta H(m)^{-sk}=H(m)^{\beta-sk}$. This signature will pass the BLS check as: $$e(H(m),pk^*)=e(H(m),g^\beta pk^{-1})=e(H(m),g^{\beta-sk})$$

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,149 @@
+pragma solidity ^0.8.0;
+
+import "solmate/utils/SafeTransferLib.sol";
+
+import { BN254 } from "bn254/BN254.sol";
+import { BLSSig } from "./libraries/BLSSig.sol";
+import "./interfaces/IStakeTable.sol";
```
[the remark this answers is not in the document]
philippecamacho: Done in 6d90e39 and 9f10be5.

contracts/test/StakeTable.t.sol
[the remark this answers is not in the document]
philippecamacho: No problem. It was good to [split the test](a4a17a2) into several ones anyways: I had missed [one case](https://github.com/EspressoSystems/espresso-sequencer/blob/a4a17a205560c842ea67a149322cbe4123070241/contracts/test/StakeTable.t.sol#L212) where the tokens cannot be transfered.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: > @philippecamacho TODO use an abstract contract instead of an interface and declare only global variables instead of getters.

Done in 87ed794.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alxiong: I change my mind. We should keep this code.

1. it's simpler and should be enough for most cases (even for the paymaster case, as long as paymaster does `delegatecall`, the `msg.sender` would still be the original initiator, so should be fine)
2. we don't accept extra `withdrawAddress` input argument in this function, so best to keep it that way.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alxiong: This is true, but again, how does this amounts to an attack? 
i agree that "BLS signature verification != proof of knowledge on the secret key", but I don't see how registering a rouge pk can do any good to the attacker.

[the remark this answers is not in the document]
philippecamacho: Once the rogue key $pk^{\star}$ is registered, the attacker can sign messages on the aggregate public key $pk^*pk=g^\beta pk^{-1}pk=g^\beta$.  In our context an attacker could register a rogue public key $g^\beta \Pi_{k=1}^N pk_k^{-1}$ so that the $N+1$ keys correspond to more than 2/3 of the stake and thus sign QCs on his own.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+contract StakeTable_register_Test is Test {
+    event Registered(bytes32, uint64, IStakeTable.StakeType, uint256);
+
+    S public stakeTable;
+    ExampleToken public token;
+    LightClient public lightClientContract;
+    address public lightClientAddress;
+    address public tokenAddress;
+    address exampleTokenCreator;
```
alxiong, line 29: why do we need the 3 address? the first two can be removed since we already have their contract instances. the last one, you can use [`makeAddr()`](https://book.getfoundry.sh/reference/forge-std/make-addr#makeaddr) to make it clear instead of using `msg.sender` everywhere.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+        private
+        returns (BN254.G2Point memory, EdOnBN254.EdOnBN254Point memory, BN254.G1Point memory)
+    {
+        depositAmount = 10;
+        validUntilEpoch = 5;
+
+        // Generate a BLS signature and other values using rust code
+        string[] memory cmds = new string[](3);
+        cmds[0] = "diff-test";
+        cmds[1] = "gen-bls-sig";
```
alxiong, line 44: rename to `gen-client-wallet`? (we are doing more than just bls signature.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+    uint256 constant INITIAL_BALANCE = 1_000;
+    uint64 depositAmount;
+    uint64 validUntilEpoch;
+
+    function computeGroupElements(address sender)
+        private
+        returns (BN254.G2Point memory, EdOnBN254.EdOnBN254Point memory, BN254.G1Point memory)
+    {
+        depositAmount = 10;
+        validUntilEpoch = 5;
```
alxiong, line 39: why are these here?
also I think they should be local variables (even if the same values are used across different tests, it's somewhat unexpected for me to see these as global)

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+    ExampleToken public token;
+    LightClient public lightClientContract;
+    address public lightClientAddress;
+    address public tokenAddress;
+    address exampleTokenCreator;
+    uint256 constant INITIAL_BALANCE = 1_000;
+    uint64 depositAmount;
+    uint64 validUntilEpoch;
+
+    function computeGroupElements(address sender)
```
alxiong, line 34: rename to `prepareWallet`? or some names like that?

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+            uint256 blsVKy1,
+            uint256 schnorrVKx,
+            uint256 schnorrVKy
+        ) = abi.decode(
+            result, (uint256, uint256, uint256, uint256, uint256, uint256, uint256, uint256)
+        );
+
+        return (
+            BN254.G2Point(blsVKx1, blsVKx0, blsVKy1, blsVKy0), // blsVK. Note: (x,y) coordinates for
+                // each field component must be inverted.
```
alxiong, line 63: personally, I find "must be reverted" bit confusing, we can simply remove that. I understand what you try to say. it's fine to leave it unexplained, I'm gonna fix it soon when we update to latest `bn254` anyway.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        IStakeTable.Node memory node;
+        node.account = msgSenderAddress;
+        node.balance = depositAmount;
+        node.stakeType = IStakeTable.StakeType.Native;
+        node.schnorrVK = schnorrVK;
+        node.registerEpoch = 1;
+
+        // The function IStakeTable._hashBlsKey is private but its code is small hence we copy it
+        // there.
```
alxiong, line 126: can we mark it as `internal` and avoid copying here?

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+            BN254.G2Point memory blsVK,
+            EdOnBN254.EdOnBN254Point memory schnorrVK,
+            BN254.G1Point memory sig
+        ) = computeGroupElements(msgSenderAddress);
+
+        // Prepare for the token transfer
+        vm.prank(msg.sender);
+        token.approve(address(stakeTable), depositAmount);
+
+        vm.prank(msgSenderAddress);
```
alxiong, line 105: 1. why do we need to have this `msgSenderAddress` variable
2. why do we prank twice? (one at L102)

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+        assertEq(token.balanceOf(msgSenderAddress), INITIAL_BALANCE);
+
+        uint256 nativeAmount;
+        uint256 restakedAmount;
+        (nativeAmount, restakedAmount) = stakeTable.totalStake();
+        assertEq(nativeAmount, 0);
+        assertEq(restakedAmount, 0);
+
+        // Check event is emitted
+        vm.expectEmit(false, false, false, true, address(stakeTable));
```
alxiong, line 117: try to declare `expectEmit/Revert` 1 line above the actual contract invocation for readability.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+        assertTrue(res);
+
+        // Balance after registration
+        assertEq(token.balanceOf(msgSenderAddress), INITIAL_BALANCE - depositAmount);
+        (nativeAmount, restakedAmount) = stakeTable.totalStake();
+        assertEq(nativeAmount, depositAmount);
+        assertEq(restakedAmount, 0);
+    }
+
+    function testInvalidBLSSig() external {
```
alxiong, line 148: we follow this [guide 2nd & 6th point](https://book.getfoundry.sh/tutorials/best-practices#general-test-guidance) for naming:

- `test[Fuzz]_revert[When|IF]_InvalidBLSSig`
- `test_Description_matches|succeeds`

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+        ) = computeGroupElements(msg.sender);
+
+        // Throw "Restaking not implemented" error
+        vm.expectRevert(S.RestakingNotImplemented.selector);
+        stakeTable.register(
+            blsVK, schnorrVK, depositAmount, IStakeTable.StakeType.Restake, sig, validUntilEpoch
+        );
+    }
+
+    function testInvalidNextRegistrationEpoch() external {
```
alxiong, line 174: we should use `testFuzz` in this case, whenever we are testing some user-input, we should use fuzzing as much as we could to test for unexpected inputs.

namely this function should take in `uint64 validUntilEpoch` as an input (which will be filled by the foundry fuzzer), if you want to restrict its value/range. try to use [`bound()`](https://book.getfoundry.sh/reference/forge-std/bound). 
(sometimes you want rejection sampling using [`vm.assume()`](https://book.getfoundry.sh/cheatcodes/assume#assume), just FYI, also both of these have examples in my other tests, so you can take a look)

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,225 @@
...
+            BN254.G2Point memory blsVK,
+            EdOnBN254.EdOnBN254Point memory schnorrVK,
+            BN254.G1Point memory sig
+        ) = computeGroupElements(msg.sender);
+
+        vm.prank(msg.sender);
+        vm.expectRevert("TRANSFER_FROM_FAILED");
+        stakeTable.register(
+            blsVK, schnorrVK, depositAmount, IStakeTable.StakeType.Native, sig, validUntilEpoch
+        );
```
alxiong, line 223: we forget to check: if the balance is correct if transfer failed.

in fact, we should add "[invariant test](https://book.getfoundry.sh/forge/invariant-testing)" w.r.t. the total balance of the StakeTable contract and of the users. 
^^ for this, we can add slightly later, not in this PR.

alxiong (commented):
Two general comments:

- the order of tests should follow the logic of contract: if contract check `StakeType` first, then first few test functions should be about fail if it's wrong
- the order of the test contract (within the same `StakeTable.t.sol` file) should follow the order of their target function in the implementation: if in `StakeTable.sol`, the order is `regiseter()` then `deposit()` then in the test file, it should be `contract StakeTable_register_Test` and `contract StakeTable_deposit_Test`. 
- whenever there's `external/public` function with potential user-input, we should add `testFuzz` on these as much as possible. For example, we never add validity check on `blsVK`, if we use fuzzing, then it could be that user is passing in an invalid G2/G1 point. which we should also be able to handle.

alxiong:
@sveitser if we are only using `solmate`'s ERC20, we could totally `forge uninstall openzeppelin`

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alxiong: you are right. we do need to sign distinct messages then.

contracts/src/StakeTable.sol
```diff
@@ -0,0 +1,147 @@
...
+        BN254.G2Point memory blsVK,
+        EdOnBN254.EdOnBN254Point memory schnorrVK,
+        uint64 amount,
+        StakeType stakeType,
+        BN254.G1Point memory blsSig,
+        uint64 validUntilEpoch
+    ) external override returns (bool) {
+        bytes32 key = _hashBlsKey(blsVK);
+        Node memory node = nodes[key];
+
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in 2783b42.

contracts/test/StakeTable.t.sol
philippecamacho: You are right. Useless fields removed in 2295635.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, renaming done in 5f9711a.

contracts/test/StakeTable.t.sol
philippecamacho: Agreed, fixed in 91f8df8.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, renamed to `genClientWallet` in df906e7.

contracts/test/StakeTable.t.sol
philippecamacho: Put vm.expectEmit closer to the call to contract in 1d85c9f, even though it looks like the expected event must be defined before the call.

contracts/test/StakeTable.t.sol
philippecamacho: Done in db06877 and d4893b9f3cab0d05d0fea65242c693a77ca06aba.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, given that this will be fixed at some point the confusing comment has been removed in e5165c2.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, done in 1c99334 even though I had to mark the function as `public` because it is called by another contract (the test).

philippecamacho:
> Two general comments:
> 
> * the order of tests should follow the logic of contract: if contract check `StakeType` first, then first few test functions should be about fail if it's wrong

Done in ce1ebbf .

> * the order of the test contract (within the same `StakeTable.t.sol` file) should follow the order of their target function in the implementation: if in `StakeTable.sol`, the order is `regiseter()` then `deposit()` then in the test file, it should be `contract StakeTable_register_Test` and `contract StakeTable_deposit_Test`.

Should not we have a different file for each important Stake table contract function like register, deposit etc ...?

> whenever there's external/public function with potential user-input, we should add testFuzz on these as much as possible. For example, we never add validity check on blsVK, if we use fuzzing, then it could be that user is passing in an invalid G2/G1 point. which we should also be able to handle.

Yes, added a TODO in the ticket #737.

contracts/test/StakeTable.t.sol
philippecamacho: Good idea, ticket #737 updated.

contracts/test/StakeTable.t.sol
philippecamacho: Done in 44cfc9e.

philippecamacho:
> General comment: all variables/struct/event, all functions (including those defined in `IStakeTable.sol`, except `constructor`) should add code doc. note that `@notice` is the general doc, `@dev` is dev-specific info, I noticed that sometimes we misuse the latter for the former. When in doubt, consult[ NatSpec](https://docs.soliditylang.org/en/latest/natspec-format.html).

Done in 3c97955.

.gitmodules
```diff
@@ -10,3 +10,9 @@
 	path = contracts/lib/bn254
 	url = https://github.com/EspressoSystems/solidity-bn254
 	branch = v0.1.1
+[submodule "contracts/lib/openzeppelin-contracts"]
```
sveitser, line 13: @philippecamacho I think this can be removed now. For the future it seems that foundry cannot handle multiple contracts with the same name being used in the project. The problem is that when creating bindings the bindings aren't namespaced so there's only one ERC20 token in the bindings. The bindings work anyway because the contracts should have the same interface. The problems are that that the 

1. The bytecode is embedded in the bindings and one seems to be chosen at random, it might not be exactly the same between different implementations.
2. The bindings files for the contracts also have some other differences. This is why we were sometimes seeing the CI job that checks the bindings fail because of a different ERC20 bindings file.

If we really need to have implementations of a contract with the same name we can probably do some workarounds to make it work but it would require some work. I think we don't need it at the moment.

cc @alxiong

philippecamacho: Done in 44fc0f0.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,272 @@
...
+        // Throw "Restaking not implemented" error
+        vm.expectRevert(S.RestakingNotImplemented.selector);
+        stakeTable.register(
+            blsVK,
+            schnorrVK,
+            depositAmount,
+            AbstractStakeTable.StakeType.Restake,
+            sig,
+            validUntilEpoch
+        );
```
alxiong, line 100: nit: this is a take-or-leave suggestion
instead of hardcoding these two variables, I would fuzz them: 
```solidity
   function testFuzz_RevertWhen_UsingRestakeToken(uint64 depositAmount, uint64 validUntilEpoch) external {
       uint64 curEpoch = stakeTable.currentEpoch();
       depositAmount = bound(depositAmount, 1, INITIAL_BALANCE);
       validUntilEpoch = bound(validUntilEpoch, curEpoch, curEpoch + 10);
     }     
```

personally i find this test much more convincing if I'm a security auditor.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,272 @@
...
+
+    function test_RevertWhen_InvalidBLSSig() external {
+        uint64 depositAmount = 10;
+        uint64 validUntilEpoch = 5;
+
+        (BN254.G2Point memory blsVK, EdOnBN254.EdOnBN254Point memory schnorrVK,) =
+            genClientWallet(exampleTokenCreator);
+
+        // Failed signature verification
+        BN254.G1Point memory badSig = BN254.P1();
```
alxiong, line 111: I would fuzz this by passing in a random scalar, then use `BN254.scalarMul` to get the random point as the badSig.

alxiong (commented):
The test code looks so much better and idiosyncratic! it's very easy to read them!! 👍 

I left some comments to improve the quality and coverage of the logic that I believe can make our future contract auditor happier.

alxiong:
>  > the order of the test contract (within the same StakeTable.t.sol file) should follow the order of their target function in the implementation: if in StakeTable.sol, the order is regiseter() then deposit() then in the test file, it should be contract StakeTable_register_Test and contract StakeTable_deposit_Test.
>
> Should not we have a different file for each important Stake table contract function like register, deposit etc ...?

this depends on preferences as the best practice guides ([1st point here](https://book.getfoundry.sh/tutorials/best-practices#general-test-guidance)) says:

> For testing MyContract.sol, the test file should be MyContract.t.sol. For testing MyScript.s.sol, the test file should be MyScript.t.sol.
> 
>  - If the contract is big and you want to split it over multiple files, group them logically like MyContract.owner.t.sol, MyContract.deposits.t.sol, etc.

So I guess it depends, we can split the file when it's getting big. but for now, for easier location, just keep it in `StakeTable.t.sol` single file.

contracts/test/StakeTable.t.sol
philippecamacho: Done in 3a3ec04.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        (
+            BN254.G2Point memory blsVK,
+            EdOnBN254.EdOnBN254Point memory schnorrVK,
+            BN254.G1Point memory sig
+        ) = genClientWallet(exampleTokenCreator);
+
+        vm.prank(exampleTokenCreator);
+        vm.expectRevert("TRANSFER_FROM_FAILED");
+        stakeTable.register(
```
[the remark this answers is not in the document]
philippecamacho: Done in 4d63381.

philippecamacho:
> > > the order of the test contract (within the same StakeTable.t.sol file) should follow the order of their target function in the implementation: if in StakeTable.sol, the order is regiseter() then deposit() then in the test file, it should be contract StakeTable_register_Test and contract StakeTable_deposit_Test.
> > 
> > 
> > Should not we have a different file for each important Stake table contract function like register, deposit etc ...?
> 
> this depends on preferences as the best practice guides ([1st point here](https://book.getfoundry.sh/tutorials/best-practices#general-test-guidance)) says:
> 
> > For testing MyContract.sol, the test file should be MyContract.t.sol. For testing MyScript.s.sol, the test file should be MyScript.t.sol.
> > 
> > * If the contract is big and you want to split it over multiple files, group them logically like MyContract.owner.t.sol, MyContract.deposits.t.sol, etc.
> 
> So I guess it depends, we can split the file when it's getting big. but for now, for easier location, just keep it in `StakeTable.t.sol` single file.

Sounds good, let us try with a single file for now.

contracts/test/StakeTable.t.sol
```diff
@@ -0,0 +1,272 @@
...
+        token.approve(address(stakeTable), depositAmount);
+
+        (
+            BN254.G2Point memory blsVK,
+            EdOnBN254.EdOnBN254Point memory schnorrVK,
+            BN254.G1Point memory sig
+        ) = genClientWallet(exampleTokenCreator);
+
+        // Invalid next registration epoch
+        if (validUntilEpoch < currentEpoch) {
```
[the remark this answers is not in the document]
alxiong: to follow-up on our call, 
one thing for sure is we should remove `blockNumber` variable, since it's not doing anything.

as for `curEpoch`, we can try the following in order:
- the [`StdStorage`](https://book.getfoundry.sh/reference/forge-std/std-storage) api seems reasonable to mutate the value of `currentEpoch`
- if you couldn't get it work, then, make a `test/mocks/LightClientHelper.sol` that inherits and expose storage mutation function (see optimism's [example here](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/test/mocks/NextImpl.sol))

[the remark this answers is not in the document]
philippecamacho: Done in 9261d5b.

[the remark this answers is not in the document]
alxiong: I rename it so that it ends with `*Test.sol` to avoid `forge bind` to generate unnecessary bindings for it. 
[968b4a8](https://github.com/EspressoSystems/espresso-sequencer/pull/769/commits/968b4a84cd1e34b3b8df2a2528094947e19f7599)

PR #770, opened 2023-11-21
Don't use `NonceManagerMiddleware` in commitment task

jbearer:
We don't need it, since we never have multiple transactions in flight at once, and it causes problems when a transaction fails and the nonce manager gets out of sync with the L1.

Also use less confusing names for the nonce manager and signer type aliases, and make it so that nonce manager is not the default L1 connection.


nomaxg (approved):

sveitser (approved):
LGTM

PR #776, opened 2023-11-22
feat: VID namespace impl QueryableBlock

### LINKED ISSUE #745: T2: implement tx proof
A tx proof includes:
* Evaluation proof for the first point in the namespace, to get the length of the metadata section
* Evaluation proof for the point within the metadata section containing both (i) the offset of the transaction, and (ii) the offset of the subsequent tx so we can deduce the tx length.
  * Edge case, if these two offsets span multiple points then need to batch 2 point proofs
* Evaluation proofs in the data section to read all the data between our transaction offset and the following transaction offset

Support only 1 namespace at this stage.

ggutoski:
close #745 

- `hotshot-query-service` dependency is pinned to a specific commit because recent changes to `main` break this repo, but I need the following fix: https://github.com/EspressoSystems/hotshot-query-service/pull/268
- Please look at new TODOs in the code. Some (such as idiomatic usize <-> u32 conversion) could be addressed immediately if we want, others flag temporary hacks, etc.
- Upstream jellyfish ADVZ `PayloadProver` currently fails if a proof request spans multiple polynomials https://github.com/EspressoSystems/jellyfish/issues/424 . Current tests in this PR do not trigger this edge case. But we will need to deal with it eventually.


sveitser:
> * Some (such as idiomatic usize <-> u32 conversion) could be addressed immediately if we want, others flag temporary hacks, etc.

@ggutoski I think if this is something we'll do anyway it might be good to do it now because it looks like it would make the code easier to read.

sequencer/src/block2.rs
```diff
@@ -58,13 +66,168 @@ impl BlockPayload {
...
+        let end = self
+            .get_value_usize(index.checked_add(1)?)?
+            .checked_add(tx_bodies_offset)?;
+
+        Some(start..end)
+    }
+
+    // Viewing `self.payload` bytes as a vec of words of type `TxIndex`,
+    // return the `index`th word.
+    fn get_value(&self, index: TxIndex) -> Option<TxIndex> {
```
sveitser, line 96: Not sure I understand correctly. It feels a bit weird because both inputs and outputs are of type `TxIndex` but depending on the value passed in for `index` (if the index is larger than the table) the function could construct a TxIndex from tx payload data?

This function seems to be designed for looking up indexes in the table in the payload. Can we make this safer by indicating it in the name and making it panic or return an error if it's used wrongly?

sequencer/src/block2.rs
```diff
@@ -58,13 +66,168 @@ impl BlockPayload {
...
+    fn get_tx_range(&self, index: TxIndex) -> Option<Range<usize>> {
+        let tx_bodies_offset = self.tx_bodies_offset()?;
+
+        // tx_table[i] is end index for the ith tx,
+        // so the range for the ith tx is tx_table[i-1..i].
+        // But tx_table starts at index 1 in the payload
+        // because index 0 in the payload is the length of the tx_table.
+        // So the range for the ith tx in the payload is actually
+        // [index..index+1].
+        let start = if index == 0 {
```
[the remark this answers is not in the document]
ggutoski: > Edit: I get it now.

Yeah, I should probably draw a picture but that takes a lot of time. 😕 

> Does `get_value(index)` not work with `index=0`?

It works at `index=0`. Indeed, that occurs explicitly in this PR:
https://github.com/EspressoSystems/espresso-sequencer/blob/451bf3e2581a1a462188ec395c99cacbbb319bc5/sequencer/src/block2.rs#L112-L120

The reason for this edge case is that each entry in the table denotes the *end* index for that tx. To get the *start* index for tx `i` you need to look at the *end* index for tx `i-1`. This entry does not exist at `i=0` but we know the first tx always starts at index 0 in the payload.

Does that make sense? If so then perhaps I should clarify this in the comments.

sequencer/src/block2.rs
ggutoski: > It feels a bit weird because both inputs and outputs are of type `TxIndex` ...

Agree it's confusing. In an ideal world we would use Rust's type system to differentiate between an index *into* the tx table and the actualy *entry* itself. (I have a crusade for [typed collection indices](https://github.com/axelarnetwork/tofn/blob/4f2a4de3c4a57f4168368b2afd02bc8daaea6822/README.md#typed-collection-indices) for which I'm attempting to recruit adherents.)

This is an internal function so I did not take the time to make it bulletproof for downstream users.

> ...but depending on the value passed in for `index` (if the index is larger than the table) the function could construct a TxIndex from tx payload data?

Correct.

> This function seems to be designed for looking up indexes in the table in the payload. Can we make this safer by indicating it in the name and making it panic or return an error if it's used wrongly?

Not a bad idea. I'll do that.

sequencer/src/block2.rs
```diff
@@ -58,13 +66,168 @@ impl BlockPayload {
...
+            .checked_add(tx_bodies_offset)?;
+
+        Some(start..end)
+    }
+
+    // Viewing `self.payload` bytes as a vec of words of type `TxIndex`,
+    // return the `index`th word.
+    fn get_value(&self, index: TxIndex) -> Option<TxIndex> {
+        // TODO idiomatic usize <-> TxIndex conversion.
+        let start = usize::try_from(index)
```
jbearer, line 98: IMO, we could treat it as an invariant that all the indices are small enough to fit in a `u32`, and panic if any of these conversions fail. We just need to check and return an error whenever we are building a block or receiving a block we don't necessarily know is well-formed (e.g. `from_transactions` and `deserialize`). Then we could simplify error handling a lot everywhere else.

sequencer/src/block2.rs
```diff
@@ -58,13 +66,168 @@ impl BlockPayload {
...
+        0..self.tx_table_len().unwrap_or(0)
+    }
+
+    fn transaction_with_proof(
+        &self,
+        index: &Self::TransactionIndex,
+    ) -> Option<(Self::Transaction, Self::InclusionProof)> {
+        let vid = boilerplate::test_vid_factory(); // TODO temporary VID construction
+        let tx_range = self.get_tx_range(*index)?;
+        let proof: SmallRangeProof<_> = vid.payload_proof(&self.payload, tx_range.clone()).unwrap();
```
jbearer, line 155: Under what conditions does `payload_proof` fail? Is it ok to panic here or should we return `None`?

sequencer/src/block2.rs
```diff
@@ -98,38 +261,88 @@ mod test {
             vec![],
         ];
 
-        for tx_payloads in test_cases {
+        let vid = test_vid_factory();
+        let num_test_cases = test_cases.len();
+        for (t, tx_bodies) in test_cases.into_iter().enumerate() {
+            println!(
```
jbearer, line 267: Better to use `tracing::info`. You can call `async_compatibility_layer::logging::{setup_logging, setup_backtrace}` at the beginning of the test to set this up to output nicely to the terminal

sequencer/src/block2.rs
```diff
@@ -58,13 +66,168 @@ impl BlockPayload {
...
+
+        Some(start..end)
+    }
+
+    // Viewing `self.payload` bytes as a vec of words of type `TxIndex`,
+    // return the `index`th word.
+    fn get_value(&self, index: TxIndex) -> Option<TxIndex> {
+        // TODO idiomatic usize <-> TxIndex conversion.
+        let start = usize::try_from(index)
+            .unwrap()
```
ggutoski, line 99: @sveitser regarding https://github.com/EspressoSystems/espresso-sequencer/pull/776#issuecomment-1827542376:

> > * Some (such as idiomatic usize <-> u32 conversion) could be addressed immediately if we want, ...
> 
> @ggutoski I think if this is something we'll do anyway it might be good to do it now because it looks like it would make the code easier to read.

Agree. My comment is a cry for help / solicitation of opinions on how best to proceed. Options I see:

1. Fail the build if `usize` is too small to hold a `u32` / `TxIndex`. Good because we don't need error handling. (Though we still might need some error handling for `usize` -> `u32` conversion.)
2. Panic if conversion fails.
3. Use `Result` / `Option` for failed conversion.

I prefer something clean with no panic and no error handling cruft such as item 1. But in general what's best practice here?

sequencer/src/block2.rs
ggutoski: Whoa thanks for catching this! Fixed in 4e73ce7d79fddb4b06e32ef2d8c533711b08ba24

sequencer/src/block2.rs
ggutoski: Introduced a newtype `TxTableEntry`, renamed `get_value` to `get_tx_table_entry`, added a check so that it can be used only to read from the tx table and nothing else.

(At first this function returned a `Option<TxTableEntry>` as one might expect but it's only use always converts to `usize` so I later moved the `TxTableEntry` -> `usize` conversion inside `gest_tx_table_entry`.)

sequencer/src/block2.rs
ggutoski: I introduced a newtype `TxTableEntry` for type safety and clarity as per https://github.com/EspressoSystems/espresso-sequencer/pull/776#discussion_r1405992221 . Now it's equally crufty to panic vs. check errors. Also, all arithmetic is paranoid with `checked_add` etc. Now someone would need to try pretty hard to accidentally introduce a bug.

(Currently my "errors" are only `Option` but this can be changed later to `Result` if desired. Frankly, I'm getting sick of fighting with `Result`.)

sequencer/src/block2.rs
```diff
@@ -30,100 +35,83 @@ impl BlockPayload {
...
-    // Viewing `self.payload` bytes as a vec of words of type `TxIndex`,
-    // return the `index`th word.
-    fn get_value(&self, index: TxIndex) -> Option<TxIndex> {
-        // TODO idiomatic usize <-> TxIndex conversion.
-        let start = usize::try_from(index)
-            .unwrap()
-            .checked_mul(size_of::<TxIndex>())?;
+    // Return the `index`th entry from the tx table as `usize`
+    // Return `None` if `index` exceeds the tx table length.
+    fn get_tx_table_entry(&self, index: TxIndex) -> Option<usize> {
```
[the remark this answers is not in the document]
ggutoski: > (At first this function returned a `Option<TxTableEntry>` as one might expect but it's only use always converts to `usize` so I later moved the `TxTableEntry` -> `usize` conversion inside `gest_tx_table_entry`.)

https://github.com/EspressoSystems/espresso-sequencer/pull/776#discussion_r1406859801

sequencer/src/block2.rs
```diff
@@ -16,11 +17,15 @@ pub struct BlockPayload {
...
     #[allow(dead_code)] // TODO temporary
-    fn build(txs: impl IntoIterator<Item = Transaction>) -> Self {
-        // tx_table[i] is the end index (exclusive) of the ith tx
-        // so that the payload bytes of the ith tx is
-        // tx_bodies[tx_table[i-1]..tx_table[i]].
-        // edge case: tx_table[-1] is defined as 0.
+    fn build(txs: impl IntoIterator<Item = Transaction>) -> Option<Self> {
+        // `tx_table` is a bytes representation of the following table:
+        // word[0]: [number n of entries in tx table]
+        // word[i>0]: [end byte index of the ith tx in the payload]
```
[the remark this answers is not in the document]
ggutoski: Yep. Thank you.

sequencer/src/block2.rs
```diff
@@ -152,16 +140,75 @@ impl QueryableBlock for BlockPayload {
...
+    }
+
+    impl TryFrom<usize> for TxTableEntry {
+        type Error = <TxTableEntryWord as TryFrom<usize>>::Error;
+
+        fn try_from(value: usize) -> Result<Self, Self::Error> {
+            TxTableEntryWord::try_from(value).map(Self)
+        }
+    }
+    impl TryFrom<TxTableEntry> for usize {
```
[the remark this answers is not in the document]
ggutoski: Could do, especially if we combine with `const_assert` to enforce it at compile time. But the improvement to readability is not very significant. I'm just going to leave it in ultra-paranoid mode for now. Really all this cruft is rust's fault. 😉

sequencer/src/block2.rs
```diff
@@ -152,16 +140,75 @@ impl QueryableBlock for BlockPayload {
...
+        type Error = <usize as TryFrom<TxTableEntryWord>>::Error;
+
+        fn try_from(value: TxTableEntry) -> Result<Self, Self::Error> {
+            usize::try_from(value.0)
+        }
+    }
+
+    impl TryFrom<super::TxIndex> for TxTableEntry {
```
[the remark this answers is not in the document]
ggutoski: Yes but they might not stay the same forever. It feels good that the size of either could be changed by a one-liner.

PR #791, opened 2023-12-01
Implementation of `deposit`, `requestExit` and `withdrawFunds` functions

philippecamacho:
Part of #737.

To test:
```
> just sol-test
```

Implements the three functions `deposit`, `requestExit` and `withdrawFunds`. Note that the queue logic for registration and exit will be addressed in another PR.


contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
     function requestExit(BN254.G2Point memory blsVK) external override returns (bool) {
-        bytes32 hash = _hashBlsKey(blsVK);
-        nodes[hash].exitEpoch = 0;
+        bytes32 key = _hashBlsKey(blsVK);
+        Node memory node = nodes[key];
+
+        // The exit request must come from the node's withdrawal account.
+        if (node.account != msg.sender) {
+            revert CallerIsNotTheNodeOwner();
+        }
```
alxiong, line 260: again, do we need this?

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+        // Cannot request to exit if an exit request is already in progress.
+        if (node.exitEpoch != 0) {
+            revert ExitRequestInProgress();
+        }
+
+        // Cannot exit before becoming an active participant. Activation happens one epoch after the
+        // node's registration epoch, due to the consensus-imposed activation waiting period.
+        if (currentEpoch() < node.registerEpoch + 1) {
+            revert PrematureExit();
+        }
```
alxiong, line 271: good find, I think we should parameterize this into `min_residency` system parameter.

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+
+        // Prepare the node to exit.
+        node.exitEpoch = this.nextExitEpoch();
+
+        nodes[key] = node;
```
alxiong, line 276: maybe just `nodes[key].exitEpoch = this.nextExitEpoch()` instead of copying the whole struct?

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+            revert PrematureDeposit();
+        }
+
+        // A node cannot deposit more tokens if an exit request is in progress.
+        if (node.exitEpoch != 0) {
+            revert ExitRequestInProgress();
+        }
+
+        node.balance += amount;
+        nodes[key] = node;
```
alxiong, line 239: maybe only changes those few fields instead of copying the whole struct?

contracts/src/StakeTable.sol
```diff
@@ -99,18 +115,28 @@ contract StakeTable is AbstractStakeTable {
...
+    /// for easier testing.
+    /// @param node node which is assigned an exit escrow period.
+    /// @return Number of epochs post exit after which funds can be withdrawn.
+    function exitEscrowPeriod(Node memory node) public pure returns (uint64) {
+        if (node.balance > 100) {
+            return 10;
+        } else {
+            return 5;
+        }
+    }
```
alxiong, line 138: this should be the logic of `nextExitEpoch`, correct? I'm not sure why this exists.

contracts/src/StakeTable.sol
```diff
@@ -212,8 +286,19 @@ contract StakeTable is AbstractStakeTable {
...
+
+        if (currentEpoch() < node.exitEpoch + exitEscrowPeriod(node)) {
+            revert PrematureWithdrawal();
+        }
+        uint64 balance = node.balance;
+        SafeTransferLib.safeTransfer(ERC20(tokenAddress), node.account, balance);
+
+        Node memory nullNode =
+            Node(address(0), StakeType.Native, 0, 0, 0, EdOnBN254.EdOnBN254Point(0, 1));
+        nodes[key] = nullNode;
```
alxiong, line 300: can simply use `delete nodes[key]` syntax.

contracts/src/StakeTable.sol
```diff
@@ -22,6 +22,22 @@ contract StakeTable is AbstractStakeTable {
     /// account.
     error NodeAlreadyRegistered();
 
+    /// Error raised when a user tries to make a deposit or request an exit but does not control the
+    /// node public key.
+    error CallerIsNotTheNodeOwner();
```
alxiong, line 27: rename to `Unauthenticated()`?

alxiong (commented):
I think pending queue logic is little tricky when we support `deposit`.
for register and exit, we could set their `registerEpoch` and `exitEpoch` to the `nextRegister/ExitEpoch()` based on the queuing logic.
However, it's currently underspecified how this queue works, or who would _clear_ the queue and when. 
and when there's deposit, the current `struct Node` is not sufficient to capture "this node currently is active and has 10 token, but he has another 5 pending token to be active in epoch X down the line" info.
^^ we need to discuss this.

note: I haven't review the tests yet, since there are a few things to settle on contract logic first.

philippecamacho:
> However, it's currently underspecified how this queue works, or who would _clear_ the queue and when

The way the [queue](https://github.com/EspressoSystems/gitbook/blob/f3c8a00297dccb54b8f53cd255564f466e9a32dc/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table.md#queueing) is implemented does not require any cleanup process: the queue just tracks two numbers that are updated when `next{Register/Exit}Epoch` is called.

> `struct Node` is not sufficient to capture "this node currently is active and has 10 token, but he has another 5 pending token to be active in epoch X down the line

Yes, I agree that the way this is currently specified allows for balances to change within an epoch and that feels wrong. Let us discuss.

cc @jbearer

contracts/src/StakeTable.sol
philippecamacho: When I try something like this, some test fails:

```solidity
[nix-shell:~/repositories/espressosys/espresso-sequencer]$ git diff
diff --git a/contracts/src/StakeTable.sol b/contracts/src/StakeTable.sol
index 54f2cd4..e24cac8 100644
--- a/contracts/src/StakeTable.sol
+++ b/contracts/src/StakeTable.sol
@@ -235,8 +235,7 @@ contract StakeTable is AbstractStakeTable {
             revert ExitRequestInProgress();
         }
 
-        node.balance += amount;
-        nodes[key] = node;
+        nodes[key].balance += amount;
         SafeTransferLib.safeTransferFrom(ERC20(tokenAddress), msg.sender, address(this), amount);
 
         emit Deposit(_hashBlsKey(blsVK), uint256(amount));
```

```
Failing tests:
Encountered 1 failing test in contracts/test/StakeTable.t.sol:StakeTable_Test
[FAIL. Reason: Assertion failed.] test_Deposits_succeeds() (gas: 467140)
```

contracts/src/StakeTable.sol
philippecamacho: Similarly to the [thread](https://github.com/EspressoSystems/espresso-sequencer/pull/791#discussion_r1415815687) above, some test fails when doing:

```solidity
[nix-shell:~/repositories/espressosys/espresso-sequencer]$ git diff
diff --git a/contracts/src/StakeTable.sol b/contracts/src/StakeTable.sol
index 54f2cd4..416f2e4 100644
--- a/contracts/src/StakeTable.sol
+++ b/contracts/src/StakeTable.sol
@@ -271,9 +271,7 @@ contract StakeTable is AbstractStakeTable {
         }
 
         // Prepare the node to exit.
-        node.exitEpoch = this.nextExitEpoch();
-
-        nodes[key] = node;
+        nodes[key].exitEpoch = this.nextExitEpoch();
 
         emit Exit(key, node.exitEpoch);
```

```
Failing tests:
Encountered 1 failing test in contracts/test/StakeTable.t.sol:StakeTable_Test
[FAIL. Reason: log != expected log] test_RequestExit_Succeeds() (gas: 453974)
```

contracts/src/StakeTable.sol
philippecamacho: No, this is different. It is an extra period of time needed when withdrawing (after the exit is complete).
From the [specs](https://github.com/EspressoSystems/gitbook/blob/f3c8a00297dccb54b8f53cd255564f466e9a32dc/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table.md#exit):
> Once the exit epoch has been recorded, HotShot replicas will observe the change at the start of the next epoch, when they next synchronize the stake table, and honest replicas will consider the node exited once consensus enters exitEpoch. The node that exited will be able to withdraw their staked tokens after the escrow period has passed, using the withdraw function:

contracts/src/StakeTable.sol
philippecamacho: Right, done in 978171a4744d748d21993b63809a51018b3642f8.

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
-        bytes32 hash = _hashBlsKey(blsVK);
-        nodes[hash].balance += amount;
-        return (0, 0);
+        bytes32 key = _hashBlsKey(blsVK);
+        Node memory node = nodes[key];
+
+        // The deposit must come from the node's registered account.
+        if (node.account != msg.sender) {
+            revert CallerIsNotTheNodeOwner();
+        }
```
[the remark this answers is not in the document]
philippecamacho: You mean, you could potentially allow donations?

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+            revert ExitRequestInProgress();
+        }
+
+        node.balance += amount;
+        nodes[key] = node;
+        SafeTransferLib.safeTransferFrom(ERC20(tokenAddress), msg.sender, address(this), amount);
+
+        emit Deposit(_hashBlsKey(blsVK), uint256(amount));
+
+        uint64 effectiveEpoch = currentEpoch() + 1;
```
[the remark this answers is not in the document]
philippecamacho: At this stage the node is already active, but this is part of the unspecified behavior we need to [discuss](https://github.com/EspressoSystems/espresso-sequencer/pull/791#issuecomment-1841007924).

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+        bytes32 key = _hashBlsKey(blsVK);
+        Node memory node = nodes[key];
+
+        // The deposit must come from the node's registered account.
+        if (node.account != msg.sender) {
+            revert CallerIsNotTheNodeOwner();
+        }
+
+        // A node cannot deposit more tokens while it waiting to register,
+        if (currentEpoch() <= node.registerEpoch) {
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in 1b09262.

contracts/src/StakeTable.sol
philippecamacho: In this case I think we should not allow anyone to request an exit for a specific node, right?

contracts/src/StakeTable.sol
philippecamacho: Why would we need this extra parameter?

contracts/src/interfaces/AbstractStakeTable.sol
```diff
@@ -23,7 +23,24 @@ abstract contract AbstractStakeTable {
...
+
+    /// @notice Signals an exit request has been granted.
+    /// @param blsVKhash hash of the BLS public key owned by the user who requested to exit.
+    /// @param exitEpoch epoch when the user will be allowed to withdraw its funds.
+    event Exit(bytes32 blsVKhash, uint64 exitEpoch);
+
+    /// @notice Signals a deposit to a BLS public key.
+    /// @param blsVKhash hash of the BLS public key that received the deposit.
+    /// @param amount amount of the deposit
+    event Deposit(bytes32 blsVKhash, uint256 amount);
```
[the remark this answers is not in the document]
philippecamacho: There is no amount involved when calling `requestExit`. However you are right, we should problably have an event `Withdrawal` that specify the epoch and amount involved.  Also `deposit` should probably mention the epoch but we still need to specify more this part of the design.

philippecamacho:
> I think pending queue logic is little tricky when we support `deposit`.

@alxiong  The queue logic (for both Registration and Exit) is implemented in #803.

philippecamacho:
> > However, it's currently underspecified how this queue works, or who would _clear_ the queue and when
> 
> The way the [queue](https://github.com/EspressoSystems/gitbook/blob/f3c8a00297dccb54b8f53cd255564f466e9a32dc/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table.md#queueing) is specified does not require any cleanup process: the queue just tracks two numbers that are updated when `next{Register/Exit}Epoch` is called.
> 
> > `struct Node` is not sufficient to capture "this node currently is active and has 10 token, but he has another 5 pending token to be active in epoch X down the line
> 
> Yes, I agree. The current [specification](https://github.com/EspressoSystems/gitbook/blob/f3c8a00297dccb54b8f53cd255564f466e9a32dc/espresso-sequencer-architecture/internal-functionality/sequencer-contract/stake-table.md#deposit) allows for balances to change within an epoch and that feels wrong. Let us discuss.
> 
> cc @jbearer

Based on the [zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/Stake.20table.20contract.3Ahow.20to.20handle.20deposits.20within.20an.20epoch.3F/near/406103539) conversation, I would add a test to ensure a user can deposit several times inside an epoch when the node is active.

contracts/src/interfaces/AbstractStakeTable.sol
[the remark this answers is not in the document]
alxiong: > There is no amount involved when calling requestExit.

why is that? imagine a top staker who wants to exit, we may not allow him to exit all of his assets (depending on the rate-limiting policy)

contracts/src/StakeTable.sol
```diff
@@ -191,18 +217,66 @@ contract StakeTable is AbstractStakeTable {
...
+            revert PrematureDeposit();
+        }
+
+        // A node cannot deposit more tokens if an exit request is in progress.
+        if (node.exitEpoch != 0) {
+            revert ExitRequestInProgress();
+        }
```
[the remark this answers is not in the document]
alxiong: I see.

maybe I'm complicating our design, but in my mind, we might want to allow staker to "partially withdraw", say I have 32 token staked, I may only want to exit with 10?

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alxiong: nvm, I was thinking maybe the one who request the exit doesn't have to be the beneficiary of the amount. 
then I realize that we still need a way to authenticate the exit request anyway. either via Bls sig verification, or via `msg.sender` check, and the latter is easier and cheaper.

contracts/src/StakeTable.sol
alxiong: Ah, thanks for the reference. let's add some code comments to explain this.

```
requestExit -- (queing) -->  Exited -- (escrow) --> Withdrawable
First phase is about waiting in queue due to rate-limiting on exit, the wait is dependent on the exit amount and currently exit traffic.
At the point of "Exited", Sequencer is officially off duty, and stops participating in consensus.
Second phase is about slashable security, the wait is dependent only on amount, during which period cryptographic evidence of misbehavior (e.g. double-voting) might still lead to the forfeit of stakes.
From the point of `Withdrawable` onwards, staker can freely withdraw.
```

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alxiong: :thinking:  also, do we only rate limit registration but not further deposit?

^^ this is somewhat reasonable? (I don't have objection, but seems a bit artificial) neither registration nor deposit affect state prover, the proving cost is fixed (As the circuit is upper-bounded) 

the rate limit on registration is motivated by two points imo:
-  the "hand-over" overhead between old and new parliament, you don't want a completely new chamber in the parliament cuz the state syncing takes time, and having significant overlaps between two chambers improve stability in practice
- mitigate attacks where a sudden inflow of malicious capital trying to hijack the consensus (think someone took a high-value short-term loan to attack our consensus)

when considering the second point, we should also consider rate-limit deposits?

contracts/src/StakeTable.sol
alxiong: as conceded in https://github.com/EspressoSystems/espresso-sequencer/pull/791#discussion_r1416734021, I agree.

contracts/src/StakeTable.sol
alxiong: instead of `+1`, we might want to enforce a staker to stay on-duty for a `min_residency` epochs before requesting exit to avoid too frequent change of consensus quorum?
This is just a design decision I guess.

WDYT @jbearer ?

contracts/src/StakeTable.sol
alxiong: maybe your test is wrong, I double checked in Remix:

```solidity
pragma solidity ^0.8.0;

contract StakeTable {
    struct Node {
        uint256 registerEpoch;
        uint256 exitEpoch;
    }
    mapping(address => Node) public nodes;
    constructor () {
        nodes[msg.sender].registerEpoch = 5;
        nodes[msg.sender].exitEpoch = 6;
    }

    function increase(address addr) public {
        Node memory n = nodes[addr];
        nodes[addr].registerEpoch = n.exitEpoch + 1;
        nodes[addr].exitEpoch = n.exitEpoch + 2;
    }
}
```
storage is updated as expected.

contracts/src/interfaces/AbstractStakeTable.sol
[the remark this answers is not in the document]
philippecamacho: It is an additional feature. I am ok with that. Is it important to implement it now? cc @jbearer

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Ok, closing thread then.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
jbearer: It seems like a good feature. I think the only downside would be more complex bookkeeping, in the same vein as incremental deposit

contracts/src/interfaces/AbstractStakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Will be addressed in #806 (see *Limit withdrawals amount w.r.t. to some queue.* item).

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Will be addressed in #806.

contracts/src/StakeTable.sol
philippecamacho: Will be addressed in #806. See *Force the stake to stay active more than one epoch before requesting an exit* item.

contracts/src/StakeTable.sol
philippecamacho: This was a bug on my side. Fixed in 8e9c15a.

contracts/src/StakeTable.sol
philippecamacho: Also a bug on my side. Fixed in 8e9c15a.

alxiong (approved):

Issue #747, opened 2023-11-13
[parent] Support arbitrary payload bytes


ggutoski:
From https://github.com/EspressoSystems/espresso-sequencer/pull/756#discussion_r1394568563 :

> ... we'll need to test with some blocks that weren't built by `BlockPayload::build`, since that function only builds "nice" blocks. We'll need to maliciously build some blocks with overlapping and truncated and out-of-order transactions.

- [x] #816 
- [x] #817

ggutoski:
closing in favour of #717

PR #796, opened 2023-12-04
feat: add tx table entry proofs to tx payload proof

### LINKED ISSUE #795: Tx payload proofs should include proofs for tx table entries
Currently a tx payload proof for the `i`th tx includes only a proof for the payload bytes for that tx. We also need to prove that these payload bytes actually correspond to the `i`th entry in the tx table.

ggutoski:
close #795 

- Allow proof generation for zero-length tx payloads. (Previously we disallowed this.)
- Add a trivial `from_bytes` implementation in preparation for #747 , which is out of scope for the current PR.
- New TODOs worthy of attention flagged with `TODO(795)`


sequencer/src/block2.rs
```diff
@@ -57,31 +57,53 @@ impl BlockPayload {
...
+        B: IntoIterator<Item = u8>,
+    {
+        Self {
+            payload: bytes.into_iter().collect(),
+        }
+    }
+
     // Return the range `r` such that the `index`th tx bytes are at `self.payload[r]`.
-    fn get_tx_range(&self, index: TxIndex) -> Option<Range<usize>> {
-        let tx_bodies_offset = self.tx_bodies_offset()?;
```
jbearer, line 62: Why did we get rid of the `tx_bodies_offset` function?

sequencer/src/block2.rs
```diff
@@ -139,17 +151,53 @@ impl QueryableBlock for BlockPayload {
...
+        let (tx_range, tx_table_range_proof) = self.get_tx_range_with_proof(*index, &vid)?;
+
         Some((
             // TODO don't copy the tx bytes into the return value
             // https://github.com/EspressoSystems/hotshot-query-service/issues/267
             Transaction::new(crate::VmId(0), self.payload.get(tx_range.clone())?.to_vec()),
-            vid.payload_proof(&self.payload, tx_range).ok()?,
+            TxInclusionProof {
+                // TODO(795) every proof will contain a copy of the tx table len proof.
+                // Should we cache it in `BlockPayload` or something?
```
jbearer, line 163: 👍

sequencer/src/block2.rs
```diff
@@ -139,17 +151,53 @@ impl QueryableBlock for BlockPayload {
...
 
-mod tx_table {
+type TxIndex = <BlockPayload as QueryableBlock>::TransactionIndex;
+
+// TODO upstream type aliases: https://github.com/EspressoSystems/jellyfish/issues/423
+type RangeProof =
+    SmallRangeProof<<UnivariateKzgPCS<Bls12_381> as PolynomialCommitmentScheme>::Proof>;
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct TxInclusionProof {
```
jbearer, line 185: Either in this PR or soon after, we should implement a method something like

```rust
impl TxInclusionProof {
  fn validate(&self, payload_comm: VidCommitment, tx: &Transaction) -> bool;
}
```

This will just help be sure we haven't missed anything in the proof generation that we would need in the verifier. Although with this change I don't think we have (except of course the namespace stuff)

jbearer (approved):
LGTM

sequencer/src/block2.rs
ggutoski: Collateral damage from a chaotic dev process. Restored in 35499ad84f8fe4d19a65665aa985330c64567eaa

sequencer/src/block2.rs
ggutoski: Done in https://github.com/EspressoSystems/espresso-sequencer/pull/796/commits/be7e66b7483df7bec9f9be22128d29bc6cdf129d via `OnceLock`. That was a learning experience.

sequencer/src/block2.rs
ggutoski: Done in https://github.com/EspressoSystems/espresso-sequencer/pull/796/commits/90d8ddd4b103dff730c7ff59fd4a1f18e87f29de .

The new `verify()` method has an obnoxious arg list. I left it that way on purpose for now so you can see all the pieces needed to verify a proof. Where should all these pieces go? eg. Should they all become fields of `TxInclusionProof`?

ggutoski: @jbearer Thinking about this a bit more: the problem is that we need both range and payload bytes for *three* subproofs in a `TxInclusionProof`. That's a lot of cruft.

I could just add all this info to `TxInclusionProof` so that the verifier doesn't need to hang on to this info and pass it in a big ugly statement to `verify` but that would conflict with the design principle we discussed in https://github.com/EspressoSystems/jellyfish/pull/389#discussion_r1383801714 . If it's okay to clean it up here then it should also be okay to clean it up in jellyfish, too.

If we're not careful we could create a cascade of copies of all this metadata at each level of the stack: a copy in the jellyfish proof plus a copy somewhere in the sequencer (maybe in the `TxInclusionProof`?) to pass to the jellyfish verification, etc.

I'm still not convinced that jellyfish verification should require the range. If we eliminate this requirement from jellyfish then it would eliminate the need to store 3 ranges in the sequencer. Moreover, I think we could include the payload bytes for the tx table length and entry inside `TxInclusionProof`, which would eliminate the need to store those things separately in the sequencer. I guess my point is that all this cruft is really just implementation details of the proof and should not be exposed to the user. wdyt?

sequencer/src/block2.rs
```diff
@@ -190,13 +218,140 @@ pub struct TxInclusionProof {
...
     Other(RangeProof, RangeProof),
     // TODO new variant `Final` with implicit `end` https://github.com/EspressoSystems/espresso-sequencer/issues/757
 }
 
+impl TxInclusionProof {
+    // TODO(795) prototype only!
+    //
+    // - Obnoxious arg list; where to store all this cruft?
+    // - Returns `None` if an error occurred.
+    // - Use of `Result<(),()>` pattern to indicate a must-use bool return type.
```
jbearer, line 233: I don't mind `Result<(), ()>` but FYI you can also put `#[must_use]` on a function returning `bool` (or anything)

sequencer/src/block2.rs
jbearer: I think the {`table_len`,`entry_start`,`entry_end`} bytes could all be part of the proof. The reason is that the verifier isn't actually interested in those things, they just want to know generally that the transaction bytes are correct, and those things are implementation details of the proof.

I think we can _probably_ also move `tx_payload_range` to the proof, since the caller is not likely interested in reading a transaction at any particular offset (how would they know?), they just want to know they have the transaction at the right index. So only `tx_index` needs to be a verifier input.

Eventually, we may create an even higher-level wrapper for the API endpoint that puts `tx_index` in the proof as well, since probably the end user queried by hash, and they can check that just by recomputing the hash of `tx`. But at the level of abstraction we're currently at, I think it makes sense to expose `tx_index`.

sequencer/src/block2.rs
ggutoski: True. I forgot to mention that it also enables use of `?` for concise abort-on-failure.

jbearer: Oh yeah that's nice

sequencer/src/block2.rs
ggutoski: Ok sounds good.

I really don't like that the range is duplicated in both the jellyfish proof and in `TxInclusionProof` and I think I finally have a good solution. We made a mistake including this range in the jellyfish proof. As a general rule, each piece of data should be in the statement *or* the proof but never both. You have argued that the range should be in the (jellyfish) statement, so it should be removed from the (jellyfish) proof. Indeed, the only use of this data in jellyfish is to check that it equals the range given by the user in the `Statement`---in other words, it's completely useless.

This is a trivial upstream change that will save a few bytes of duplicated data inside proofs. Make sense?

jbearer: Yeah, sound sgood

PR #798, opened 2023-12-05
Feat/sql data source

### LINKED ISSUE #584: Set up SQL backend for query service in local demo
- [ ] https://github.com/EspressoSystems/hotshot-query-service/issues/212
- [ ] Replace file system command line flags with SQL connection parameters (or add new options, if we want to support both backends at runtime)
- [ ] Add postgres container(s) to docker-compose.yaml
- [ ] Configure query service sequencer nodes to connect to the SQL containers

jbearer:
Support SQL backend for query service

The timestamp window endpoints don't work yet, because they depend on reading the timestamp out of the application-defined block header, which isn't yet being populated in the generic query service. This should be fixed after the latest HotShot update, which exposes the header type to consensus and the query service.

Closes #584


jbearer:
Not sure why Github isn't triggering CI for this PR. I ran [build](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/7176704845) and [lint](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/7176705744) manually, both passed

.github/workflows/build.yml
```diff
@@ -23,9 +23,7 @@ env:
 
 jobs:
   build:
-    runs-on: [self-hosted, X64]
-    container:
-      image: ghcr.io/espressosystems/devops-rust:stable
+    runs-on: ubuntu-latest
```
sveitser, line 26: Because we need to run docker for the tests now, or?

sequencer/src/api.rs
```diff
@@ -115,10 +122,7 @@ mod test {
             handle.hotshot.start_consensus().await;
         }
 
-        let mut options = Options::from(options::Http { port }).submit(Default::default());
-        if let Some(query) = query_opt {
-            options = options.query_fs(query);
-        }
+        let options = opt(Options::from(options::Http { port }).submit(Default::default()));
```
sveitser, line 125: This seems a bit tricky to reason about together with the function/closure argument. Maybe add a comment about what the closure should and should not (i guess change the port (?)) do?

.github/workflows/build.yml
jbearer: Yeah, and because we don't really need the self-hosted runner for the tests in this repo.

sequencer/src/api/options.rs
```diff
@@ -152,6 +162,30 @@ pub struct Submit;
 #[derive(Parser, Clone, Copy, Debug, Default)]
 pub struct Status;
 
+/// Options for the query API module backed by a Postgres database.
+#[derive(Parser, Clone, Debug, Default)]
+pub struct Sql {
+    /// Hostname for the remote Postgres database server.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_POSTGRES_HOST")]
```
sveitser, line 169: For later, but I think we could use a URL instead ( https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html#url ).

.github/workflows/build.yml
sveitser: It's surprising the build is slightly faster on the public runner.

sequencer/src/api/options.rs
jbearer: Hmm I'm not sure I like that. I think having a separate setting for each parameter makes it easier to read and search for in e.g. .env files or terraform.

Maybe we could support both: ie the config is first parsed from a URL and then individual settings can be overridden via env vars

sveitser: I just thought since we usually go with URLs for clients. I agree postgres URLs can be a bit complex. I'm fine keeping what we have.

sequencer/src/api/endpoints.rs
```diff
@@ -23,22 +23,22 @@ pub struct NamespaceProofQueryData {
...
-            window: vec![],
-            prev: None,
-            next: None,
-        }
+    /// The block height of the block that starts the window.
+    ///
+    /// If the window is empty, this is the height of the block that ends the window.
+    pub fn from(&self) -> Option<u64> {
+        self.window
+            .last()
```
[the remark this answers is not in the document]
jbearer: I believe you're right. This is a mistake

sequencer/src/options.rs
```diff
@@ -197,8 +201,12 @@ enum SequencerModule {
     /// * query: add query service endpoints
     /// * submit: add transaction submission endpoints
     Http(Module<api::options::Http>),
-    /// Alias for query-fs.
-    Query(Module<api::options::Fs>),
+    /// Alias for query-sql.
```
sveitser, line 204: Should we give @Ancient123's a heads up that the default query service backend will change to postgres after this is merged?

Ancient123: Not just me. But also other people.

jbearer: Yes. It's easy enough to just change `query` to `query-fs` to get the old behavior. But we should switch our infra over to Postgres soon

sequencer/src/api/endpoints.rs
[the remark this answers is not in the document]
sveitser: If we don't add a regression test now I think we should open an issue for it.

sveitser (approved):
LGTM

sequencer/src/api/endpoints.rs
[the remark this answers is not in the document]
jbearer: https://github.com/EspressoSystems/espresso-sequencer/issues/841

PR #803, opened 2023-12-06
Feat/t737 queue logic

philippecamacho:
Part of #737.
Implements the queue (both registration and exit) logic.
Built on top of #791.

### DIFF
--- a/contracts/test/StakeTable.t.sol
+++ b/contracts/test/StakeTable.t.sol
@@ -332,7 +332,7 @@ contract StakeTable_Test is Test {
         node.registerEpoch = 1;
 
         // Check event is emitted after calling successfully `register`
-        vm.expectEmit(false, false, false, true, address(stakeTable));
+        vm.expectEmit(true, true, true, true, address(stakeTable));
         emit Registered(
             stakeTable._hashBlsKey(blsVK), node.registerEpoch, node.stakeType, node.balance
         );


contracts/src/interfaces/AbstractStakeTable.sol
```diff
@@ -69,6 +69,14 @@ abstract contract AbstractStakeTable {
         EdOnBN254.EdOnBN254Point schnorrVK;
     }
 
+    /// @notice Struct representing a queue
+    /// @param firstAvailableEpoch epoch where the queue is located.
+    /// @param pendingRequests current size of the queue.
+    struct Queue {
+        uint64 firstAvailableEpoch;
+        uint64 pendingRequests;
+    }
```
alxiong, line 78: I prefer flatten this. 
struct helps with mental load by introducing abstraction at the cost of slight overhead (in terms of abi encoded bytes).

contracts/src/StakeTable.sol
```diff
@@ -101,27 +111,48 @@ contract StakeTable is AbstractStakeTable {
         return nodes[_hashBlsKey(blsVK)];
     }
 
+    /// @dev Helper function to avoid code duplication. Contains the core logic of a queue for
+    /// finding the next epoch.
+    /// @param queueType can be Registration or Exit.
+    /// @return Number of the epoch when a user can register/exit.
+    function _nextEpoch(QueueType queueType) private returns (uint64) {
```
alxiong, line 118: 1. remove `QueueType` abstraction, simply flatten them
2. specialize the logic into the original `nextRegisration/ExitEpoch` function

while I understand the intent is for code dedup, I still find these structs introducing unnecessary overheads.
since the logic is simple, won't be lots of duplications and IMO the code will clearer

contracts/src/StakeTable.sol
```diff
@@ -101,27 +111,48 @@ contract StakeTable is AbstractStakeTable {
...
+        if (queueType == QueueType.Registration) {
+            queue = registrationQueue;
+        } else {
+            queue = exitQueue;
+        }
+        if (queue.firstAvailableEpoch < currentEpoch() + 1) {
+            queue.firstAvailableEpoch = currentEpoch() + 1;
+            queue.pendingRequests = 1;
+        } else if (queue.pendingRequests >= MAX_CHURN_RATE) {
+            queue.firstAvailableEpoch += 1;
```
alxiong, line 129: One problem I see is that, we never increment this `firstAvaialbeEpoch` when we move on to the a new epoch. 

what happens if there's no registration for many epochs, and `currentEpoch > registrationQueue.firstAvailableEpoch`?

contracts/src/StakeTable.sol
philippecamacho: Ok, done in bc88335. Note that the `QueueType` enum has been moved to the tests. I personally prefer avoiding code duplication in the test as there is a risk to introduce error if the logic of the queue changes.

contracts/src/interfaces/AbstractStakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Yes, this _struct_ is redundant. Removed in dfb3b9b.

contracts/test/mocks/StakeTableMock.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: UNLICENSED
+
+pragma solidity ^0.8.0;
+
+import { StakeTable } from "../../src/StakeTable.sol";
+
+/// @dev A helper that wraps LightClient contract for testing
+contract StakeTableMock is StakeTable {
```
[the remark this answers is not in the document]
philippecamacho: `nextRegistrationEpoch` cannot be public as it alters the state. So I think we still need a wrapper contract.

contracts/src/StakeTable.sol
```diff
@@ -47,21 +47,31 @@ contract StakeTable is AbstractStakeTable {
...
     LightClient public lightClient;
 
+    /// Registration queue
+    uint64 firstAvailableRegistrationEpoch;
+    uint64 pendingRegistrations;
+
+    /// Exit queue
+    uint64 firstAvailableExitEpoch;
+    uint64 pendingExits;
+
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in 5cd2aa2.

alxiong (commented):
- rename `pendingReg/Exit` to `pendingReg/ExitInFirstAvailableEpoch` (I think clearer, more explicit names are always better in smart contracts)
- `nextXXEpoch` should be view only, it's simply a calculator embedding the logic of rate limiting based on the current state
- add `appendRegistration/ExitQueue() private` which is the *only* functions that will mutate the `firstAvailableXX` and `pendingXXInFirstAvail..` variables. 
  - also remember to only invoke after the `<validUntil` check

contracts/test/StakeTable.t.sol
```diff
@@ -537,3 +538,121 @@ contract StakeTable_Test is Test {
...
+        uint64 expectedPendingRequests
+    ) private {
+        (uint64 v, uint64 w) = stakeTable.getQueueParameters(queueType);
+
+        assertEq(v, expectedFirstAvailableEpoch);
+        assertEq(w, expectedPendingRequests);
+    }
+
+    /// @dev Helper function to be able to fuzz with the type of queue
+    function convertIntToQueueType(uint256 n) private view returns (SM.QueueType) {
```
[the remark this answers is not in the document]
philippecamacho: Done in 66513f0e819c10027bb674a974daadb021400d0f.

contracts/src/StakeTable.sol
philippecamacho: My impression is that we don't need to update anything in the stake table when the light client increments the epoch number: `currentEpoch` that tracks the latest epoch number from the light client is enough to compute the next registration/exit epoch.

contracts/src/StakeTable.sol
```diff
@@ -47,21 +47,31 @@ contract StakeTable is AbstractStakeTable {
...
+    uint256 public constant MAX_CHURN_RATE = 20;
+
     constructor(address _tokenAddress, address _lightClientAddress) {
         tokenAddress = _tokenAddress;
         lightClient = LightClient(_lightClientAddress);
+
+        firstAvailableRegistrationEpoch = 0;
+        pendingRegistrations = 0;
+
+        firstAvailableExitEpoch = 0;
```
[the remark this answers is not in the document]
philippecamacho: Good catch. Fixed in 580464b.

contracts/test/mocks/StakeTableMock.sol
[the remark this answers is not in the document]
philippecamacho: See https://github.com/EspressoSystems/espresso-sequencer/pull/803#pullrequestreview-1774819372.

contracts/src/StakeTable.sol
philippecamacho: See https://github.com/EspressoSystems/espresso-sequencer/pull/803#pullrequestreview-1774819372.

contracts/src/StakeTable.sol
```diff
@@ -103,25 +113,41 @@ contract StakeTable is AbstractStakeTable {
...
-        return currentEpoch() + 1;
+    function nextRegistrationEpoch() internal override returns (uint64) {
+        if (firstAvailableRegistrationEpoch < currentEpoch() + 1) {
+            firstAvailableRegistrationEpoch = currentEpoch() + 1;
+            pendingRegistrations = 1;
+        } else if (pendingRegistrations >= MAX_CHURN_RATE) {
+            firstAvailableRegistrationEpoch += 1;
+            pendingRegistrations = 1;
+        } else {
+            pendingRegistrations += 1;
```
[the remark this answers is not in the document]
philippecamacho: See https://github.com/EspressoSystems/espresso-sequencer/pull/803#pullrequestreview-1774819372.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -395,12 +395,17 @@ fn main() {
...
-            if cli.args.len() != 1 {
-                panic!("Should provide arg1=senderAddress");
+            if cli.args.len() != 2 {
+                panic!("Should provide arg1=senderAddress arg2=seed");
             }
 
+            // Use seed from cli to generate different bls keys
+            let seed_value: u8 = cli.args[1].parse::<u8>().unwrap();
+            let mut seed_bytes_length_32 = [0u8; 32];
+            seed_bytes_length_32[0] = seed_value;
```
alxiong, line 412: super nit: why don't we simply: 
`let seed = [seed_value; 32]`?

contracts/rust/src/bin/diff_test.rs
```diff
@@ -35,6 +34,7 @@ use jf_primitives::signatures::schnorr::KeyPair as SchnorrKeyPair;
 use jf_relation::{Arithmetization, Circuit, PlonkCircuit};
 use num_bigint::BigUint;
 use num_traits::Num;
+use std::str::FromStr;
```
alxiong:
```suggestion
```
also remove the empty line above (L13)

contracts/src/interfaces/AbstractStakeTable.sol
```diff
@@ -81,14 +81,14 @@ abstract contract AbstractStakeTable {
...
-    /// @notice Get the next available epoch for new registration
-    function nextRegistrationEpoch() external view virtual returns (uint64);
+    /// @notice Get the next available epoch and queue size after new registration
+    function nextRegistrationEpoch() external view virtual returns (uint64, uint64);
     /// @notice Get the number of pending registration requests in the waiting queue
     function numPendingRegistrations() external view virtual returns (uint64);
-    /// @notice Get the next available epoch for exit
-    function nextExitEpoch() external view virtual returns (uint64);
+    /// @notice Get the next available epoch for exit and queue size after new exit request
+    function nextExitEpoch() external view virtual returns (uint64, uint64);
```
alxiong, line 89: not "queue size" but "queue size *in that epoch*",

by default, ppl would assume it's overall queue size.

contracts/src/StakeTable.sol
```diff
@@ -47,21 +47,35 @@ contract StakeTable is AbstractStakeTable {
...
     LightClient public lightClient;
 
-    constructor(address _tokenAddress, address _lightClientAddress) {
+    /// Registration queue
+    uint64 public firstAvailableRegistrationEpoch;
+    uint64 public pendingRegistrationsInFirstAvailableRegistrationEpoch;
+
+    /// Exit queue
+    uint64 public firstAvailableExitEpoch;
+    uint64 public pendingExitsInFirstAvailableExitEpoch;
```
alxiong, line 62:
```suggestion
    /// @notice the first available epoch for registration, please use `nextRegistrationEpoch()` to get the correct epoch
    uint64 private _firstAvailableRegistrationEpoch;
    /// @notice number of pending registrations in the `firstAvailableRegistrationEpoch` (not the total pending queue size!)
    uint64 private _numPendingRegistrations;

    /// @notice the first available epoch for exit, please use `nextExitEpoch()` to get the correct epoch
    uint64 private _firstAvailableExitEpoch;
    /// @notice number of pending exits in the `firstAvailableExitEpoch` (not the total pending queue size!)
    uint64 private _numPendingExits;
```

contracts/src/StakeTable.sol
```diff
@@ -102,26 +116,68 @@ contract StakeTable is AbstractStakeTable {
     }
 
     /// @notice Get the next available epoch for new registration.
-    /// @return Number of the epoch when the user can register.
-    function nextRegistrationEpoch() external view override returns (uint64) {
-        // TODO implement queue logic
-        return currentEpoch() + 1;
+    /// @return Number of the epoch when the user can register and size of the queue in that epoch.
+    function nextRegistrationEpoch() external view override returns (uint64, uint64) {
```
alxiong, line 120: imo, we shouldn't return "the size after the user register" cuz this function should be totally oblivious of whether or when the caller will register, it could be people are just calling this function for fun/other purposes.

it should return the number of people already in the line!

contracts/src/StakeTable.sol
```diff
@@ -281,10 +338,12 @@ contract StakeTable is AbstractStakeTable {
         }
 
         // Prepare the node to exit.
-        uint64 nextEpoch = this.nextExitEpoch();
-        nodes[key].exitEpoch = nextEpoch;
+        (uint64 epoch, uint64 queueSize) = this.nextExitEpoch();
```
alxiong, line 341: `epoch` -> `exitEpoch`

contracts/test/StakeTable.t.sol
```diff
@@ -110,9 +133,9 @@ contract StakeTable_Test is Test {
             stakeTableAmountComm: 0,
             threshold: 0
         });
-        lightClientContract = new LightClientTest(genesis,10);
+        lightClientContract = new LightClientTest(genesis, 10);
         address lightClientAddress = address(lightClientContract);
-        stakeTable = new S(address(token),lightClientAddress);
+        stakeTable = new S(address(token), lightClientAddress, 10);
```
alxiong: let's remove L137, no point declaring a new variable here.

contracts/test/StakeTable.t.sol
```diff
@@ -89,12 +104,20 @@ contract StakeTable_Test is Test {
...
+            assertFalse(res);
+        }
 
         return (blsVK, depositAmount);
     }
 
+    /// @dev  Helper function to simulate a successful registration
+    function runSuccessfulRegistration() private returns (BN254.G2Point memory, uint256) {
+        return registerWithSeed(exampleTokenCreator, 34, 10, false);
+    }
```
alxiong, line 113: I saw almost all subsequent functions starts with this, why don't we add this in `setUp()` so that we don't have to manually call this in each test case?

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+    function asQueueType(uint256 n) private view returns (QueueType) {
+        QueueType queueType;
+        uint256 typeOfQueueInt = bound(n, 0, 1);
+        if (typeOfQueueInt == 0) {
+            queueType = QueueType.Registration;
+        } else if (typeOfQueueInt == 1) {
+            queueType = QueueType.Exit;
+        } else {
+            revert("Queue type not supported");
+        }
```
alxiong, line 595: we will never hit this, as `typeOfQueueInt` is already bound to be `[0,1]`

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+                uint64 currentEpoch = lightClientContract.currentEpoch();
+                uint64 nextEpoch = currentEpoch + 1;
+                lightClientContract.setCurrentEpoch(nextEpoch);
+                assertEq(stakeTable.currentEpoch(), nextEpoch);
+            }
+
+            // Global invariants
+            assertLe(stakeTable.numPendingRegistrations(), stakeTable.maxChurnRate());
+            assertLe(stakeTable.numPendingExits(), stakeTable.maxChurnRate());
+            assertLe(numExits, numRegistrations);
```
alxiong, line 711: why is this an invariant?

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+                // Registrations
+                bool res = handleRegistrations(i, rands, registeredKeys, isKeyActive);
+                if (res) {
+                    numRegistrations++;
+                }
+            } else if (ev == 1) {
+                // Exits
+                if (numRegistrations == 0) {
+                    break;
+                }
```
alxiong, line 666: break or continue?

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+            } else {
+                // Advance epoch
+                // ev == 2
+                uint64 currentEpoch = lightClientContract.currentEpoch();
+                uint64 nextEpoch = currentEpoch + 1;
+                lightClientContract.setCurrentEpoch(nextEpoch);
+                assertEq(stakeTable.currentEpoch(), nextEpoch);
+            }
+
+            // Global invariants
```
alxiong, line 708: I think another one would be, giving `numRegistration` and `Churn_rate`, we should calculate the first available epoch and pending queue size, we can check if contract also computed that correctly

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+                if (canExit) {
+                    vm.prank(sender);
+                    bool res = stakeTable.requestExit(blsVK);
+                    assertTrue(res);
+                    numExits++;
+                    exitRequestSuccessful = true;
+
+                    // Invariants specific to a successful exit
+                    assertGe(stakeTable.firstAvailableExitEpoch(), stakeTable.currentEpoch() + 1);
+                    assertGe(stakeTable.numPendingExits(), 1);
```
alxiong, line 692: I think this invariant should be, we first record down before exit , this exit queue size, and then now, make sure this queue size increment by 1

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
...
+        }
+
+        return queueType;
+    }
+
+    uint256 private constant ARRAY_SIZE = 20;
+
+    /// Helper function to handle registrations in testFuzz_SequencesOfEvents
+    /// This function was extracted to make sol-lint happy by reducing cyclotomic complexity
+    function handleRegistrations(
```
alxiong, line 604: it's kind of asymmetric to see these logic being separated into a function while there's no `handleExits` which is also quite complex.

I'd say be consistent, either create both functions or none.

alxiong (commented):
tbh, I don't see any simplifying benefits of `enum QueueType`... I'd propose we remove it. (and all associated helper functions)

contracts/rust/src/bin/diff_test.rs
philippecamacho: Done in d7bc82e.

contracts/rust/src/bin/diff_test.rs
philippecamacho: Looks like we need this import. Otherwise we get the following compilation error:

```
 just sol-test
cargo build --bin diff-test --release
   Compiling hotshot_contract v0.1.0 (/home/leloup/repositories/espressosys/espresso-sequencer/contracts/rust)
error[E0405]: cannot find trait `FromStr` in this scope
   --> contracts/rust/src/bin/diff_test.rs:574:6
    |
574 | impl FromStr for ParsedTranscript {
    |      ^^^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
1   + use ark_std::str::FromStr;
    |
1   + use core::str::FromStr;
    |
1   + use std::str::FromStr;
    |

error[E0405]: cannot find trait `FromStr` in this scope
   --> contracts/rust/src/bin/diff_test.rs:629:6
    |
629 | impl FromStr for ParsedVerifyingKey {
    |      ^^^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
1   + use ark_std::str::FromStr;
    |
1   + use core::str::FromStr;
    |
1   + use std::str::FromStr;
....
```

The empty line has been removed in 202638475ff911f53406be798c63a07f744a9ad5

contracts/src/interfaces/AbstractStakeTable.sol
philippecamacho: Yes, done in e0c43fd.

contracts/src/StakeTable.sol
philippecamacho: Done in e76d0d5.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, done in 06b93b3 .

contracts/src/StakeTable.sol
philippecamacho: Yes, done in f30159c.

contracts/test/StakeTable.t.sol
philippecamacho: It makes each test harder to understand I think, especially because some tests would not need to use `runSuccessfulRegistration` (e.g. `test_RevertWhen_TransferFailed`).

philippecamacho:
> tbh, I don't see any simplifying benefits of enum QueueType... I'd propose we remove it. (and all associated helper functions)

Yes, this is not needed anymore. Deleted in 523acbe.

contracts/test/StakeTable.t.sol
```diff
@@ -530,4 +553,162 @@ contract StakeTable_Test is Test {
 
         assertEq(abi.encode(node), abi.encode(nullNode));
     }
+
+    // Queue logic
+
+    function getQueueParameters(QueueType queueType) public view returns (uint64, uint64) {
```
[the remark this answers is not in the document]
philippecamacho: This function is not needed anymore. Deleted in 8038961f5e4df59b51bcaab10f3581eba25f8e89.

contracts/test/StakeTable.t.sol
philippecamacho: The corresponding function has been deleted in 8038961f5e4df59b51bcaab10f3581eba25f8e89.

contracts/test/StakeTable.t.sol
philippecamacho: Because one needs to register before exiting and thus the number of exits is always lower or equal to the number of registrations.

contracts/test/StakeTable.t.sol
philippecamacho: Good catch! Fixed in 674c13e.

contracts/test/StakeTable.t.sol
philippecamacho: Finding such a general formula seems hard because epochs are not necessarily filled up with registrations as some epochs can be arbitrarily skipped.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, done in 870adea.

contracts/test/StakeTable.t.sol
philippecamacho: This is not always the case. For example we may have the queue size equal to 4 during epoch 10. Then nothing happens during epochs 11,12,13. The next queue size will be 1 instead of 5 (assuming `max_churn_rate == 10` for example).

contracts/rust/src/bin/diff_test.rs
alxiong: I see! I forgot that I have moved those to a different library file in my own PR i think. let's leave it then, I will deal with merge conflict in my PR later.

contracts/test/StakeTable.t.sol
alxiong: i see, but this invariant is more of an artifact of how we write our tests, not exactly how we expect the Queue to uphold in general.
I'd prefer to remove it as it's a little confusing.

contracts/test/StakeTable.t.sol
alxiong: why? just having some prior successful registration is perfectly fine? it's equivalent of initializing a ledger with a non-empty stake table to start with (before the logic of each tests start) 

`test_RevertWhen_TransferFailed` shouldn't be affected, right? it'll still pass even with the addition in `setUp()`, correct?

contracts/test/StakeTable.t.sol
alxiong: I see, what I describe may be a Registration-specific invariant, not a global one.
when I register, I first query the `_firstAvailable` and `pendingXX` variable, and do the calculation in the test, and then compare with what I actually got after I successfully register. 

- also, can we change visibility of `_firstAvailable` etc from `public` to `internal` instead?

contracts/test/StakeTable.t.sol
philippecamacho: Ok, invariant removed in ed12203.

contracts/test/StakeTable.t.sol
philippecamacho: Here is  a way to make the "formula" work: The idea is to introduce a fuzzing variable `skipEpoch`. When `skipEpoch` is false, we don't skip the epochs and thus the queue of each epochs are filled up. This allows us to check the relationship between `numRegistration` , max churn rate and `pendingRegistrations` (same for next registration epoch). See 93264d8.

philippecamacho: > also, can we change visibility of _firstAvailable etc from public to internal instead?

We then would need a getter funcion?

contracts/test/StakeTable.t.sol
philippecamacho: As mentioned [above](https://github.com/EspressoSystems/espresso-sequencer/pull/803#discussion_r1427899546), with the `skipEpoch` fuzzing variable we can implement your idea.

contracts/test/StakeTable.t.sol
alxiong: i think we can calculate w/o `skipEpoch` flag, 
simply by knowing the `currentEpoch`, `_firstAvailable` and `pendingXX` we should be able to compute the expected epoch for the new registration and the post-registration `pendingXX` value. 

there are two parts:
- logic of `nextRegistrationEpoch()`: given `currentEpoch`, `_firstAvail` and `_pendingXX`, does it return the expected value
- update of `_firstAvail` and `_pendingXX` after registration assuming a correct `nextRegistrationEpoch()`

To test the first, if we spell out the formula, it would be verbatim of the actual function, thus no need. Maybe we could add another test function with a few hardcoded instances (esp edge cases such as pending is just full, or the `firstAvail` is behind `currentEpoch`, and see if the output is expected

To test the second, we simply record down the output of `nextRegistrationEpoch()` before calling `register`, then inspect the variable again to see if they are properly updated.

^^ does this make sense?

philippecamacho: I am ok with implementing the second check. Now I think checking what happens when no skipping happens captures part of the behaviour of the queue.  Especially because it could happen in practice.

philippecamacho: > I am ok with implementing the second check

Done in ec56a87 for the registration queue. Is that what you had in mind?

contracts/test/StakeTable.t.sol
philippecamacho: The test test_`RevertWhen_TransferFailed` fails on this condition:

```
assertEq(ERC20(token).balanceOf(exampleTokenCreator), INITIAL_BALANCE);
```

It is easy to fix (by replacing INITIAL_BALANCE with `INITIAL_BALANCE-10` but it would be harder to read/understand I think.  

> it's equivalent of initializing a ledger with a non-empty stake table to start with (before the logic of each tests start)

What I don't like too much about this is that the test needs to know about the setup. For example you cannot reuse the same seed to generate the wallet in `test_Registration_succeeds`, otherwise the registration will fail.
It feels simpler to start with a blank state and prepare this state inside the test function. I would prioritize being explicit in the test function over minimizing code duplication in this case.

alxiong: okay, sure. 

>What I don't like too much about this is that the test needs to know about the setup.

altho for this point, you are supposed to know what happens to `setUp`, the first thing you read each test is to read setUp. so I don't think that's a problem.
That said, it's not too much code duplication. i don't have strong opinion here.

alxiong (approved):

PR #808, opened 2023-12-07
Feat/update hotshot

jbearer:
Areas with substantial changes:
* lib.rs -- mostly related to HotShot instantiation (`SeqTypes`, `Node`) and initialization (`init_node`). Review this, but these changes are mostly of the "if it compiles and passes tests it is probably correct" variety
* block.rs -- biggest change to conform to the new header/payload model. Needs review
* block2.rs -- mostly just a placeholder implementation of `BlockPayload`
* `api/`, `hotshot-commitment.rs` -- mostly superficial renaming and removing of type parameters


sequencer/src/block.rs
```diff
@@ -81,45 +69,147 @@ pub struct Metadata {
...
+
+    fn tag() -> String {
+        "BLOCK".into()
+    }
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, Hash, PartialEq, Eq)]
-pub struct Block {
-    metadata: Metadata,
+impl Committable for NMTRoot {
```
ggutoski, line 95: Do we really need to re-commit the NMT root?

sequencer/src/block.rs
```diff
@@ -81,45 +69,147 @@ pub struct Metadata {
...
+    fn metadata(&self) -> NMTRoot {
+        self.transactions_root
+    }
+}
+
+impl Header {
+    fn from_parent(
+        payload_commitment: VidCommitment,
+        transactions_root: NMTRoot,
+        parent_height: Option<u64>,
```
ggutoski, line 151: Nit: sucks that we need `Option` just to handle the genesis block. Couldn't we just construct the genesis block directly without calling `from_parent`, thus avoiding the need for `Option` here?

sequencer/Cargo.toml
```diff
@@ -62,3 +61,6 @@ typenum = { version = "1.15.0", default-features = false, features = [
     "no_std",
 ] }
 url = "2.3"
+
+# Dependencies for feature `testing`
```
ggutoski, line 65: ew

ggutoski (commented):
- block2.rs: lgtm
- block.rs: I don't know what I'm looking at but lgtm

I'll let others approve.

sequencer/src/block.rs
```diff
@@ -81,45 +69,147 @@ pub struct Metadata {
...
+        let height = match parent_height {
+            Some(height) => height + 1,
+            None => 0,
+        };
+
+        // Sample a timestamp, ensuring that it does not decrease.
+        let mut timestamp = OffsetDateTime::now_utc().unix_timestamp() as u64;
+        if timestamp < parent_timestamp {
+            tracing::warn!(
+                "Espresso timestamp {timestamp} behind parent {}",
```
sveitser, line 180: We are emitting a warning here but not giving instructions what to check or do about it. Would it make sense to ask the operator to check if their servers clock is synced and if yes, ignore the warning?

sveitser:
We can set `TestMetadata.overall_safety_properties.num_sucessful_views` to a low enough number like 10.

sequencer/src/block.rs
jbearer: I guess we could just fixed_size_bytes it in the header commitment. Not sure if there's a good conversion to bytes for that one. Anyways we should be getting rid of the NMT completely soon 🤞 so I'm not too worried about it

sequencer/src/block.rs
jbearer: Sure. I wanted to avoid duplicating the logic for getting the L1 snapshot, since it's a bit complex. But all the rest of the complexity (computing the height, adjusting the timestamp) doesn't apply to the genesis block. So there is probably a better way to structure this.

jbearer: Realized this is a bug anyways. Everyone needs to have the same genesis header, so this needs to be deterministic. I think it is only down to luck and the 1-second timestamp resolution that the tests pass

ggutoski: Presumably we'll eventually hard-code the genesis block or read it from disk, no? I guess that could get tricky if we have wall-clock sync constraints.

PR #815, opened 2023-12-08
Fix local demo

sveitser:
- Cargo update
- Fix: https://github.com/EspressoSystems/HotShot/releases/tag/0.5.2


jbearer (approved):
Much cleaner!

PR #818, opened 2023-12-09
Enable status API without full query API

### LINKED ISSUE #792: Enable status API without full query API
Currently, we have these optional modules related to APIs: `http` and `query`. The former just runs a minimal HTTP server with healthcheck and version endpoints, nothing more. The latter runs the full HotShot query service with status API and the full archival API. Something in between that just runs the status API but not the full archival API would be useful for many node operators, who want the former for monitoring but don't want to bother setting up persistence for the latter.

We should create a new optional module `status`. If `status` is enabled but `query` is not, we run just the status API module using the minimal status data source, with no persistent storage. If `status` _and_ `query` are enabled, we run the status and availability APIs, both sharing the same archival data source (file system or SQL).

Depends on https://github.com/EspressoSystems/hotshot-query-service/issues/285

jbearer:
Closes #792

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -181,7 +181,7 @@ mod test {
         options.serve(init_handle).await.unwrap();
         client.connect(None).await;
 
-        // The status API is will tested in the query service repo. Here we are just smoke testing
+        // The status API is well tested in the query service repo. Here we are just smoke testing
         // that we set it up correctly. Wait for a (non-genesis) block to be sequenced and then
         // check the success rate metrics.
         tracing::info!(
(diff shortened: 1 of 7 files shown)


sveitser (approved):
LGTM

sequencer/src/api.rs
```diff
@@ -137,6 +139,76 @@ mod test {
...
+        };
+
+        let mut options = Options::from(options::Http { port }).status(Default::default());
+        if let Some(query) = query_opt {
+            options = options.query_fs(query);
+        }
+        options.serve(init_handle).await.unwrap();
+        client.connect(None).await;
+
+        // The status API is will tested in the query service repo. Here we are just smoke testing
```
sveitser, line 184:
```suggestion
        // The status API is well tested in the query service repo. Here we are just smoke testing
```

PR #829, opened 2023-12-11
Build and push a prebuilt geth docker

Ancient123:


jbearer (approved):
I don't think this has any contracts deployed. Is that alright?

Ancient123:
> I don't think this has any contracts deployed. Is that alright?

I thought that was what the genesis file being imported was for? This is just replicating the `docker-compose.yml` that is in this repo.

jbearer:
When we run the demo locally we run `commitment-task --deploy` which deploys the contract at runtime. I believe the genesis file only has funded accounts, but no deployed contracts

Ancient123:
or is that handled by the commitment task running like `commitment-task --deploy` ?

PR #846, opened 2023-12-13
feat: Support tx proofs for truncated tx payloads and negative-length txs

### LINKED ISSUE #816: Support tx proofs for truncated tx payloads
A tx payload might be truncated if the block payload is too small to hold everything declared in the tx table. It should still be possible to create and verify a `TxInclusionProof` for such a tx.

ggutoski:
close #816 

- Support *any* block payload for which the tx table len fits inside the payload.
  - In other words: support a completely arbitrary block payload except the first 4 bytes are a number small enough that the tx table fits in the payload.
- Update jellyfish dependency
- I confess that these test cases are uninspired. Suggestions welcome. I suggest we do not aim for completely exhaustive tests in this PR, as we have other priorities.


ggutoski:
CI failing. I can fix some failures locally with various fmt commands but I don't know how to fix `contract-bindings` failure.

sequencer/src/block2.rs
```diff
@@ -161,6 +161,8 @@ impl QueryablePayload for BlockPayload {
         // This correctness proof requires a range of its own, which we read into `tx_table_range_proof_[start|end]`.
         //
         // Edge case--the first transaction: tx payload range `start` is implicitly 0 and we do not include this item in the correctness proof.
+        //
+        // TODO why isn't cargo fmt wrapping these comments?
```
jbearer, line 165: It doesn't do that. I use an editor extension for this.

sequencer/src/block2.rs
```diff
@@ -506,53 +522,46 @@ mod test {
...
-            vec![vec![0, 1, 2, 3, 4], vec![5, 6, 7, 8, 9, 10, 11, 12], vec![]],
-            // 1 nonempty tx
-            vec![vec![0, 1, 2, 3, 4]],
-            // 1 empty tx
-            vec![vec![]],
-            // zero txs
-            vec![],
+            vec![5, 13, 21],        // 3 non-empty txs
+            vec![0, 8, 16],         // 1 empty tx at the beginning
+            vec![5, 5, 13],         // 1 empty tx in the middle
```
jbearer, line 533:
```suggestion
            vec![5, 0, 13],         // 1 empty tx in the middle
```
?

sequencer/src/block2.rs
```diff
@@ -506,53 +522,46 @@ mod test {
...
+            vec![5, 13, 13],        // 1 empty tx at the end
```
jbearer, line 534:
```suggestion
            vec![5, 13, 0],        // 1 empty tx at the end
```

sequencer/src/block2.rs
```diff
@@ -614,4 +622,153 @@ mod test {
...
+            tx_table.extend(TxTableEntry::from_usize(*entry).to_bytes());
+        }
+        tx_table
+    }
+
+    fn random_tx_payloads_flat<R>(entries: &[usize], rng: &mut R) -> Vec<u8>
+    where
+        R: RngCore,
+    {
+        random_tx_payloads_flat_truncated(entries, 0, rng)
```
jbearer, line 693: By setting `max_tx_payloads_bytes_len` to 0 doesn't the result always get truncated to length 0?

sequencer/src/block2.rs
```diff
@@ -614,4 +622,153 @@ mod test {
...
+        }
+        assert_eq!(result.len(), entries.len());
+        result
+    }
+
+    fn random_payload<R>(entries: &[usize], rng: &mut R) -> Vec<u8>
+    where
+        R: RngCore,
+    {
+        random_payload_truncated(entries, 0, rng)
```
jbearer, line 736: Same here, why 0?

jbearer (commented):
Few questions but the testing framework here is great. We can easily extend this with more edge cases as we go.

sequencer/src/block2.rs
ggutoski: Because I'm stupid and changed the function without changing its call sites. I'll fix that.

sequencer/src/block2.rs
ggutoski: hmm. It does in jellyfish. But perhaps it's some weird interaction between that repo and one of the many vscode extensions I have.

sequencer/src/block2.rs
ggutoski: These vecs are verbatim tx table entries. So `5,5` means that the tx starts and ends at index 5. I did it this way to allow for negative tx lengths in the tx table. But perhaps it is a bit confusing. Maybe it's better to have 2 different helpers: one that takes *lengths* as input and another that takes *table entries* as input.

ggutoski: added `entries_from_lengths` in https://github.com/EspressoSystems/espresso-sequencer/pull/846/commits/ab2a58b8e7bc9ac779722b3b28f3c108ac6b60b5

sequencer/src/block2.rs
ggutoski: addressed in https://github.com/EspressoSystems/espresso-sequencer/pull/846#discussion_r1425946213

sequencer/src/block2.rs
ggutoski: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/846/commits/ab2a58b8e7bc9ac779722b3b28f3c108ac6b60b5

sequencer/src/block2.rs
ggutoski: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/846/commits/ab2a58b8e7bc9ac779722b3b28f3c108ac6b60b5

ggutoski:
Seems that the CI break is caused by failure to fetch git submodules. @sveitser maybe you could help here?

sequencer/src/block2.rs
jbearer: Oh yeah I figured this out later in my reading and forgot to come back to this comment

jbearer (approved):

sequencer/src/block2.rs
sveitser: Jellyfish uses rustfmt from nightly and has a rustformat config file.

https://github.com/EspressoSystems/jellyfish/blob/main/flake.nix#L104
https://github.com/EspressoSystems/jellyfish/blob/main/rustfmt.toml

sveitser:
The PR
- #742

updated solc via updating the flake. This seems to have changed the generated bytecode which is embedded in the bindings. I think the PR was merged without running the CI so it broke the contracts CI job on main.

https://github.com/EspressoSystems/espresso-sequencer/actions/workflows/contracts.yml?query=branch%3Amain

sveitser:
@ggutoski https://github.com/EspressoSystems/espresso-sequencer/pull/853 should fix the CI.

PR #864, opened 2023-12-15
fix port forwarding in docker_compose

jonas089:
Port forwarding seemed incorrect, all nodes were mapped to the same port.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -82,7 +82,7 @@ services:
   sequencer1:
     image: ghcr.io/espressosystems/espresso-sequencer/sequencer:main
     ports:
-      - "$ESPRESSO_SEQUENCER1_API_PORT:$ESPRESSO_SEQUENCER_API_PORT"
+      - "$ESPRESSO_SEQUENCER1_API_PORT:$ESPRESSO_SEQUENCER1_API_PORT"
     # Run the availability API only, using the local file system for persistence
     command: sequencer -- http -- query-fs
     environment:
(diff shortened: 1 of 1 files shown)


jonas089:
On a separate note:

I have noticed a "connection refused" error when trying to reach the WebServer @ port 40000. 
Could someone take a minute to explain to me in this thread how the different APIs work? Is it only possible to query a block / transaction from the Sequencer API or is it also possible to query a block / transaction through the HotShot API directly? If so, how could I query HotShot directly when running the docker image(s)?

jbearer:
Those are the ports internal to each docker container, so it is fine that they are all the same. The first port in each pair is the host port they are mapped to, which are all different.

I guess it would be possible to get a block directly from the HotShot CDN, but it's not really recommended or intended, as the CDN is meant to be internal to consensus: it's only real role is facilitating the exchange of messages between consensus nodes. So we don't expose the CDN port in this demo. It is better to just get the block from the query service, which is meant to be public facing.

jonas089:
But how can I run a HotShot node in isolation, submit a transaction to that HotShot node and query that Transaction? Also, is the transaction format the same for HotShot as is for the Sequencer e.g. {vm, payload}?

Would be great if you could help me better understand the differences between a HotShot node and a Sequencer Node and how they communicate with one another

jonas089:
https://github.com/EspressoSystems/HotShot/blob/main/crates/web_server/api.toml

I am referring to this HotShot API where you can POST a transaction and query a transaction. The structure of a transaction and the context are unclear to me

PR #869, opened 2023-12-15
Fix Header::new and add tests

jbearer:
I noticed a bug where we were not preventing the l1_head or l1_finalized block from decreasing, due to L1 clients between nodes being out of sync.

Added a bunch of new tests for the monotonicity properties of various header fields. Note that the test `test_header_invariants` (more of an integration test) is not very stringent yet, because all the nodes are using the same (fake) L1 client. However, once HotShot's stateful header stuff is complete, it should be possible to give each node its own L1 client, even within the same process, and then this test will become much more useful. In the meantime, we now have good unit test coverage.

Added a utility program `verify-headers` to check the consistency of headers in an existing deployment. Our current staging and Blockdaemon deployments are all good 😄


sequencer/src/bin/verify-headers.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    let opt = Arc::new(Options::parse());
+    let seq = Arc::new(SequencerClient::new(opt.url.clone()));
+    seq.connect(None).await;
+
+    let block_height: usize = seq.get("status/latest_block_height").send().await.unwrap();
+    let from = opt.from.unwrap_or(0);
+    let to = max(opt.to.unwrap_or(block_height), from + 1);
+
+    let ok = join_all(
```
sveitser, line 177: I think it would be nice if this logged something about the progress on INFO level or had something like a progress bar. Currently there's no way to distinguish the long running working output from the one where the URL is wrong (both log nothing).

sequencer/src/block.rs
```diff
@@ -454,3 +492,209 @@ mod reference {
...
+    struct HeaderInfo {
+        timestamp: u64,
+        l1_head: u64,
+        l1_finalized: Option<L1BlockInfo>,
+    }
+
+    #[derive(Debug, Default)]
+    #[must_use]
+    struct TestCase {
+        parent: HeaderInfo,
```
sveitser, line 512: Not super important but I think the test cases would be shorter easier to read (less nesting and `..Default::default()`) if the struct here was flattened instead of nested.

sveitser (approved):
LGTM

sequencer/src/bin/verify-headers.rs
jbearer: I did something slightly different ([919f753](https://github.com/EspressoSystems/espresso-sequencer/pull/869/commits/919f7534f69516ac5a50cfd431c1316c348d56e6)). I got rid of the `seq.connect()` call, which blocked until a successful connection, so it now panics if the URL is wrong. After that, it will print out an INFO level log telling you the block range it is going to process, and if anything goes wrong it logs something at WARN level. The per-header logs remain at DEBUG level so it's not too noisy and it's easy to see if there are errors.

PR #870, opened 2023-12-15
feat: Support block payloads that are too small to hold the entire tx table

### LINKED ISSUE #817: Support block payloads that are too small to hold the entire tx table
If the tx table declares a large length (eg. 4 billion) then the block payload might end before the tx table is complete. In that case every tx in the block has zero length, which is fine. But if anyone calls `QueryablePayload::iter()` then they'll get an iterator over 4 billion empty txs, which is bad.

The solution is to define the tx table length as `min(declared length, number of words that can fit into the payload)`.

ggutoski:
close #817 

- Includes support for extremely small block payloads (< 4 bytes) that are too small to hold the tx table length.
- See `TODO(817)` for comments about overflow for very large blocks.
- Refactor test infra just a bit.


ggutoski:
[CI failure](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/7227421249/job/19695047362?pr=870) seems to be just a timeout initializing the nix environment:
```
copying path '/nix/store/skscsvd5nsga8gm9x7yjryrqb1925jyq-source' from 'https://espresso-systems-private.cachix.org/'...
warning: error: unable to download 'https://espresso-systems-private.cachix.org/acccsbfdyyapqz9196z0kq6qnq9xkmp8.narinfo': HTTP error 530; retrying in 285 ms
warning: error: unable to download 'https://espresso-systems-private.cachix.org/acccsbfdyyapqz9196z0kq6qnq9xkmp8.narinfo': HTTP error 530; retrying in 509 ms
warning: error: unable to download 'https://espresso-systems-private.cachix.org/acccsbfdyyapqz9196z0kq6qnq9xkmp8.narinfo': HTTP error 530; retrying in 1081 ms
warning: error: unable to download 'https://espresso-systems-private.cachix.org/acccsbfdyyapqz9196z0kq6qnq9xkmp8.narinfo': HTTP error 530; retrying in 2038 ms
error: unable to download 'https://espresso-systems-private.cachix.org/acccsbfdyyapqz9196z0kq6qnq9xkmp8.narinfo': HTTP error 5[30](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/7227421249/job/19695047362?pr=870#step:8:31)
Error: Process completed with exit code 1.
```

sequencer/src/block2.rs
```diff
@@ -99,7 +111,7 @@ impl BlockPayload {
     fn get_tx_table_len_proof(&self, vid: &impl PayloadProver<RangeProof>) -> Option<&RangeProof> {
         self.tx_table_len_proof
             .get_or_init(|| {
-                vid.payload_proof(&self.payload, 0..TxTableEntry::byte_len())
+                vid.payload_proof(&self.payload, self.tx_table_len_range())
```
jbearer, line 114: I think we need to change the proof verification to require that if this range is less than `TxTableEntry::byte_len`, then the payload length equals the range length.

ggutoski: By "proof verification" I guess you mean `TxInclusionProof::verify()`, right? Perhaps you're referring to the following line in `TxInclusionProof::verify()`?

https://github.com/EspressoSystems/espresso-sequencer/blob/e95618c607ce3228f3b4bad60abe882f78741693/sequencer/src/block2.rs#L323-L323

If the payload is that small then it must contain zero txs, and any purported tx inclusion proof will fail anyway (as it should). We do have some tests which confirm `transaction_with_proof()` fails to create a proof for such small blocks.

But we are missing a test for maliciously-generated `TxInclusionProof`s. I do think such a proof would be correctly rejected by current code, even without your suggestion. But I probably should add a test for this anyway.

jbearer: Ah, you're right...there is no valid inclusion proof in this case. And it does seem that there is nothing a malicious prover can do to convince a verifier otherwise, since the verifier can always see that the length of the payload is too short. Still a good test to add indeed

jbearer (approved):

PR #874, opened 2023-12-18
Parameterize the WS polling interval

### LINKED ISSUE #873: Parameterize Webserver polling interval
This allows us to decrease the polling frequency for nodes with high latency.

rob-maron:
Closes #873 

Allows us to increase the polling interval for nodes with higher latency.

### DIFF
--- a/sequencer/src/options.rs
+++ b/sequencer/src/options.rs
@@ -64,6 +64,11 @@ pub struct Options {
     #[clap(short, long, env = "ESPRESSO_SEQUENCER_CONFIG_PATH")]
     pub config_path: Option<PathBuf>,
 
+    /// The amount of time to wait between each request to the HotShot 
+    /// consensus or DA web servers during polling, in milliseconds.
+    #[clap(short, long, env = "ESPRESSO_SEQUENCER_WEBSERVER_POLL_INTERVAL_MS", default_value = "100")]
+    pub webserver_poll_interval_ms: u64,
+
     /// Add optional modules to the service.
     ///
     /// Modules are added by specifying the name of the module followed by it's arguments, as in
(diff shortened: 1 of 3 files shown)


sequencer/src/options.rs
```diff
@@ -64,6 +64,16 @@ pub struct Options {
...
 
+    /// The amount of time to wait between each request to the HotShot
+    /// consensus or DA web servers during polling, in milliseconds.
+    #[clap(
+        short,
+        long,
+        env = "ESPRESSO_SEQUENCER_WEBSERVER_POLL_INTERVAL_MS",
+        default_value = "100"
+    )]
+    pub webserver_poll_interval_ms: u64,
```
jbearer, line 75: Nit: I prefer to make this `Duration`, like [in the orchestrator](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/bin/orchestrator.rs#L25-L26), parsed  [like this](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/bin/orchestrator.rs#L116-L122). This gives us type safety in the code and lets users/terraform write things like `100ms` or `1s`, which can be easier to read and understand

jbearer (approved):

sequencer/src/options.rs
```diff
@@ -87,6 +102,19 @@ impl Options {
     }
 }
 
+#[derive(Clone, Debug, Snafu)]
+struct ParseDurationError {
+    reason: String,
+}
+
+fn parse_duration(s: &str) -> Result<Duration, ParseDurationError> {
```
rob-maron, line 110: cc @jbearer - is there a better place I can put this?

jbearer: I think this is a fine place for it. You could make it `pub` and import it from the orchestrator, so it's not duplicated. But not a big deal

jbearer (approved):

PR #878, opened 2023-12-19
Refactor/t805 stake table improvements

### LINKED ISSUE #805: Stake table contract improvements
- [x] [Implement ParsedG2Point struct](https://github.com/EspressoSystems/espresso-sequencer/pull/769#discussion_r1401536761) (done in [`solidity-bn254` crate](https://github.com/EspressoSystems/solidity-bn254/blob/main/diff-test/src/lib.rs#L95))
- [ ] emit `event EpochChanged` so that the prover service can subscribe the even stream and know when to pull new stake table snapshot from L1
- [x] ~Update to the latest version of BN254.sol~ (already cover by another issue)
- [x] ~Delete HotShot.sol contract which is obsolete~ (Done as part of #775)
- [ ] Use testFuzz to ensure the BLS key is validated correctly. See https://github.com/EspressoSystems/espresso-sequencer/pull/769#pullrequestreview-1755093893
- [ ] Add invariant test relative to the balances. See https://github.com/EspressoSystems/espresso-sequencer/pull/769#discussion_r1409223555

philippecamacho:
Closes #805. 

A couple of comments:

* `Use testFuzz to ensure the BLS key is validated correctly. See https://github.com/EspressoSystems/espresso-sequencer/pull/769#pullrequestreview-1755093893`: as mentioned in the comments, we cannot validate a point in G2 (too expensive).

* `Add invariant test relative to the balances. See https://github.com/EspressoSystems/espresso-sequencer/pull/769#discussion_r1409223555*`: invariant testing seems to be a powerful tool indeed. ~~I thought we could use it for our tests about queues but so far only the global invariants are captured (see function `invariant_Queue`). Will revisit this tomorrow.~~ Indeed it is possible to also capture the invariants relative to a function like `register`. However I could not see any clean way to migrate the invariants defined when no epoch is skipped. Maybe we could drop these invariants and get rid of `testFuzz_SequencesOfEvents`? WDYT @alxiong ?


alxiong (commented):
imo, the handler should we written slightly differently.

- all 4 APIs should add necessary setup to ensure success call (in contrast to our original `registerWithSeed(bool expectRevert)`
  - use a dynamic list of `users` (can create using seed that deries from `users.length` so that we won't generate the same wallet or user address) @philippecamacho 
  - `register` should start with `genClientWallet` each time, and credit some token before registering to ensure success, and upon registration add the bls key to a list
  - `deposit` should take a random index and get the corresponding bls key, credit with more token before depositing
  - `requestExit` should take a random index and get the bls key and submit, then store the assigned exit epoch in a data struct (most likely a mapping `key => exitEpoch` + a list of requested keys)
  - `withdraw` should only withdraw on fully exited 
  - `advanceEpoch() public` leave as it is

^^ this also avoids the conditional invariant anti-pattern [mentioned in forge doc](https://book.getfoundry.sh/forge/invariant-testing#conditional-invariants).

I think we can remove `SequenceOfEvent` test with this invariant test.

Besides, some tech-debt, we should change `_variable` visibility from `public` to `internal/private` in `StakeTable.sol`

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+        LightClient.LightClientState memory genesis = LightClient.LightClientState({
+            viewNum: 0,
+            blockHeight: 0,
+            blockCommRoot: BN254.ScalarField.wrap(0),
+            feeLedgerComm: BN254.ScalarField.wrap(0),
+            stakeTableBlsKeyComm: BN254.ScalarField.wrap(0),
+            stakeTableSchnorrKeyComm: BN254.ScalarField.wrap(0),
+            stakeTableAmountComm: BN254.ScalarField.wrap(0),
+            threshold: 0
+        });
```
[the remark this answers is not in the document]
philippecamacho: Right, fixed in 5570ee29ef9adc87921b0f11bfa96c202d9f6400.

contracts/test/BLSSig.t.sol
```diff
@@ -61,21 +62,57 @@ contract BLSSig_Test is Test {
...
+
+        BN254.G2Point memory g2 = BN254.P2();
+        (uint256 x0, uint256 x1, uint256 y0, uint256 y1) = BN256G2.ECTwistMul(
+            exp,
+            BN254.BaseField.unwrap(g2.x0),
+            BN254.BaseField.unwrap(g2.x1),
+            BN254.BaseField.unwrap(g2.y0),
+            BN254.BaseField.unwrap(g2.y1)
+        );
+
```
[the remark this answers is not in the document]
philippecamacho: I agree. Done in 4adf465a795b3646b0d0bdb7b8e36b34cc773340.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+import { ExampleToken } from "../src/ExampleToken.sol";
+
+// Target contract
+import { StakeTable as S } from "../src/StakeTable.sol";
+
+import { CommonBase } from "forge-std/Base.sol";
+import { StdCheats } from "forge-std/StdCheats.sol";
+import { StdUtils } from "forge-std/StdUtils.sol";
+
+contract StakeTableHandler is CommonBase, StdCheats, StdUtils {
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in a7a3251c1c1bd929e509d1bdf7e744233dae15ca.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+        exampleTokenCreator = makeAddr("tokenCreator");
+        vm.prank(exampleTokenCreator);
+        token = new ExampleToken(INITIAL_BALANCE);
+
+        address userAddress;
+
+        // Distribute tokens to users
+        for (uint256 i = 0; i < NUM_USERS; i++) {
+            userAddress = makeAddr(string(abi.encode(i)));
+            users.push(userAddress);
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in 86bbe51.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+
+// Target contract
+import { StakeTable as S } from "../src/StakeTable.sol";
+
+import { CommonBase } from "forge-std/Base.sol";
+import { StdCheats } from "forge-std/StdCheats.sol";
+import { StdUtils } from "forge-std/StdUtils.sol";
+
+contract StakeTableHandler is CommonBase, StdCheats, StdUtils {
+    StakeTable_Test private testStakeTableContract;
```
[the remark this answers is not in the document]
philippecamacho: Fixed in 9fcbb54

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+            blockHeight: 0,
+            blockCommRoot: BN254.ScalarField.wrap(0),
+            feeLedgerComm: BN254.ScalarField.wrap(0),
+            stakeTableBlsKeyComm: BN254.ScalarField.wrap(0),
+            stakeTableSchnorrKeyComm: BN254.ScalarField.wrap(0),
+            stakeTableAmountComm: BN254.ScalarField.wrap(0),
+            threshold: 0
+        });
+        lightClientContract = new LightClientTest(genesis, 10);
+        stakeTable = new S(address(token), address(lightClientContract), 10);
```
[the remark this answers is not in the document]
philippecamacho: Done in 50f3ddb.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,242 @@
...
+        handler = new StakeTableHandler(
+            stakeTable, exampleTokenCreator, token, lightClientContract, users
+        );
+
+        // Only test the handler
+        targetContract(address(handler));
+    }
+
+    function invariant_BalancesAreConsistent() external {
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in 01888ed610ef1cbf4a16ee6ce3c2246a6a2042d7.

contract-bindings/src/stake_table_handler.rs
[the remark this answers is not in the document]
philippecamacho: I could not find the right regexp either. I just renamed `StakeTableHandler` to `StakeTableHandlerTest`. See d05f97d.

philippecamacho:
> imo, the handler should we written slightly differently.
> 
> - all 4 APIs should add necessary setup to ensure success call (in contrast to our original `registerWithSeed(bool expectRevert)`

Overall I agree that we should shape the handler in order to reduce the number of reverts. However it is hard to eliminate all reverts without making the tests harder to read (e.g.: for `register` one would need to keep track of users who have registered already to avoid a `NodeAlreadyRegistered` error). Also I think it is good to let some calls revert as it is more realistic and allows for finding bugs (imagine for example a failed call to `register` prevents other users to `register` at a later stage).

In all cases now the number of reverts has decreased significantly:
BEFORE:
```
Running 4 tests for contracts/test/StakeTableInvariant.t.sol:StakeTableInvariant_Tests
[PASS] invariant_BalancesAreConsistent() (runs: 256, calls: 3840, reverts: 1970)
[PASS] invariant_Queue() (runs: 256, calls: 3840, reverts: 1931)
[PASS] invariant_Register() (runs: 256, calls: 3840, reverts: 1941)
[PASS] invariant_RequestExit() (runs: 256, calls: 3840, reverts: 1906)
```

AFTER:
```
Running 4 tests for contracts/test/StakeTableInvariant.t.sol:StakeTableInvariant_Tests
[PASS] invariant_BalancesAreConsistent() (runs: 256, calls: 3840, reverts: 621)
[PASS] invariant_Queue() (runs: 256, calls: 3840, reverts: 580)
[PASS] invariant_Register() (runs: 256, calls: 3840, reverts: 639)
[PASS] invariant_RequestExit() (runs: 256, calls: 3840, reverts: 685)
``` 

>   - `register` should start with `genClientWallet` each time, and credit some token before registering to ensure success, and upon registration add the bls key to a list

The function `registerWithSeed` called by `register` already calls `genClientWallet` and credits some token. As discussed I prefer to let some calls to register to fail.

>   - `deposit` should take a random index and get the corresponding bls key, credit with more token before depositing

Function `deposit` added. 

>   - `requestExit` should take a random index and get the bls key and submit, then store the assigned exit epoch in a data struct (most likely a mapping `key => exitEpoch` + a list of requested keys)
>   - `withdraw` should only withdraw on fully exited 

I agree. This helps a lot to reduce the number of reverts for the `withdrawFunds` function. Both functions have been updated.

> ^^ this also avoids the conditional invariant anti-pattern [mentioned in forge doc](https://book.getfoundry.sh/forge/invariant-testing#conditional-invariants).
> 

Looks more like a tradeoff to me than an antipattern. Happy to discuss further.

> I think we can remove `SequenceOfEvent` test with this invariant test.

Yes.

> Besides, some tech-debt, we should change `_variable` visibility from `public` to `internal/private` in `StakeTable.sol`

Done in c04a2625ae7c3b0cbaa7f497017abe7385d448ab.

contracts/test/StakeTable.t.sol
```diff
@@ -63,6 +49,22 @@ contract StakeTable_Test is Test {
             blsSig
         );
     }
+}
+
+contract StakeTable_Test is Test, StakeTableCommonTest {
```
alxiong, line 54: btw, you don't need extra `is Test`, just `is StakeTableCommonTest` is enough. 

Same applied to `Handler is CommonTest`.

contracts/test/StakeTable.t.sol
```diff
@@ -63,6 +49,22 @@ contract StakeTable_Test is Test {
...
+    event Registered(bytes32, uint64, AbstractStakeTable.StakeType, uint256);
+    event Deposit(bytes32, uint256);
+    event Exit(bytes32, uint64);
```
alxiong, line 57: maybe we don't need to re-declare these?
I have example when I try to call `expectEmit`, I can just use the original definition: https://github.com/EspressoSystems/espresso-sequencer/blob/main/contracts/test/LightClient.t.sol#L208-L210

alxiong (commented):
As discussed, we should remove `returns (bool)` on `register` and `requestExit` since they are not used, those functions either pass through or revert.

philippecamacho:
> As discussed, we should remove `returns (bool)` on `register` and `requestExit` since they are not used, those functions either pass through or revert.

Done in 7fb6c9f.

philippecamacho:
> imo, the handler should we written slightly differently.
> 
> * all 4 APIs should add necessary setup to ensure success call (in contrast to our original `registerWithSeed(bool expectRevert)`
>   
>   * use a dynamic list of `users` (can create using seed that deries from `users.length` so that we won't generate the same wallet or user address) @philippecamacho
>   * `register` should start with `genClientWallet` each time, and credit some token before registering to ensure success, and upon registration add the bls key to a list
>   * `deposit` should take a random index and get the corresponding bls key, credit with more token before depositing
>   * `requestExit` should take a random index and get the bls key and submit, then store the assigned exit epoch in a data struct (most likely a mapping `key => exitEpoch` + a list of requested keys)
>   * `withdraw` should only withdraw on fully exited
>   * `advanceEpoch() public` leave as it is

Now the invariant tests do not generate any revert.
```
Running 4 tests for contracts/test/StakeTableInvariant.t.sol:StakeTableInvariant_Tests
[PASS] invariant_BalancesAreConsistent() (runs: 256, calls: 3840, reverts: 0)
[PASS] invariant_Queue() (runs: 256, calls: 3840, reverts: 0)
[PASS] invariant_Register() (runs: 256, calls: 3840, reverts: 0)
[PASS] invariant_RequestExit() (runs: 256, calls: 3840, reverts: 0)
Test result: ok. 4 passed; 0 failed; 0 skipped; finished in 5.36s
 
Ran 1 test suites: 4 tests passed, 0 failed, 0 skipped (4 total tests)
```

contracts/test/StakeTable.t.sol
philippecamacho: Done in aeaf0c0.

contracts/test/StakeTable.t.sol
philippecamacho: Yes, done in ced31cd.

contracts/rust/src/bin/diff_test.rs
```diff
@@ -441,6 +446,19 @@ fn main() {
...
         }
+        Action::GenRandomG2Point => {
+            if cli.args.len() != 1 {
+                panic!("Should provide arg1=exponent");
+            }
+
+            let exponent: u64 = cli.args[0].parse::<u64>().unwrap();
+            let mut point = G2Affine::generator();
+            let exp_big_int: BigInt<4> = BigInt::from(exponent);
+            point = point.mul_bigint(&exp_big_int).into_affine();
```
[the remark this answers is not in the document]
philippecamacho: Done in 3519048e6f75570c6786ec169031d114a52ca2b6.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,341 @@
...
+            viewNum: 0,
+            blockHeight: 0,
+            blockCommRoot: BN254.ScalarField.wrap(1),
+            feeLedgerComm: BN254.ScalarField.wrap(1),
+            stakeTableBlsKeyComm: BN254.ScalarField.wrap(1),
+            stakeTableSchnorrKeyComm: BN254.ScalarField.wrap(1),
+            stakeTableAmountComm: BN254.ScalarField.wrap(1),
+            threshold: 10
+        });
+        uint32 numBlocksPerEpoch = 10;
```
alxiong, line 285: i feel like if we were to set it to lower value, then we can test more "cross-epoch" behavior/scenarios.
can we use 4 instead?

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,341 @@
...
+    // Variables for testing invariant relative to requestExit and withdrawFunds
+    uint64 public nextExitEpochBefore;
+    uint64 public pendingExitsBefore;
+    uint64 public stakeTableFirstAvailableExitEpoch;
+    uint64 public stakeTableNumPendingExits;
+    uint64 public currentEpoch;
+    bool public requestExitCalledAtLeastOnce;
+
+    mapping(bytes32 blsKeyHash => uint64 exitEpoch) public exitEpochForBlsVK;
+    BN254.G2Point[] public requestExitKeys;
```
alxiong, line 52: if `requestExitKeys` -> `uint256[] public requestExitKeyIndexes`, then we wouldn't need to build `userIndexFromVk` map, instead, we would use the fuzzy-random index to select a exit key _index_, then use `vk[idx]` to get the vk, and use `exitEpochForBlsVk(vkHash)` to get exit epoch.

contracts/test/StakeTableInvariant.t.sol
```diff
@@ -0,0 +1,341 @@
...
+            assertEq(handler.stakeTableFirstAvailableExitEpoch(), 0);
+            assertEq(handler.stakeTableNumPendingExits(), 0);
+        }
+    }
+
+    function invariant_Queue() external {
+        // Global invariants
+        assertLe(stakeTable.numPendingRegistrations(), stakeTable.maxChurnRate());
+        assertLe(stakeTable.numPendingExits(), stakeTable.maxChurnRate());
+    }
```
alxiong, line 340: what about: keep a counter for the total registration & total exit in the handler, and make sure our `total reg = churn rate * nextEpoch + numPending`?

alxiong (approved):
This is great @philippecamacho Good work! 
left some improvement suggestions, but giving my principle approval now.

contracts/test/StakeTableInvariant.t.sol
philippecamacho: I don't follow as `userIndexFromVk` is not used in the context of exit requests.

contracts/test/StakeTableInvariant.t.sol
philippecamacho: This is true only if we don't skip epochs.

contracts/test/StakeTableInvariant.t.sol
philippecamacho: Yes, done in 548207f9432ce77e994bcaeda5f20b580e310c38.

contracts/test/StakeTableInvariant.t.sol
alxiong: can we also document the number of skipped epoch?

philippecamacho: I don't think that would work either because an epoch can be skipped when the first available epoch is partially filled. So we cannot derive a clean equation between the total number of registrations and the number of epochs (skipped and not skipped) and the churn rate.

contracts/test/StakeTableInvariant.t.sol
philippecamacho: Ah right, I understand now. Suggested change done in https://github.com/EspressoSystems/espresso-sequencer/pull/1008.

Issue #801, opened 2023-12-05
[light-client-integration] support to light client state prover


mrain:
Replica:

- When register on the stake table in L1, generate a Schnorr key pair (besides the HotShot’s consensus key which is `bls-over-bn254`)
- On every new finalized block, if it leads to a new finalized `LightClientState` then the replica sign the new state and store the Schnorr signature locally and send one copy to the CDN
    - The copy are sent alongside the normal consensus vote in that round (i.e. no extra round is required)
- Garbage collect: replicas also subscribe event stream from `LightClient.sol` on L1, when `event NewFinalizedState` is emitted, replicas delete all signatures on that state and even older ones. We could also add a worst-case storage cap on how long these signatures would be kept before being pruned.
    - Assume a worst-case delay on L1 light client update of 5 hour (and assume 3 sec block time), each replica only keeps ~6k * 64 = 0.38MB of signature data, ~6k * 208 = 1.25MB of state/message data locally.
- Implicit CDN support

mrain:
currently blocked by #807

PR #879, opened 2023-12-19
[light-client-integration] support for light client state signature

mrain:
closes: #801 

**What does this PR do**
* Watches the hotshot event stream, if a `LeafDecided` event happens, it triggers the `light_client_signature_hook` to generate a new state signature
* `SystemContextHandle` is now wrapped in `SequencerContext`, where the later includes a light client state signature pool that contains the most recent 100 signatures
* Adds a endpoint for fetching the light client state signature. It's enabled when http option is available.


sequencer/api/state_signature.toml
```diff
@@ -0,0 +1,4 @@
+[route.getstatesignature]
+PATH = ["block/:height/statesignature"]
```
jbearer, line 2: Nit-pick, I know this is the style but I'm trying to transition these to being more readable:

```suggestion
[route.get_state_signature]
PATH = ["block/:height/state-signature"]
```

jbearer: Actually since this is going to be in a `state-signature` API module we could maybe just make the route `block/:height`, so that overall it looks like `query.espresso.network/state-signature/block/42`

sequencer/src/api/signature_pool.rs
```diff
@@ -0,0 +1,24 @@
+use hotshot_types::light_client::StateSignature;
+use std::collections::{HashMap, VecDeque};
+
+pub(crate) const SIGNATURE_POOL_CAPACITY: usize = 100;
+
+#[derive(Debug, Default)]
+pub struct StateSignaturePool {
+    pool: HashMap<u64, StateSignature>,
+    deque: VecDeque<u64>,
```
jbearer, line 9: Can we name this something better? What is it storing?

jbearer: Come to think of it, we can combine these into `VecDeque<StateSignature>` and only store one height: the height of the first entry in the deque. Then the height of any entry `i` is `first_height + i`, and the index of any `height >= first_height` is `height - first_height`.

jbearer: Also do we want persistence for this? What should be the behavior if the node crashes and restarts? Is it ok if it loses its cached signatures?

sequencer/src/enriched_handle.rs
```diff
@@ -0,0 +1,55 @@
+use hotshot::{traits::NodeImplementation, types::SystemContextHandle};
+use hotshot_types::{light_client::StateSignature, traits::node_implementation::NodeType};
+use std::{
+    ops::{Deref, DerefMut},
+    sync::{Arc, RwLock},
+};
+
+use crate::api::signature_pool::StateSignaturePool;
```
jbearer, line 8: I wasn't convinced `signature_pool` belonged under the API module when I first saw it, and this convinces me. Conceptually, the API is like a leaf in the dependency graph of modules, so it's a bit weird to have a top-level module like this one importing things from the API module.

I would recommend making `src/state_signature/{signature_pool, enriched_handle}`, and then in the API we can import things from `crate::state_signature` as needed, rather than the other way around. It doesn't make any difference practically but I think it better captures the conceptual structure we're going for.

sequencer/src/enriched_handle.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub struct EnrichedSystemContextHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
```
jbearer, line 10: Can we think of a better name for this?

sequencer/src/enriched_handle.rs
```diff
@@ -0,0 +1,55 @@
...
+    handle: SystemContextHandle<TYPES, I>,
+
+    state_signature_pool: Arc<RwLock<StateSignaturePool>>,
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> Deref
```
jbearer, line 16: Nit: this usage of `Deref` is [frowned upon](https://rust-lang.github.io/api-guidelines/predictability.html?highlight=deref#only-smart-pointers-implement-deref-and-derefmut-c-deref). Better to use `AsRef` or just an inherent method like `fn hotshot(&self) -> &SystemContextHandle`

sequencer/src/enriched_handle.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EnrichedSystemContextHandle<TYPES, I> {
+    pub fn get_state_signature(&self, height: u64) -> Option<StateSignature> {
+        let pool_guard = self.state_signature_pool.read().unwrap();
+        pool_guard.get_signature(height)
+    }
+
+    pub fn add_new_state_signature(&mut self, height: u64, signature: StateSignature) {
```
jbearer, line 51: Not sure what the correct thing should be, but there's no point in using an `RwLock` if we only ever obtain write access through a `&mut`. Seems like this should be `&self` (if this is actually going to be shared between threads) or we should get rid of the `RwLock` (if it's not shared)

sequencer/src/api.rs
```diff
@@ -1,26 +1,30 @@
...
 pub use options::Options;
 
+use self::data_source::StateSignatureDataSource;
+
 type NodeIndex = u64;
 
-pub type Consensus<N> = SystemContextHandle<SeqTypes, Node<N>>;
+pub type Consensus<N> = EnrichedSystemContextHandle<SeqTypes, Node<N>>;
 
 pub struct SequencerNode<N: network::Type> {
```
jbearer, line 23: I feel like `EnrichedSystemContextHandle` and `SequencerNode` are both kind of serving the same purpose: wrapping a `SystemContextHandle` with some extra state. What do you think about flattening it out and adding the signature pool directly to `SequencerNode`?

sequencer/src/api/options.rs
```diff
@@ -79,7 +82,7 @@ impl Options {
     pub async fn serve<N, F>(mut self, init_handle: F) -> anyhow::Result<SequencerNode<N>>
     where
         N: network::Type,
-        F: FnOnce(Box<dyn Metrics>) -> BoxFuture<'static, (Consensus<N>, NodeIndex)>,
+        F: FnOnce(Box<dyn Metrics>) -> BoxFuture<'static, (Consensus<N>, NodeIndex, StateKeyPair)>,
```
jbearer, line 85: Getting to the point where we might consider replacing the return type tuple with a struct

sequencer/src/api/endpoints.rs
```diff
@@ -121,3 +124,52 @@ where
...
+    S::State: Send + Sync + StateSignatureDataSource<N>,
+{
+    let toml = toml::from_str::<toml::Value>(include_str!("../../api/state_signature.toml"))?;
+    let mut api = Api::<S, Error>::new(toml)?;
+
+    api.get("getstatesignature", |req, state| {
+        async move {
+            let height = req
+                .integer_param("height")
+                .map_err(|err| Error::internal(err.to_string()))?;
```
jbearer, line 141: Can you just do `?` here without the `map_err`? I think `Error` should have a `From<RequestError>` impl, and we want this to return 400 something when there's a bad request, not 500 for internal error

If `?` doesn't work on it's own, you could also try [`from_request_error`](https://tide-disco.docs.espressosys.com/tide_disco/error/trait.Error.html#method.from_request_error)

sequencer/src/api/endpoints.rs
```diff
@@ -121,3 +124,52 @@ where
...
+            state
+                .handle()
+                .get_state_signature(height)
+                .ok_or(Error::internal("Signature not found."))
```
jbearer, line 145: Probably not great to return an internal error here, we should return something with a 404 error code. You should be able to do `Error::catch_all(StatusCode::NotFound, "signature not found")`

jbearer (commented):
Question: is it ok for a node to choose not to run the state signature API? In other words isn't it considered malicious behavior for a node not to respond to this query? I'm wondering if we should make the `Http` option required and always run at least the state signature API.

.docker-compose.yaml.swp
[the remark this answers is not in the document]
mrain: Yes. Will remove it.

sequencer/src/api/signature_pool.rs
mrain: No persistence required. 
It's safer to just use a `HashMap` instead. 
I thought about the `VecDeque<StateSignature>` idea. The thing that I was worried about is that, we have to make sure to populate a signature for every block height.

mrain: Also it's okay to lose all cached signatures, at least for now.

sequencer/src/enriched_handle.rs
mrain: Good idea. Will do

sequencer/src/api/signature_pool.rs
jbearer: > The thing that I was worried about is that, we have to make sure to populate a signature for every block height

Good point, I didn't think about what happens if we miss a block

sequencer/src/api/endpoints.rs
mrain: But the `Error` here is `hotshot_query_service::Error`.

mrain:
> Question: is it ok for a node to choose not to run the state signature API? In other words isn't it considered malicious behavior for a node not to respond to this query? I'm wondering if we should make the `Http` option required and always run at least the state signature API.

This is worth discussing. IMHO
First, a light client state prover should be able to collect the signatures, at least from all honest nodes. If there's no data(signature) availability guarantee/mechanism, then yes, honest nodes must respond to these queries.
Second, signature collection doesn't have to be through this `Http`. This is just an option that works, if it's okay to make `Http` a default option.

cc @jbearer @chancharles92 @alxiong

sequencer/src/api.rs
mrain: Sure that'll be the best.
I was a little frustrated by the current initialization workflow. Let me spend more time to find a way.

sequencer/src/api/endpoints.rs
jbearer: Yes but `?` does an implicit `.into()`. And `hotshot_query_service::Error` implements `tide_disco::Error` which provides `from_request_error`

sequencer/src/enriched_handle.rs
mrain: I'm thinking about calling it `context::SequencerContext` that includes a consensus handle, node index and the collection of state signatures. Then the `SequencerNode` would contains only a sequencer context and the list of update tasks. WDYT?

sequencer/src/api/endpoints.rs
mrain: using `from_request_error` in [2f4ab54](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/2f4ab5484d02c57dbfc0dca96ca4b183d5f355d6)

sequencer/src/api/endpoints.rs
mrain: using this in [f7033f7](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/f7033f7222880be8e53eeee798c12ca2691f496c)

sequencer/src/api/options.rs
mrain: I changed this to `SequencerContext`. [f7033f7](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/f7033f7222880be8e53eeee798c12ca2691f496c)

sequencer/api/state_signature.toml
mrain: Changed in [f7033f7](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/f7033f7222880be8e53eeee798c12ca2691f496c)

sequencer/src/enriched_handle.rs
mrain: I'm not familiar with the tide-disco apps. For safety, I kept the `RwLock` and remove those `mut`s. [f7033f7](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/f7033f7222880be8e53eeee798c12ca2691f496c)

sequencer/src/api/endpoints.rs
```diff
@@ -121,3 +124,30 @@ where
...
+    let toml = toml::from_str::<toml::Value>(include_str!("../../api/state_signature.toml"))?;
+    let mut api = Api::<S, Error>::new(toml)?;
+
+    api.get("get_state_signature", |req, state| {
+        async move {
+            let height = req
+                .integer_param("height")
+                .map_err(Error::from_request_error)?;
+            state
+                .get_state_signature(height)
```
philippecamacho, line 143: This function returns an `Option`.  Should not we use `unwrap_or` instead?

philippecamacho: Sorry, this does not work.

philippecamacho:
Do we plan to write tests for this feature in another PR?

sequencer/src/state_signature.rs
```diff
@@ -0,0 +1,27 @@
...
+pub use hotshot_stake_table::vec_based::config::FieldType as BaseField;
+
+pub async fn state_signature_hook<N: network::Type>(
+    context: &SequencerContext<SeqTypes, Node<N>>,
+    leaf: &Leaf,
+) {
+    let new_light_client_state = LightClientState::<BaseField> {
+        view_number: leaf.get_view_number().get_u64() as usize,
+        block_height: leaf.get_height() as usize,
+        block_comm_root: BaseField::default(),
```
philippecamacho, line 17: We should add a TODO here as well right?

mrain:
> Do we plan to write tests for this feature in another PR?

I'm writing tests now. Dealing with more problems :(

sequencer/src/state_signature.rs
mrain: I just created 2 new issues. https://github.com/EspressoSystems/espresso-sequencer/issues/923, https://github.com/EspressoSystems/espresso-sequencer/issues/924

jbearer (commented):
This looks good, but are we going to change it to POST signatures to another service, rather than just providing a query API for signatures?

mrain:
> This looks good, but are we going to change it to POST signatures to another service, rather than just providing a query API for signatures?

Yes we are. But I think it's better to keep this query API, maybe as a backup in the future.

sequencer/src/api.rs
```diff
@@ -152,12 +184,65 @@ mod test_helpers {
...
+        client
+            .post::<()>("submit/submit")
+            .body_json(&txn)
+            .unwrap()
+            .send()
+            .await
+            .unwrap();
+
+        // Wait for a Decide event containing transaction matching the one we sent
+        wait_for_decide_on_handle(&mut events, &txn).await.unwrap();
```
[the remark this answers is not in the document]
mrain: Sure. Let me see how to implement this.

sequencer/src/api.rs
```diff
@@ -152,12 +184,65 @@ mod test_helpers {
...
+
+        let height = context.consensus().get_decided_leaf().await.get_height();
+        sleep(std::time::Duration::from_secs(1)).await;
+        let _signature = client
```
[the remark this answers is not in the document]
mrain: I don't think it's needed here.
cc @alxiong

sequencer/src/context.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::state_signature::{self, StateSignatureMemStorage};
+
+/// The sequencer context contains a consensus handle and other sequencer specific information.
+#[derive(Clone)]
+pub struct SequencerContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
+    /// The consensus handle
+    handle: SystemContextHandle<TYPES, I>,
+
+    /// Index of this sequencer node
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
mrain: Not from my side, it was one of the returns in the previous `init_node`

sequencer/src/api.rs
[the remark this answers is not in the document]
mrain: This reminds me that we should have an endpoint for fetching light client state. 
But current light client state is not serde serializable. Will do it in next PR.

sequencer/src/api.rs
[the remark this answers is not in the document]
mrain: Done in [58effbb](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/58effbbf62bbf42b7002c70cc8a35a2514a3a4b4)

sequencer/src/context.rs
```diff
@@ -0,0 +1,76 @@
...
+    traits::node_implementation::NodeType,
+};
+use jf_primitives::signatures::SignatureScheme;
+use std::sync::{Arc, RwLock};
+
+use crate::state_signature::{self, StateSignatureMemStorage};
+
+/// The sequencer context contains a consensus handle and other sequencer specific information.
+#[derive(Clone)]
+pub struct SequencerContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
```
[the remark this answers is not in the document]
mrain: Done in [3b7b0d1](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/3b7b0d181b23a6c3ca379c6d544e60445a3116cd)

sequencer/src/context.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Return a signature of a light client state at given height.
+    pub fn get_state_signature(&self, height: u64) -> Option<StateSignature> {
+        let pool_guard = self.state_signatures.read().unwrap();
+        pool_guard.get_signature(height)
+    }
+
+    /// Sign the light client state at given height and store it.
+    pub fn sign_new_state(&self, state: &LightClientState<state_signature::BaseField>) {
+        let state_msg: [state_signature::BaseField; 7] = state.into();
```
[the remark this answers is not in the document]
mrain: Is there a better solution other than converting it to `Vec`? 
I'm also thinking about moving all light client related things to the sequencer repo, that makes future upgrade a lot easier. Although that's not likely to happen.

sequencer/src/context.rs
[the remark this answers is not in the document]
jbearer: Huh. If we're not using it I guess it's fine to delete it. Not a big deal though

sequencer/src/context.rs
```diff
@@ -25,13 +30,20 @@ pub struct SequencerContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     state_signatures: Arc<RwLock<StateSignatureMemStorage>>,
 }
 
-impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SequencerContext<TYPES, I> {
+impl<N: network::Type> Clone for SequencerContext<N> {
```
jbearer, line 33: I assume this is implemented manually because the derived implementation requires `N: Clone`? You can use `Derivative` to get around this:

```rust
#[derive(Derivative)]
#[derivative(Clone(bound = ""))]
```

sequencer/src/context.rs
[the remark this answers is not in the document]
jbearer: If it doesn't compile without the annotation, don't worry about it. No reason to convert to `Vec`.

jbearer (approved):
Looks good!

sequencer/src/context.rs
mrain: Yes `Derivative` works! [ddd5882](https://github.com/EspressoSystems/espresso-sequencer/pull/879/commits/ddd588239450fbc6ff311db35354315a475272f0)

PR #880, opened 2023-12-19
Update query service and improve SQL backend

### LINKED ISSUE #843: Add code to initialize and/or reset the Postgres database when using SQL persistence
Given a database name and credentials for a user with permissions over that database, when creating a `SqlDataSource` we should
* Remove all existing tables if requested (`--reset-store`)
* Run migrations to create necessary tables

jbearer:
* Use new Query API for database queries
* Make database resettable
* Fix some tests, which were failing based on timing

Closes #843

### DIFF
--- a/sequencer/src/api/options.rs
+++ b/sequencer/src/api/options.rs
@@ -184,6 +184,10 @@ pub struct Sql {
     /// Password for Postgres user.
     #[clap(long, env = "ESPRESSO_SEQUENCER_POSTGRES_PASSWORD")]
     pub password: Option<String>,
+
+    /// Reset database upon connecting.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_RESET_STORE")]
+    pub reset_store: bool,
 }
 
 /// Options for the query API module backed by the file system.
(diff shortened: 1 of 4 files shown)


sequencer/src/api/options.rs
```diff
@@ -184,6 +184,10 @@ pub struct Sql {
     /// Password for Postgres user.
     #[clap(long, env = "ESPRESSO_SEQUENCER_POSTGRES_PASSWORD")]
     pub password: Option<String>,
+
+    /// Reset database upon connecting.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_RESET_STORE")]
```
sveitser, line 189: This is a bit annoying because it only supports `true` (not `1`, `True`, ...).

sveitser: @Ancient123 what's your preferred way to use this feature?

sveitser (approved):
LGTM. A question about UX.

sequencer/src/api/options.rs
Ancient123: true is fine, but supporting more options would be better.. Not worth holding up the PR though honestly.

PR #882, opened 2023-12-20
Reduce max_blocks in the commit task if we exceed the block gas limit

### LINKED ISSUE #857: Commit task gets stuck forever if number of leaves to send exceeds block gas limit

nomaxg:
Closes #857

### DIFF
--- a/sequencer/src/hotshot_commitment.rs
+++ b/sequencer/src/hotshot_commitment.rs
@@ -74,7 +74,7 @@ pub async fn run_hotshot_commitment_task(opt: &CommitmentTaskOptions) {
 
     // Get the maximum number of blocks the contract will allow at a time.
     let max = match contract.max_blocks().call().await {
-        Ok(max) => max.as_u64(),
+        Ok(max) => max.as_usize(),
         Err(err) => {
             tracing::error!("unable to read max_blocks from contract: {}", err);
             panic!("hotshot commitment task will exit");
@@ -83,7 +83,7 @@ pub async fn run_hotshot_commitment_task(opt: &CommitmentTaskOptions) {
     sequence(max, hotshot, contract).await;
 }
 
-async fn sequence(hard_block_limit: u64, hotshot: HotShotClient, contract: HotShot<Signer>) {
+async fn sequence(hard_block_limit: usize, hotshot: HotShotClient, contract: HotShot<Signer>) {
     // This is the number of blocks we attempt to sequence
     // If we fail to submit soft_block_limit leaves, we assume we have hit
     // A gas limit exception and decrease the limit
@@ -145,12 +145,15 @@ impl HotShotDataSource for HotShotClient {
 
 #[derive(Debug)]
 enum SyncError {
-    TransactionFailed { err: anyhow::Error, num_leaves: u64 },
+    TransactionFailed {
+        err: anyhow::Error,
+        num_leaves: usize,
+    },
     Other(anyhow::Error),
 }
 
 async fn sync_with_l1(
-    max_blocks: u64,
+    max_blocks: usize,
     hotshot: &impl HotShotDataSource,
     contract: &HotShot<Signer>,
 ) -> Result<(), SyncError> {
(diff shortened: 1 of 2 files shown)


sequencer/src/hotshot_commitment.rs
```diff
@@ -83,10 +83,15 @@ pub async fn run_hotshot_commitment_task(opt: &CommitmentTaskOptions) {
...
 }
 
-async fn sequence(max_blocks: u64, hotshot: HotShotClient, contract: HotShot<Signer>) {
+async fn sequence(mut max_blocks: u64, hotshot: HotShotClient, contract: HotShot<Signer>) {
     loop {
         if let Err(err) = sync_with_l1(max_blocks, &hotshot, &contract).await {
             tracing::error!("error synchronizing with HotShot contract: {err}");
+            // If we are erroring because of a block gas limit exception, we can try again
+            // with fewer blocks next time
+            if err.to_string().contains("exceeds block gas limit") && max_blocks > 1 {
```
jbearer, line 92: This seems very fragile. What if something in the L1 or RPC or or ethers changes the wording slightly? But I guess there's not really a good way to extract the error cause from the error. What do you think about this instead:

If there is any error sending the transaction to L1, and the number of leaves being sent is `max_blocks`, assume we hit a gas error and decrease max blocks. `sync_with_l1` would have to be changed to return something like

```rust
enum SyncError {
  TransactionFailed { err: anyhow::Error, num_leaves: usize },
  Other(anyhow::Error),
}
```

but this would let us reliably detect this case without parsing error strings.

sequencer/src/hotshot_commitment.rs
```diff
@@ -83,10 +83,15 @@ pub async fn run_hotshot_commitment_task(opt: &CommitmentTaskOptions) {
...
+                max_blocks /= 2
```
jbearer, line 93: This means that we might end up underestimating the maximum number of blocks we can send. I don't think this is really big deal but in the extreme if `max_blocks` is too small we can get in a situation where we can't keep up with HotShot, since HotShot almost always produces multiple blocks per L1 block.

As an optimization we could keep track of like `mut soft_block_limit` and `hard_block_limit`. Whenever we _succeed_ in syncing with L1, increase `soft_block_limit` as long as it doesn't exceed `hard_block_limit`.

jbearer (approved):
Generally looks good, I have some suggestions. The changes in utils are a nice improvement, thanks for that

sequencer/src/hotshot_commitment.rs
nomaxg: This makes sense, `max_leaves` is the only parameter here we can really control here so seems reasonable for us to reduce if we max out and error

sequencer/src/hotshot_commitment.rs
nomaxg: oh yes, good call

sequencer/src/hotshot_commitment.rs
```diff
@@ -165,7 +199,10 @@ async fn sync_with_l1(
     // example, if there are multiple commitment tasks racing.
     contract_send(&txn)
         .await
-        .ok_or_else(|| anyhow::Error::msg("failed to send transaction"))?;
+        .map_err(|e| SyncError::TransactionFailed {
+            err: e,
+            num_leaves: max_blocks,
```
nomaxg, line 204: Need to make this num_leaves

sequencer/src/hotshot_commitment.rs
```diff
@@ -132,20 +143,37 @@ impl HotShotDataSource for HotShotClient {
...
     hotshot: &impl HotShotDataSource,
     contract: &HotShot<Signer>,
-) -> Result<(), anyhow::Error> {
-    let contract_block_height = contract.block_height().call().await?.as_u64();
+) -> Result<(), SyncError> {
+    let contract_block_height = contract
+        .block_height()
+        .call()
+        .await
+        .map_err(|e| SyncError::Other(e.into()))?
```
jbearer, line 161: Nit: implement `From<anyhow::Error> for SyncError` and you can just use `?` without this `map_err`

sequencer/src/hotshot_commitment.rs
```diff
@@ -168,7 +197,12 @@ async fn sync_with_l1(
     // error. We will retry, and may end up changing the transaction we send if the contract state
     // has changed, which is one possible cause of the transaction failure. This can happen, for
     // example, if there are multiple commitment tasks racing.
-    contract_send(&txn).await?;
+    contract_send(&txn)
+        .await
+        .map_err(|e| SyncError::TransactionFailed {
+            err: e,
+            num_leaves: max_blocks,
```
jbearer, line 204: We should set `leaves.len()` here, rather than `max_blocks`. Otherwise we might think we have reached `max_blocks` and hit a gas limit error, when we haven't and have actually failed for some other transient error.

Note that as long as we keep failing, more leaves will accumulate until we hit `max_blocks`, so we will always _eventually_ start decreasing the soft limit, and make progress

jbearer (approved):
Looks good

sequencer/src/hotshot_commitment.rs
nomaxg: But this underlying error isn't an anyhow:Error, so we'd still need to do a double mapping somehow right?

PR #883, opened 2023-12-20
Fix tests that do not run locally by waiting for anvil endpoints to be up.

philippecamacho:
The solution is similar to https://github.com/EspressoSystems/espresso-polygon-zkevm-demo/blob/main/zkevm-contract-bindings/src/bin/deploy.rs#L483-L514

### DIFF
--- a/utils/src/lib.rs
+++ b/utils/src/lib.rs
@@ -59,12 +59,40 @@ impl AnvilOptions {
     pub async fn spawn(self) -> Anvil {
         let state_dir = TempDir::new().unwrap();
         let (child, url) = Anvil::spawn_server(&self, Some(state_dir.path())).await;
-        Anvil {
+        let anvil: Anvil = Anvil {
             child,
             url,
             state_dir,
             opt: self,
+        };
+
+        // When we are running a local Anvil node, as in tests, some endpoints (e.g. eth_feeHistory)
+        // do not work until at least one block has been mined. Send a transaction to force the
+        // mining of a block.
+        anvil
+            .provider()
+            .send_transaction(
+                TransactionRequest {
+                    to: Some(Address::zero().into()),
+                    ..Default::default()
+                },
+                None,
+            )
+            .await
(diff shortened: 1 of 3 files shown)


utils/src/lib.rs
```diff
@@ -404,10 +405,36 @@ async fn wait_for_transaction_to_be_mined<P: JsonRpcClient>(
...
+            None,
+        )
+        .await
+        .unwrap()
+        .await
+        .unwrap();
+
+    // When we are running a local Anvil node, as in tests, some endpoints (e.g. eth_feeHistory)
+    // do not work until at least one block has been mined. Send a transaction to force the
+    // mining of a block.
```
jbearer, line 431: This comment is a duplicate of the above?

utils/src/lib.rs
```diff
@@ -404,10 +405,36 @@ async fn wait_for_transaction_to_be_mined<P: JsonRpcClient>(
         sleep(interval).await;
     }
 
-    tracing::error!("contraact call {hash:?}: not mined after {retries} retries");
+    tracing::error!("contract call {hash:?}: not mined after {retries} retries");
     false
 }
 
+pub async fn wait_for_anvil_endpoints<P: JsonRpcClient>(provider: &Provider<P>) {
```
jbearer, line 412: Should we make this part of `spawn`, so that `AnvilOptions::spawn()` doesn't return until Anvil is ready? That way we can't forget to call it in some new test or something

utils/src/lib.rs
philippecamacho: Right, removed in ba28632ad033e86795ff111859009832bf12b38a.

utils/src/lib.rs
philippecamacho: I agree. Done in f05cad023f3daea69388b63d2991f9ffe341b823.

jbearer (approved):

nomaxg (approved):

PR #907, opened 2023-12-26
Add a program to submit random transactions

jbearer:


sequencer/src/bin/submit-transactions.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// This can be used to apply backpressure so that the tasks submitting transactions do not get
+    /// too far ahead of the task processing results.
+    #[clap(long, default_value = "1000")]
+    channel_bound: usize,
+
+    /// Number of parallel tasks to run.
+    #[clap(short, long, default_value = "1")]
+    jobs: usize,
+
+    /// URL of the query service.
```
sveitser, line 52: This doesn't have the clap derive. What does that do?

sequencer/src/bin/submit-transactions.rs
```diff
@@ -0,0 +1,161 @@
...
+        if let Some(delay) = opt.delay {
+            sleep(delay).await;
+        }
+    }
+}
+
+fn random_transaction(min_size: usize, max_size: usize) -> Transaction {
+    let mut rng = thread_rng();
+
+    let vm = rng.next_u64();
```
sveitser, line 154: Wondering if we want to narrow this down to a few VM IDs instead of sampling from the whole range which means we will probably never use the same VM ID twice.

sequencer/src/bin/submit-transactions.rs
jbearer: Makes it a (required) positional argument

sequencer/src/bin/submit-transactions.rs
```diff
@@ -0,0 +1,161 @@
...
+        sender.send(hash).await.ok();
+
+        if let Some(delay) = opt.delay {
+            sleep(delay).await;
+        }
+    }
+}
+
+fn random_transaction(min_size: usize, max_size: usize) -> Transaction {
+    let mut rng = thread_rng();
```
sveitser, line 152: How does this work with multithreading? It might be nice to allow passing in a seed to make the generated transactions deterministic, but we can also add that later.

sveitser (approved):
LGTM

PR #911, opened 2023-12-28
Fix local demo

jbearer:
* Do not set HotShot contract address. This causes the commitment task to ignore the --deploy flag and use an address which has no contract deployed
* Add healthchecks to codify startup order for native demo


philippecamacho:
How can we check that the demo works properly?
I ran `just demo-native` and I get this screen:

image

sveitser (approved):
LGTM

philippecamacho:
> It should look something like this:
> 

Is this normal? (Looks like the commitment task cannot start)

image


image

image

sveitser:
@philippecamacho the logs mention something about a duplicate docker network so I would remove the docker networks (`docker network rm espresso-sequencer`) and try again.

philippecamacho:
> docker network rm espresso-sequencer

Thanks!

philippecamacho (approved):
LGTM

Issue #925, opened 2024-01-05
Add block merkle tree to sequencer state


nomaxg:
For the hotshot light client, we need sequencer nodes to store and correctly update a block merkle root. 
- You can use [lightweight merkle tree implementation](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/merkle_tree/light_weight.rs) in jellyfish for this. 
- Add a merkle tree root (and since we don't have validated state yet, the merkle tree itself) to the [header](https://github.com/EspressoSystems/espresso-sequencer/blob/bef803f5831b1078a153db7d4ec34025991127c3/sequencer/src/block.rs#L176) 
- in [`<Header as BlockHeader>::new()`](https://github.com/EspressoSystems/espresso-sequencer/blob/bef803f5831b1078a153db7d4ec34025991127c3/sequencer/src/block.rs#L176), derive the new merkle tree from the parent header's tree and the new leaf, which is the new payload commitment 
- Write a standalone function (something like validate_block_state) that asserts that a new payload commitment applied to an existing merkle tree hashes to some state root + write a unit test
  e.g.
  ```rust
  fn validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<SHA3MerkleTree>;
  ```
  This function should check that `proposal` is a valid successor of `parent`, e.g.
  * `height` increases by 1
  * `timestamp` is within some delta of the current time and not decreasing
  * L1 head is within some delta of the current L1 head and not decreasing
  * **`block_merkle_tree_root` is correctly computed in `proposal`**
  and return the new validated state, ie the new merkle tree frontier.

  For now, we can ignore the unrelate stuff (height, timestamp, L1 head, etc) and just focus on the merkle tree stuff. We can add validation for everything else incrementally.

PR #961, opened 2024-01-09
add block_merkle_tree to `Header`

tbro:
WIP to address #925. `block_merkle_tree` field has been added to `Header` struct. MerkleTree details have been lifted from upstream examples.

### DIFF
--- a/sequencer/Cargo.toml
+++ b/sequencer/Cargo.toml
@@ -43,6 +43,7 @@ lazy_static = "1.4"
 rand_chacha = "0.3"
 serde_json = "1.0"
 sha2 = "0.10" # TODO temporary, used only for VID, should be set in hotshot
+sha3 = "0.10.8"
 time = "0.3"
 tokio-postgres = { version = "0.7", default-features = false, features = [ # disabling the default features removes dependence on the tokio runtime
 	"with-serde_json-1",
@@ -73,4 +74,3 @@ url = "2.3"
 
 # Dependencies for feature `testing`
 hotshot-testing = { workspace = true, optional = true }
-sha3 = "0.10.8"
--- a/sequencer/Cargo.toml
+++ b/sequencer/Cargo.toml
@@ -43,7 +43,6 @@ lazy_static = "1.4"
 rand_chacha = "0.3"
 serde_json = "1.0"
 sha2 = "0.10" # TODO temporary, used only for VID, should be set in hotshot
-sha3 = "0.10.8"
 time = "0.3"
 tokio-postgres = { version = "0.7", default-features = false, features = [ # disabling the default features removes dependence on the tokio runtime
 	"with-serde_json-1",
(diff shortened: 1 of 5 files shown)


sequencer/src/block.rs
```diff
@@ -2,26 +2,101 @@ use crate::{
...
     env,
     fmt::{Debug, Display},
     time::Duration,
 };
 use time::OffsetDateTime;
+use typenum::U3;
+
+/// Update the array length here
+#[derive(Default, Eq, PartialEq, Clone, Copy, Debug, Ord, PartialOrd, Hash)]
+pub struct Sha3Node(pub(crate) [u8; 32]);
```
nomaxg, line 37: @jbearer @mrain is this the right hash function for the block merkle tree?

sequencer/src/block.rs
```diff
@@ -2,26 +2,101 @@ use crate::{
...
+        let mut hasher = Sha3_256::new();
+        for value in data {
+            hasher.update(value);
+        }
+        Ok(Sha3Node(hasher.finalize().into()))
+    }
+
+    fn digest_leaf(_pos: &I, elem: &E) -> Result<Sha3Node, PrimitivesError> {
+        let mut writer = Vec::new();
+        elem.serialize_compressed(&mut writer).unwrap();
```
mrain, line 92: serialization may not be needed if the underlying element is already bytes.

sequencer/Cargo.toml
```diff
@@ -73,3 +73,4 @@ url = "2.3"
 
 # Dependencies for feature `testing`
 hotshot-testing = { workspace = true, optional = true }
+sha3 = "0.10.8"
```
jbearer, line 76: Try to keep dependencies grouped together by feature flag. This looks like a required dependency but it's listed under the comment "Dependencies for feature `testing`". Should be moved up to the big group of dependencies above.

sequencer/src/block.rs
jbearer: SHA3 should be fine but it looks like Jellyfish already provides a SHA3 hash scheme: https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/examples/struct.Sha3Digest.html. Not sure why it's in examples, though? @mrain?

In any case I don't think we should be implementing our own hash scheme here since it's not anything sequencer-specific. We should use the one from Jellyfish or update it if necessary.

sequencer/src/block.rs
```diff
@@ -72,6 +149,10 @@ pub struct Header {
 
     pub payload_commitment: VidCommitment,
     pub transactions_root: NMTRoot,
+    /// Root Commitment of Block Merkle Tree
+    pub block_merkle_tree_root: <SHA3MerkleTree as MerkleTreeScheme>::Commitment,
```
jbearer, line 153: Nit: might be good to make a type alias for this (e.g. `SHA3MerkleCommitment` or something)

jbearer (approved):
LGTM

PR #976, opened 2024-01-11
Add `validate_proposal` fn to `block.rs`

### LINKED ISSUE #925: Add block merkle tree to sequencer state

tbro:
This commit adds a standalone `_validate_proposal` function and related unit tests to `block.rs`. The intent is to validate block proposal with the primary goal of validating block merkle tree root. To avoid clippy errors `_` is appended to the function name. This should be removed in the future. Closes #925

### DIFF
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -20,8 +20,8 @@ use std::{
 };
 use time::OffsetDateTime;
 
-pub type SHA3MerkleTree = LightWeightSHA3MerkleTree<Commitment<Header>>;
-pub type SHA3MerkleCommitment = <SHA3MerkleTree as MerkleTreeScheme>::Commitment;
+pub type BlockMerkleTree = LightWeightSHA3MerkleTree<Commitment<Header>>;
+pub type BlockMerkleCommitment = <BlockMerkleTree as MerkleTreeScheme>::Commitment;
 
 /// A header is like a [`Block`] with the body replaced by a digest.
 #[derive(Clone, Debug, Deserialize, Serialize, Hash, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


sequencer/src/block.rs
```diff
@@ -178,6 +178,31 @@ impl Header {
     }
 }
 
+fn _validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<SHA3MerkleTree> {
```
nomaxg, line 181: Nice! Just a small nit: maybe we can make the merkle tree type a public type alias (something like `pub type BlockMerkleTree = SHA3MerkleTree`. Just so don't need to bake the hash function into the type name.

nomaxg (approved):
LGTM, one small comment

nomaxg (approved):

jbearer (approved):

PR #980, opened 2024-01-12
fixed reetrancy exploit surface on withdraw function

alysiahuggins:
used the Checks -> Effects -> Interactions (CEI) pattern to prevent re-entrancy attacks on the `withdrawFunds` function which could have resulted in the funds being drained from the Stake Table contract.


philippecamacho (approved):
Good catch!

PR #988, opened 2024-01-15
chore: update hotshot, hotshot-query-service, jellyfish

ggutoski:
@mrain @akonring what a marathon!

### DIFF
--- a/sequencer/src/api/fs.rs
+++ b/sequencer/src/api/fs.rs
@@ -33,9 +33,9 @@ impl SequencerDataSource for DataSource {
         let mut index = Index::default();
 
         // Index blocks by timestamp.
-        let mut blocks = data_source.get_block_range(..).await?;
+        let mut blocks = data_source.get_block_range(..).await;
         while let Some(block) = blocks.next().await {
-            index_block_by_time(&mut index.blocks_by_time, &block?);
+            index_block_by_time(&mut index.blocks_by_time, &block.await);
         }
         drop(blocks);
 
@@ -50,15 +50,12 @@ impl SequencerDataSource for DataSource {
         // one block into memory, and rarely very many.
         let blocks: Vec<_> = self
             .get_block_range(from_block..)
-            .await?
+            .await
             .enumerate()
             .collect()
             .await;
-        for (i, block) in blocks {
-            let Ok(block) = block else {
-                tracing::warn!("missing block {}, index may be out of date", from_block + i);
-                continue;
-            };
+        for (_, block) in blocks {
+            let block = block.await;
             index_block_by_time(&mut self.as_mut().blocks_by_time, &block);
         }
 
@@ -88,22 +85,22 @@ impl SequencerDataSource for DataSource {
     {
         let first_block = match from.into() {
             ResourceId::Number(n) => n,
-            ResourceId::Hash(h) => self.get_block(h).await?.height() as usize,
+            ResourceId::Hash(h) => self.get_block(h).await.await.height() as usize,
         };
 
         let mut res = TimeWindowQueryData::default();
 
         // Include the block just before the start of the window, if there is one.
         if first_block > 0 {
-            let prev = self.get_block(first_block - 1).await?;
+            let prev = self.get_block(first_block - 1).await.await;
             res.prev = Some(prev.header().clone());
         }
 
         // Add blocks to the window, starting from `first_block`, until we reach the end of the
         // requested time window.
-        let mut blocks = self.get_block_range(first_block..).await?;
+        let mut blocks = self.get_block_range(first_block..).await;
         while let Some(block) = blocks.next().await {
-            let block = block?;
+            let block = block.await;
             let header = block.header().clone();
             if header.timestamp >= end {
                 res.next = Some(header);
--- a/sequencer/src/api/fs.rs
+++ b/sequencer/src/api/fs.rs
@@ -55,7 +55,7 @@ impl SequencerDataSource for DataSource {
         // function is called every time a new block is added so this usually requires loading only
         // one block into memory, and rarely very many.
         let blocks: Vec<_> = self
-            .get_block_range(from_block..)
+            .get_block_range(from_block..self.block_height().await?)
             .await
             .enumerate()
             .collect()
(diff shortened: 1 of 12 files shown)


sequencer/src/api/fs.rs
```diff
@@ -50,15 +50,12 @@ impl SequencerDataSource for DataSource {
...
             .enumerate()
             .collect()
             .await;
-        for (i, block) in blocks {
-            let Ok(block) = block else {
-                tracing::warn!("missing block {}, index may be out of date", from_block + i);
-                continue;
-            };
+        for (_, block) in blocks {
+            let block = block.await;
```
ggutoski, line 58: @jbearer is this change acceptable? Seems that it's no longer possible to have a missing block. Seems good but I want to be sure I've not overlooked something.

mrain (commented):
LGTM on the signature key related parts. Don't have knowledge on other API changes.

mrain (approved):

sequencer/src/api/fs.rs
```diff
@@ -33,9 +33,9 @@ impl SequencerDataSource for DataSource {
         let mut index = Index::default();
 
         // Index blocks by timestamp.
-        let mut blocks = data_source.get_block_range(..).await?;
```
jbearer, line 36: Semantics of `get_block_range` have changed with an unbounded range. It will now continue to fetch blocks as long as you keep polling the stream, which is not what we want here. Instead you'll want something like `data_source.get_block_range(..data_source.block_height().await?).await?`.

sequencer/src/api/fs.rs
```diff
@@ -33,9 +33,9 @@ impl SequencerDataSource for DataSource {
...
+        let mut blocks = data_source.get_block_range(..).await;
         while let Some(block) = blocks.next().await {
-            index_block_by_time(&mut index.blocks_by_time, &block?);
+            index_block_by_time(&mut index.blocks_by_time, &block.await);
```
jbearer, line 38: This `await` will block until the block is fetched, if the block was not available immediately. For the FS backend, fetching isn't really supported at all. It's probably better here to keep the old behavior of failing if the block is not available, with `&block.try_resolve()?`

sequencer/src/api/fs.rs
jbearer: Same as above, if the block is missing this will block until it can be retrieved. I think it would be better here to use `let Ok(block) = block.try_resolve() else {`, to mirror the previous behavior.

sequencer/src/api/fs.rs
```diff
@@ -88,22 +85,22 @@ impl SequencerDataSource for DataSource {
     {
         let first_block = match from.into() {
             ResourceId::Number(n) => n,
-            ResourceId::Hash(h) => self.get_block(h).await?.height() as usize,
+            ResourceId::Hash(h) => self.get_block(h).await.await.height() as usize,
```
jbearer, line 88: Same

sequencer/src/api/fs.rs
```diff
@@ -88,22 +85,22 @@ impl SequencerDataSource for DataSource {
...
         };
 
         let mut res = TimeWindowQueryData::default();
 
         // Include the block just before the start of the window, if there is one.
         if first_block > 0 {
-            let prev = self.get_block(first_block - 1).await?;
+            let prev = self.get_block(first_block - 1).await.await;
```
jbearer, line 95: Same

sequencer/src/api/fs.rs
```diff
@@ -88,22 +85,22 @@ impl SequencerDataSource for DataSource {
...
             res.prev = Some(prev.header().clone());
         }
 
         // Add blocks to the window, starting from `first_block`, until we reach the end of the
         // requested time window.
-        let mut blocks = self.get_block_range(first_block..).await?;
+        let mut blocks = self.get_block_range(first_block..).await;
         while let Some(block) = blocks.next().await {
-            let block = block?;
+            let block = block.await;
```
jbearer, line 103: Same

sequencer/src/api/fs.rs
```diff
@@ -88,22 +85,22 @@ impl SequencerDataSource for DataSource {
...
         if first_block > 0 {
-            let prev = self.get_block(first_block - 1).await?;
+            let prev = self.get_block(first_block - 1).await.await;
             res.prev = Some(prev.header().clone());
         }
 
         // Add blocks to the window, starting from `first_block`, until we reach the end of the
         // requested time window.
-        let mut blocks = self.get_block_range(first_block..).await?;
+        let mut blocks = self.get_block_range(first_block..).await;
```
jbearer, line 101: `first_block..self.block_height().await?`

sequencer/src/api/sql.rs
```diff
@@ -128,7 +128,7 @@ impl SequencerDataSource for DataSource {
         // Find the specific block that starts the requested window.
         let first_block = match from.into() {
             ResourceId::Number(n) => n,
-            ResourceId::Hash(h) => self.get_block(h).await?.height() as usize,
+            ResourceId::Hash(h) => self.get_block(h).await.await.height() as usize,
```
jbearer, line 131: Same

sequencer/src/api/sql.rs
```diff
@@ -157,7 +157,7 @@ impl SequencerDataSource for DataSource {
 
         // Find the block just before the window.
         let prev = if first_block > 0 {
-            let prev = self.get_block(first_block - 1).await?;
+            let prev = self.get_block(first_block - 1).await.await;
```
jbearer, line 160: Same

sequencer/src/bin/submit-transactions.rs
```diff
@@ -128,7 +128,7 @@ async fn main() {
         }
 
         // Clear pending transactions from the block.
-        for (_, tx) in block.payload().enumerate() {
+        for (_, tx) in block.payload().enumerate(block.metadata()) {
```
jbearer, line 131: You can use `block.enumerate()` to be a bit more concise

sequencer/src/block2.rs
```diff
@@ -886,31 +889,31 @@ mod test {
             );
 
             let block = BlockPayload::from_bytes(test_case.payload);
-            assert_eq!(block.len(), test_case.num_txs);
+            // assert_eq!(block.len(), test_case.num_txs);
```
philippecamacho, line 892: Why is this assert commented?

ggutoski: Because it was broken by recent upstream API changes. It is a work in progress and will be fixed in the near future. The purpose of this PR is to update dependencies.

philippecamacho: Got it, thanks.

Issue #943, opened 2024-01-08
Implement fee state


nomaxg:
Add a [SparseMerkleTree](https://github.com/EspressoSystems/jellyfish/blob/main/primitives/src/merkle_tree/universal_merkle_tree.rs) to the block header representing fee state. The key will be an ECSDA-style[ Ethereum address](https://github.com/EspressoSystems/sequencer-example-l2/blob/main/src/state.rs#L8), and the value will be the balance (this can be u64 for now). 

Add some data structure like:
```
struct FeeReceipt {
   recipient: Address, 
   amount: Amount
}
```

Eventually, we will have logic that scans the L1 to create an array of FeeTransfer data for a certain HotShot block type. For now, you can mock this function with a stub like:

```
fetch_fee_receipts() -> Vec<FeeReceipt> // This can return mock data for now
```

So using this mock function, when creating a new header, update the fee state tree with updated balances.

PR #989, opened 2024-01-15
Add `fetch_fee_receipts` add call it from `Header` methods

tbro:
Rought draft implementation of `FeeState` (#943). As of yet, we are using naive types for key and values in order to implement the basic structure. It appears that `ethers::abi::Address` and `ethers::types::U256` will require impls for
`CanonicalSerialize`. Assuming the general structure here is correctish, that would be the next step. In addtion, a clippy annotation was added to work around `too_many_arguments` error for `Header::from_info()`. Should be encapsulate the state headers in a type to avoid this error, or is the clippy annotation good enough?


sequencer/src/block.rs
```diff
@@ -210,6 +232,22 @@ impl BlockHeader for Header {
         block_merkle_tree.push(parent.commit()).unwrap();
         let block_merkle_tree_root = block_merkle_tree.commitment();
 
+        // fetch receipts from the l1
+        let receipts = fetch_fee_receipts();
```
philippecamacho, line 236: Is the ethereum block number the espresso block is tied to known before calling this function?

sequencer/src/block.rs
```diff
@@ -210,6 +232,22 @@ impl BlockHeader for Header {
...
+
+        let mut fee_merkle_tree = parent.fee_merkle_tree.clone();
+        for FeeReceipt { recipient, amount } in receipts {
+            // get the balance in order to add amount. currently we
+            // are ignoring the proof, but we should probably verify
+            // it.
```
philippecamacho, line 242: Why do we need to use the proof? The replica trusts its local data right?

sequencer/src/block.rs
tbro: I believe you are referring to `l1_head` field on the `Header`. Which if I understand should be available here as `parent.l1_head`. I may be mistaken though.

sequencer/src/block.rs
```diff
@@ -210,6 +231,23 @@ impl BlockHeader for Header {
...
+            // Get the balance in order to add amount. currently we
+            // are ignoring the proof. Should we verify it?
+            match fee_merkle_tree.universal_lookup(recipient) {
+                LookupResult::Ok(balance, _) => {
+                    fee_merkle_tree.update(recipient, balance + amount).unwrap()
+                }
+                // Handle `NotFound` and `NotInMemory` by initializing
+                // state. Should these be handled in different ways?
+                _ => fee_merkle_tree.update(recipient, amount).unwrap(),
+            };
```
jbearer, line 248: Not a big deal now, but this incurs an unnecessary tree traversal: one to lookup and one to update. Maybe we could add an API to jellyfish that takes a closure to update the existing value and a default value to insert if not existing? Then this update could all be done in a single pass. Not urgent. cc @mrain

mrain: How about an interface like
```Rust
fn update(&mut self, f: F)
where
    F: FnOnce(Option<&E>) -> Option<E>;
```

`None` means that the key doesn't exist or we are removing it.

sequencer/src/block.rs
jbearer: Yes, although we should use `l1_finalized` instead, so we don't have to deal with L1 reorgs

sequencer/src/block.rs
```diff
@@ -30,13 +30,27 @@ struct FeeReceipt {
...
 // ethers::types::U256
 pub type FeeMerkleTree = UniversalMerkleTree<u64, Sha3Digest, u64, U3, Sha3Node>;
+pub type FeeMerkleCommtment = <FeeMerkleTree as MerkleTreeScheme>::Commitment;
 
-fn fetch_fee_receipts(_parent: &Header) -> Vec<FeeReceipt> {
-    Vec::from([FeeReceipt::default()])
+#[derive(Hash, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ValidatedState {
+    /// Root Commitment of Block Merkle Tree
+    pub block_merkle_tree_root: BlockMerkleCommitment,
```
jbearer, line 47: We want to keep these roots in the header itself...since ValidatedState will eventually be moved out of the header but we want the roots in there

sequencer/src/block.rs
```diff
@@ -227,17 +230,21 @@ fn _validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<Bloc
...
-        let mut block_merkle_tree = parent.block_merkle_tree.clone();
+        let ValidatedState {
+            mut block_merkle_tree,
+            mut fee_merkle_tree,
+            ..
+        } = parent.validated_state.clone();
         block_merkle_tree.push(parent.commit()).unwrap();
+
         let block_merkle_tree_root = block_merkle_tree.commitment();
+        let fee_merkle_tree_root = fee_merkle_tree.commitment();
```
jbearer, line 241: We need to get the fee root _after_ inserting all the new fee receipts, right?

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+}
+impl std::convert::From<u64> for FeeIndex {
+    fn from(item: u64) -> Self {
+        FeeIndex(Address::from_low_u64_le(item))
+    }
+}
+
+impl<U2: Unsigned> ToTraversalPath<U2> for FeeIndex {
+    fn to_traversal_path(&self, height: usize) -> Vec<usize> {
+        let mut pos = Address::to_low_u64_le(&self.0);
```
mrain, line 124: How do we interpret this `Address` type? 
Are the lowest 64 bits sufficient to represent an ether address?

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+        let mut ret = vec![];
+        for _i in 0..height {
+            let x = (pos % (U2::to_u64())) as usize;
```
mrain, line 127: Since the arity is 2 here. We may replace this repeated mod and division by a simple `to_le_bits` function. However, I could only find `to_le_bytes`.

This is already good enough though.

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+// New Type for `U256` in order to implement `CanonicalSerialize` and
+// `CanonicalDeserialize`
+#[derive(Default, Hash, Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct FeeElement(U256);
+// New Type for `Address` in order to implement `CanonicalSerialize` and
+// `CanonicalDeserialize`
+#[derive(
+    Default, Hash, Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord,
+)]
+pub struct FeeIndex(Address);
```
jbearer, line 41: Nit: the `Index` here refers to only one usage of this type, where we use it as an index in a key-value map. But we may well end up using this type elsewhere, any time we deal with fees. WDYT about naming this something more descriptive like `FeeAccount` or `FeeAddress`?

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+use typenum::U2;
 
 pub type BlockMerkleTree = LightWeightSHA3MerkleTree<Commitment<Header>>;
 pub type BlockMerkleCommitment = <BlockMerkleTree as MerkleTreeScheme>::Commitment;
 
-#[derive(Default)]
+// New Type for `U256` in order to implement `CanonicalSerialize` and
+// `CanonicalDeserialize`
+#[derive(Default, Hash, Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct FeeElement(U256);
```
jbearer, line 35: Same comment here, maybe we could call this like `FeeAmount` or something

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+// New Type for `Address` in order to implement `CanonicalSerialize` and
+// `CanonicalDeserialize`
+#[derive(
+    Default, Hash, Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord,
+)]
+pub struct FeeIndex(Address);
+
+impl Add for FeeElement {
```
jbearer, line 43: Nit: could use [derive_more](https://docs.rs/derive_more/latest/derive_more/) for this

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+}
+
+impl CanonicalSerialize for FeeElement {
+    fn serialize_with_mode<W: std::io::prelude::Write>(
+        &self,
+        mut writer: W,
+        _compress: Compress,
+    ) -> Result<(), SerializationError> {
+        let mut bytes = [0u8; core::mem::size_of::<U256>()];
+        self.0.to_little_endian(&mut bytes);
```
jbearer, line 70: Nit: maybe cleaner to do `let bytes = self.0.0`?

sequencer/src/block.rs
```diff
@@ -17,28 +19,131 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+        _compress: Compress,
+        _validate: Validate,
+    ) -> Result<Self, SerializationError> {
+        let mut bytes = [0u8; core::mem::size_of::<Address>()];
+        reader.read_exact(&mut bytes)?;
+        let value = Address::from_slice(&bytes);
+        Ok(Self(value))
+    }
+}
+impl std::convert::From<u64> for FeeIndex {
```
jbearer, line 116: Why do we want this conversion?

sequencer/src/block.rs
jbearer: Yeah, we should use [to_fixed_bytes](https://docs.rs/ethers/latest/ethers/core/abi/type.Address.html#method.to_fixed_bytes) instead

sequencer/src/block.rs
tbro: It is a requirement of `UniversalMerkleTreeScheme`: https://github.com/EspressoSystems/jellyfish/blob/bf1a6dfb63e2ce9b409a682d2793b836f1659097/primitives/src/merkle_tree/universal_merkle_tree.rs#L34

sequencer/src/block.rs
```diff
@@ -112,16 +111,13 @@ impl std::convert::From<u64> for FeeAccount {
...
     fn to_traversal_path(&self, height: usize) -> Vec<usize> {
-        let mut pos = Address::to_low_u64_le(&self.0);
-        let mut ret = vec![];
-        for _i in 0..height {
-            let x = (pos % (U2::to_u64())) as usize;
-            ret.push(x);
-            pos /= U2::to_u64();
-        }
-        ret
+        Address::to_fixed_bytes(self.0)
```
mrain, line 116: LGTM

sequencer/src/block.rs
tbro: I'm not sure if it is really much cleaner as as the internal type of `U256` is `&[u64: 4]`.

sequencer/src/block.rs
```diff
@@ -19,10 +20,25 @@ use std::{
...
+    recipient: u64, // FIXME `Address`
+    amount: u64,    // FIXME `U256`
+}
+
+// TODO use U256. We need to impl `ark_serialize::CanonicalDeserialize`
+// ethers::types::U256
+pub type FeeMerkleTree = UniversalMerkleTree<u64, Sha3Digest, u64, U3, Sha3Node>;
+
+fn fetch_fee_receipts(_parent: &Header) -> Vec<FeeReceipt> {
+    Vec::from([FeeReceipt::default()])
```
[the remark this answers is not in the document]
tbro: Is the spirit of this comment just to inform us of future changes that will be necessary, or should we go ahead and move the mock function to where the real function will eventually live?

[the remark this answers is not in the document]
jbearer: Nope, just wanted to let you know. No action needed for now.

sequencer/src/block.rs
jbearer: Oh right. Never mind

sequencer/src/block.rs
jbearer: Ah right. The thing that needs to be implemented but doesn't really get used. Np

sequencer/src/block.rs
jbearer: @tbro can you move these roots back to the header proper, and add them to the `impl Committable for Header`? After that I think we can merge this PR

tbro: Will do.

jbearer (approved):
The change look good, though there seems to be a bug in some merkle tree indexing. I think we are good to merge as soon as the tests are passing.

Issue #923, opened 2024-01-04
[light-client-integration] add stake table commitment in light client state.


mrain:
After the signature key type is stable in HotShot repo.

mrain:
Waiting for https://github.com/EspressoSystems/HotShot/issues/2328

PR #995, opened 2024-01-16
[light-client-integration] add stake table commitment

mrain:
closes: #923 

Integrity test shall be included when we implement the light client prover service.

### DIFF
--- a/sequencer/src/state_signature.rs
+++ b/sequencer/src/state_signature.rs
@@ -2,8 +2,13 @@
 
 use crate::context::SequencerContext;
 use crate::{network, Leaf, SeqTypes};
+use ethers::types::U256;
 use futures::stream::{Stream, StreamExt};
-use hotshot::types::Event;
+use hotshot::types::{Event, SignatureKey as _};
+use hotshot_stake_table::vec_based::StakeTable;
+use hotshot_types::light_client::StateVerKey;
+use hotshot_types::signature_key::BLSPubKey;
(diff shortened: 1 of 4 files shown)


alxiong (approved):
I have no issue, LGTM.

sequencer/src/state_signature.rs
```diff
@@ -79,3 +84,25 @@ impl StateSignatureMemStorage {
...
+/// Helper function for stake table commitment
+pub(crate) fn static_stake_table_commitment(
+    seed: [u8; 32],
+    num_nodes: usize,
+    capacity: usize,
+) -> (FieldType, FieldType, FieldType) {
+    let mut st = StakeTable::<BLSPubKey, StateVerKey, FieldType>::new(capacity);
+    (0..num_nodes).for_each(|i| {
+        let bls_key = BLSPubKey::generated_from_seed_indexed(seed, i as u64).0;
+        let state_key = StateKeyPair::generate_from_seed_indexed(seed, i as u64).ver_key();
```
jbearer, line 100: This might break when we change key management, so that nodes pass in their keys instead of generating them from seeds. This function should probably just take `known_nodes_with_stake` (ie the stake table itself) and compute the comitment from that

jbearer (commented):
One change request, everything else looks good

sequencer/src/state_signature.rs
mrain: Changed in 96d2af1

PR #996, opened 2024-01-16
Get rid of etherscan warnings

### LINKED ISSUE #979: Get rid of etherscan warning when running the tests

philippecamacho:
Closes #979 

The solution is a bit hacky but I think there is not much that we can do about it for now.
It consists of specifying an invalid etherscan api key when calling `forge test`.


alysiahuggins:
lgtm, we can include an api key for etherscan when the contracts are deployed

alysiahuggins (approved):

alxiong:
locally when I run, I still see 

> WARN etherscanidentifier: rate limit exceeded on attempt

Do I need to configure differently? or adding anything to my `.env` file?

philippecamacho:
> locally when I run, I still see
> 
> > WARN etherscanidentifier: rate limit exceeded on attempt
> 
> Do I need to configure differently? or adding anything to my `.env` file? @philippecamacho

That should not be needed. Maybe run `forge cache clean` or clone again the repository.

Locally the only warning that I see on my machine is
```
2024-01-22T15:49:01.238168Z  WARN etherscanidentifier: invalid api key
```

Which is ok/expected.

alxiong:
for the record, the root of the problem comes from us specifying a `ETHERSCAN_API_KEY=` in `.env` file, 
if we don't add this line, there won't be any warning. (which is why I can't find much issue reported on foundry-rs repo, most ppl don't have that env var set)

since we want to leave it there, we can set it to some "invalid key/placeholder value" for now. which I did in #1017 .

PR #1001, opened 2024-01-16
Update Jellyfish

jbearer:
This fixes the linearly growing frontier bug.


Ayiga:
Looks Better now:
image

PR #1007, opened 2024-01-19
Bump solc version

jbearer:
I was running `just bind` in [espresso-sequencer-go](https://github.com/EspressoSystems/espresso-sequencer-go), which does `forge build --force` in this repo (as a submodule). It wasn't working, and I realized that when I run `forge build` manually from a Nix shell, it uses solc 0.8.23, which works, but when I run it via `just`, it uses 0.8.20, which doesn't work. I'm not exactly sure what's going on, but this change seems to fix the problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,9 +31,6 @@ jobs:
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
-        # TODO: remove version pinning once our tests work again
-        with:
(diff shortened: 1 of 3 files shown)


philippecamacho:
I don't think this might cause any problem as foundry uses the solc binary defined in the flake.nix file as well:
```
[nix-shell:~/repositories/espressosys/espresso-sequencer]$ echo $FOUNDRY_SOLC 
/nix/store/90xsdmak5f2s0kq54qg9rzf4bmw781fb-solc-bin-0.8.23/bin/solc
```

philippecamacho (approved):
LGTM

PR #1008, opened 2024-01-19
Simplify stake table invariant test

philippecamacho:
See https://github.com/EspressoSystems/espresso-sequencer/pull/878#discussion_r1452947444


alxiong (approved):
thanks!

Issue #1011, opened 2024-01-20
Equivalent of QueryableBlock::transaction_with_proof for namespaces


ggutoski:
Like #1010 except for namespaces. Next step after #1009 . As usual, this new method will not be added upstream to `QueryableBlock`. Instead just add it to the concrete `BlockPayload` struct in this repo. Suggested name: `namespace_with_proof`

philippecamacho:
Wait until https://github.com/EspressoSystems/espresso-sequencer/issues/1011 is done.

Issue #1009, opened 2024-01-20
Equivalents of QueryableBlock::{len, iter} for namespaces


ggutoski:
Like #991, #931 except for namespaces. Should be easy given what we've already done for those issues. Indeed, we should probably factor out the namespace-specific code from whatever code we write for those issues and re-use it here.

Suggested names, but feel free to think of something better:
- len -> num_namespaces
- iter -> namespace_iter

As per https://github.com/EspressoSystems/hotshot-query-service/issues/328#issuecomment-1883785794 these new methods will not be added upstream to `QueryableBlock`. Instead just add them to the concrete `BlockPayload` struct in this repo. (Currently that's in `block2.rs` along with everything else VID.)

Issue #1012, opened 2024-01-20
`Payload` redesign


ggutoski:
`TxTableEntry` is horrible. It was a quick-and-dirty solution devised before we knew what we needed. I now have a better idea of what's needed and am hatching a better design.

# Proposal
- 3 types of table word: `TableLen`, `Offset`, `NsId`
- 2 of these (`TableLen`, `Offset`) must be convertible to/from `usize` for arithmetic and indexing. `NsId` is an arbitrary identifier.
- It should be possible to vary these 3 types independently.
- Lean on the type system to help us avoid confusing the word types.
- Make ns/tx tables generic over these types, like
```rust
TableLen: CanonicalSerialize + CanonicalDeserialize + TryFrom<usize> + TryInto<usize>,
Offset: CanonicalSerialize + CanonicalDeserialize + TryFrom<usize> + TryInto<usize>,
NsId: CanonicalSerialize + CanonicalDeserialize,
```
- All 3 types might be the same (4 bytes) or different: 3-bytes `TableLen`, 4-bytes `Offset`, 32-bytes `NsId`.
- Downstream users could use any type that imples these traits, like
```rust
TableLen = [u8; 3];
Offset = u32;
NsId = Output<H>; // or [u8; 32] or whatever
```
- To avoid generic parameter hell we could have a single generic param `BlockPayloadTableWords` (better name?) that has 1 assoc type for each of the 3 word types listed above.
- `TryFrom<usize>` is far less ergonomic than `From<usize>`. Example: we use it inside an impl of `Iterator`, which does not afford us the luxury to propagate errors up the stack. For these cases we'll just use `unwrap` anyway, which defeats the entire point.
  - Sadly, `From<usize>` is not supported by most primitive types, so we cannot use it as a trait bound because of foreign trait rules.
  - It would be nice if `TryFrom` would use infallible `!` error type for conversions that cannot fail, because then we could exploit this fact in trait bounds. Alas, that is [not to be](https://users.rust-lang.org/t/shouldnt-the-implementations-of-tryfrom-for-integer-types-be-infallible-for-unbounded-conversions/79635).
  - Our remaining options: panic when conversion fails, or use `as` keyword and tell users you're on your own. I propose we just panic on failure. Anything else would lead to horrible bugs.

ggutoski:
> > * 2 of these (`TableLen`, `Offset`) must be convertible to/from `usize` for arithmetic and indexing. `NsId` is an arbitrary
> 
[...]

Issue #951, opened 2024-01-08
[priority code clean-up] Split `block2.rs` into multiple files


ggutoski:
The correct way to split will depend on how various helpers are re-used. Candidate breakdown:
## Root file
- keep it short? eg. Define `BlockPayload` struct and maybe a couple widely used helpers?
- maybe impl the `hotshot::BlockPayload` trait for `BlockPayload` here, too.
## Separate files
- `TxTableEntry` ick #1012
- impl `QueryablePayload` and `TxIterator`. We could furthersplit `TxIterator` into another file if desired.
- Tests are tricky: it's useful to have a big "everything" test that iterates through a bunch of test payloads and does everything to them. Where to put this thing?

PR #1014, opened 2024-01-22
feat: namespace-aware tx iterator

### LINKED ISSUE #931: Support tx iterator with namespaces
Blocker for #892 because we have no easy way to iterate through the txs in a block. We need this to test transaction proofs via `verify()`.

From https://github.com/EspressoSystems/espresso-sequencer/issues/892#issuecomment-1878978395 :

> Ugh the impl for `QueryablePayload` needs a complete rewrite for this issue:
> - `len()` is more complex now because we need to sum the number of txs over all namespaces
> - `iter()` is more complex: need to iterate over `([namespace id], [tx index])`

akonring:
closes #931

### DIFF
--- a/sequencer/src/block2.rs
+++ b/sequencer/src/block2.rs
@@ -330,7 +330,7 @@ impl QueryablePayload for BlockPayload {
         result
     }
 
-    fn iter(&self, meta: &Self::Metadata) -> Self::Iter<'_> {
+    fn iter<'a>(&'a self, meta: &'a Self::Metadata) -> Self::Iter<'a> {
         TxIterator::new(meta, self)
     }
 
@@ -638,25 +638,21 @@ pub struct TxIndex {
 }
 
 pub struct TxIterator<'a> {
-    ns_idx: usize, // TODO should be able to eliminate this field
+    ns_idx: usize, // simpler than using `Peekable`
     ns_iter: Range<usize>,
     tx_iter: Range<usize>,
     block_payload: &'a BlockPayload,
-
-    // TODO should be `&'a NsTable`, but that requires upstream change to `QuearyablePayload::iter`:
-    // fn iter<'a>(&'a self, meta: &'a Self::Metadata)
-    //        ++++  ++              ++
-    ns_table: NsTable,
+    ns_table: &'a NsTable,
 }
 
 impl<'a> TxIterator<'a> {
-    fn new(ns_table: &NsTable, block_payload: &'a BlockPayload) -> Self {
+    fn new(ns_table: &'a NsTable, block_payload: &'a BlockPayload) -> Self {
         Self {
-            ns_idx: 0,
+            ns_idx: 0, // arbitrary value, changed in first call to next()
             ns_iter: 0..get_ns_table_len(ns_table),
             tx_iter: 0..0, // empty range
             block_payload,
-            ns_table: ns_table.clone(),
+            ns_table,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


ggutoski (commented):
Thanks for this! A few notes below.

sequencer/src/block2.rs
```diff
@@ -555,6 +638,104 @@ mod tx_table_entry {
     }
 }
 
+type NsTable = <BlockPayload as hotshot::traits::BlockPayload>::Metadata;
+type TxIndexUnsigned = u32; // need a better name, want `TxIndex` but that's already taken.
+
+struct TxIndex2 {
+    ns_id: VmId,
```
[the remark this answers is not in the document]
ggutoski:
```suggestion
    ns_idx: usize,
```

sequencer/src/block2.rs
```diff
@@ -170,6 +171,91 @@ fn get_table_len(table_bytes: &[u8], offset: usize) -> TxTableEntry {
...
+// TODO comment
+fn get_tx_table_len(ns_bytes: &[u8]) -> usize {
+    std::cmp::min(
+        get_table_len(ns_bytes, 0).try_into().unwrap_or(0),
+        (ns_bytes.len() - TxTableEntry::byte_len()) / TxTableEntry::byte_len(),
+    )
+}
+
+// returns (ns_id, payload_offset)
+// payload_offset is not checked, could be anything
```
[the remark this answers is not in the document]
ggutoski: It means that this function parses whatever's in the namespace table as an offset. So if that data is corrupted then the returned offset is corrupted.

This should never happen now that the namespace table is part of the header, so consensus nodes enforce a well-formed namespace table. But I want this code / docs to be paranoid.

sequencer/src/block2.rs
```diff
@@ -555,6 +627,83 @@ mod tx_table_entry {
     }
 }
 
+type NsTable = <BlockPayload as hotshot::traits::BlockPayload>::Metadata;
+
+/// TODO do we really need `PartialOrd`, `Ord` here?
+/// Could the `Ord` bound be removed from `QueryablePayload::TransactionIndex`?`
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
```
ggutoski, line 634: @jbearer question for you

sequencer/src/block2.rs
```diff
@@ -555,6 +627,83 @@ mod tx_table_entry {
...
+pub struct TxIterator<'a> {
+    ns_idx: usize, // TODO should be able to eliminate this field
+    ns_iter: Range<usize>,
+    tx_iter: Range<usize>,
+    block_payload: &'a BlockPayload,
+
+    // TODO should be `&'a NsTable`, but that requires upstream change to `QuearyablePayload::iter`:
+    // fn iter<'a>(&'a self, meta: &'a Self::Metadata)
+    //        ++++  ++              ++
+    ns_table: NsTable,
```
ggutoski, line 649: @jbearer looks like we need another tweak to `QueryablePayload` 😕

ggutoski (commented):
I rebased main and force-pushed new commits to this branch. Need approval from @jbearer and/or @nomaxg .

We are currently pushing to close #1009 , #1011 . Some code from this PR will be used to close these subsequent issues.

Code still needs significant tidying. We already have a few issues posted about this #856 #951 #1012 that will be addressed after we close #1011 . In the meantime, I ask that you be forgiving about code cleanliness.

Flagged a couple notes for @jbearer in comments below.

sequencer/src/block2.rs
jbearer: Ah, that looks like a good change. I can do it, should be quick 🤞

jbearer (approved):
LGTM

Issue #933, opened 2024-01-08
[light-client-integration] schnorr signature relay server


mrain:
We need a schnorr signature relay server that:

- All nodes submit their signatures to it.
- Watch and maintain a stake table. (Currently it's fixed)
- Serve the latest light client state and signatures whose total weight exceeds the threshold.

PR #1016, opened 2024-01-23
feat(light-client-intergration): light client state signature relay server

mrain:
closes: #933 

What's included in this PR:

- implements a relay server that collects signatures from the nodes and serves it to the prover
- relay server integration: add logic for each node that sends their light client state and signature to the relay server
- environment variables and docker building script for the relay server

What's not included:

- The relay server is not maintaining a stake table, thus no authentication.


sequencer/api/state_relay_server.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[route.poststatesignature]
+PATH = ["state/post"]
+METHOD = "POST"
+DOC = """
+Post a light client state and its signature for a given block height.
+"""
+
+[route.getlateststate]
+PATH = ["state/getlatest"]
```
alxiong, line 14: paths for both apis can be `PATH = ["state/"]`, seem clear enough?

docker-compose.yaml
```diff
@@ -53,6 +53,16 @@ services:
       orchestrator:
         condition: service_healthy
 
+  state-relay-server:
+    image: ghcr.io/espressosystems/espresso-sequencer/state-relay-server:main
+    ports:
+      - "$ESPRESSO_STATE_RELAY_SERVER_PORT:$ESPRESSO_STATE_RELAY_SERVER_PORT"
+    environment:
+      - $ESPRESSO_STATE_RELAY_SERVER_PORT
+      - $ESPRESSO_STATE_SIGNATURE_WEIGHT_THRESHOLD
```
jbearer, line 62: I may just not be familiar with the syntax but I don't think this should have a `$` in front?

process-compose.yaml
```diff
@@ -66,6 +67,18 @@ processes:
         port: $ESPRESSO_CONSENSUS_SERVER_PORT
         path: /healthcheck
 
+  state-relay-server:
+    command: state-relay-server
+    environment:
+      - ESPRESSO_STATE_RELAY_SERVER_PORT:$ESPRESSO_STATE_RELAY_SERVER_PORT
+      - ESPRESSO_STATE_SIGNATURE_WEIGHT_THRESHOLD:$ESPRESSO_STATE_SIGNATURE_WEIGHT_THRESHOLD
```
jbearer, line 74: If we're just passing through the value from the environment I think you can just put its name, no need to give the value

sequencer/api/state_relay_server.toml
```diff
@@ -0,0 +1,19 @@
+[meta]
+NAME = "hotshot_state_relay_server"
+DESCRIPTION = "A relay server for storing and serving the light client states and their signatures"
+FORMAT_VERSION = "0.1.0"
+
+[route.poststatesignature]
+PATH = ["state/post"]
```
jbearer, line 7: Nit: it would be more conventional for the route to just be `state` with a method of `POST`. Putting `post` also in the route path is redundant

sequencer/api/state_relay_server.toml
jbearer, line 14: Same nit: could just be GET `state`

sequencer/src/bin/state-relay-server.rs
```diff
@@ -0,0 +1,43 @@
...
+    setup_logging();
+    setup_backtrace();
+
+    let args = Args::parse();
+
+    tracing::info!("starting state relay server on port {}", args.port);
+    run_relay_server(
+        None,
+        args.threshold,
+        format!("http://0.0.0.0:{}", args.port).parse().unwrap(),
```
jbearer, line 39: Remove `http://`, the server uses HTTP no matter what, and some platforms don't like this when they parse the socket address

sequencer/src/state_signature.rs
```diff
@@ -39,13 +48,36 @@ pub(super) async fn state_signature_loop<N>(
         } = event
         {
             if let Some(leaf) = leaf_chain.first() {
-                tracing::debug!("New leaves decided. Newest leaf: {:?}", leaf);
-                let new_state = form_light_client_state(leaf, stake_table_comm);
-                context.sign_new_state(&new_state);
+                let state = form_light_client_state(leaf, stake_table_comm);
+                let signature = context.sign_new_state(&state).await;
+                tracing::info!(
```
jbearer, line 53: This should probably also be `debug!` since it happens every decide

sequencer/src/state_signature.rs
```diff
@@ -39,13 +48,36 @@ pub(super) async fn state_signature_loop<N>(
...
+                    "New leaves decided. Latest block height: {}, posting to relay server {:?}",
+                    leaf.get_height(),
+                    state_relay_server_url,
+                );
+
+                if let Some(state_relay_server_url) = state_relay_server_url {
+                    let client: Client<ServerError> = Client::new(state_relay_server_url.clone());
+                    client.connect(None).await;
```
jbearer, line 61: Would be good to create this client and `connect` once before entering the loop. You could create an `Option<Client>` at the top and then just `if let Some(client) = client_opt` to avoid reconnecting each time

sequencer/src/state_signature.rs
```diff
@@ -39,13 +48,36 @@ pub(super) async fn state_signature_loop<N>(
...
+                        .await
+                    {
+                        tracing::warn!("Error posting signature to the relay server: {:?}", error);
+                    }
+                }
             }
         }
     }
-    tracing::debug!("And now his watch has ended.");
+    tracing::info!("And now his watch has ended.");
```
jbearer, line 80: Incredible thing to print as the node shuts down, 10/10 no notes

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+struct StateRelayServerState {
+    /// Minimum weight to form an available state signature package
+    threshold: U256,
+    /// Stake table
+    known_nodes: HashMap<StateVerKey, U256>,
+    /// Signatures packages for each block height
+    packages: HashMap<u64, StateSignaturesPackage>,
+
+    /// The latest state signatures package whose total weight exceeds the threshold
+    latest_available_package: Option<StateSignaturesPackage>,
```
jbearer, line 49: Could we remove this? Is it equivalent to `packages[latest_block_height]`?

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        &mut self,
+        key: StateVerKey,
+        state: LightClientState,
+        signature: StateSignature,
+    ) -> Result<(), Error> {
+        if (state.block_height as u64) <= self.latest_block_height.unwrap_or(0) {
+            return Err(tide_disco::error::ServerError::catch_all(
+                StatusCode::Locked,
+                "The posted signature is no longer needed.".to_owned(),
+            ));
```
jbearer, line 120: Should this be considered an error vs just returning with no effect? The client didn't do anything wrong, they were just slow, and the thing the client cares about -- generating a signature package for this block height or later -- has in fact happened. So there would not be any need e.g. for a client to log an error or retry the request.

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        let one = U256::one();
+        let weight = self.known_nodes.get(&key).unwrap_or(&one);
```
jbearer, line 123: Shouldn't the weight be 0 if the node is not in the stake table?

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+            self.queue.insert(block_height);
+            StateSignaturesPackage {
+                state,
+                signatures: Default::default(),
+                accumulated_weight: U256::from(0),
+            }
+        });
+        let incoming_bytes = to_bytes!(&package.state).unwrap();
+        if bytes != incoming_bytes {
+            return Err(tide_disco::error::ServerError::catch_all(
```
jbearer, line 150: Is there an attack here where if the first node to post a signature signed the wrong state, all subsequent signatures on the correct state will be rejected? I think to prevent this we would have to keep track of all states we have seen for a given block height until one of them has a quorum of signatures (or at least f+1 signatures). It's a bit complicated but maybe necessary?

jbearer (changes requested):

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+    api::ApiError,
+    error::ServerError,
+    method::{ReadState, WriteState},
+    Api, App, Error as _, StatusCode,
+};
+use url::Url;
+
+/// The state signatures package is a light client state and its signatures collected
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct StateSignaturesPackage {
```
alxiong, line 28: Package sounds like a network package, what about `StateSignatureQuorum` or sth?

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        self
+    }
+}
+
+// TODO(Chengyu): move this `RwLock` inside `StateRelayServerState` so that when nodes are submitting
+//                signatures, it won't block the prover from fetching the available signatures.
+type State = RwLock<StateRelayServerState>;
+type Error = ServerError;
+
+pub trait StateRelayServerDataSource {
```
alxiong, line 82: why do we need this trait?

sequencer/src/state_signature.rs
mrain: Does it mean that the server has to maintain many active connections? Although it might be okay for now.
Also will it reconnect if the connection is lost?

sequencer/src/state_signature/relay_server.rs
mrain: `known_nodes` is empty here, so we can't do authentication. Once we have a stake table maintained, we should use the code commented below.

sequencer/src/state_signature/relay_server.rs
mrain: I'm hoping in the future we could push `RwLock` inside this state struct. So that when nodes are submitting signatures and locking `packages`, prover could still read the latest package here.

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+                accumulated_weight: U256::from(0),
+            }
+        });
+        let incoming_bytes = to_bytes!(&package.state).unwrap();
+        if bytes != incoming_bytes {
+            return Err(tide_disco::error::ServerError::catch_all(
+                StatusCode::BadRequest,
+                "The posted light client is not compatible with the one previously stored."
+                    .to_owned(),
+            ));
```
alxiong, line 154: what if the first posted signature is attached with the wrong state for that block height, (i.e. signed by a dishonest replica to occupy the seat for that block height, which invalidate all subsequent honest submissions)?

sequencer/src/state_signature/relay_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        match package.signatures.entry(key) {
+            std::collections::hash_map::Entry::Occupied(_) => {
+                return Err(tide_disco::error::ServerError::catch_all(
+                    StatusCode::BadRequest,
+                    "A signature is already posted at this block height for this key.".to_owned(),
+                ))
```
alxiong, line 161: should we simply ignore here?

sequencer/src/state_signature/relay_server.rs
mrain: Yes you are right. I was a bit lazy before and ignored this attack.

alxiong (commented):
> What's not included:
>  *  The relay server is not maintaining a stake table, thus no authentication.

We only have a static stake table for cappuccino anyway, as long as we have 2 APIs reserved at the relay server side:
1) the stake table to start with
2) an async function to sync stake table which is left as trivially returning the same static table for now
(^^ WDYT?)

mrain:
> > What's not included:
> > 
> > * The relay server is not maintaining a stake table, thus no authentication.
> 
> We only have a static stake table for cappuccino anyway, as long as we have 2 APIs reserved at the relay server side:
> 
> 1. the stake table to start with
> 2. an async function to sync stake table which is left as trivially returning the same static table for now
>    (^^ WDYT?)

For 1, it's not easy to configure the stake table in the current deployment. I assume `orchestrator` should do the job, but not for now.

sequencer/src/state_signature/relay_server.rs
jbearer: Ah,  good idea

sequencer/src/state_signature/relay_server.rs
jbearer: oh lol

sequencer/src/state_signature/relay_server.rs
mrain: Similar structures as https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/api/data_source.rs

sequencer/src/state_signature/relay_server.rs
mrain: Rename it to bundle in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/api/state_relay_server.toml
mrain: changed in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/state_signature/relay_server.rs
mrain: Fix this by adding another HashMap: [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/state_signature/relay_server.rs
mrain: Fixed by another HashMap: [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/state_signature/relay_server.rs
mrain: Ignored: [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/state_signature/relay_server.rs
mrain: Returning `Ok(())` now: [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

docker-compose.yaml
mrain: Right, removed in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/api/state_relay_server.toml
mrain: changed in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/api/state_relay_server.toml
mrain: changed in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/bin/state-relay-server.rs
mrain: removed in [1dde57c](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/1dde57c61e4683b1fde67bb5bde9e666a563311b)

sequencer/src/state_signature/relay_server.rs
```diff
@@ -136,49 +133,41 @@ impl StateRelayServerDataSource for StateRelayServerState {
...
-        match package.signatures.entry(key) {
+            });
+        match bundle.signatures.entry(key) {
             std::collections::hash_map::Entry::Occupied(_) => {
-                return Err(tide_disco::error::ServerError::catch_all(
-                    StatusCode::BadRequest,
-                    "A signature is already posted at this block height for this key.".to_owned(),
-                ))
+                // A signature is already posted for this key with this state
+                return Ok(());
```
jbearer, line 150: In this case I think we could return an error or at least log something, since it is the client's own fault that they sent a duplicate signature

jbearer (approved):
LGTM

sequencer/src/bin/state-relay-server.rs
mrain: Wait, server wont start after removing `http://`
```bash
state-relay-server-1   | Message:  called `Result::unwrap()` on an `Err` value: RelativeUrlWithoutBase
state-relay-server-1   | Location: sequencer/src/bin/state-relay-server.rs:39
state-relay-server-1   | 
state-relay-server-1   | Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
state-relay-server-1   | Run with RUST_BACKTRACE=full to include source snippets.

```

mrain: I think we should keep this `http://`

sequencer/src/state_signature/relay_server.rs
mrain: Added back.
To further justify this, this error will trigger only if a **VALID** signature of this state is posted more than once. Could be some network issue, or your signing key is leaked.

sequencer/src/state_signature.rs
mrain: addressed in [88a46f6](https://github.com/EspressoSystems/espresso-sequencer/pull/1016/commits/88a46f68497a8124094d2eba688e9a4cbf476815)

sequencer/src/bin/state-relay-server.rs
jbearer: Oh maybe since you're parsing it as a URL, it's ok. The place where this was causing problems was just using a string, which was getting interpreted (wrongly) somewhere inside the web server library

jbearer (approved):

PR #1017, opened 2024-01-23
chore: placeholder for etherscan api key in .env

alxiong:
Follow up on #996 

Instead of having to add `--etherscan-api-key` flag on every command (not just running `just sol-test`, we might as well set the env var to our nix shell with a placeholder key. 

All you need to do is exit and re-enter nix shell with the updated env var. warning should be suppressed as expected.


philippecamacho (approved):
Thanks for the improvement!

PR #1019, opened 2024-01-23
Improvements for `submit-transactions` utility

jbearer:
* Create Docker image and add to local demo
* Improve default options somewhat, particularly handling of delays between transactions. I chose a very low minimum delay and a very high max delay, so we get a lot of variety, and today we encountered a bug which only happened when we had both non-empty blocks and very long sequences of empty blocks.
* Track and log latency of transactions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,7 @@ jobs:
             target/release/sequencer
             target/release/cli
             target/release/commitment-task
+            target/release/submit-transactions
             target/release/reset-storage
 
   build-arm:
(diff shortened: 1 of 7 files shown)


Ancient123:
CI parts look good to me. A list of what env variables it actually needs would be useful for me deploying this to staging.

jbearer:
Only `ESPRESSO_SUBMIT_TRANSACTIONS_SUBMIT_URL=https://query.stating.aws.espresso.network`. Defaults should be fine for everything else

nomaxg (approved):
LGTM

PR #1029, opened 2024-01-24
Iterate over and count namespaces in `BlockPayload`

### LINKED ISSUE #1009: Equivalents of QueryableBlock::{len, iter} for namespaces

akonring:
closes #1009

### DIFF
--- a/sequencer/src/block2.rs
+++ b/sequencer/src/block2.rs
@@ -117,6 +117,14 @@ impl BlockPayload {
         }
     }
 
+    fn num_namespaces(&self, ns_table_bytes: &[u8]) -> usize {
+        get_ns_table_len(ns_table_bytes)
+    }
+
+    fn namespace_iter(&self, ns_table_bytes: &[u8]) -> impl Iterator<Item = usize> {
+        0..get_ns_table_len(ns_table_bytes)
+    }
+
     /// Return a range `r` such that `self.payload[r]` is the bytes of the tx table length.
     ///
     /// Typically `r` is `0..TxTableEntry::byte_len()`.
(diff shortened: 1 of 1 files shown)


sequencer/src/block2.rs
```diff
@@ -117,6 +117,14 @@ impl BlockPayload {
         }
     }
 
+    fn num_namespaces(&self, ns_table_bytes: &[u8]) -> usize {
```
ggutoski, line 120: We don't need `self` here. This could be an assoc function instead. However... we may wish to keep `self` anyway to avoid an API change in case we want it in the future. Also, it's more consistent with the corresponding `QueryableBlock` trait methods for tx proofs to keep `self`.

But you'll need to write `_self` in ordrer to pacify clippy. Same for `namespace_iter`.

ggutoski (approved):
LGTM. Please see my comment and fix lint. Then I think we're good to go.

sequencer/src/block2.rs
ggutoski: ah these methods should also be `pub`

PR #1031, opened 2024-01-24
Update query service

jbearer:
The new version includes fetching. The sequencer node can be configured with a list of peer URLs to fetch missing data from.

This change also changes the SQL persistence to use the underlying storage type (SqlStorage) instead of reusing the SQL data source. This eliminates unnecessary query-service-specific stuff from minimal persistence, including not initializing fetches for missing data when all we are trying to do is connect to a database to store a config file.


philippecamacho (approved):
LGTM

PR #1037, opened 2024-01-25
Adjust light client contract to fit Decaf scope.

### LINKED ISSUE #940: Update contracts to reflect Decaf scope
* Set blocks per epoch to a very high value to ensure the epoch will remain 0 during the execution.
* Do not deploy the stake table contract

philippecamacho:
Closes #940.

Note: I did not see any deployment of the stake table contract in demo scripts. Currently demos are using the legacy contract `HotShot.sol.`

To test the deployment script:
```
> anvil
> forge script contracts/script/HotShot.s.sol --fork-url http://localhost:8545
```


contracts/src/LightClient.sol
```diff
@@ -18,7 +18,10 @@ contract LightClient {
     // === Constants ===
     //
     /// @notice System parameter: number of blocks per epoch
-    uint32 public immutable BLOCKS_PER_EPOCH;
+    // TODO make the variable immutable and format it as using  as BLOCKS_PER_EPOCH once we can use
+    // epochs. See
+    // https://github.com/EspressoSystems/espresso-sequencer/issues/940
+    uint64 public blocksPerEpoch;
```
alxiong, line 24: why do we need to change this? we still leave this as immutable and set it to `type(uint64).max` in the constructor, right?

also why change type?

contracts/src/LightClient.sol
```diff
@@ -118,7 +127,7 @@ contract LightClient {
         ) {
             revert OutdatedState();
         }
-        uint64 epochEndingBlockHeight = currentEpoch * BLOCKS_PER_EPOCH;
+        uint64 epochEndingBlockHeight = currentEpoch * blocksPerEpoch;
         bool isNewEpoch = finalizedState.blockHeight == epochEndingBlockHeight;
         if (!isNewEpoch && newState.blockHeight > epochEndingBlockHeight) {
             revert MissingLastBlockForCurrentEpoch(epochEndingBlockHeight);
```
alxiong, line 133: do you think we should package all these into a `if else` clause and skip it completely when there's no epoch change (i.e. when EPOCH_PER_BLOCK is max) ?

I'm hesitant. if we skip this in `else`, then users don't have to pay for the extra gas of "computing `isNewEpoch`" but our code looks slightly more complicated; OTOH, it's not too much gas, it's negligible compared to the verifier cost, so maybe it's perfectly fine for simpler logic sake.

wdyt?

contracts/src/LightClient.sol
```diff
@@ -76,7 +79,9 @@ contract LightClient {
     /// @notice Wrong plonk proof or public inputs.
     error InvalidProof();
 
-    constructor(LightClientState memory genesis, uint32 numBlockPerEpoch) {
+    /// @dev Note that numBlockPerEpoch is ignore for now. See
+    /// https://github.com/EspressoSystems/espresso-sequencer/issues/940
+    constructor(LightClientState memory genesis, uint64 numBlockPerEpoch) {
```
alxiong, line 84: npoe, we shouldn't ignore, we should simply set it to max during deployment script. 

if we set it to max, then lots of our tests make no sense anymore. (cuz they are supposed to test when epoch changes, the stake table commitments gets updated etc)

alxiong (changes requested):
We should add deployment script for `LightClient.sol` in this PR (so that we will ensure setting the `blockPerEpoch` properly) ?

contracts/src/LightClient.sol
philippecamacho: Yes, I understand now. It is better to set type(uint32).max inside the deployment script.

> also why change type?

Did not realize that $2^{32}$ is a lot of seconds... :)
Fixed in eede4e8.

contracts/src/LightClient.sol
philippecamacho: Fixed in eede4e8.

philippecamacho: Comment removed in 0c54374.

contracts/src/LightClient.sol
philippecamacho: I think we can keep it as is. I added a TODO in 0c54374 so that we can come back to it when working on  gas optimization.

contracts/script/HotShot.s.sol
```diff
@@ -2,14 +2,38 @@
 pragma solidity ^0.8.0;
 
 import "forge-std/Script.sol";
-import { HotShot } from "../src/HotShot.sol";
+import { LightClient as LC } from "../src/LightClient.sol";
+import { LightClient } from "../src/LightClient.sol";
```
alxiong, line 6:
```suggestion
import { LightClient as LC } from "../src/LightClient.sol";
```

contracts/script/HotShot.s.sol
```diff
@@ -2,14 +2,38 @@
...
-        new HotShot();
+
+        //// LightClient contract deployment
+
+        // For Decaf there will be only one epoch
+        uint32 blocksPerEpoch = type(uint32).max;
+
+        // TODO change when we start using epochs. See
+        // https://github.com/EspressoSystems/espresso-sequencer/issues/940
+        uint256 stakeTableCapacity = 10;
```
alxiong, line 21: this comment is a little misleading, technically we don't need this at all. here we only need it since we are generating a "mock genesis", 
in actual production deployment script, this will be removed (even when we are using epochs)

this value will only be used in `StakeTable.s.sol`, not in this file

alxiong (commented):
Generally LGTM, another thing (maybe through another PR) is to update the VK, which right now is supporting 2^20 degrees (targeting 200 nodes) 

if we end up having a smaller stake table (say 128), then we should update that VK.

alxiong (approved):
given my provisional appoval conditioned on the above nit-comments addressed

contracts/script/HotShot.s.sol
philippecamacho: I see, misleading comment removed in ae07794ae91771bf150b36f95348da0e34f8743e.

Issue #937, opened 2024-01-08
[light-client] prover service implementation


mrain:
Move `hotshot-state-prover` here, the light client related structs should be integrated in the sequencer crate. Others, circuit implementations, etc should be inside the prover service crate.

Implements a prover service that

- Read the stake table from contract.
- Read the current updated state from contract.
- Fetch the current light client state and the signatures from the relay server.
- Generate and submit the proof if there's an update.

alxiong:
I can probably help on two sub-items: 1) using ethers-rs to read contract state; 2) serialize generated proofs and public inputs and send to contract.

jbearer:
You can probably take the basic structure of `hotshot-commitment.rs` for this. We have made a number of improvements in how it handles synchronizing with the contract state, retry loop, etc that make it very reliable. Nothing special to do with the actual logic of the contract.

Issue #939, opened 2024-01-08
[light-client] final integration and deployment


mrain:
A cronjob that periodically launches the prover service.
Also update the example, dockers, documentations, etc..

PR #1041, opened 2024-01-25
feat(light-client): altruistic hotshot state prover

mrain:
closes: #937, #939

This PR does:
- Move the `hotshot-state-prover` crate from `HotShot` repo here.
- Implement a prover service.
- Deployment script for prover service and demo docker


hotshot-state-prover/src/state.rs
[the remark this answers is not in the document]
mrain: Good point.
Also our specific signature related types and structs should also stay here.

alxiong:
ah sorry, I accidentally pushed one commit to this branch, I should have worked on my own branch and then merge into this PR. pls safely ignore, the commit above are some minor api and doc improvement mostly.

alxiong:
@mrain 

I have reorg in [75584aa](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/75584aab0adf15192d8692359b52f1a653e478a8) which contains a HUGE amount of file changes and apologize for making reviewers' life harder here. 

changes include:
- Break up `contracts/rust/` into 3 crates: 
  - `hotshot-contract-adapters` only deals with adapting types across domains (i.e. Solidity and Rust), and depends on `hotshot-types`
  - `gen-vk-contract` is a binary that depends on both `hotshot-stake-table` and `hotshot-state-prover` (to preprocess the circuit)
  - `diff-test-hotshot` is a binary (and a lib for testing mocks) that helps with contract's differential test (matching outputs from solidity and jellyfish), this depends on both `adapters` and `state-prover`
 - update all other crates about the change in imports, both due to HotShot's reorg https://github.com/EspressoSystems/HotShot/pull/2491, and due to our own reorg above
 

## Blockers

Currently, I commented out `sequencer` member from workspace, since it has the following compiler error, due to our upgrade to `hotshot:0.5.10` which contains breaking changes in `types/traits/block_content.rs` and seems to be addressed by Theodore in https://github.com/EspressoSystems/espresso-sequencer/pull/1059, 
(I tried some quick fix, but don't want to overlap the work especially I have limited understanding of the upstream change)

```
error[E0050]: method `new` has 3 parameters but the declaration in trait `hotshot_types::traits::block_contents::BlockHeader::new` has 4
   --> sequencer/src/block.rs:342:32
    |
342 |     fn new(payload_commitment: VidCommitment, transactions_root: NMTRoot, parent: &Self) -> Self {
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 4 parameters, found 3
    |
    = note: `new` from trait: `fn(<Advz<ark_ec::models::bls12::Bls12<ark_bls12_381::Config>, CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>>> as VidScheme>::Commit, <<Self as BlockHeader>::Payload as BlockPayload>::Metadata, &Self, &<Self as BlockHeader>::State) -> Self`

error[E0046]: not all trait items implemented, missing: `State`
   --> sequencer/src/block.rs:340:1
    |
340 | impl BlockHeader for Header {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `State` in implementation
    |
    = help: implement the missing item: `type State = /* Type */;`

error[E0046]: not all trait items implemented, missing: `from_header`
  --> sequencer/src/state.rs:15:1
   |
15 | impl HotShotState for State {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `from_header` in implementation
   |
   = help: implement the missing item: `fn from_header(_: &<Self as hotshot::traits::State>::BlockHeader) -> Self { todo!() }`

Some errors have detailed explanations: E0046, E0050.
For more information about an error, try `rustc --explain E0046`.
```

## Action now

meanwhile, I still think you can start reviewing my implementation in `read_contract_state()`, `submit_state_and_proof()`, and `struct StateProverConfig` and see if you agree. 

As discussed, there are some complication in writing unit tests, but I will add it in a future PR.

hotshot-state-prover/src/bin/gen-demo-contract.rs
```diff
@@ -0,0 +1,157 @@
...
+        
+                // For Decaf there will be only one epoch
+                uint32 blocksPerEpoch = type(uint32).max;
+        
+                uint64 viewNum = 0;
+                uint64 blockHeight = 0;
+                BN254.ScalarField blockCommRoot = BN254.ScalarField.wrap(0);
+                BN254.ScalarField feeLedgerComm = BN254.ScalarField.wrap(0);
+                BN254.ScalarField stakeTableBlsKeyComm = BN254.ScalarField.wrap(
+                    {}
```
sveitser, line 126: I don't quite understand why we have to generate this code. Couldn't we use env vars to pass arguments to the deploy script instead?

contracts/script/LightClientTest.s.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "forge-std/Script.sol";
+import { BN254 } from "bn254/BN254.sol";
+import { LightClient as LC } from "../src/LightClient.sol";
+import { LightClientTest as LCTest } from "../test/mocks/LightClientTest.sol";
+
+contract DeployLightClientScript is Script {
```
sveitser, line 9: There now seem to be 3 scripts that do about the same thing. Would it be feasible to have one script and pass different inputs (via env vars, JSON file, ...) to achieve the same? If not I think there should be at least some comments explaining why we do it this way.

hotshot-state-prover/src/bin/gen-demo-contract.rs
alxiong: >Couldn't we use env vars to pass arguments to the deploy script instead?

that's a good point, maybe we can use [`vm.envString()` ](https://book.getfoundry.sh/cheatcodes/env-string?#envstring) to load instead. 

i guess the complication comes from: the desire to do "one-click" demo deployment, whereas the mock genesis state requires running rust to determine. 
^^ can we deterministically compute this and hardcode them into `.env`, maybe?

contracts/script/LightClientTest.s.sol
alxiong: agree, we should remove the duplication, maybe just leave the one for deploying the LightClientTest, for now?

scripts/launch-prover-service
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
sveitser, line 1: I think this script is a bit risky because it's easy to trigger a deployment (for example by forgetting or misspelling the env var). This could potentially cost a lot of money. I think it would be preferable if there was an explicit argument to trigger the deployment or maybe better having a separate docker image with foundry and deployment (with for example a `-dev` suffix in the name).

sveitser: Not saying we need to fix it immediately but ideally before we deploy anything to mainnet.

mrain: good point, let me add a `--deploy` argument just like commitment-task

mrain: or have a env variable

hotshot-state-prover/src/bin/state-prover.rs
```diff
@@ -0,0 +1,132 @@
...
+            .with_chain_id(chain_id)
+            .signer()
+            .clone(),
+        num_nodes: args.num_nodes,
+        seed: args.keygen_seed,
+    };
+
+    if args.daemon {
+        // Launching the prover service daemon
+        run_prover_service(config).await;
```
sveitser, line 127: I haven't reviewed everything in detail so might have overlooked it. Does the prover currently have a health check API? It's used to confirm that the service is running and restart it if necessary. If not, adding a tide-disco server will provide a simple health check.

https://github.com/EspressoSystems/espresso-sequencer/blob/58852cbd989606f103a41664ef4c728859de007d/sequencer/src/bin/commitment-task.rs#L138

hotshot-state-prover/src/bin/gen-demo-contract.rs
alxiong: the initial motivation is: we could have some different genesis state (different num_nodes, diff stake table config), which would result in variable stake table commitments, 
thus, Chengyu wrote this generation file to automate/simplify the hardcoding process.

hotshot-state-prover/src/bin/state-prover.rs
mrain: health check added [8601b08](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/8601b080bd0c086bbfd0e7f2920e289940b5f36f)

scripts/launch-prover-service
mrain: env `ESPRESSO_SEQUENCER_DEPLOY_LIGHTCLIENT_CONTRACT` added

contracts/rust/adapter/src/archived.rs
philippecamacho, line 1: Do we still need this file?

contracts/rust/adapter/src/jellyfish.rs
```diff
@@ -350,6 +347,17 @@ impl From<ParsedPlonkProof> for Proof<Bn254> {
     }
 }
 
+// NOTE: this extra "middle-man" conversion is the unfortunate result of type in rust binding is auto-generated,
+// and are mutually oblivious of the types in Jellyfish, thus no From can be implemented at either side.
+// Since we cannot implement foreign traits on foreign types, we can only transit via our own `ParsedPlonkProof`.
+impl From<ParsedPlonkProof> for contract_bindings::i_plonk_verifier::PlonkProof {
+    fn from(p: ParsedPlonkProof) -> Self {
+        // parsed_proof is our own defined type, which share exactly the same structure and field types
```
philippecamacho, line 355: Do we have a test to detect and mismatch between the bindings and our own type?

contracts/rust/diff-test/src/lib.rs
```diff
@@ -351,54 +364,115 @@ pub(crate) fn stake_table_for_testing(
...
-            u256_to_field(s.block_comm_root),
-            u256_to_field(s.fee_ledger_comm),
-            u256_to_field(s.bls_key_comm),
-            u256_to_field(s.schnorr_key_comm),
-            u256_to_field(s.amount_comm),
-        ];
-        Self::from(fields)
+// Different `m`s lead to different circuits.
+// Different `a0`s lead to different witness values.
+pub fn gen_circuit_for_test<F: PrimeField>(m: usize, a0: usize) -> Result<PlonkCircuit<F>> {
```
philippecamacho, line 446: What does this circuit do?

contracts/script/LightClient.s.sol
```diff
@@ -0,0 +1,46 @@
...
+
+        // For Decaf there will be only one epoch
+        uint32 blocksPerEpoch = type(uint32).max;
+
+        uint64 viewNum = 0;
+        uint64 blockHeight = 0;
+        BN254.ScalarField blockCommRoot = BN254.ScalarField.wrap(0);
+        BN254.ScalarField feeLedgerComm = BN254.ScalarField.wrap(0);
+        BN254.ScalarField stakeTableBlsKeyComm = BN254.ScalarField.wrap(
+            21279528985922333501911071464328386841445872547720180015435547221260402565727
```
philippecamacho, line 22: How were these values computed?

mrain: These are fixed stake table commitment, computed here: https://github.com/EspressoSystems/espresso-sequencer/pull/1041/files/651cee18d9f34b7ee4f4250da678f53b6ae6c2a1#diff-b1dabd045ed0acc2bfdf354d6189dd9bc005b2d20030283b4c7b732626d79767

contracts/script/LightClientTest.s.sol
philippecamacho: Agreed. I suggest we create a ticket as deployment becomes a bit more involved with upgradeable contracts.

contracts/script/LightClientTest.s.sol
```diff
@@ -0,0 +1,48 @@
...
+        // For Decaf there will be only one epoch
+        uint32 blocksPerEpoch = type(uint32).max;
+
+        // TODO for a production deployment provide the right genesis state
+        uint64 viewNum = 0;
+        uint64 blockHeight = 0;
+        BN254.ScalarField blockCommRoot = BN254.ScalarField.wrap(42);
+        BN254.ScalarField feeLedgerComm = BN254.ScalarField.wrap(42);
+        BN254.ScalarField stakeTableBlsKeyComm = BN254.ScalarField.wrap(
+            0x0d168a6fe13a8b344caea06d1fd2a2f1e62eeb7b2b00c6284e094a819a1f1790
```
philippecamacho, line 24: How are these values computed?

hotshot-state-prover/src/bin/gen-demo-contract.rs
philippecamacho: I guess this answers my question above regarding how commitment values are generated.

hotshot-state-prover/src/bin/gen-demo-contract.rs
```diff
@@ -0,0 +1,157 @@
...
+    // overwrite the file
+    let mut file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(path)
+        .unwrap();
+
+    let code = format!(
+        "// SPDX-License-Identifier: UNLICENSED
```
philippecamacho, line 105: Add a comment about the fact that this code is generated automatically.

mrain: added here [73e8fb4](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/73e8fb4b637e01ecc045b21cfa7c85463117d3a9)

contracts/rust/adapter/src/jellyfish.rs
philippecamacho: Or will something break in an obvious way if there is a mismatch?

hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,541 @@
...
+/// A function that takes as input:
+/// - a list of stake table entries (`Vec<(SchnorrVerKey, Amount)>`)
+/// - a bit vector indicates the signers
+/// - a list of schnorr signatures of the updated states (`Vec<SchnorrSignature>`), default if the node doesn't sign the state
+/// - updated light client state (`(view_number, block_height, block_comm_root, fee_ledger_comm, stake_table_comm)`)
+/// - a quorum threshold
+/// Lengths of input vectors should not exceed the `STAKE_TABLE_CAPACITY`.
+/// The list of stake table entries, bit indicators and signatures will be padded to the `STAKE_TABLE_CAPACITY`.
+/// It checks that
+/// - the signer's accumulated weight exceeds the quorum threshold
```
philippecamacho, line 132:
```suggestion
/// - the signers' accumulated weight exceeds the quorum threshold
```

hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,541 @@
...
+/// - the stake table corresponds to the one committed in the light client state
+/// - all signed Schnorr signatures are valid
```
philippecamacho, line 134:
```suggestion
/// - all Schnorr signatures are valid
```

philippecamacho, line 134: Also that:  "The vector that indicates who signed is a bit vector."

hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,541 @@
...
+/// and returns
+/// - A circuit for proof generation
+/// - A list of public inputs for verification
+/// - A `PlonkError` if any error happens when building the circuit
+#[allow(clippy::too_many_lines)]
+pub(crate) fn build<F, P, STIter, BitIter, SigIter, const STAKE_TABLE_CAPACITY: usize>(
+    stake_table_entries: STIter,
+    signer_bit_vec: BitIter,
```
philippecamacho, line 142: Is it possible to pass a vector of field elements instead of booleans in order to ensure that the circuit enforces the each element of this vector to be a boolean?

hotshot-state-prover/src/snark.rs
```diff
@@ -0,0 +1,277 @@
...
+/// Given a proving key and
+/// - a list of stake table entries (`Vec<(BLSVerKey, Amount, SchnorrVerKey)>`)
+/// - a list of schnorr signatures of the updated states (`Vec<SchnorrSignature>`), default if the node doesn't sign the state
+/// - updated light client state (`(view_number, block_height, block_comm_root, fee_ledger_comm, stake_table_comm)`)
+/// - a bit vector indicates the signers
+/// - a quorum threshold
+/// Returns error or a pair `(proof, public_inputs)` asserting that
+/// - the signer's accumulated weight exceeds the quorum threshold
+/// - the stake table corresponds to the one committed in the light client state
+/// - all signed schnorr signatures are valid
```
philippecamacho, line 51:
```suggestion
/// - all schnorr signatures over the light client state are valid
```

hotshot-state-prover/src/service.rs
```diff
@@ -0,0 +1,575 @@
...
+    let mut signatures = vec![Default::default(); entries.len()];
+    let mut accumulated_weight = U256::zero();
+    entries.iter().enumerate().for_each(|(i, (key, stake))| {
+        if let Some(sig) = bundle.signatures.get(key) {
+            signer_bit_vec[i] = true;
+            signatures[i] = sig.clone();
+            accumulated_weight += *stake;
+        }
+    });
+
```
philippecamacho, line 218: Maybe throw an error if `accumulated_weight < threshold` instead of getting the error while generating the proof.

hotshot-state-prover/src/service.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// If daemon and provided, the service will run a basic HTTP server on the given port.
+    ///
+    /// The server provides healthcheck and version endpoints.
+    pub port: Option<u16>,
+}
+
+pub async fn init_stake_table(
+    config: &StateProverConfig,
+) -> StakeTable<BLSPubKey, StateVerKey, CircuitField> {
+    // We now initialize a static stake table as what hotshot orchestrator does.
```
philippecamacho, line 73: Why can't we reuse the code of the orchestrator?

hotshot-state-prover/src/circuit.rs
mrain: This enforcement is auto-applied when you create a boolean variable: https://github.com/EspressoSystems/jellyfish/blob/af4dddd573d8590fe7793412b5334e23335b1a3a/relation/src/constraint_system.rs#L105

hotshot-state-prover/src/service.rs
mrain: As the design of orchestrator seems not stable for now. And the current orchestrator cannot provide the needed information, `known_node_with_stakes` doesn't contain the Schnorr ver key.

hotshot-state-prover/src/service.rs
mrain: Done in [f005900](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/f0059003d42ca79309513fe94c1ffc647d3e474b)

contracts/rust/adapter/src/archived.rs
alxiong: not really, i didn't remove it for references purposes, it's not even included in `lib.rs` anyway. it doesn't really hurt IMO.
I'm also fine removing it.

contracts/rust/adapter/src/jellyfish.rs
alxiong: the tests in `service.rs` where we submit a proof to contract should fail (if there's a mismatch) since are calling this conversion there.
but i agree it might be better to add a dedicated test for this unsafe operation for explicit detection

hotshot-state-prover/src/circuit.rs
alxiong: i think it's implicit that bit vector would be constrained to be a boolean vector.
don't think we need to include as part of the circuit description here.

hotshot-state-prover/src/bin/gen-demo-contract.rs
alxiong: you already have some deployment script in https://github.com/EspressoSystems/espresso-sequencer/pull/1079,

may I suggest:
- relocate this code to `contracts/rust/gen-light-client-deploy` (basically at the same level as `gen-vk-contract` binary), as it's harder to find this code that supposed to generate script in `contracts/scripts` 
- in your 1079, Philippe you try to adapt this generated by rust approach instead of `diff-test mock-gensis`?

cc @philippecamacho @mrain

contracts/script/LightClientTest.s.sol
alxiong: rn, we only have two scripts, which make sense, one is for deploying LightClientTest, the other is for demo and for actual production/staging.

hotshot-state-prover/src/bin/gen-demo-contract.rs
philippecamacho: @sveitser's approach looks more straightforward and flexible (we could have different tests with different initial states for the light client contract) to me.

contracts/rust/adapter/src/jellyfish.rs
philippecamacho: Thanks.

hotshot-state-prover/src/bin/gen-demo-contract.rs
mrain: The problem of environment variable way is, computation of the genesis state is not that straightforward.

hotshot-state-prover/src/circuit.rs
philippecamacho: Right but we cannot test that a malicious non-binary vector is rejected by the circuit.

hotshot-state-prover/src/circuit.rs
philippecamacho: As mentioned [below](https://github.com/EspressoSystems/espresso-sequencer/pull/1041#discussion_r1489942751), the current implementation is correct, yet ensuring the vector is composed by booleans is a core requirement of the circuit that is worth mentioning I think.

hotshot-state-prover/src/bin/gen-demo-contract.rs
philippecamacho: Why not using a forge deploy script and generate the genesis state using `diff-test mock-genesis`? That way in the test you can generate the state and then call the deploy script. See example here: https://github.com/EspressoSystems/espresso-sequencer/blob/0e4a83e4486de1580711643d53eb8cc236d07aff/contracts/test/LightClientContractUpgrade.t.sol#L26 

In this example we generate the genesis state inside the [`deployer.run()`](https://github.com/EspressoSystems/espresso-sequencer/blob/0e4a83e4486de1580711643d53eb8cc236d07aff/contracts/script/DeployLightContract.s.sol#L14) function but that could be done outside.

hotshot-state-prover/src/circuit.rs
mrain: Done in [24bb979](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/24bb97974fd20657011734e458dd589788ed1765)

hotshot-state-prover/src/circuit.rs
mrain: Done in [24bb979](https://github.com/EspressoSystems/espresso-sequencer/pull/1041/commits/24bb97974fd20657011734e458dd589788ed1765)

philippecamacho: Thanks!

hotshot-state-prover/src/circuit.rs
philippecamacho: Thanks!

philippecamacho (approved):
Great work!

hotshot-state-prover/src/bin/gen-demo-contract.rs
philippecamacho: As agreed with @mrain , we will revisit how the light contract is initialized in PR #1079.

contracts/script/LightClientTest.s.sol
philippecamacho: As discussed with  @mrain we will unify how the light contract is initialized/deployed for testing purposes. See #1092.

Issue #1040, opened 2024-01-25
 VID-namespace error handling should be pure-`anyhow`


ggutoski:
Error handling in `block2.rs` is currently downright disgusting. As discussed, we'll switch to pure-`anyhow` error handling for everything inside `block2.rs`. Recommended reading: the `anyhow` readme: https://github.com/dtolnay/anyhow

ggutoski:
On hold pending agreement on a proper design pattern for error-handling.

PR #1042, opened 2024-01-26
refactor block2.rs

### LINKED ISSUE #856: [priority code clean-up] VID namespace use new hotshot BlockPayload API
Current namespace code in `block2.rs` has a bunch of mock code because hotshot's `BlockPayload` API was not complete at the time it was written. When this API is stabilized in hotshot we need to update `block2.rs` to use it.

akonring:
closes #856 

- rename the struct `BlockPayload` to `Payload` to avoid confusion with `hotshot_types::traits::BlockPayload`

- remove the boilerplate module and merge the content

- TODO: fix error handling (https://github.com/EspressoSystems/espresso-sequencer/issues/1040)

### DIFF
--- a/sequencer/src/block2.rs
+++ b/sequencer/src/block2.rs
@@ -803,7 +803,7 @@ mod tx_table_entry {
     }
 }
 
-type NsTable = <Payload as hotshot::traits::BlockPayload>::Metadata;
+type NsTable = <Payload as BlockPayload>::Metadata;
 
 /// TODO do we really need `PartialOrd`, `Ord` here?
 /// Could the `Ord` bound be removed from `QueryablePayload::TransactionIndex`?`
--- a/sequencer/src/block2.rs
+++ b/sequencer/src/block2.rs
@@ -35,16 +35,6 @@ pub struct Payload {
 }
 
 impl Payload {
-    fn from_bytes<B>(bytes: B) -> Self
-    where
-        B: IntoIterator<Item = u8>,
-    {
-        Self {
-            payload: bytes.into_iter().collect(),
-            tx_table_len_proof: Default::default(),
-        }
-    }
-
     // TODO dead code even with `pub` because this module is private in lib.rs
     #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


sequencer/src/block2.rs
```diff
@@ -108,92 +230,85 @@ impl BlockPayload {
...
+    fn from_bytes<I>(encoded_transactions: I, _metadata: &Self::Metadata) -> Self
     where
-        B: IntoIterator<Item = u8>,
+        I: Iterator<Item = u8>,
     {
-        Self {
-            payload: bytes.into_iter().collect(),
-            tx_table_len_proof: Default::default(),
-        }
+        Self::from_bytes(encoded_transactions)
```
ggutoski, line 241: can we eliminate the delegation to `Self::from_bytes` and simply put that function body in here?

sequencer/src/block2.rs
```diff
@@ -688,7 +803,7 @@ mod tx_table_entry {
     }
 }
 
-type NsTable = <BlockPayload as hotshot::traits::BlockPayload>::Metadata;
+type NsTable = <Payload as hotshot::traits::BlockPayload>::Metadata;
```
ggutoski, line 806:
```suggestion
type NsTable = <Payload as BlockPayload>::Metadata;
```
Now that we are directly importing `BlockPayload` we should replace all occurrences of `hotshot::traits::BlockPayload` with `BlockPayload`.

ggutoski (commented):
Looks good! Just a couple minor suggestions.

ggutoski:
Regarding #1040 don't do anything there until the team has come to consensus on it.

sequencer/src/block2.rs
ggutoski: ...just like you did for `from_transactions`?

akonring: yes, I think it is doable. But seems like our `malformed_payloads`-test needs changing since we need to make the metadata (ns_table) available.

ggutoski (approved):

PR #1045, opened 2024-01-26
splitting up block2.rs

### LINKED ISSUE #951: [priority code clean-up] Split `block2.rs` into multiple files

akonring:
close #951


sequencer/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 pub mod api;
 mod block;
-mod block2;
+pub mod block2;
```
ggutoski, line 3: Why `pub` here? I note that `block` is not `pub` and my default is to mimic `block` in `block2`.

ggutoski (commented):
Looks good so far. This should be a quick PR because we don't want to make non-trivial changes while moving around so much code. So I think we can merge Monday morning before we move on to #1046 .

Some notes:
- maybe rename `iterator.rs` -> `tx_iterator.rs`
- Tests are huge. Do you think we should split them into a separate `test.rs` file?
- see comment about `pub mod block2`

sequencer/src/lib.rs
akonring: yes, good point

akonring:
> * maybe rename `iterator.rs` -> `tx_iterator.rs`
Yes, agree. This is updated now.


> * Tests are huge. Do you think we should split them into a separate `test.rs` file?
I don't have strong opinions regarding this. I believe the idiomatic Rust way would dictate placing the unit tests in the same file as the unit. But then again the tests now relates to all the files in the sub-directory (`iterator`, `entry`, `queryable`).

Let me know if I should change and create it a new `block/test.rs`

philippecamacho:
LGTM

Regarding the tests we may want to move only the helper functions / structs to another file (e.g. `TestCase`) . I have no strong opinion about it though.

philippecamacho:
Should `TxInclusionProof` be part of `queryable.rs`?

ggutoski:
> Should `TxInclusionProof` be part of `queryable.rs`?

Great catch! Yes, we should move `TxInclusionProof` to `queryable.rs`.

> Let me know if I should change this and create a new `block/test.rs`

I think we can leave it alone for now. Let's wait until we have strong inspiration and agreement on what (if anything) to do about tests.

akonring:
> Should `TxInclusionProof` be part of `queryable.rs`?

Yes, that makes sense to me at least. I can change this now.

ggutoski (approved):
LGTM

philippecamacho (approved):
LGTM

PR #1051, opened 2024-01-29
Refactor in preparation for restartability

jbearer:
* Add interfaces to the persistence trait to save and restore state after each consensus event
* Refactor initialization code and test infrastructure, because the previous mess was making it almost impossible to add in the new persistence stuff
* Add a test for restartability

This PR does not yet enable restartability. For it to work properly, we need either validated state persistence or catchup. The restart test is disabled until we have one of those (or a mock version). However, this PR contains cleanup that will be beneficial to other work, so we should merge it now and finish restarting later.


tbro (approved):
LGTM

PR #1057, opened 2024-01-29
935 fee token l1 contract

alysiahuggins:
- created the fee contract
- made it upgradeable
- added tests


philippecamacho:
@alysiahuggins I made some minor changes:
 * 687d1e0b9329de01ee1288bdd16d911fe26f2601 
   This strange value for `ETHERSCAN_API_KEY` is needed to avoid some annoying warnings. I added a [comment](47a9565121127966cdf3396c34e42799dc444ce3).
   * ac55d525b55ecdfb1c7b462e56bf520941ad8992
   By default the contracts which name contain "Test" are omitted when generating bindings.
   
   Now a question: do we need the bindings for the new contracts like `uups_upgradeable.rs`?

contracts/src/FeeContract.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    /// @notice Revert if no method name was called
+    receive() external payable {
+        revert NoFunctionCalled();
+    }
+
+    /// @notice Allows anyone to deposit an ETH balance for any user
+    /// @dev the deposit amount is less than a specified threshold to prevent fat finger errors
+    function deposit(address user) public payable {
+        if (msg.value == 0) {
```
philippecamacho, line 70: This if-then statement is redundant with the next one, maybe we can remove it?

contracts/src/FeeContract.sol
```diff
@@ -0,0 +1,98 @@
...
+        if (msg.value < MIN_DEPOSIT_AMOUNT) {
+            revert DepositTooSmall();
+        }
+        if (msg.value > MAX_DEPOSIT_AMOUNT) {
+            revert DepositTooLarge();
+        }
+        if (user == address(0)) {
+            revert InvalidUserAddress();
+        }
+        balances[user] += msg.value;
```
philippecamacho, line 82: I think we do not need to track the balances of each user in the contract as the accounting will happen inside HotShot. Is the `balances` mapping for testing purposes only?

contracts/src/FeeContract.sol
alysiahuggins: ah yes, very true, i had this if statement before I added the other

alysiahuggins: fix: https://github.com/EspressoSystems/espresso-sequencer/commit/fa35ff3bdb59e938fc0895b3dd6b7d4d0cba5e64

contracts/src/FeeContract.sol
alysiahuggins: Oh I see, but would hotshot need to read the balance deposited from this contract initially? Also when the user deposits more

contracts/script/DeployFeeContract.s.sol
```diff
@@ -0,0 +1,39 @@
...
+
+        FeeContract feeContract = new FeeContract(); //Our implementation(logic).Proxy will point
+            // here to delegate
+
+        // Encode the initializer function call
+        bytes memory data = abi.encodeWithSelector(
+            FeeContract(feeContract).initialize.selector,
+            msg.sender // Initial owner/admin of the contract
+        );
+
```
alxiong, line 31: I actually didn't fully understand how this works. 
[`ERC1967Proxy.constructor`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Proxy.sol#L27) will call `ERC1967Utils.upgradeToAndCall` which [would call ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Utils.sol#L86)`Address.functionDelegateCall`, which [would call ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol#L105)the constructor on fee contract with `msg.sender`

but our fee contract constructor doesn't accept any input argument

what did I miss?

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        assertEq(balance, 0);
+    }
+}
+
+contract FeeContractUpgradabilityTest is Test {
+    address payable public proxy;
+    address public admin;
+    FeeContract public feeContractProxy;
+    DeployFeeContract public deployer = new DeployFeeContract();
```
alxiong, line 169: if we set the `admin` to what's derived from env.mnemonic, then we wouldn't need this deployer contract variable, just storing the `address proxy` is sufficient.

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        feeContractProxy.upgradeToAndCall(address(feeContractV2), "");
+    }
+
+    function testFailUpgradeToWithWrongAdmin() public {
+        FeeContractV2Test feeContractV2 = new FeeContractV2Test();
+
+        //start the upgrade with another address which isn't the admin
+        vm.prank(makeAddr("otherUser"));
+        feeContractProxy.upgradeToAndCall(address(feeContractV2), "");
```
alxiong, line 254: we should add `expectRevert` ?

also should we test that if the we can upgrade to a `V2` with no admin? and/or become non-upgradable?

alxiong (commented):
👍 good job!

- when expected revert, we should also try to match the thrown error type to make sure we are catching the expected revert, not accidentally caused by other reasons.
- remember to remove commented out code (such as `//console.log`, or some lines in deployment scripts

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,278 @@
...
+    function testFuzz_noFunction() public payable {
+        address fcAddress = address(feeContract);
+        (bool success,) = fcAddress.call{ value: msg.value }("");
+
+        //assert that the transaction was not successful
+        assertFalse(success);
+
+        //assert that the balance of the fee contract is still zero
+        assertEq(address(feeContract).balance, 0);
+    }
```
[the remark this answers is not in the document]
alysiahuggins: yea good point, I'll update these

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,278 @@
...
+/// @title FeeContract Test
+contract FeeContractTest is Test {
+    FeeContract public feeContract;
+
+    function setUp() public {
+        feeContract = new FeeContract();
+    }
+
+    //test deposits work
+    function testFuzz_deposit(address user, uint256 amount) public payable {
```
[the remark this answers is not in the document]
alysiahuggins: yea that makes sense, i'll remove testFuzz_depositTwice(), replace with depositMany which involves a for loop. good catch

contracts/script/DeployFeeContract.s.sol
```diff
@@ -0,0 +1,39 @@
...
+import "forge-std/Script.sol";
+
+import { FeeContract } from "../src/FeeContract.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+
+contract DeployFeeContract is Script {
+    function run() external returns (address payable proxy, address admin) {
+        // string memory seedPhrase = vm.envString("MNEMONIC");
+        // (address admin,) = deriveRememberKey(seedPhrase, 0);
+        (proxy, admin) = deployFeeContract();
```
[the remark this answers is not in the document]
alysiahuggins: well actually when I deployed to testnet, i passed the private key in as an env variable, however, would we use a multisig for the actual deployment?

contracts/test/FeeContract.t.sol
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/132ced07ad629b957f13633a42a7937cd337230b

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,278 @@
...
+        address user = address(0);
+        uint256 amount = 0.5 ether;
+
+        vm.expectRevert(FeeContract.InvalidUserAddress.selector);
+
+        //deposit for the user
+        feeContract.deposit{ value: amount }(user);
+    }
+
+    function testFuzz_depositEvent(address user, uint256 amount) public payable {
```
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/428d568966fadfb87338b9535d0430b88df40419

contracts/script/DeployFeeContract.s.sol
```diff
@@ -0,0 +1,39 @@
...
+
+        return (proxy, admin);
+    }
+
+    /// @notice deploys the impl, proxy & initializes the impl
+    // @return address of proxy and admin
+    function deployFeeContract() public returns (address payable proxyAddress, address admin) {
```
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/ca547ffc8dbcab42a8830ab3f5e8b193d6d6864d

contracts/script/DeployFeeContract.s.sol
```diff
@@ -0,0 +1,39 @@
...
+        vm.startBroadcast();
+
+        FeeContract feeContract = new FeeContract(); //Our implementation(logic).Proxy will point
+            // here to delegate
+
+        // Encode the initializer function call
+        bytes memory data = abi.encodeWithSelector(
+            FeeContract(feeContract).initialize.selector,
```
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/ca547ffc8dbcab42a8830ab3f5e8b193d6d6864d

contracts/script/DeployFeeContract.s.sol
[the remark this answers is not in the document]
alysiahuggins: > why separate function, should inline it

https://github.com/EspressoSystems/espresso-sequencer/commit/fd5bdd1b09d29146c1df649d7a64d100ba5e783f

contracts/test/FeeContract.t.sol
alysiahuggins: true, there is no need for the deployer to be a state variable https://github.com/EspressoSystems/espresso-sequencer/commit/3c7c6fe7fd90130907e29bd90863354a7cd58bea

contracts/test/FeeContract.t.sol
alysiahuggins: > also should we test that if the we can upgrade to a V2 with no admin? and/or become non-upgradable?
so test whether it's possible to upgrade to another contract that isn't upgradeable?

alysiahuggins: > we should add expectRevert ?
I used the `testFail` syntax which means that the foundry test fails when the contract call reverts. Currently checking with the team to see if there is a preferred testing method for expected failures.

alysiahuggins: > > we should add expectRevert ?
> 
> I used the `testFail` syntax which means that the foundry test fails when the contract call reverts. Currently checking with the team to see if there is a preferred testing method for expected failures.

since we decided to go with expectRevert, I've updated the test https://github.com/EspressoSystems/espresso-sequencer/commit/d4e477df94a8b75a1dd3ec1a30d75363afc5fb4d

contracts/src/FeeContract.sol
alysiahuggins: any thoughts on this @alxiong @jbearer ?

contracts/script/DeployFeeContract.s.sol
alysiahuggins: Yea the FeeContract's constructor is actually disabled, so the function that's actually being called is the `initialize` method on `FeeContract.sol`

contracts/src/FeeContract.sol
philippecamacho: While it is possible, I don't think it is needed as it should be enough to read from the deposit events. That said it depends on the implementation of the fee accounting inside HotShot cc @jbearer.

philippecamacho: Anyways, this function cannot create any harm and makes testing simpler so I am ok to keep it.

alysiahuggins: very true @ being able to read from the deposit events. Right now, we assume that Hotshot takes care of all the accounting and also storage. So maybe it's worth considering removing it so that the L1 smart contract storage doesn't increase which can impact gas fees over time. Unless we want to consider things like onchain (Ethereum L1) integrity and a possible future where we handle ERC20 tokens and then give the users a wrapped version that's represented in their wallet?

alxiong: personally, i prefer to keep the map in contract, it's not _very_ expensive as it's minimally simple.

having this map makes upgrade with "less" trust, instead of completely at the arbitration of the admin (i'd imagine ppl trusting Espresso, but not when there's many millions deposited and our admin and just "oops, I upgrade a contract that can claim a different mapping, users can prove historical event to claim their coin if they are so keen")

contracts/script/DeployFeeContract.s.sol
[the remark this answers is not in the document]
alxiong: if we are using multisig, we should also be loading the multisig account/address from env var, right?

if so, here loading from `envString` is closer to actual code?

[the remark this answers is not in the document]
alysiahuggins: So it depends on how we manage it. for example, if we use Openzeppelin Defender for managing the deployment of the contracts the script would look like this https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/blob/main/test/Defender.s.sol (though at the moment this is not sufficient since deployments with defender via Foundry doesn't support constructor arguments, but I think Hardhat + Defender does). So I can uncomment this code for now but it's likely to change.

contracts/script/DeployFeeContract.s.sol
alysiahuggins: oh yeaaa, good point, it just uses the msg.sender that calls the method, so I can comment that code out

alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/7df709c6ea1e027957ebd43e507e080eb4a7e2d0

alxiong (commented):
✅ 
The code looks good, but pls wait until #997 is merged, then sync this with main, as they would exclude all the unnecessary bindings generated. 

btw, should we also remove the `broadcasts/*` artifacts? @alysiahuggins

contracts/script/DeployFeeContract.s.sol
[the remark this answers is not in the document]
alxiong: thanks for the info! we can leave it as is for now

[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/65917037d9433e95af5fe4c54c7060535b7ebe74

alysiahuggins:
> ✅ The code looks good, but pls wait until #997 is merged, then sync this with main, as they would exclude all the unnecessary bindings generated.
> 
> btw, should we also remove the `broadcasts/*` artifacts? @alysiahuggins

Hey, let me know if all is ready to merge especially since we excluded the forge bindings issue from #997 and put it into a new issue, #1068. As for the broadcast folder, are you sure you want to remove it? It was a real deployment to sepolia testnet?

alysiahuggins:
> > ✅ The code looks good, but pls wait until #997 is merged, then sync this with main, as they would exclude all the unnecessary bindings generated.
> > btw, should we also remove the `broadcasts/*` artifacts? @alysiahuggins
> 
> Hey, let me know if all is ready to merge especially since we excluded the forge bindings issue from #997 and put it into a new issue, #1068. As for the broadcast folder, are you sure you want to remove it? It was a real deployment to sepolia testnet?

sepolia deployment broadcast logs removed since it's not a part of the public testnet release https://github.com/EspressoSystems/espresso-sequencer/commit/2e193c09908c2a4f02bccff8218228656cfaaf9d

contracts/test/FeeContract.t.sol
```diff
@@ -0,0 +1,336 @@
...
+        address user = makeAddr("user");
+
+        //fund this account
+        vm.deal(user, 1 ether);
+
+        uint256 totalAmountDeposited = 0;
+
+        for (uint256 i = 0; i < 10; i++) {
+            uint256 amount = i / 100 ether + feeContract.MIN_DEPOSIT_AMOUNT();
+
```
alxiong, line 87: i think this will always be `MIN_DEPOSIT_AMOUNT` as the integer division will always round to zero?

alxiong (approved):
LGTM

contracts/test/FeeContract.t.sol
alysiahuggins: oh that's right!! the test still proves what it sets out to prove but I'll remove the division to make it easier https://github.com/EspressoSystems/espresso-sequencer/commit/b1884c4bea3246a86cbd430853a5a69e4def0f0a

philippecamacho (approved):
LGTM

Issue #1024, opened 2024-01-24
Implement validated state


nomaxg:
Implement the `validate_block` function, performing all necessary merkle-tree validation (I believe we have some mock functions that currently do this) and move the block/fee merkle tree frontier out of the header.

tbro:
This ticket is a little vague so I've tried to tie down some of the pieces. In  #1059 there are some first steps toward addressing this issue.

  * replace mock impl of `State` with logic that uses fee and block merkle trees for validation
  * add fee based validation to `fn _validate_proposal()`

tbro:
closed in #1066

PR #1059, opened 2024-01-29
Update `state.rs` to use fee and block state

tbro:
Update `state.rs` with first steps for `State` implementation. Addresses #1024

### DIFF
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -313,7 +313,7 @@ impl Header {
     }
 }
 
-fn _validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<BlockMerkleTree> {
+pub fn _validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<BlockMerkleTree> {
     anyhow::ensure!(
         proposal.height == parent.height + 1,
         anyhow::anyhow!(
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -373,15 +373,8 @@ impl BlockHeader for Header {
     ) {
         let (payload, transactions_root) = Payload::genesis();
         let payload_commitment = vid_commitment(&payload.encode().unwrap().collect(), 1);
-        let block_merkle_tree =
-            BlockMerkleTree::from_elems(32, Vec::<Commitment<Header>>::new()).unwrap();
+        let ValidatedState { fee_merkle_tree, block_merkle_tree } = ValidatedState::default();
         let block_merkle_tree_root = block_merkle_tree.commitment();
-        // Words of wisdom from @mrain:
-        // "capacity = arity^height
-        // For index space 2^160, arity 256 (2^8),
-        // you should set the height as 160/8=20"
-        let fee_merkle_tree =
-            FeeMerkleTree::from_kv_set(20, Vec::<(FeeAccount, FeeAmount)>::new()).unwrap();
         let fee_merkle_tree_root = fee_merkle_tree.commitment();
 
         let header = Self {
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -1,6 +1,7 @@
 use crate::{
     l1_client::{L1Client, L1ClientOptions, L1Snapshot},
-    L1BlockInfo, NMTRoot, NamespaceProofType, Transaction, TransactionNMT, VmId, MAX_NMT_DEPTH,
+    L1BlockInfo, NMTRoot, NamespaceProofType, Transaction, TransactionNMT, ValidatedState, VmId,
+    MAX_NMT_DEPTH,
 };
 use ark_serialize::{
     CanonicalDeserialize, CanonicalSerialize, Compress, Read, SerializationError, Valid, Validate,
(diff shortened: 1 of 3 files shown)


sequencer/src/state.rs
```diff
@@ -1,19 +1,27 @@
...
-use std::fmt::Debug;
-
-#[derive(Default, Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
-pub struct State {
-    chain_variables: ChainVariables,
-    ethereum_block_tag: BlockNumber,
+use hotshot_types::data::{ViewNumber, BlockError};
+use jf_primitives::merkle_tree::{MerkleTreeScheme, UniversalMerkleTreeScheme};
+
+impl Default for ValidatedState {
```
jbearer, line 10: Where is `ValidatedState` defined? IMO it would make sense to define it right here before the trait ipls

sequencer/src/state.rs
```diff
@@ -51,8 +65,19 @@ impl hotshot_types::traits::state::TestableState for State {
     }
 }
 
-impl Committable for State {
+impl Committable for ValidatedState {
```
jbearer, line 68: I don't think this impl is, or should be, required anymore. Since we really only deal with the header commitment.

If we are going to implement it, we should commit to the roots, not the entire trees.

sequencer/src/state.rs
```diff
@@ -24,24 +32,30 @@ impl HotShotState for State {
...
+        parent_header: &Self::BlockHeader,
         _view_number: &Self::Time,
     ) -> Result<Self, Self::Error> {
-        todo!("Tracking issue: [https://github.com/EspressoSystems/espresso-sequencer/issues/968]")
+
+         // TODO validate fee state
+         if let Err(_) = _validate_proposal(parent_header, proposed_header) {
+             Err(BlockError::InvalidBlockHeader)
+         } else {
+             Ok(proposed_header.validated_state)
```
jbearer, line 44: It should now be possible to remove `validated_state` from the header, compute the updates to the state here using `self`, and simply verify the results against the merkle tree roots in `parent_header`/`proposed_header`

sequencer/src/state.rs
tbro: It is currently a bound on the trait: https://github.com/EspressoSystems/HotShot/blob/160e8c52e63a5bacb94751fa6777d1703b94986d/crates/types/src/traits/state.rs#L38

sequencer/src/state.rs
tbro: Yea I was just using the one we previously defined in `block.rs` to make minimum changes. but I will move it here.

sequencer/src/state.rs
tbro: 182ffde should do as described.

sequencer/src/block.rs
```diff
@@ -308,43 +298,25 @@ impl Header {
             transactions_root,
             fee_merkle_tree_root,
             block_merkle_tree_root,
-            validated_state,
         }
     }
 }
 
-fn _validate_proposal(parent: &Header, proposal: &Header) -> anyhow::Result<BlockMerkleTree> {
```
tbro, line 316: Removed in favor of `ValidatedState::validate_proposal()`. The signature has changed there somewhat so please verify this is heading in the right direction.  See below comment.

sequencer/src/state.rs
```diff
@@ -1,47 +1,141 @@
...
+            FeeMerkleTree::from_kv_set(20, Vec::<(FeeAccount, FeeAmount)>::new()).unwrap();
+        Self {
+            block_merkle_tree,
+            fee_merkle_tree,
+        }
+    }
+}
+
+impl ValidatedState {
+    fn validate_proposal(&self, parent: &Header, proposal: &Header) -> anyhow::Result<Self> {
```
tbro, line 38: This seemed like a good place to do the validation. All of this could just as easily be in `self.validate_and_apply()` but I liked keeping the trait implementation clean for the moment. Contents here are the previous `_validate_proposal` standalone function plus addition validation against fee and block merkle trees in state. The signature differs from that of `_validate_proposal` due to addition content. Maybe I am deviating from some original intention that I am not aware of? Original `_validate_proposal` returned `Result<BlockMerkleTree>`.

jbearer: Yeah I don't think there was any great intent in the original signature. I think the new signature is a result of just getting a clearer understanding of the problem. So I think this is the right direction 👍

sequencer/src/state.rs
```diff
@@ -1,47 +1,146 @@
...
+        let mut block_merkle_tree = self.block_merkle_tree.clone();
+        let block_merkle_tree_root = block_merkle_tree.commitment();
+        anyhow::ensure!(
+            parent.block_merkle_tree_root == block_merkle_tree_root,
+            anyhow::anyhow!(
+                "Invalid Block Merkle Tree Error: {}, {}",
+                parent.block_merkle_tree_root,
+                block_merkle_tree_root
+            )
+        );
```
jbearer, line 74: I don't think we need to validate this, though. We are keeping track of `parent` and the corresponding state (`self`) locally, so we should be able to trust that the correspond

sequencer/src/state.rs
```diff
@@ -1,47 +1,146 @@
...
+        proposed_header: &Self::BlockHeader,
+        parent_header: &Self::BlockHeader,
         _view_number: &Self::Time,
     ) -> Result<Self, Self::Error> {
-        todo!("Tracking issue: [https://github.com/EspressoSystems/espresso-sequencer/issues/968]")
+        match self.validate_proposal(parent_header, proposed_header) {
+            // Note that currently only block state is updated.
+            Ok(validated_state) => Ok(validated_state),
+            Err(e) => {
+                debug!("Invalid Proposal: {}", e);
```
jbearer, line 116: This should be at warn or error level

jbearer (commented):
Looking good

sequencer/src/state.rs
jbearer: We should be able to remove this now

jbearer: Whoops commented on the wrong review

PR #1064, opened 2024-01-31
Add builder info and fee

tbro:
Add builder info and fee amount to block header. Closes https://github.com/EspressoSystems/espresso-sequencer/issues/1044

So far this PR adds builder related fields to Header and implements rudimentary logic for signing the header. Some remaining tasks are:

  * get the mnemonic from config
  * deduct fee from account
  * more things probably


sequencer/src/block.rs
```diff
@@ -81,6 +88,10 @@ pub struct Header {
     pub block_merkle_tree_root: BlockMerkleCommitment,
     /// Root Commitment of `FeeMerkleTree`
     pub fee_merkle_tree_root: FeeMerkleCommitment,
+    /// Account (etheruem address) of builder
+    pub builder_address: Option<FeeAccount>,
+    pub builder_signature: Option<Signature>,
+    pub builder_fee_amount: Option<FeeAmount>,
```
jbearer, line 94: These three fields have to be present together or not at all, so it would be cleaner to have `fee_info: Option<(FeeAccount, Signature, FeeAmount)>`. Maybe we can make a struct like
```rust
struct FeeInfo {
  account: FeeAccount,
  signature: Signature,
  amount: FeeAmount,
}
```

sequencer/src/block.rs
```diff
@@ -190,6 +202,19 @@ impl Header {
             transactions_root,
             fee_merkle_tree_root,
             block_merkle_tree_root,
+            builder_address: None,
+            builder_signature: None,
+            builder_fee_amount: None,
+        };
+
+        let hash =
+            ethers::utils::hash_message(serde_json::to_string(&header).unwrap().into_bytes());
```
jbearer, line 211: I think it's better to use the commitment scheme we use everywhere else to hash the header (so just `header.commit()`. Then you can use the low-level [`sign`](https://docs.rs/ethers/latest/ethers/core/k256/ecdsa/signature/trait.Signer.html#method.sign) method to sign the commitment (which is just a byte vector)

sequencer/src/block.rs
```diff
@@ -190,6 +202,19 @@ impl Header {
             transactions_root,
             fee_merkle_tree_root,
             block_merkle_tree_root,
+            builder_address: None,
+            builder_signature: None,
+            builder_fee_amount: None,
```
jbearer, line 207: Very important that the address and fee are set _before_ signing, so that an adversary can't change the builder or cause the builder to pay more than intended.

sequencer/src/block.rs
```diff
@@ -270,7 +296,7 @@ impl BlockHeader for Header {
         let ValidatedState {
             fee_merkle_tree,
             block_merkle_tree,
-        } = ValidatedState::default();
+        } = instance_state.validated_state.clone();
```
jbearer, line 299: Nice, now we can actually write tests with non-empty state lol 👍

sequencer/src/options.rs
```diff
@@ -95,6 +95,13 @@ pub struct Options {
     /// Multiple modules can be specified, provided they are separated by --
     #[clap(raw = true)]
     modules: Vec<String>,
+
+    /// Mnemonic phrase for builder account.
+    ///
+    /// This is the address fees will be charged to.
+    /// It must be funded with ETH on the layer 1.
```
jbearer, line 102:
```suggestion
    /// It must be funded with ETH in the Espresso fee ledger
```

sequencer/src/state.rs
```diff
@@ -116,10 +114,31 @@ impl HotShotState for ValidatedState {
...
+        instance: &Self::Instance,
         parent_header: &Self::BlockHeader,
         proposed_header: &Self::BlockHeader,
     ) -> Result<Self, Self::Error> {
+        // check header signagure
+        // in the signed message these are set to `None`.
+        let mut verifiable_header = proposed_header.clone();
+        let builder_signature = verifiable_header.builder_signature.take();
+        let _builder_address = verifiable_header.builder_address.take();
+        let _builder_fee_amount = verifiable_header.builder_fee_amount.take();
```
jbearer, line 126: This shouldn't be necessary, since the presence of `signature` should not affect the commitment scheme, and the address and amount actually should be included in the signed message (see earlier comment)

sequencer/src/state.rs
```diff
@@ -116,10 +114,31 @@ impl HotShotState for ValidatedState {
...
+        let header_bytes = serde_json::to_string(&verifiable_header)
+            .unwrap()
+            .into_bytes();
+
+        if builder_signature
+            .unwrap()
```
jbearer, line 132: Should return errors here rather than panicking

sequencer/src/state.rs
```diff
@@ -178,7 +197,7 @@ pub struct FeeAmount(U256);
 #[derive(
     Default, Hash, Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq, PartialOrd, Ord,
 )]
-pub struct FeeAccount(Address);
+pub struct FeeAccount(pub Address);
```
jbearer, line 200: Nit: could also derive `from`, `into` instead of exposing this field

sequencer/src/block.rs
tbro: Humm they can't *really* be there "together or none at all" ... can they? I have to be able to construct the type w/out the signature, sign it, and then add the signature. Wouldn't it have to be something like:
```
FeeInfo {
  account: FeeAccount,
  signature: Option<Signature>,
  amount: FeeAmount
}
Header {
  fee_info: Option<FeeInfo>,
  ..snip..
}
```
Or maybe to think of it another way... Signature isn't really the signature of fee stuff, its the signature of the entire Header. So I might prefer this:

```
FeeInfo {
  account: FeeAccount,
  amount: FeeAmount
}
Header {
  fee_info: Option<FeeInfo>,
  signature: Option<Signature>,
  ..snip..
}
```
But then that makes me wonder if we can't just always have the FeeInfo and so it would be more like this:
```
FeeInfo {
  account: FeeAccount,
  amount: FeeAmount
}
Header {
  fee_info: FeeInfo,
  signature: Option<Signature>,
  ..snip..
}
```

jbearer: Huh yeah I kind of like that last idea. `FeeInfo` could just be default for the genesis with no signature.

sequencer/src/state.rs
```diff
@@ -45,63 +43,102 @@ impl Default for ValidatedState {
...
-        })
+/// Fetch receipts from the l1 and add them to local balance.
+fn update_balance(fee_merkle_tree: &mut FeeMerkleTree, parent: &Header) {
+    let receipts = fetch_fee_receipts(parent);
+    for FeeReceipt { recipient, amount } in receipts {
+        // Get the balance in order to add amount, ignoring the proof.
+        match fee_merkle_tree.universal_lookup(recipient) {
+            LookupResult::Ok(balance, _) => fee_merkle_tree
+                .update(recipient, balance.add(amount))
+                .unwrap(),
```
jbearer, line 98: Can use `update_with` here to avoid the extra lookup

sequencer/src/state.rs
```diff
@@ -45,63 +43,102 @@ impl Default for ValidatedState {
...
+            // Handle `NotFound` and `NotInMemory` by initializing
```
jbearer, line 99: `NotInMemory` we will need to handle differently, by looking up the value from a peer. Because it is possible that there is some value for this account, we just don't know it, in which case re-initializing the account is incorrect. Can you just leave a `TODO` here so we remember to fix this when we implement catchup?

sequencer/src/state.rs
```diff
@@ -45,63 +43,102 @@ impl Default for ValidatedState {
...
     }
 }
+/// Validate builder account by verifiying signature and charging the account.
+fn validate_builder(
+    fee_merkle_tree: &mut FeeMerkleTree,
+    proposed_header: &Header,
+) -> anyhow::Result<()> {
+    let mut verifiable_header = proposed_header.clone();
+    // These unraps should be safe since Header::new() must have
+    // set these fields.
```
jbearer, line 112: We cannot trust `Header::new`, it was executed by a different node (the builder, whereas we are acting as a replica here)

sequencer/src/state.rs
```diff
@@ -45,63 +43,102 @@ impl Default for ValidatedState {
...
+    );
+
+    // charge the fee to the builder
+    let mut fee_merkle_tree = fee_merkle_tree.clone();
+    match fee_merkle_tree.universal_lookup(builder_address) {
+        LookupResult::Ok(balance, _) => fee_merkle_tree
+            .update(builder_address, balance.sub(builder_fee_amount))
+            .unwrap(),
+        // `NotFound` or `NotInMemory` is a ghost account that
+        // somehow signed the block so it must be a BUG.
```
jbearer, line 135: Not true, any malicious builder can sign with an account that doesn't exist in an attempt to avoid paying a fee. For `NotFound`, returning an error is the correct thing to do (it's the same as not having enough balance). For `NotInMemory`, we will have to look up the account as part of catchup

sequencer/src/state.rs
```diff
@@ -45,63 +43,102 @@ impl Default for ValidatedState {
...
+            .verify(header_bytes, builder_address.address())
+            .is_ok(),
+        "Invalid Builder Signature"
+    );
+
+    // charge the fee to the builder
+    let mut fee_merkle_tree = fee_merkle_tree.clone();
+    match fee_merkle_tree.universal_lookup(builder_address) {
+        LookupResult::Ok(balance, _) => fee_merkle_tree
+            .update(builder_address, balance.sub(builder_fee_amount))
```
jbearer, line 132: The `sub` here will panic if the balance isn't enough to pay the fee. You can use `checked_sub` and return an error if the balance isn't sufficient, or just check the balance before subtracting.

jbearer (commented):
I like the way the validation is broken up!

sequencer/src/state.rs
tbro: fixed in 95fe801091f35ff6874efe52d24bd8215121872c

sequencer/src/state.rs
tbro: fixed in https://github.com/EspressoSystems/espresso-sequencer/commit/95fe801091f35ff6874efe52d24bd8215121872c

sequencer/src/state.rs
tbro: added in c48d5cbe7f1daf0e9a0dfcb0463a11c67fcc0dfa

sequencer/src/state.rs
tbro: fixed in f5e45de1b12025ecaf5ae4eab0f7a90c90c6c29e

sequencer/src/state.rs
tbro: fixed in 6ea06031645765d2cdcc6542bb4d73afa45f9823

sequencer/src/state.rs
tbro: :thinking: If only `update_with` had documentation :disappointed:

jbearer: You give it a closure to take the old value and replace it with the new value. Old value will be `None` if not present (ie becomes an insert in that case)

sequencer/src/header.rs
```diff
@@ -96,6 +97,9 @@ impl Committable for Header {
         self.fee_merkle_tree_root
             .serialize_with_mode(&mut fmt_bytes, ark_serialize::Compress::Yes)
             .unwrap();
+        let mut fee_info_bytes = vec![];
+        self.fee_info
+            .serialize_with_mode(&mut fee_info_bytes, ark_serialize::Compress::Yes);
```
jbearer, line 102: Hmm I think this serialization makes the commitment needlessly opaque (and difficult to port to other languages). Better to implement `Committable for FeeInfo` and then we can just use `.field("fee_info", self.fee_info.commit())`

sequencer/src/state.rs
```diff
@@ -229,9 +227,35 @@ impl hotshot_types::traits::states::TestableState for ValidatedState {
...
+    CanonicalDeserialize,
+)]
+/// `FeeInfo` holds data related to builder fees.
+pub struct FeeInfo {
+    account: FeeAccount,
+    amount: FeeAmount,
+}
+impl FeeInfo {
+    pub fn new(account: FeeAccount) -> Self {
+        let amount = FeeAmount::default(); // TODO grab from config (instance_state?)
```
jbearer, line 250: Leaving this 0 is fine for now. Eventually we will have third party builders and each one will decide how to set it.

sequencer/src/state.rs
```diff
@@ -251,6 +251,21 @@ impl FeeInfo {
         Self { account, amount }
     }
 }
+
+impl Committable for FeeInfo {
+    fn commit(&self) -> Commitment<Self> {
+        let mut comm_bytes = vec![];
+        self.serialize_with_mode(&mut comm_bytes, ark_serialize::Compress::Yes)
+            .unwrap();
```
jbearer, line 259: We're still making this a bit awkward to port by serializing here. Better to do the individual fields as `var_size_field`. Serializing the fields themselves is fine because for both `FeeAccount` and `FeeAmount`, the serialization just converts directly to bytes in the intuitive way

jbearer (commented):
Looks good. Can we make a similar `as_bytes` function for `FeeAmount` using `to_little_endian` as we do in `serialize` (and then replace the usage in `serialize` with `as_bytes`)?

sveitser:
What's the reason for having this commit https://github.com/EspressoSystems/espresso-sequencer/pull/1064/commits/a63874f4ed14c9644d4b9c66725a8461287da288 as part of this PR?

tbro:
> What's the reason for having this commit [a63874f](https://github.com/EspressoSystems/espresso-sequencer/commit/a63874f4ed14c9644d4b9c66725a8461287da288) as part of this PR?

It is a squashed version of merge from origin/main. Needed for conflict resolution.

sequencer/src/header.rs
```diff
@@ -183,6 +192,23 @@ impl Header {
...
+            builder_signature: None,
+        };
+
+        // Sign Header with builder wallet from state
+        let signing_key: &SigningKey = builder_address.signer();
+        let header_signature: ecdsa::Signature = signing_key.sign(header.commit().as_ref());
+        // In order to use `type::Signature` for verification we do a
+        // `TryFrom` conversion on the signature. But first we need to
+        // append Etherium's `Electurm` value (can be 27 or 28). There
+        // must be a better way to do this.
```
jbearer, line 205: There does not seem to be a better way to do this. I can't believe ethers doesn't just have a function to sign arbitrary bytes and get a `Signature` 🙄 

Anyways, I think the choice of `v` is not arbitrary (has to be either 27 or 28 depending on the signing key). We can a different overloading of `sign` which gives an `ecdsa::Signature` and a `RecoveryId`  (which tells you what `v` is supposed to be) and then copy what ethers does to reconstruct the `types::Signature` [here](https://docs.rs/ethers-signers/2.0.13/src/ethers_signers/wallet/mod.rs.html#152)

sveitser:
> > What's the reason for having this commit [a63874f](https://github.com/EspressoSystems/espresso-sequencer/commit/a63874f4ed14c9644d4b9c66725a8461287da288) as part of this PR?
> 
> It is a squashed version of merge from origin/main. Needed for conflict resolution.

Interestingly github was showing me all the changes in that commit as part of the diff here. Now this is no longer the case :)

sequencer/src/header.rs
sveitser: This might be why there's no such functionality in ethers-rs: https://github.com/gakonst/ethers-rs/pull/2369#issuecomment-1520607943

jbearer: Ah, I guess that decision makes sense. In our case it should be fine, since we're not signing arbitrary data, but it's up to the software to be sure of that. We may have to revisit this decision in the future if we ever want to support hardware signers, but that's probably a ways off

sequencer/src/header.rs
```diff
@@ -218,18 +218,19 @@ impl Header {
...
-        let header_signature = [header_signature.to_vec(), vec![27]].concat();
+        // Sign our header using its `Committment` as a prehash.
+        let builder_signature = builder_address
+            .sign_hash(types::H256(header.commit().into()))
+            .unwrap();
+
+        assert_eq!(
+            builder_address.address(),
+            parent.fee_info.account().address()
+        );
```
jbearer, line 229: What is this assert? It's fine for the parent and the current header to be built/signed by different builders

sequencer/src/header.rs
```diff
@@ -218,18 +218,19 @@ impl Header {
...
-        let header_signature: ecdsa::Signature = signing_key.sign(header.commit().as_ref());
-        // In order to use `types::Signature` for verification we do a
-        // `TryFrom` conversion on the signature. But first we need to
-        // append Etherium's `Electurm` value (can be 27 or 28). There
-        // must be a better way to do this.
-        let header_signature = [header_signature.to_vec(), vec![27]].concat();
+        // Sign our header using its `Committment` as a prehash.
+        let builder_signature = builder_address
+            .sign_hash(types::H256(header.commit().into()))
+            .unwrap();
```
jbearer, line 224: Not sure about this unwrap. I can't think of why a signature would fail, but it isn't documented and maybe there is some weird case where it does. Anyways, since `Header::new` currently doesn't return a result, there's not much we can do here *but* panic. Although maybe it would be cleaner to make `from_info` return a result and then just have a top level `unwrap` in `Header::new` with a comment explaining we should fix this when HotShot is updated to [handle failures in block building](https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/New.20failure.20modes)

jbearer:
Come to think of it, the validation helper functions in `state.rs` could be reworked a bit so they just update the state but don't validate it against the proposed header, and then the caller could just update the state and check against the proposed header at the end. This would allow us to reuse the state update functions in `Header::new`, which would reduce the risk of bugs where `Header::new` and `validate_and_apply_header` are inconsistent. This is not necessary though, could be a followup tech debt issue

sequencer/src/header.rs
```diff
@@ -191,16 +191,21 @@ impl Header {
...
             // Get the balance in order to add amount, ignoring the proof.
-            match fee_merkle_tree.universal_lookup(recipient) {
-                LookupResult::Ok(balance, _) => fee_merkle_tree
-                    .update(recipient, balance.add(amount))
-                    .unwrap(),
+            match fee_merkle_tree.update_with(account, |balance| {
+                Some(balance.cloned().unwrap_or_default().add(amount))
+            }) {
+                Ok(LookupResult::Ok(..)) => (),
                 // Handle `NotFound` and `NotInMemory` by initializing
```
jbearer, line 203: `NotFound` is now handled by `update_with` (the case where we `unwrap_or_default` to initialize the balance. `NotInMemory` will be handled in future work by catchup

jbearer (approved):
I think we're good once CI passes

tbro:
> Come to think of it, the validation helper functions in `state.rs` could be reworked a bit so they just update the state but don't validate it against the proposed header, and then the caller could just update the state and check against the proposed header at the end. This would allow us to reuse the state update functions in `Header::new`, which would reduce the risk of bugs where `Header::new` and `validate_and_apply_header` are inconsistent. This is not necessary though, could be a followup tech debt issue

Should we create an issue for this so we don't forget?

Issue #1046, opened 2024-01-26
[cappuccino milestone] VID namespace proofs in live production


ggutoski:
Currently all the VID stuff is unused in `block2.rs`. Current namespace proofs are NMT-based in `block.rs`. We need to swap out NMT and use VID.

To this end we need to switch the namespace / tx proof query endpoints to use a VID struct instead of the NMT struct. In principle it should be just a matter of following the compiler.

@philippecamacho @akonring let's coordinate on who will take on this issue.

- [x] #1067 
- [x] #1080

akonring:
Brief note after discussion on Zulip:
We might not want to have `Header`-related stuff in `block2.rs`.
Preliminary suggestion is to split `Header` into a separate file.

PR #1066, opened 2024-02-01
update to pin hostshot version to 0.5.11

tbro:
Update hotshot version. This is needed for sequencer to implement ValidatedState. This PR will track changes needed to incorporate current `main` branch of hotshot into sequencer.

Addresses #1074


sequencer/src/block.rs
```diff
@@ -446,8 +404,8 @@ impl BlockHeader for Header {
         self.payload_commitment
     }
 
-    fn metadata(&self) -> &NMTRoot {
-        &self.transactions_root
+    fn metadata(&self) -> &<Self::Payload as BlockPayload>::Metadata {
+        unimplemented!("TODO");
```
jbearer, line 408: I think this should be changed to return the namespace table. Maybe it can just return `()` for now if the VID integration is still in progress. I'm a bit unsure how that change is going to interact with this change. cc @ggutoski

sequencer/src/lib.rs
```diff
@@ -184,22 +184,21 @@ impl<N: network::Type> NodeImplementation<SeqTypes> for Node<N> {
...
+
 impl NodeType for SeqTypes {
     type Time = ViewNumber;
     type BlockHeader = Header;
     type BlockPayload = Payload;
     type SignatureKey = PubKey;
     type Transaction = Transaction;
     type ElectionConfigType = ElectionConfig;
-    type StateType = State;
+    type InstanceState = Instance;
```
jbearer, line 200: Should this be `NodeState`? Or what is `Instance`?

sequencer/src/lib.rs
```diff
@@ -315,7 +316,7 @@ pub async fn init_node(
     let state_ver_keys = (0..num_nodes)
         .map(|i| StateKeyPair::generate_from_seed_indexed(config.seed, i as u64).ver_key())
         .collect::<Vec<_>>();
-    let state_key_pair = config.config.my_own_validator_config.state_key_pair.clone();
+    let state_key_pair = StateKeyPair::generate_from_seed_indexed(config.seed, node_index);
```
jbearer, line 319: Why this change?

sequencer/src/state.rs
```diff
@@ -1,47 +1,164 @@
...
+                parent.height,
+                proposal.height
+            )
+        );
+
+        // validate parent fee merkle tree root against state
+        let fee_merkle_tree = self.fee_merkle_tree.clone();
+        let fee_merkle_tree_root = fee_merkle_tree.commitment();
+        anyhow::ensure!(
+            parent.fee_merkle_tree_root == fee_merkle_tree_root,
```
jbearer, line 56: We shouldn't have to check this, `parent` and `fee_merkle_tree` are both local state so we can trust them

sequencer/src/state.rs
```diff
@@ -1,47 +1,164 @@
...
+                fee_merkle_tree_root
+            )
+        );
+
+        // validate parent block merkle tree root against state
+        // `clone` to avoid possible side-effects
+        let mut block_merkle_tree = self.block_merkle_tree.clone();
+        let block_merkle_tree_root = block_merkle_tree.commitment();
+        anyhow::ensure!(
+            parent.block_merkle_tree_root == block_merkle_tree_root,
```
jbearer, line 69: Same

sequencer/src/state.rs
```diff
@@ -1,47 +1,164 @@
...
+        let block_merkle_tree_root = block_merkle_tree.commitment();
+        anyhow::ensure!(
+            proposal.block_merkle_tree_root == block_merkle_tree_root,
+            anyhow::anyhow!(
+                "Invalid Block Merkle Tree Error: {}, {}",
+                block_merkle_tree_root,
+                proposal.block_merkle_tree_root
+            )
+        );
+
```
jbearer, line 88: We also need to update `fee_merkle_tree` here the same way we do in `Header::new` (e.g. by getting the deposits and inserting them)

sequencer/src/lib.rs
tbro: I'm building on top of #1041. 

https://github.com/EspressoSystems/espresso-sequencer/pull/1041/files#diff-d10b6700d92426a01587eb55cefd08511231a87147b0616810bded2dcc2dbb74R318

But I guess that makes it hard to review. I can move the target branch or make another PR with #1099 as the target. or some other approach...

sequencer/src/block.rs
ggutoski: VID migration has not yet begun. I don't think it matters much what we do here because it'll change soon in the VID integration.  (At that time it'll return the namespace table.) We cannot return `()` here unless we change the `Metadata` assoc type from `NMTRoot` to `()` and I see no reason to bother with that.

sequencer/src/lib.rs
jbearer: Ooh yeah moving the target branch so we merge into #1041 instead of accumulating the changes would make it much easier to review

jbearer: I do have this question though for @alxiong or @mrain : how come we're re-generating the key pair now instead of using the one from `my_own_validator_config`?

sequencer/src/block.rs
jbearer: Well, the reason would be to remove this `unimplemented!()`. Nothing will work with that there

sequencer/src/lib.rs
mrain: I was planing to remove the `StateKeyPair` from the consensus scope. But it's not fully decided yet.

jbearer: Oh gotcha

sequencer/src/block.rs
tbro: I have a quick fix in c274c49 that should be good enough for now.

sequencer/src/lib.rs
tbro: Yes it should be. Fixed in c958648

sequencer/src/lib.rs
mrain: Oh I really don't recommend basing on #1041 for this hotshot upgrade, b/c we have so many breaking changes there. Since you have already done it, remember to revert the changes here: [f48af9a](https://github.com/EspressoSystems/espresso-sequencer/pull/1066/commits/f48af9a7bb64cf26ed113c48e130af9a8cbede68)

tbro: :thinking:

alxiong: > I do have this question though for @alxiong or @mrain : how come we're re-generating the key pair now instead of using the one from my_own_validator_config?
> 
> [mrain](https://github.com/mrain) [Feb 9, 2024](https://github.com/EspressoSystems/espresso-sequencer/pull/1066#discussion_r1483666648)
> 
> I was planing to remove the StateKeyPair from the consensus scope. But it's not fully decided yet.

I don't understand the change as well. but judging from the latest tip of this PR, this particular change is reverted?

also @mrain wdym by "removing the StateKeyPair from the consensus scope?

tbro: > Oh I really don't recommend basing on #1041 for this hotshot upgrade, b/c we have so many breaking changes there. Since you have already done it, remember to revert the changes here: [f48af9a](https://github.com/EspressoSystems/espresso-sequencer/pull/1066/commits/f48af9a7bb64cf26ed113c48e130af9a8cbede68)

I spent some time trying to disentangle our PRs. The problem is that we have mutual inter-dependencies. The #1041 is making updates that this PR needs in order to build. I think what we can do is integrate this PR with #1041 first, then do the merge to main. Any objection @jbearer @alxiong @mrain ?

sequencer/src/block.rs
tbro: e0d9f3b4c60d361f96c4228b0c95efce400e2642 should fix it.

sequencer/src/lib.rs
```diff
@@ -520,18 +522,18 @@ mod test {
...
         setup_backtrace();
-
-        TestMetadata {
-            overall_safety_properties: OverallSafetyPropertiesDescription {
-                num_successful_views: 10,
-                ..Default::default()
-            },
-            ..Default::default()
-        }
-        .gen_launcher::<SeqTypes, Node<network::Memory>>(0)
```
tbro, line 531: It's unclear what needs to be implemented here in order to satisfy the compiler.

jbearer:
This still seems to have changes in it not related to the HotShot upgrade (all the stuff in contracts, e.g.). Not sure what happened

sequencer/src/lib.rs
mrain: No objection.

tbro:
> This still seems to have changes in it not related to the HotShot upgrade (all the stuff in contracts, e.g.). Not sure what happened

Ok I think I fixed all the weirdness. the remaining seemingly unrelated changes under `contracts` is in order to work around state-prover dependecy by @mrain at f48af9a7

sequencer/src/block.rs
```diff
@@ -752,29 +723,30 @@ mod test_headers {
...
             let FeeReceipt { recipient, amount } = FeeReceipt::default();
             let fee_merkle_tree =
                 FeeMerkleTree::from_kv_set(20, Vec::from([(recipient, amount)])).unwrap();
             let fee_merkle_tree_root = fee_merkle_tree.commitment();
 
-            let validated_state = ValidatedState {
+            let _validated_state = ValidatedState {
                 block_merkle_tree,
                 fee_merkle_tree,
             };
```
jbearer, line 745: If this is not used anywhere, should we delete it?

sequencer/src/block.rs
```diff
@@ -917,19 +890,22 @@ mod test_headers {
...
     #[test]
     fn test_validate_proposal_error_cases() {
-        let (mut header, ..) = Header::genesis();
-        let mut block_merkle_tree = header.validated_state.block_merkle_tree.clone();
+        let (mut header, ..) = Header::genesis(&NodeState {});
+        let mut validated_state = ValidatedState::from_header(&header);
+        let mut block_merkle_tree = validated_state.block_merkle_tree.clone();
 
         // Populate the tree with an initial `push`.
         block_merkle_tree.push(header.commit()).unwrap();
```
jbearer, line 898: I think this is going to be a problem, because `ValidatedState::from_header` creates a partial view of the merklized state (just the root hashes, not any leaves) so we wouldn't be able to push to that. In actual operation we would have to fetch the missing state if we encountered this scenario. But for test purposes, we know what the state should be and so we might just have to create it separately from the header and assert that it's consistent.

sequencer/src/block.rs
```diff
@@ -938,30 +914,46 @@ mod test_headers {
...
     fn test_validate_proposal_success() {
-        let (mut header, ..) = Header::genesis();
-        let mut block_merkle_tree = header.validated_state.block_merkle_tree.clone();
+        let state = NodeState {};
+        let (mut header, _, metadata) = Header::genesis(&state);
+        let mut parent_state = ValidatedState::from_header(&header);
+        let mut block_merkle_tree = parent_state.block_merkle_tree.clone();
 
         // Populate the tree with an initial `push`.
         block_merkle_tree.push(header.commit()).unwrap();
```
jbearer, line 932: Same here

sequencer/src/lib.rs
jbearer: The first argument of `gen_launcher` is `TYPES: NodeType`, so it seems like `SeqTypes` ought to be correct What error are you getting?

sequencer/src/state.rs
```diff
@@ -1,47 +1,155 @@
...
+                tracing::warn!("Invalid Proposal: {}", e);
+                Err(BlockError::InvalidBlockHeader)
+            }
+        }
+    }
+    /// Construct the state with the given block header.
+    ///
+    /// This can also be used to rebuild the state for catchup.
+    fn from_header(_block_header: &Self::BlockHeader) -> Self {
+        ValidatedState::default()
```
jbearer, line 123: It will be hard to tell if the tests are working with this dummy implementation in. I'm a bit afraid we might updating tests in ways that they shouldn't be updated, and then have to fix them again once we implement this for real. So I think it might be a good idea to go ahead and implement this now. You can use [`from_commitment`](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/trait.ForgetableMerkleTreeScheme.html#tymethod.from_commitment) to generate completely sparse views of the merkle trees from teh commitments contained in `block_header`

sequencer/src/state.rs
```diff
@@ -1,47 +1,155 @@
...
+    }
+    /// Construct a genesis validated state.
+    #[must_use]
+    fn genesis(instance: &Self::Instance) -> Self {
+        Self::from_header(&Self::BlockHeader::genesis(instance).0)
+    }
+}
+
+// FIXME remove when `Commitable` is removed from the trait
+impl Committable for ValidatedState {
```
jbearer, line 133: I think we can remove this now

sequencer/src/lib.rs
tbro:
```
error[E0599]: the method `run_test` exists for struct `TestRunner<SeqTypes, Node<Memory>>`, but its trait bounds were not satisfied
   --> sequencer/src/lib.rs:539:10
    |
171 |   pub struct SeqTypes;
    |   ------------------- doesn't satisfy `<_ as NodeType>::InstanceState = TestInstanceState`
...
530 | /         TestMetadata {
531 | |             overall_safety_properties: OverallSafetyPropertiesDescrip...
532 | |                 num_successful_views: 10,
533 | |                 ..Default::default()
...   |
538 | |         .launch()
539 | |         .run_test()
    | |         -^^^^^^^^ method cannot be called on `TestRunner<SeqTypes, Node<Memory>>` due to unsatisfied trait bounds
    | |_________|
    | 
    |
    = note: the following trait bounds were not satisfied:
            `<SeqTypes as hotshot_types::traits::node_implementation::NodeType>::InstanceState = hotshot_testing::state_types::TestInstanceState`

```

sequencer/src/state.rs
tbro: removed in 488f71e106c6785ba85000721b03d42b1c3a9699

sequencer/src/state.rs
tbro: fixed in c8440bfff33f85094ae1635a4c0761a4be741527

sequencer/src/block.rs
tbro: fixed in c8440bfff33f85094ae1635a4c0761a4be741527

sequencer/src/block.rs
tbro: fixed in c8440bfff33f85094ae1635a4c0761a4be741527

sequencer/src/state.rs
tbro: fixed in 488f71e106c6785ba85000721b03d42b1c3a9699

sequencer/src/state.rs
tbro: fixed in 1ad224b1dfae9c120f2505ba5f19fda15f5786a0

jbearer (approved):

sequencer/src/lib.rs
jbearer: Ohhh it seems we can only run the HotShot tests if we use their test types. TBH I'm not even sure this test is worth maintaining then. We have plenty of our own tests that run HotShot. You can delete this test.

sequencer/src/block.rs
```diff
@@ -210,7 +210,9 @@ impl BlockHeader for Header {
             mut fee_merkle_tree,
             ..
         } = parent_state.clone();
-        block_merkle_tree.push(parent_header.commit()).unwrap();
+        if let Err(e) = block_merkle_tree.push(parent_header.commit()) {
+            dbg!(e);
```
jbearer, line 214: I would prefer to panic! here for now, since it shouldn't happen in any of our current tests and if it does happen the state is incorrect. Eventually we will replace the panic with fetching the missing frontier

tbro: Ok. I'll put it back. I'm currently trying to gain insight on the failing tests.

tbro: Rereading your comment... in our current tests we are panicking here with the error: `"Given part of merkle tree is not in memory"`

sequencer/src/lib.rs
```diff
@@ -583,8 +564,13 @@ mod test {
 
             // Check that each successive header satisfies invariants relative to its parent: all
             // the fields which should be monotonic are.
-            for leaf in leaf_chain.iter().rev() {
+            'inner: for (i, leaf) in leaf_chain.iter().rev().enumerate() {
                 let header = leaf.block_header.clone();
+                if i == 0 {
+                    parent = header;
+                    continue 'inner;
```
jbearer, line 571: Is the `'inner` label necessary? Wouldn't just `continue;` work here?

tbro: Yes I thought explicitness might be nice there. But it is unneeded so removed in d55614106.

sequencer/src/state.rs
```diff
@@ -1,47 +1,162 @@
...
+                Err(BlockError::InvalidBlockHeader)
+            }
+        }
+    }
+    /// Construct the state with the given block header.
+    ///
+    /// This can also be used to rebuild the state for catchup.
+    fn from_header(block_header: &Self::BlockHeader) -> Self {
+        let fee_merkle_tree = FeeMerkleTree::from_commitment(block_header.fee_merkle_tree_root);
+        let block_merkle_tree = BlockMerkleTree::from_commitment(block_header.fee_merkle_tree_root);
```
sveitser, line 137: @tbro isn't this supposed to be `block_merkle_tree_root`?

sequencer/src/state.rs
```diff
@@ -51,8 +166,111 @@ impl hotshot_types::traits::state::TestableState for State {
...
+        _compress: Compress,
+        _validate: Validate,
+    ) -> Result<Self, SerializationError> {
+        let mut bytes = [0u8; core::mem::size_of::<Address>()];
+        reader.read_exact(&mut bytes)?;
+        let value = Address::from_slice(&bytes);
+        Ok(Self(value))
+    }
+}
+impl std::convert::From<u64> for FeeAccount {
```
sveitser, line 247: @tbro is this impl used? If not, what is it for?

sequencer/src/state.rs
tbro: No. from_commitment()` returns a merkle tree that contains only the commitment to be updated later (through catch up I believe).

sequencer/src/state.rs
tbro: It is only there to comply with the bounds of UniversalMerkleTree.  I believe that bound has since been removed though, so we can probably remove it.

ggutoski: @sveitser I'm currently working on #1046 , which will render some of these questions irrelevant. Now that this PR is merged I suggest you save your time for future reviews.

PR #1073, opened 2024-02-06
feat: RPC-friendly verification of namespace proofs

### LINKED ISSUE #1067: New `NamespaceProof` struct with sequencer-friendly methods
Currently `NamespaceProof` is a simple type alias into jellyfish:
https://github.com/EspressoSystems/espresso-sequencer/blob/e8f4ddc359ae98b5df769324cd51131b1d14238f/sequencer/src/block2.rs#L307-L308

We should make it into a struct that is easily mapped to convenient RPC endpoints. Current plan looks like the following (except in actual Rust not some illegitimate Rust-JSON hybrid):
```rust
struct NamespaceProof{
  "ns_payload": [...],
  "proof": {
    "vid_common": { ... },
    "vid_namespace_proof" { ... },
}
```
and a method like
```rust
fn verify(&self) -> Result<(Vec<Transaction>, NamespaceId, BlockCommitment), Err>;
```
@nomaxg please correct any misunderstanding.

ggutoski:
close #1067 

- type alias `NamespaceProof` replaced with a new struct with that name
- new method `NamespaceProof::verify` with API from #1067 + additional args `vid`, `commit`.
- test demos use of `verify`
- tech debt to refactor in the future:
  - resurrect function `_get_tx_table_entry`, make it like `get_ns_table_entry`
  - new function `parse_ns_payload`

### DIFF
--- a/sequencer/src/block2/payload.rs
+++ b/sequencer/src/block2/payload.rs
@@ -9,7 +9,7 @@ use jf_primitives::pcs::prelude::UnivariateKzgPCS;
 use jf_primitives::pcs::{checked_fft_size, PolynomialCommitmentScheme};
 use jf_primitives::vid::advz::payload_prover::{LargeRangeProof, SmallRangeProof};
 use jf_primitives::vid::advz::Advz;
-use jf_primitives::vid::payload_prover::PayloadProver;
+use jf_primitives::vid::{payload_prover::PayloadProver, VidScheme};
 use num_traits::PrimInt;
 use serde::{Deserialize, Serialize};
 use snafu::OptionExt;
@@ -288,6 +288,17 @@ pub(super) type RangeProof =
 /// but that's still pretty crufty.
 pub type JellyfishNamespaceProof =
     LargeRangeProof<<UnivariateKzgPCS<Bls12_381> as PolynomialCommitmentScheme>::Evaluation>;
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct NamespaceProof<V>
+where
+    V: PayloadProver<JellyfishNamespaceProof>,
+{
+    ns_payload_flat: Vec<u8>,
+    ns_index: u32,
+    ns_proof: JellyfishNamespaceProof,
+    vid_common: <V as VidScheme>::Common,
+}
 
 // TODO find a home for this function
 fn parse_ns_payload(ns_payload_flat: &[u8], ns_id: VmId) -> Vec<Transaction> {
--- a/sequencer/src/block2/payload.rs
+++ b/sequencer/src/block2/payload.rs
@@ -1,6 +1,6 @@
 use crate::block2::entry::{TxTableEntry, TxTableEntryWord};
 use crate::block2::payload;
-use crate::{BlockBuildingSnafu, Error, VmId};
+use crate::{BlockBuildingSnafu, Error, Transaction, VmId};
 use ark_bls12_381::Bls12_381;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use commit::Committable;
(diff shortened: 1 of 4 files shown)


nomaxg (commented):
Looks good, just one question about how we authenticate ns_id

sequencer/src/block2.rs
```diff
@@ -304,9 +314,52 @@ pub(super) type RangeProof =
...
+            },
+            &self.ns_proof,
+        )
+        .ok()?
+        .ok()?;
+
+        // verification succeeded, return some data
+        Some((
+            parse_ns_payload(&self.ns_payload_flat, self.ns_id),
+            self.ns_id,
```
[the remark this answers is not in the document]
ggutoski: Jellyfish doesn't know anything about namespaces, so `payload_verify` has no way to check `ns_id`.

Ugh if you want to authenticate `ns_id` then `verify` needs additional args: the namespace table (aka `metadata`) and the index for the namespace (so we know where to look in the namespace table to check `ns_id`).

I suppose you'll also want to authenticate the namespace table, too, as otherwise an attacker could make up any namespace table it needs to pass verification. I don't even know whether we currently have infra for that.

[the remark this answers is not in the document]
ggutoski: In principle we could eliminate the need for these additional args to `verify` (that is, the namespace table and index) by storing them in the namespace proof like we do for everything else. But we'd still need some way to authenticate this data against a block commitment.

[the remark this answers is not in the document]
ggutoski: Really all that is accomplished in the present PR is that we prove that *some* specific subslice of the payload has some specific bytes. We prove nothing about the ID of the namespace corresponding to that subslice, or even whether that subslice corresponds to a namespace at all.

This is going to require more work. We should meet to discuss.

sequencer/src/block2.rs
```diff
@@ -647,28 +727,31 @@ mod test {
                 );
 
                 // test ns proof
-                let (ns_payload_flat_from_proof, ns_proof) = block
-                    .namespace_with_proof(&actual_ns_table, ns_idx)
+                let ns_proof2 = block
```
philippecamacho, line 730: Why the variable name `ns_proof_2`?

sequencer/src/block2.rs
```diff
@@ -647,28 +727,31 @@ mod test {
...
-                        payload_subslice: &ns_payload_flat_from_proof,
-                        range: actual_ns_payload_range,
-                        commit: &disperse_data.commit,
-                        common: &disperse_data.common,
-                    },
-                    &ns_proof,
-                )
-                .unwrap()
-                .unwrap_or_else(|_| panic!("namespace {} proof verification failure", ns_id.0));
+                let (ns_proof2_txs, ns_proof2_ns_id) = ns_proof2
```
philippecamacho, line 743: I guess there was some discussion about this API, but it seems unnatural for a proof verification to return a list of txs and a namespace id instead of a `boolean / Result<(),Error>`

sequencer/src/block2.rs
ggutoski: Sorry I forgot to change that. I'll fix it.

sequencer/src/block2.rs
ggutoski: Yeah, there was discussion offline. I'm told that it makes life easy for use as an RPC endpoint. I don't ask questions. 😉

sequencer/src/block2.rs
ggutoski: done in https://github.com/EspressoSystems/espresso-sequencer/pull/1073/commits/5c461b4331ef71db1ffe9362ce5ea2677fb51407

ggutoski:
@nomaxg I think I got everything---you happy? It remains only to merge with main, which is nontrivial so I'll do it tomorrow.

ggutoski:
@nomaxg @philippecamacho force-pushed a manual merge main.

nomaxg (approved):
verify looks good! I'll let someone else check the cryptography

philippecamacho (approved):
LGTM

PR #1078, opened 2024-02-08
945 smart contract specifications

alysiahuggins:
- expanded on comments on the lightclient contract


contracts/src/LightClient.sol
```diff
@@ -7,7 +7,19 @@ import { IPlonkVerifier } from "./interfaces/IPlonkVerifier.sol";
...
 import { LightClientStateUpdateVK as VkLib } from "./libraries/LightClientStateUpdateVK.sol";
 
-/// @notice A light client for HotShot consensus. Keeping track of its finalized states in safe,
+/// @title Light Client Contract
+/// @notice This contract serves as an always-on client
+/// that verifies HotShot's state (Espresso's consensus state) which can be used by
+/// Rollup contracts on L1 (Ethereum).
+/// This state is submitted by any state-prover with evidence which is
+/// a SNARK proof that proves consensus.
+/// This contract also keeps track of the current epoch and stake table.
```
philippecamacho, line 16: Implicitly the HotShot state will refer to some stake table state but the light client contract does not interact nor read from the stake table contract.

philippecamacho: Well, my observation above is a bit confusing. I think no change is needed in the code comment. Just wanted to highlight that the light client proactively updates the epoch for other contracts to read (e.g. stake table), but only receives a commitment to the stake table state.

philippecamacho: So this is important because it means the light client contract MUST be updated at least once per epoch. We should document this both in the gitbook (did not check yet maybe it's already written down!) and in the comment of the function `newFinalizedState`.

This is not relevant for Decaf but we can easily forget about this implicit property. Let us document it.

alysiahuggins: Thanks so much for highlighting this `the light client contract MUST be updated at least once per epoch` is not currently documented so i'll add to both the function comments and the gitbook

alysiahuggins: I can also remove `and stake table` if that makes it less confusing - don't want to add it if it can cause an incorrect  assumptions

philippecamacho: Ok, let us remove it for now, especially because the stake table is fixed for Decaf. Thank you.

contracts/src/LightClient.sol
```diff
@@ -18,11 +30,13 @@ contract LightClient {
     // === Constants ===
     //
     /// @notice System parameter: number of blocks per epoch
+    /// - not used in Decaf to set to type(uint32).max when deployed
     uint32 public immutable BLOCKS_PER_EPOCH;
 
     // === Storage ===
     //
     /// @notice genesis block commitment
+    /// @dev the genesis block contains a single Genesis transaction
```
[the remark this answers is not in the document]
alysiahuggins: ok I can update it to `the genesis block contains an empty block merkle tree and empty fee ledger state`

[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/dd362f7ab0db1761c763427661bdf2a2703cc8d6

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/dd362f7ab0db1761c763427661bdf2a2703cc8d6

contracts/src/LightClient.sol
```diff
@@ -106,10 +124,16 @@ contract LightClient {
 
     // === State Modifying APIs ===
     //
-    /// @notice Update the latest finalized light client state. It is expected to be updated
-    /// periodically, especially an update for the last block for every epoch has to be submitted
+    /// @notice Update the latest finalized light client state. It must be updated
+    /// periodically (atleast once per epoch), especially an update for the last block for every
```
philippecamacho, line 128:
```suggestion
    /// periodically (at least once per epoch), especially an update for the last block for every
```

jbearer (approved):
LGTM

PR #1079, opened 2024-02-09
Upgradable light client contract

philippecamacho:
WIP

### DIFF
--- a/contracts/src/LightClient.sol
+++ b/contracts/src/LightClient.sol
@@ -2,6 +2,12 @@
 
 pragma solidity ^0.8.0;
 
+import { OwnableUpgradeable } from
+    "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import { UUPSUpgradeable } from
+    "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
 import { BN254 } from "bn254/BN254.sol";
 import { IPlonkVerifier } from "./interfaces/IPlonkVerifier.sol";
 import { PlonkVerifier } from "./libraries/PlonkVerifier.sol";
@@ -9,16 +15,19 @@ import { LightClientStateUpdateVK as VkLib } from "./libraries/LightClientStateU
 
 /// @notice A light client for HotShot consensus. Keeping track of its finalized states in safe,
 /// authenticated ways.
-contract LightClient {
+contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     // === Events ===
     //
     // @notice Notify a new epoch is starting
     event EpochChanged(uint64);
 
+    /// @notice upgrade event when the proxy updates the implementation it's pointing to
+    event Upgrade(address implementation);
+
     // === Constants ===
     //
     /// @notice System parameter: number of blocks per epoch
-    uint32 public immutable BLOCKS_PER_EPOCH;
+    uint32 public BLOCKS_PER_EPOCH; // TODO how to make this variable immmutable while using the UUPS pattern?
 
     // === Storage ===
     //
@@ -76,7 +85,29 @@ contract LightClient {
     /// @notice Wrong plonk proof or public inputs.
     error InvalidProof();
 
-    constructor(LightClientState memory genesis, uint32 numBlockPerEpoch) {
+    /// @notice since the constuctor initializes storage on this contract we disable it
+    /// @dev storage is on the proxy contract since it calls this contract via delegatecall
+    constructor() {
+        _disableInitializers();
+    }
+
+    /// @notice This contract is called by the proxy when you deploy this contract
+    function initialize(LightClientState memory genesis, uint32 numBlockPerEpoch)
+        public
+        initializer
+    {
+        __Ownable_init(msg.sender); //sets owner to msg.sender
+        __UUPSUpgradeable_init();
+        _initializeState(genesis, numBlockPerEpoch);
+    }
+
+    /// @notice only the owner can authorize an upgrade
+    function _authorizeUpgrade(address newImplementation) internal override onlyOwner {
+        emit Upgrade(newImplementation);
+    }
+
+    // TODO are the decorators of this function correctly set?
+    function _initializeState(LightClientState memory genesis, uint32 numBlockPerEpoch) internal onlyOwner {
         // stake table commitments and threshold cannot be zero, otherwise it's impossible to
         // generate valid proof to move finalized state forward.
         // Whereas blockCommRoot can be zero, if we use special value zero to denote empty tree.
--- a/contracts/test/LightClient.t.sol
+++ b/contracts/test/LightClient.t.sol
@@ -100,7 +100,7 @@ contract LightClient_constructor_Test is LightClientCommonTest {
         init();
     }
 
-    /// @dev Test the constructor has initialized the contract state properly, espeically genesis
+    /// @dev Test the constructor has initialized the contract state properly, especially genesis
     /// block.
     function test_CorrectInitialization() external {
         assert(lc.BLOCKS_PER_EPOCH() == BLOCKS_PER_EPOCH_TEST);
--- a/contracts/test/mocks/LightClientTest.sol
+++ b/contracts/test/mocks/LightClientTest.sol
@@ -10,8 +10,10 @@ import { LightClientStateUpdateVKTest as VkLib } from "./LightClientStateUpdateV
 /// @dev A helper that wraps LightClient contract for testing
 contract LightClientTest is LC {
     constructor(LC.LightClientState memory genesis, uint32 numBlockPerEpoch)
-        LC(genesis, numBlockPerEpoch)
-    { }
+    {
+        _initializeState(genesis, numBlockPerEpoch);
+    }
+
 
     /// @dev Directly mutate `currentEpoch` variable for test
     function setCurrentEpoch(uint64 newEpoch) public {
--- a/contracts/script/HotShot.s.sol
+++ b/contracts/script/HotShot.s.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "forge-std/Script.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import { HotShot } from "../src/HotShot.sol";
 import { LightClient as LC } from "../src/LightClient.sol";
 
@@ -28,7 +29,14 @@ contract DeployHotShotScript is Script {
             abi.decode(result, (LC.LightClientState, bytes32, bytes32));
 
         LC.LightClientState memory genesis = state;
-        new LC(genesis, blocksPerEpoch);
+        LC lightClientContract = new LC();
+
+        // Encode the initializer function call
+        bytes memory data = abi.encodeWithSelector(LC.initialize.selector, genesis, blocksPerEpoch);
+
+        // TODO how to test this is really working?
+        // Proxy
+        ERC1967Proxy proxy = new ERC1967Proxy(address(lightClientContract), data);
 
         //// Legacy HotShot contract deployment
         new HotShot();


contracts/src/LightClient.sol
```diff
@@ -9,16 +16,24 @@ import { LightClientStateUpdateVK as VkLib } from "./libraries/LightClientStateU
...
     // === Constants ===
     //
     /// @notice System parameter: number of blocks per epoch
-    uint32 public immutable BLOCKS_PER_EPOCH;
+    /// @dev This variable cannot be made immutable due to how UUPS contracts work. See
+    /// https://forum.openzeppelin.com/t/upgradable-contracts-instantiating-an-immutable-value/28763/2#why-cant-i-use-immutable-variables-1
+    uint32 public blocksPerEpoch;
+
+    /// @notice A simple way to track contract versions
+    uint32 public immutable VERSION = 1;
```
alxiong, line 36: should we adopt major, minor, patch semver like [this one](https://github.com/ethereum-optimism/optimism/blob/v1.1.4/packages/contracts-bedrock/src/universal/Semver.sol)?

contracts/test/FeeContract.t.sol
```diff
@@ -5,7 +5,8 @@
 pragma solidity ^0.8.0;
 
 // Libraries
-import { Test, console2 } from "forge-std/Test.sol";
+import { Test } from "forge-std/Test.sol";
+//import { console2 } from "forge-std/Test.sol";
```
alxiong, line 9:
```suggestion
```

contracts/script/UpgradeLightClient.s.sol
```diff
@@ -0,0 +1,38 @@
...
+
+import { LightClientV2 as LCV2 } from "../src/LightClientV2.sol";
+import { LightClient as LC } from "../src/LightClient.sol";
+
+contract UpgradeLightClientScript is Script {
+    /// @notice runs the upgrade
+    /// @param mostRecentlyDeployedProxy address of deployed proxy
+    /// @return address of the proxy
+    /// TODO get the most recent deployment from the devops tooling
+
```
alxiong, line 14:
```suggestion
```

contracts/script/UpgradeLightClient.s.sol
```diff
@@ -0,0 +1,38 @@
...
+    function run(address admin, address mostRecentlyDeployedProxy) external returns (address) {
+        LCV2 newLC = new LCV2();
+        address proxy = upgradeLightClient(admin, mostRecentlyDeployedProxy, address(newLC));
```
alxiong, line 17: why separate out as a function, it'll just be extra `JUMP` without much readability benefit here?

contracts/test/LightClient.t.sol
```diff
@@ -15,12 +15,17 @@ import { BN254 } from "bn254/BN254.sol";
...
+    LCTest public lc;
     uint32 public constant BLOCKS_PER_EPOCH_TEST = 3;
     LC.LightClientState public genesis;
     // this constant should be consistent with `hotshot_contract::light_client.rs`
     uint64 internal constant STAKE_TABLE_CAPACITY = 10;
 
+    function initLC(LC.LightClientState memory _genesis, uint32 _blocksPerEpoch) public {
+        lc = new LCTest();
+        lc.initializeForTests(_genesis, _blocksPerEpoch);
+    }
```
alxiong, line 27: let's add explanation to this function.
saying that our LightClient(Test) is written with upgradable contracts in mind, thus doesn't use normal constructor, but `initalize()` instead.
here this function works like calling constructors without going through proxy deployment etc.

^^ pls reword this

contracts/test/mocks/LightClientTest.sol
```diff
@@ -9,9 +9,13 @@ import { LightClientStateUpdateVKTest as VkLib } from "./LightClientStateUpdateV
...
-    constructor(LC.LightClientState memory genesis, uint32 numBlockPerEpoch)
-        LC(genesis, numBlockPerEpoch)
-    { }
+    constructor() LC() { }
+
+    function initializeForTests(LC.LightClientState memory genesis, uint32 numBlockPerEpoch)
+        public
+    {
+        _initializeState(genesis, numBlockPerEpoch);
+    }
```
alxiong, line 18: why couldn't we call `_initializeState` inside constructor here?
I know that LightClient couldn't call constructor, but this contract can, right?

this is only used for non-upgradabed tests only anyway.

contracts/src/LightClientV2.sol
```diff
@@ -0,0 +1,254 @@
...
+    bytes32 public votingStakeTableCommitment;
+    /// @notice The quorum threshold for the stake table used in current voting
+    uint256 public votingThreshold;
+    /// @notice The commitment of the stake table frozen for change (i.e. snapshot at the start of
+    /// last epoch)
+    bytes32 public frozenStakeTableCommitment;
+    /// @notice The quorum threshold for the frozen stake table
+    uint256 public frozenThreshold;
+
+    // TODO adding this field to the LightClientState struct does not seem to work. Why? See
```
sveitser, line 56: @philippecamacho I think this makes sense because the members of the struct variable are tightly packed in storage so adding a new field to a struct it's similar to adding a new state variable, and is only safe at the end. because that memory is already used by the variable that follows the variable of type LightClientState.

philippecamacho: So in order to be able to add new field to the struct I guess we should add empty slots to this struct to be used in the future using the [gap](https://github.com/EspressoSystems/espresso-sequencer/pull/1079#discussion_r1489886269) approach.

Or we may put the struct into a map as contract updates work well in this [case](https://github.com/EspressoSystems/espresso-sequencer/blob/fb11109a5c14a48eba4aa2dd115865a55f43ad79/contracts/demo/upgradeDemo/DemoBoxV2.sol#L30).

contracts/src/LightClientV2.sol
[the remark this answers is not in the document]
philippecamacho: > LCV2.sol shouldn't be in src/ but in test/LightClientUpgrade.t.sol instead? as it is only for testing, not part of actual production code.

Right, the light client contract V2 has been moved to test directory in 02d2386.

> (nit, right now it's called LightClientContractUpgrade.t.sol which I think .sol already indicates it's a contract?)

I don't follow. `LightClientContractUpgrade.t.sol` is a test contract to check that the light client contract update works as expected.

contracts/script/DeployLightContract.s.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+pragma experimental ABIEncoderV2;
```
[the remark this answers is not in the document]
philippecamacho: Yes, done in d186df5.

contracts/src/LightClientV2.sol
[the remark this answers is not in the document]
philippecamacho: > > (nit, right now it's called LightClientContractUpgrade.t.sol which I think .sol already indicates it's a contract?)
> 
> I don't follow. `LightClientContractUpgrade.t.sol` is a test contract to check that the light client contract update works as expected.

Oh I see, you mean [this](464a8ad)?

contracts/script/UpgradeLightClient.s.sol
philippecamacho: Agreed. Change done in 6c8e4cd.

contracts/test/mocks/LightClientTest.sol
philippecamacho: Good catch. Change done in 135b6a1.

contracts/src/LightClientV2.sol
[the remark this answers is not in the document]
alxiong: Yup, sry the unclear comment. Very nit, thanks!

contracts/test/LightClient.t.sol
philippecamacho: `initalizeForTests` has been remove in 6c8e4cdeeafe5ac87bb4032176453996a2c12279  as you suggested.

contracts/src/LightClient.sol
philippecamacho: Good idea. Done in aac21637b0aeb05bc3c0f33fbe375206e0043551. Tell me if it is what you had in mind and I will apply it to the Fee contract. Note that using a Semver contract does not quite work because it messes up with the order of the variables and also one cannot use the constructor.

contracts/src/LightClientV2.sol
philippecamacho: Comment updated in f8c63d42830c9f24a11fd81a1ccf8d38a2db15af.

contracts/src/LightClient.sol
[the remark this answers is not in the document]
philippecamacho: If the `LightClientState` changes, this will likely mean that the statement to be validated by the plonk verifier will change as well. In practice the prover will remain the same (the list of public inputs might increase or decrease in size but its type will remain `uint256[]`). What will change though is the plonk proving (and verifying) key.
So in summary no code changes  for the plonk prover will be required but the proving/verifying keys will need to be updated. This in particular will also affect the Plonk verifier library as those keys are hard-coded in the contract code. The sequencer code though might require some minor change to convert the LightClientState into the list of public inputs.

contracts/test/LightClientUpgrade.t.sol
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    // that the data remains the same after upgrading the implementation
+    function testUpgradeSameData() public {
+        // Upgrade LightClient and check that the genesis state is not changed and that the new
+        // field
+        // of the upgraded contract is set to 0
+        lcV2Proxy = LCV2(upgrader.run(admin, proxy));
+
+        assertEq(lcV2Proxy.newField(), 0);
```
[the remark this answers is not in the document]
philippecamacho: Yes, however I tried that and it did not work. I think the reason is that genesisState and finalizedState are at the beginning of storage. So it seems we have too options in order to be able to extend LightClientState struct with new fields:

* Add new (and upgraded) `LightClientState` fields to the contract  and provide some conversion method for reading old states.
* Put the `LightClientState` variable (`genesis` and `finalizedState`) into a map (looks weird but seems to be the cleanest/most flexible approach).

WDYT?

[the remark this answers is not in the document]
alysiahuggins: Ok that makes sense - the mapping approach can look weird but sounds cleaner for sure. From reading, using a storage gap would lead to higher gas costs too so I think the mapping approach may be best based on what we know right now.

contracts/src/LightClient.sol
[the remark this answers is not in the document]
philippecamacho: Done in 0118f89e7f7f9768cd68450ceb4b7c943e76ffe7

contracts/test/LightClientUpgrade.t.sol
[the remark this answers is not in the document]
philippecamacho: Done in 0118f89e7f7f9768cd68450ceb4b7c943e76ffe7.

contracts/src/LightClientV2.sol
philippecamacho: Finally, aftet discussing with @alysiahuggins we decided to store the light client state struct into a map. See https://github.com/EspressoSystems/espresso-sequencer/pull/1079#discussion_r1494738481

contracts/test/LightClientUpgrade.t.sol
```diff
@@ -0,0 +1,92 @@
...
+        assertEq(lcV2Proxy.blocksPerEpoch(), 10);
+        assertEq(lcV2Proxy.currentEpoch(), 0);
+
```
alysiahuggins, line 55: I think it would be worthwhile to assert the extra field added to the LightClientState struct is also equal to what we expect (0) - `assertEq(lcV2Proxy.getFinalizedState().extraField, 0);`

alysiahuggins (approved):

contracts/test/LightClientUpgrade.t.sol
philippecamacho: As discussed, this equality is already checked in https://github.com/EspressoSystems/espresso-sequencer/blob/96a03a0ba34fbaf080bc3020ab21a4b8258ac0d0/contracts/test/LightClientUpgrade.t.sol#L65

Issue #1047, opened 2024-01-26
Proper `VidScheme` constructor shared with hotshot


ggutoski:
- Hotshot needs to initialize a `VidScheme` because hotshot is responsible for VID dispersal.
- The sequencer (this repo) needs it too because we're responsible for namespace/transaction proofs.
- Construction of this `VidScheme` should be done in one place and shared between hotshot and this repo. (Analogy: the `BlockPayload` trait is defined in hotshot and impl'ed in this repo.)
- Wherever it's constructed, it should be as opaque as possible. Example: a tiny "VID factory" module that exposes a function whose return type is `impl VidScheme`. (Actually `impl PayloadProver` but forget that.) This tiny module is the only place that "knows" the underlying impl for `VidScheme`; everything else talks to VID only via the `VidScheme` trait.

# Who hosts the factory?

VID scheme construction requires `num_storage_nodes`, `erasure_code_rate`, and `kzg_params`. Both hotshot and sequencer have (or can have) this data. Should the `VidScheme` be constructed in this repo or in hotshot?

## Reasons to do it here in the sequencer

- The sequencer is the application, hotshot is the library. As such, the params `num_storage_nodes`, etc are naturally set in the sequencer, not hotshot. (eg. source-of-truth for `num_storage_nodes` is the stake table, which is hosted in the sequencer.)

## Reasons to do it in hotshot

- Currently the sequencer needs `VidScheme` *only* for namespace/transaction proofs. It's conceivable that in the future a `VidScheme` instance is not needed to compute these proofs. By contrast, VID is core to hotshot and will always be there. If for some reason in the future the sequencer has no need for `VidScheme` at all then it's a hotshot-specific implementation detail that is now exposed to the sequencer, breaking modularity.

# Appendix: re-think VID scheme construction in jellyfish?

In order to make the ns/tx proofs we need, under the hood the ADVZ VID scheme in jellyfish needs:
1. `payload_chunk_size` (ie. polynomial degree), and
2. KZG params.

Is there a more ergonomic design for `PayloadProver` API in jellyfish? (eg. assoc function?) Or are we stuck constructing a fully-functional `VidScheme`?

jbearer:
Gonna push back slightly:

> the params num_storage_nodes, etc are naturally set in the sequencer, not hotshot.

[...]

PR #1087, opened 2024-02-13
feat: switch namespace from NMT to VID

### LINKED ISSUE #1046: [cappuccino milestone] VID namespace proofs in live production

ggutoski:
close #1046 

- minimal changes required to switch NMT to VID
- still lots of VID tech debt to repay; please don't judge
- removed as much NMT stuff as I could.


nomaxg:
When I run a sequencer network locally, I am unable to fetch a namespace proof. The server crashes with this error:
```
The application panicked (crashed).
Message:  called `Option::unwrap()` on a `None` value
Location: sequencer/src/api/endpoints.rs:84
```

Here is a [branch](https://github.com/EspressoSystems/espresso-sequencer/tree/namespace-query-test) with a failing test case. To run the test:
```
cargo test --package sequencer --lib -- api::fs::generic_tests::test_namespace_query --nocapture
```

data/header.json
```diff
@@ -8,8 +8,9 @@
         "hash": "0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
     },
     "payload_commitment": "HASH~1yS-KEtL3oDZDBJdsW51Pd7zywIiHesBZsTbpOzrxOfu",
-    "transactions_root": {
-        "root": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+    "ns_table": {
+        "raw_payload": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],
+        "phantom": null
```
jbearer, line 13: Sucks that this gets serialized automatically. Can we use `#[serde(skip)]` on the field to elide it so it's not an annoyance for other languages?

data/ns_table.json
```diff
@@ -0,0 +1,4 @@
+{
+    "raw_payload": [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],
+    "phantom": null
```
jbearer, line 3: Also remove this if you manage to remove the other one

sequencer/src/api/endpoints.rs
```diff
@@ -21,7 +22,7 @@ use tide_disco::{
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct NamespaceProofQueryData {
-    pub proof: NamespaceProofType,
+    pub proof: crate::block::payload::NamespaceProof<Advz<Bls12_381, sha2::Sha256>>,
```
jbearer, line 25: We should import [`VidScheme`](https://hotshot.docs.espressosys.com/hotshot_types/data/type.VidScheme.html) from HotShot rather than defining `Advz` ourselves, to ensure we are using the same one.

Actually, `NamespaceProof` may not need to be generic on VID scheme at all, since we are always going to want to use the one HotShot is using. If it is generic, though, can we make a type alias that plugs in the generic parameter to make it friendlier for downstream stuff?

sequencer/src/api/endpoints.rs
```diff
@@ -65,9 +66,35 @@ where
                 resource: height.to_string(),
             })?;
 
-            let proof = block.payload().get_namespace_proof(namespace.into());
+            // TODO fake VidScheme construction
+            let vid = crate::block::payload::test_vid_factory();
+            use hotshot::traits::BlockPayload;
+            let disperse_data = vid
+                .disperse(block.payload().encode().unwrap().collect::<Vec<u8>>())
```
jbearer, line 73: Can you create an issue so we don't forget: we should replace this with fetching the VID common data for this block (once the query service supports it) and using that to construct the VID scheme (once HotShot supports it), thus avoiding the need to disperse here

sequencer/src/bin/count-transactions.rs
```diff
@@ -79,8 +79,8 @@ async fn main() {
                         }
                     }
                 };
-                for i in 0..block.len() {
-                    let txn = block.transaction(&(i as u64)).unwrap();
+                for t in block.payload().iter(block.payload().get_ns_table()) {
+                    let txn = block.transaction(&t).unwrap();
```
jbearer, line 83: Shorthand for this: `for (t, txn) in block.enumerate() {`

sequencer/src/bin/orchestrator.rs
```diff
@@ -195,7 +195,7 @@ async fn main() {
     config.config.known_nodes_with_stake = known_nodes_with_stake;
     config.config.max_transactions = args
         .max_transactions
-        .unwrap_or(NonZeroUsize::new(2_usize.pow(MAX_NMT_DEPTH as u32)).unwrap());
+        .unwrap_or(NonZeroUsize::new(1000).unwrap()); // TODO what to set here?
```
jbearer, line 198: Something arbitrary is fine. This is going away (getting moved into the buidler) soon anyways, and what we will want to enforce instead is a maximum _size_.

Also, now that we're just using a constant instead of an expression, we can set this using `default_value` on the clap field and not bother with the `unwrap_or`

sequencer/src/block.rs
```diff
@@ -1,144 +1,87 @@
...
+use snafu::OptionExt;
+
+pub mod entry;
+pub mod payload;
+pub mod queryable;
+pub mod tables;
+pub mod tx_iterator;
+
+use crate::block::entry::TxTableEntryWord;
+use crate::block::tables::NameSpaceTable;
```
jbearer, line 16: Nit: can just `use entry::...` or `use tables::....` since the modules are defined here

sequencer/src/block.rs
```diff
@@ -1,144 +1,87 @@
...
+use payload::Payload;
+
+impl BlockPayload for Payload<TxTableEntryWord> {
```
jbearer, line 19: Can we make a type alias for this with the canonical choice of generic parameter?

sequencer/src/header.rs
```diff
@@ -251,14 +248,15 @@ impl BlockHeader for Header {
         )
     }
 
+    // TODO return metadata must be cloned from returned Payload; don't return metadata??
```
jbearer, line 251: Yeah, we could remove this and have HotShot get the metadata by calling `payload.metadata()`. But I feel like it's going to end up cloning the metadata anyways. We could make metadata an `Arc<NamespaceTable>` if we're really concerned. But we're also assuming metadata is small

sequencer/src/api/endpoints.rs
ggutoski: can we punt this to #1047 ?

sequencer/src/api/endpoints.rs
ggutoski: #1093

sequencer/src/block.rs
ggutoski: I blame rust-analyzer for that. 😁

sequencer/src/block.rs
ggutoski: We could, but I have plans to remove this generic param altogether as part of tech debt repayment. Do you mind if we leave it as-is for now as a poignant reminder of the debt that is owed?

sequencer/src/bin/count-transactions.rs
ggutoski: That's much cleaner but still not squeaky clean 3347b8aef8b2a98121d8bc847c000bef82e87fd8.  I had to replace `.hash()` with `.commit()` and it looks like
```rust
                for (_, txn) in block.enumerate() {
                    total += 1;
                    unique.insert(txn.commit());
                }
```

sequencer/src/api/endpoints.rs
ggutoski: nvm fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1087/commits/baa6c6935bb72c24249c8af25785aac026846b16

nomaxg (approved):
LGTM

ggutoski:
@nomaxg I'm getting nondeterministic test failure on 
```
cargo test --package sequencer --lib -- api::fs::generic_tests::test_namespace_query --nocapture
```

The failure occurs here https://github.com/EspressoSystems/espresso-sequencer/blob/fcb936e2ad25f79cf85adf92e4159f95101dbf00/sequencer/src/api.rs#L531-L535

Until now I've observed it only locally, typically after I've modified source files to trigger a rebuild. But now for the first time I saw it in CI: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/7934874194/job/21666779930#step:8:248002

This test passes for me locally (most of the time). Perhaps it's a timing issue?

Issue #934, opened 2024-01-08
Sequencer key management


nomaxg:
Derive sequencer keys from file instead of generating them from seed, integrating @dailinsubjam 's work.

dailinsubjam:
This needs to wait for the PR to be closed: https://github.com/EspressoSystems/HotShot/pull/2247, which requires re-review from @rob-maron  & @shenkeyao. It would be great if you also take a look at the PR's "Comments requested" (especially the first two).

dailinsubjam:
After discussion in today's meeting, we also need to get this done before sequencer integration: https://github.com/EspressoSystems/HotShot/issues/2416.

nomaxg:
> After discussion in today's meeting, we also need to get this done before sequencer integration: [EspressoSystems/HotShot#2416](https://github.com/EspressoSystems/HotShot/issues/2416).

thank you @dailinsubjam !

mrain:
After this issue, what's the ideal way to initialize a stake table as a non-participant of consensus?
Our light client prover service needs it.

dailinsubjam:
> After this issue, what's the ideal way to get a stake table as a non-participant of consensus? Our light client prover service needs it.

By `getting a stake table`, do you mean light client prover need an endpoint to get the stake table? In Hotshot, stake value is now saved in `HotShotConfig::known_nodes_with_stake` which I guess you already learnt. From Consensus side, the stake value is still set to 1 as default. In short term it could also be imported from the config file [hotshot/crates/orchestrator/run-config.toml](https://github.com/EspressoSystems/HotShot/blob/main/crates/orchestrator/run-config.toml) right after `node_index` and `seed`. 
Would you guys love to set up a meeting to talk about this if there might be too many details included?

sveitser:
@imabdulbasit I think there's nothing left to do here so closing it. Please re-open if necessary.

PR #1091, opened 2024-02-14
Get sequencer public keys from HotShot Orchestrator config

imabdulbasit:
This pull request addresses [issue #934](https://github.com/EspressoSystems/espresso-sequencer/issues/934).

The hotshot orchestrator was updated (https://github.com/EspressoSystems/HotShot/issues/2416) to allow the nodes to post their public keys rather than the orchestrator generating the public keys. 

This PR 
- Updates hotshot crates to tag 0.5.13
- uses the `post_and_wait_all_public_keys` function within the Orchestrator client to fetch the updated config and retrieve the public keys directly from the config. This eliminates the need for generating public keys from the seed.

key file to review: sequencer/src/lib.rs


sequencer/src/lib.rs
```diff
@@ -286,38 +290,46 @@ pub async fn init_node(
...
         None => {
             tracing::info!("loading network config from orchestrator");
-            let config = orchestrator_client.get_config(public_ip.to_string()).await;
+            let mut config: NetworkConfig<VerKey, StaticElectionConfig> =
+                orchestrator_client.get_config(public_ip.to_string()).await;
+
+            config = orchestrator_client
+                .post_and_wait_all_public_keys(
+                    config.node_index,
+                    config.config.my_own_validator_config.public_key,
```
jbearer, line 306: Here, we're still using the key that the orchestrator originally generated for us, which is known to all nodes and not secure. What we've changed is, using this key is now voluntary.

We should add a `private_staking_key` option to the command line/environment variables defined in `sequencer/options.rs`, and use that key instead. That way, each node operator can generate and use their own private key, and they never have to reveal it to other nodes.

sequencer/src/lib.rs
```diff
@@ -286,38 +290,46 @@ pub async fn init_node(
...
             persistence.save_config(&config).await?;
             (config, true)
         }
     };
     let node_index = config.node_index;
 
-    // Generate public keys and this node's private keys.
+    // Generate node's private keys.
     //
     // These are deterministic keys suitable *only* for testing and demo purposes.
```
jbearer, line 318: We can change this comment now. We're no longer generating the keys. And they are no longer unsafe (once we fix the above)

sequencer/src/lib.rs
```diff
@@ -286,38 +290,46 @@ pub async fn init_node(
...
+    let pub_keys = known_nodes_with_stake
+        .iter()
+        .map(|entry| entry.stake_key)
         .collect::<Vec<_>>();
-    let known_nodes_with_stake: Vec<<PubKey as SignatureKey>::StakeTableEntry> = (0..num_nodes)
-        .map(|id| pub_keys[id].get_stake_table_entry(1u64))
-        .collect();
     let state_ver_keys = (0..num_nodes)
         .map(|i| StateKeyPair::generate_from_seed_indexed(config.seed, i as u64).ver_key())
         .collect::<Vec<_>>();
```
jbearer, line 331: We now have the same problem with the state signing schnorr keys, where they have to be generated deterministically so that we know all the public keys. So we may have to do something similar using the orchestrator to get these. @dailinsubjam @bfish713 maybe we can add `state_pub_key` to `StakeTableEntry`, and have nodes post both their public staking key and public schnorr key?

sequencer/src/options.rs
```diff
@@ -80,6 +81,9 @@ pub struct Options {
     )]
     pub webserver_poll_interval: Duration,
 
+    // Sequencer Private Staking key
+    #[clap(short, long, env = "ESPRESSO_SEQUENCER_PRIVATE_STAKING_KEY")]
+    pub private_staking_key: BLSPrivKey,
```
sveitser, line 86: I think we need to add some code to generate such keys for the docker compose demo, or have a "demo mode" where the keys are fetched from the orchestrator.

imabdulbasit: I can make it optional and use the generated key from the orchestrator config if the arg is not provided

jbearer: I think we should add a tool like `bin/keygen` to generate keys. Then we can use that to create some random keys that we hardcode into the docker-compose file. I would feel slightly better about doing the demo that way and not having the insecure "use orchestrator keys" as a code path in the main sequencer node at all

sequencer/src/bin/keygen.rs
```diff
@@ -0,0 +1,54 @@
...
+
+const DEFAULT_SEED : &str = "[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]";
+
+/// Utility program to generate keypairs
+#[derive(Clone, Debug, Parser)]
+struct Options {
+    /// Seed for generating keypair.
+    /// Default is all zeros.
+    #[clap(long, short = 's',  default_value = DEFAULT_SEED, value_parser = parse_seed)]
+    seed: [u8; 32],
```
jbearer, line 18: IMO the default should be generate a seed from entry (ie if you use the program with no arguments it generates a cryptographically random key).

You can accomplish this using `let mut rng = rand_chacha::ChaChaRng::from_entropy()`, and then `rng.fill_bytes` to populate a 32-byte array with random data for passing to `generated_from_seed_indexed`

jbearer: Oh and if we generate a random seed it would be nice to print it out along with the private key

jbearer (commented):
Looks good. We should get back on `main` of `hotshot-query-service` before merging this. To do that, I'm going to push a commit to the query service branch that updates to HotShot 0.5.14, which should fix the CI there. Then you'll need to update this branch to 0.5.14 (shouldn't be any breaking changes. @imabdulbasit

jbearer:
Ah, there is one breaking change in HotShot, but it shouldn't be too bad to fix. `CommChannel` has been removed so where we create comm channels by wrapping networks when we're initializing HotShot, you should be able to just use the underlying network directly instead, and delete the comm channel stuff.

See [this commit](https://github.com/EspressoSystems/hotshot-query-service/pull/417/commits/3544d0217d963348d502a6a2d9432772640adbc3) doing the equivalent thing in the query service

jbearer (approved):
Looks good. Before merging, can you approve https://github.com/EspressoSystems/hotshot-query-service/pull/417 and then point the hotshot-query-service dependency here back at main?

sveitser:
@imabdulbasit I think this currently breaks the nix dev env because the rust toolchain is too old. 

We can merge https://github.com/EspressoSystems/espresso-sequencer/pull/905 first once that CI passes to update the nix rust toolchain to 1.76.

sequencer/src/lib.rs
```diff
@@ -286,53 +293,64 @@ pub async fn init_node(
...
         None => {
             tracing::info!("loading network config from orchestrator");
-            let config = orchestrator_client.get_config(public_ip.to_string()).await;
+            let mut config: NetworkConfig<VerKey, StaticElectionConfig> =
+                orchestrator_client.get_config(public_ip.to_string()).await;
+
+            // Get updated config from orchestrator containing all peer's public keys
+            config = orchestrator_client
+                .post_and_wait_all_public_keys(config.node_index, public_staking_key)
+                .await;
```
dailinsubjam, line 312: Where is this `public_staking_key` assigned?  I think here it should be `config = orchestrator_client
                .post_and_wait_all_public_keys(config.node_index, config.config.my_own_validator_config.public_key)`.

sequencer/src/lib.rs
```diff
@@ -286,53 +293,64 @@ pub async fn init_node(
...
+
+            config.config.my_own_validator_config.private_key = private_staking_key.clone();
+            config.config.my_own_validator_config.public_key = public_staking_key;
+
```
dailinsubjam, line 316: And also here, I don't think `config.config.my_own_validator_config` need to be assigned again if we're getting config from orchestrator.

sequencer/src/lib.rs
imabdulbasit: @dailinsubjam

sequencer/src/lib.rs
sveitser: The two are the same.

https://github.com/EspressoSystems/espresso-sequencer/blob/78dfc4cd5f170626f51e9ce18e63e257eb86492f/sequencer/src/lib.rs#L242-L261

sequencer/src/lib.rs
sveitser: Since they are the same this shouldn't break anything though.

jbearer: At least the private key does need to be assigned right? The orchestrator wouldn't be able to set that itself

Issue #924, opened 2024-01-04
[light-client-integration] add block tree commitment in light client state


mrain:
After #925

Issue #929, opened 2024-01-05
[light-client-integration] add fee ledger commitment in the light client state.


mrain:
as title says.

PR #1094, opened 2024-02-15
feat(light-client): complete light client state

mrain:
closes: #924 , #929 

This PR adds fee ledger commitment and block tree commitment to the light client state.

### DIFF
--- a/sequencer/src/state_signature.rs
+++ b/sequencer/src/state_signature.rs
@@ -2,6 +2,8 @@
 
 use crate::context::SequencerContext;
 use crate::{network, Leaf, SeqTypes};
+use ark_ff::PrimeField;
+use ark_serialize::CanonicalSerialize;
 use futures::stream::{Stream, StreamExt};
 use hotshot::types::{Event, SignatureKey};
 use hotshot_stake_table::vec_based::StakeTable;
@@ -12,6 +14,8 @@ use hotshot_types::signature_key::BLSPubKey;
 use hotshot_types::traits::node_implementation::ConsensusTime;
 use hotshot_types::traits::signature_key::StakeTableEntryType;
 use hotshot_types::traits::stake_table::{SnapshotVersion, StakeTableScheme as _};
+use jf_primitives::crhf::{VariableLengthRescueCRHF, CRHF};
+use jf_primitives::errors::PrimitivesError;
 use std::collections::{HashMap, VecDeque};
 
 /// A relay server that's collecting and serving the light client state signatures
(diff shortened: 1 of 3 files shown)


jbearer (approved):

sequencer/src/state_signature.rs
```diff
@@ -38,47 +42,73 @@ pub(super) async fn state_signature_loop<N>(
...
     stake_table_comm: &StakeTableCommitmentType,
-) -> LightClientState {
-    // TODO(Chengyu): fill these `default()` with actual value
-    LightClientState {
+) -> Result<LightClientState, PrimitivesError> {
+    let header = leaf.get_block_header();
+    let mut block_comm_root_bytes = vec![];
+    header
+        .block_merkle_tree_root
+        .serialize_compressed(&mut block_comm_root_bytes)?;
```
alxiong, line 99: why is the block header being the block commitment root?
we are supposed to build a merkle tree out of all the block commitment, right?

sequencer/src/state_signature.rs
```diff
@@ -38,47 +42,73 @@ pub(super) async fn state_signature_loop<N>(
...
 
+fn hash_bytes_to_field(bytes: &[u8]) -> Result<CircuitField, PrimitivesError> {
+    // make sure that `mod_order` won't happen.
+    let bytes_len = ((<CircuitField as PrimeField>::MODULUS_BIT_SIZE + 7) / 8 - 1) as usize;
+    let elem = bytes
+        .chunks(bytes_len)
+        .map(CircuitField::from_le_bytes_mod_order)
+        .collect::<Vec<_>>();
+    Ok(VariableLengthRescueCRHF::<_, 1>::evaluate(elem)?[0])
+}
```
alxiong, line 89: are there reasons to use Rescue Hash here? why can't we just hash down to 32 bytes, and then use `jf_utils::bytes_to_field_elements`?

sequencer/src/state_signature.rs
mrain: Yes and this is the Merkle root. Just naming

sequencer/src/state_signature.rs
mrain: No specific reason, We could use others as well

alxiong: yeah, let's try to minimize rescue usage wheneve possible, it's slow and not as time-tested. 
also try to reuse jf_utils, cuz converting bytes to fields is a common util. 
would you agree?

sequencer/src/state_signature.rs
alxiong: i think the BlockHeader is [more than just block_commit_root](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/header.rs#L79), right? even though I don't see a problem, but it's not technically this bytes is committing to an superset of system states than just block_comm_root.

PR #1099, opened 2024-02-16
bugfix: use correct stake table commitments for SNARK verification

### LINKED ISSUE #1090: Add missing checks on stake table commitments in public inputs
As reported by @philippecamacho 

> It seems to me that some verification is missing in the function LightClient.newFinalizedState.
>
> Currently this [function](https://github.com/EspressoSystems/espresso-sequencer/blob/58852cbd989606f103a41664ef4c728859de007d/contracts/src/LightClient.sol#L113) checks that:
>
> Some conditions are satisfied regarding the viewNumber and blockNumber.
> The public inputs are valid field elements. The proof is valid. However there is no linking between the commitment of the stake table part of the public input and some value we can trust. In practice an attacker can create a proof with a new set of signers (all under his control) that will pass all the checks of LightClient.newFinalizedState .
> 
> I think we need to ensure that the stake table commitments part of the public inputs are equal to the stake table commitments of the previous epoch in the general case (post Decaf).
> 
> For Decaf it is easy, the stake table commitments can be hard coded.

This ticket aims to fix this by adding the checks on the public inputs against stored stake table commitments

alxiong:
Closes #1090 

Fix the error I introduce, and add tests to make sure attacks won't succeed.


philippecamacho:
Looks good. I thought initially that we needed to check that the epoch  of `newState` and  `finalizedState` differ by one. That is 
`newState.epoch =  finalizedState.epoch+1`. But an adversary trying to skip an epoch when updating the light client state would not be able to produce the signatures on the new state unless there is an overlap that is big enough between the signers in epoch `e` and epoch `e+2`. 
That said for mental sanity we might want to enforce having epoch number differ by one in the `newFinalizedState` function.
WDYT @alxiong?

alxiong:
but there's no way you can "skip an epoch", because 

```solidity
        uint64 epochEndingBlockHeight = currentEpoch * BLOCKS_PER_EPOCH;

        bool isNewEpoch = finalizedState.blockHeight == epochEndingBlockHeight;
        if (!isNewEpoch && newState.blockHeight > epochEndingBlockHeight) {
            revert MissingLastBlockForCurrentEpoch(epochEndingBlockHeight);
        }
```
enforces that we update at least once every epoch, specifically the last block in each epoch.

is that sufficient? do we need the extra differ by 1 epoch check?

contracts/src/LightClient.sol
```diff
@@ -161,9 +166,9 @@ contract LightClient {
         publicInput[2] = uint256(state.blockHeight);
         publicInput[3] = BN254.ScalarField.unwrap(state.blockCommRoot);
         publicInput[4] = BN254.ScalarField.unwrap(state.feeLedgerComm);
-        publicInput[5] = BN254.ScalarField.unwrap(state.stakeTableBlsKeyComm);
-        publicInput[6] = BN254.ScalarField.unwrap(state.stakeTableSchnorrKeyComm);
-        publicInput[7] = BN254.ScalarField.unwrap(state.stakeTableAmountComm);
+        publicInput[5] = BN254.ScalarField.unwrap(finalizedState.stakeTableBlsKeyComm);
```
philippecamacho, line 169: Are we assuming there is exactly one state update per epoch?

alxiong: we assume at least one, not exactly one

philippecamacho:
> but there's no way you can "skip an epoch", because
> 
> ```solidity
>         uint64 epochEndingBlockHeight = currentEpoch * BLOCKS_PER_EPOCH;
> 
>         bool isNewEpoch = finalizedState.blockHeight == epochEndingBlockHeight;
>         if (!isNewEpoch && newState.blockHeight > epochEndingBlockHeight) {
>             revert MissingLastBlockForCurrentEpoch(epochEndingBlockHeight);
>         }
> ```
> 
> enforces that we update at least once every epoch, specifically the last block in each epoch.
> 
> is that sufficient? do we need the extra differ by 1 epoch check?

You are right, no need to add an extra check.

philippecamacho (approved):
LGTM

PR #1107, opened 2024-02-19
Fix a bug in orchestrated startup and add more CI tests

jbearer:
* Use the node index returned by `get_config` rather than `post_and_wait_all_public_keys`, where the node ID is not set
* Add a new script for quickly checking the basic functionality of the demo (docker- or process-compose)
* Enable the node API module, which we should have done already and which the script depends on
* Create new CI jobs to smoketest the process compose and docker compose demos, since it is currently really easy to make changes that affeect these demos without realizing it


scripts/smoke-test-demo
```diff
@@ -0,0 +1,38 @@
...
+sleep 10
+
+new_block_height=`curl -s $SERVER/node/block-height`
+new_num_tx=`curl -s $SERVER/node/transactions/count`
+
+echo "Final state:"
+echo "  block_height: $new_block_height"
+echo "  transactions: $new_num_tx"
+
+if ! [[ $new_block_height > $block_height ]]; then
```
sveitser, line 31:
```suggestion
if ! [[ $new_block_height -gt $block_height ]]; then
```

scripts/smoke-test-demo
```diff
@@ -0,0 +1,38 @@
...
+	echo "Block height is not increasing!"
+	exit 1
+fi
+if ! [[ $new_num_tx > $num_tx ]]; then
```
sveitser, line 35:
```suggestion
if ! [[ $new_num_tx -gt $num_tx ]]; then
```

scripts/smoke-test-demo
jbearer: what's the issue with this? with `[[ ]]` it seems to work using `>`

sveitser: I does string comparison https://stackoverflow.com/questions/57137789/bash-if-condition-gt-vs-when-using

jbearer: omg

sveitser (approved):

PR #1110, opened 2024-02-20
Make block.rs public for arbitrum integration

nomaxg:
This exposes `NamespaceProof` and other related types


sequencer/src/block/tables.rs
```diff
@@ -121,6 +121,10 @@ impl<TableWord: TableWordTraits> NameSpaceTable<TableWord> {
         std::cmp::min(left, right)
     }
 
+    pub fn is_empty(&self) -> bool {
```
nomaxg, line 124: Fixes a clippy error

ImJeremyHe (approved):

PR #1111, opened 2024-02-20
Pin jellyfish to specific commit.

tbro:
Reference jellyfish to a specific revision to avoid implicit updates. Please note that even though we are pinning to the same revision declared in the `Cargo.lock` of the previous commit, simply adding this to `Cargo.toml` causes compiler failures (presumably b/c cargo is resolving dependencies differently than we are). If we can fix these then we can have a jellyfish version declared in `Cargo.toml` and avoid a lot of head aches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2257,10 +2257,10 @@ dependencies = [
  "hotshot-state-prover",
  "hotshot-types",
  "itertools 0.12.1",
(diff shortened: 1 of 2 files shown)


tbro:
Build errors appear to be related to `hotshot-state-prover` so maybe @alxiong or @mrain have some insight?

Issue #966, opened 2024-01-10
Set up "instance state"


jbearer:
Each HotShot instance running a sequencer node should have some immutable global state used in header creation and state validation: 

- [x]  Eth L1 client (no more process-global environment variables!). (should be done with #1138 )
- [x] Fetcher for missing validated state. In theory this can just be a list of URLs of peers to query missing state from. However, it might be good to make this generic behind a trait, so that we can do something simpler for tests (should be complete in #1151)

PR #1138, opened 2024-02-22
Add l1_client to instance state

### LINKED ISSUE #1115: simplify L1 client
As part of this change and the catchup changes, we are going to have to commit to `Header::new` being fallible or divergent. Right now HotShot will not behave well if `Header::new` doesn't return, but this shouldn't happen in normal usage, and the bad case will be fixed by the end of this sprint.

  One nice consequence of this is we can do away with the complexity of having a separate background task for reading from L1. The point of this was to move the fallible, blocking L1 RPC calls to their own task, so that `Header::new` would never block or fail. But now `Header::new` is allowed to block, so we can move the async retry loops in the L1 client directly into `Header::new`, and ditch the L1 client background task. This means `L1Client::snapshot` will directly call RPC methods to get the latest and finalized L1 head (retrying when necessary) and we can remove the L1 client background task

tbro:
Move l1 client  from global var to `NodeState`. The decision structure around optional configuration has been maintained in the new format while we determine how much of is needed going forward. L1 is now queried directly from `Header::new` in place of the previous long running background task. Configuration options have been moved to CLI.

  * Take options from cli
  * Rework `L1ClientOptions` to deal with optional values.
  * remove the lazy static configuration
  * remove l1_client `update_loop`

Addresses #966, Closes #1115


sequencer/src/options.rs
```diff
@@ -125,6 +125,20 @@ pub struct Options {
...
+    /// feature.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_L1_WS_PROVIDER")]
+    pub l1_provider_url: Option<Url>,
+
+    /// For testing with a pre-merge geth node that does not support
+    /// the finalized block tag we allow setting an environment
+    /// variable to use the latest block instead. This feature is used
+    /// in the OP devnet which uses the docker images built in this
+    /// repo. Therefore it's not hidden behind the testing flag.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_L1_USE_LATEST_BLOCK_TAG")]
```
tbro, line 140: Due to updates in more recent geth versions we may no longer need this variable.

sequencer/src/l1_client.rs
```diff
@@ -87,17 +82,30 @@ impl Committable for L1BlockInfo {
...
+#[derive(Clone, Debug, Default)]
 pub struct L1ClientOptions {
-    url: Url,
+    url: Option<Url>,
     finalized_block_tag: BlockNumber,
     retry_delay: Duration,
 }
-
 impl L1ClientOptions {
+    pub fn new(url: Option<Url>) -> Self {
```
tbro, line 92: Trying to shoe horn new things into `L1ClientOptions` may have over complicated it a bit. I plan on revisiting this along with `L1Client` instantiation.

sequencer/src/header.rs
```diff
@@ -233,6 +234,31 @@ impl Header {
     }
 }
 
+// Function that can be called from `Header::new` for interactiong with
+// l1. When `Header::new` becomes async this logic can be moved there.
+async fn async_header_new(instance_state: &NodeState) -> anyhow::Result<L1Snapshot> {
```
jbearer, line 239: I don't know if this function should return a result. I think everything that fails should be retried indefinitely, until HotShot cancels the task because the view has timed out. Basically failing to propose when it's our slot is the worst thing that can happen here so there's no point in failing early if we still have time to retry.

jbearer (commented):
Boy it feels good to get this cleaned up finally

sequencer/src/header.rs
tbro: Ok. recent commits changed this to retry indefinably.

sequencer/src/l1_client.rs
```diff
@@ -88,167 +85,56 @@ impl Committable for L1BlockInfo {
...
-            snapshot.head = head;
-            if let Some(finalized) = finalized {
-                snapshot.finalized = finalized;
+    /// Proxy to `Provider.get_block_number`.
+    async fn get_block_number(&self) -> u64 {
+        loop {
+            if let Ok(n) = self.agent().get_block_number().await {
+                return n.as_u64();
+            } else {
+                sleep(self.retry_delay).await;
```
jbearer, line 118: We should log when errors happen here

tbro: fixed in dd5425e8e9a1cfb897a0fe2b374dacec06ef7498

sequencer/src/l1_client.rs
```diff
@@ -88,167 +85,57 @@ impl Committable for L1BlockInfo {
...
+                }
             }
         }
-
-        tracing::error!("L1 block stream ended unexpectedly, trying to re-establish");
     }
 }
 
-async fn get_finalized_block<P: JsonRpcClient>(
+pub async fn get_finalized_block<P: JsonRpcClient>(
```
jbearer, line 135: Why does this need to be `pub`? Seems like anyone outside this module should just be calling `get_snapshot` on the client

sequencer/src/l1_client.rs
```diff
@@ -276,140 +163,70 @@ async fn get_finalized_block<P: JsonRpcClient>(
...
-                if expected_finalized_number == snapshot_finalized.number {
-                    assert_eq!(expected_finalized.hash.unwrap(), snapshot_finalized.hash);
-                    assert_eq!(expected_finalized.timestamp, snapshot_finalized.timestamp);
-                    tracing::info!("L1 finalized block fully synchronized");
-                }
-            } else {
-                tracing::info!("no finalized L1 block yet");
-                assert_eq!(snapshot.finalized, None);
-            }
+        assert_eq!(7, expected_head);
```
jbearer, line 190: Where does 7 come from?

sequencer/src/l1_client.rs
```diff
@@ -276,140 +163,70 @@ async fn get_finalized_block<P: JsonRpcClient>(
...
-            } else {
-                assert_eq!(prev_finalized, None);
-            }
+        let finalized = get_finalized_block(&provider).await.unwrap().unwrap();
 
-            prev_head = snapshot.head;
-            prev_finalized = snapshot.finalized;
+        // Though we are asserting a values, we do so to prove
+        // structural correctness.
+        assert_eq!(0, finalized.number);
```
jbearer, line 196: Don't understand the comment. Also not sure about hard-coding 0 here. Isn't it based on timing?

sequencer/src/l1_client.rs
```diff
@@ -276,140 +163,70 @@ async fn get_finalized_block<P: JsonRpcClient>(
...
-            .unwrap();
-        let anvil_client = Provider::try_from(anvil.url().to_string()).unwrap();
-
-        let mut prev_head = 0;
-        let mut prev_finalized: Option<L1BlockInfo> = None;
 
-        let start = Instant::now();
-        while start.elapsed() < test_duration {
-            let snapshot = l1_client.snapshot().await;
+        let l1_client = L1Client::new(anvil.url());
```
jbearer, line 184: We don't actually use `l1_client` in the rest of the test, so this seems to be testing the ethers-rs provider rather than testing our `L1Client`

sequencer/src/l1_client.rs
```diff
@@ -88,167 +85,57 @@ impl Committable for L1BlockInfo {
...
-    }
-}
-
-#[derive(Clone, Debug)]
-pub struct L1Client {
-    latest_snapshot: Arc<RwLock<L1Snapshot>>,
-}
-
-impl L1Client {
+    /// Get a snapshot from the l1, or a fake snapshot if "testing"
```
sveitser, line 103: Is it still true that we provide a fake snapshot if the "testing" feature is enabled?

sequencer/src/l1_client.rs
tbro: Not sure. I just observed that it always appeared to be be 7.

sequencer/src/l1_client.rs
tbro: In ef28a828d766970c14c8d2b42b1b2afbadd4347e I've removed all the timing based assertions which also removes this comment.

sequencer/src/l1_client.rs
tbro: ef28a828d766970c14c8d2b42b1b2afbadd4347e does a better job of testing `L1Client`.

sequencer/src/l1_client.rs
tbro: It is not true. Fixed in d7963d1e42805a3f5257fba6515457409ccfa8d4

sequencer/src/l1_client.rs
tbro: this was also removed in ef28a828d766970c14c8d2b42b1b2afbadd4347e

sequencer/src/l1_client.rs
tbro: `pub` removed in ef28a828d766970c14c8d2b42b1b2afbadd4347e

sequencer/src/l1_client.rs
```diff
@@ -88,167 +85,62 @@ impl Committable for L1BlockInfo {
...
+        loop {
+            match get_finalized_block(&self.provider).await {
+                Ok(block) => return block,
+                Err(e) => {
+                    tracing::warn!("Finalized block error: {}", e);
+                    sleep(self.retry_delay).await;
+                }
             }
+            assert!(count < 6);
+            count += 1;
```
jbearer, line 135: ?

sequencer/src/l1_client.rs
```diff
@@ -276,140 +168,50 @@ async fn get_finalized_block<P: JsonRpcClient>(
...
-            .spawn()
-            .await;
-        let l1_client = L1ClientOptions::with_url(anvil.ws_url())
-            .start()
-            .await
-            .unwrap();
+        // compare response of underlying provider w/ `get_block_number`
+        let expected_head = provider.get_block_number().await.unwrap().as_u64();
+        let head = l1_client.get_block_number().await;
+        assert_eq!(expected_head, head);
```
jbearer, line 199: Technically these can return different numbers if the L1 block changes in between the two calls. Maybe this is so unlikely that we don't care. Not sure.

sequencer/src/l1_client.rs
tbro: Its a retry limit so tests fail instead of retrying indefinably. It will be removed once I figure out what is causing tests to fail. Strangely, causing these to panic (instead of retrying indefinably) causes tests to pass. I can see the panics in the test output, but they are being ignored for some reason and since the tests no longer fail on timeout, they are marked as successful.

sequencer/src/l1_client.rs
tbro: Yes I thought of that. Its the kind of thing that may cause occasional failures requiring us to rerun the tests to get them to pass. The window is 1 sec wide, so I don't *think* it will happen very frequently. If these methods had fail cases we could just `unwrap` and remove the assert statements. As they are, I'm not thinking of a better way to test them.

sequencer/src/l1_client.rs
tbro: removed in edb82e76b40af46867714115d16373285d074c8b

hotshot-state-prover/src/service.rs
```diff
@@ -556,7 +560,11 @@ mod test {
 
         let (genesis, _qc_keys, state_keys, st) = init_ledger_for_test();
 
-        let anvil = Anvil::new().spawn();
+        let anvil = {
+            let port = portpicker::pick_unused_port().expect("No ports free");
```
sveitser, line 564: Ethers should already handle this https://docs.rs/ethers-core/2.0.13/src/ethers_core/utils/anvil.rs.html#222

sequencer/src/l1_client.rs
```diff
@@ -276,140 +162,54 @@ async fn get_finalized_block<P: JsonRpcClient>(
...
-        // continue to make progress.
-        let snapshot = l1_client.snapshot().await.head;
-        sleep(Duration::from_secs(5)).await;
+        // If we drop `anvil` the same request will fail and we will
+        // get an error as below.
+        drop(anvil);
+        let error = provider.client_version().await.unwrap_err();
+        assert!(error
+            .to_string()
+            .contains("Connection reset by peer (os error 104)"));
```
jbearer, line 211: This error message varies across OSes, and is also subject to change in future versions of the anvil client. IMO what we're testing here is not important enough to justify such a brittle check. The `unwrap_err` above on its own should be sufficient

tbro: fixed in 447115188afc4855b2a1f6e56535af1675dc9356

hotshot-state-prover/src/service.rs
tbro: removed all those in: https://github.com/EspressoSystems/espresso-sequencer/commit/447115188afc4855b2a1f6e56535af1675dc9356

jbearer (approved):
LGTM

process-compose.yaml
```diff
@@ -108,6 +108,7 @@ processes:
       - ESPRESSO_SEQUENCER_ETH_MNEMONIC
       - ESPRESSO_SEQUENCER_ETH_ACCOUNT_INDEX
       - ESPRESSO_SEQUENCER_PREFUNDED_BUILDER_ACCOUNTS
+      - ESPRESSO_SEQUENCER_L1_PROVIDER
```
sveitser, line 111: Should add this to `docker-compose.yaml` as well.

hotshot-state-prover/src/service.rs
```diff
@@ -241,10 +241,10 @@ pub async fn sync_state(
     }
 
     // TODO this assert fails. See https://github.com/EspressoSystems/espresso-sequencer/issues/1161
```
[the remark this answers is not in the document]
tbro: reapplied in 26f6621740d9a05298c01d2f67c3850a41c6b73c

docker-compose.yaml
```diff
@@ -188,7 +190,7 @@ services:
       - ESPRESSO_SEQUENCER_DA_SERVER_URL
       - ESPRESSO_SEQUENCER_CONSENSUS_SERVER_URL
       - ESPRESSO_SEQUENCER_API_PORT
-      - ESPRESSO_SEQUENCER_L1_WS_PROVIDER
+      - ESPRESSO_SEQUENCER_L1_PROVIDER
```
sveitser, line 193: Good find, there are probably more unused env vars: https://github.com/EspressoSystems/espresso-sequencer/issues/1181

sveitser (approved):

PR #1106, opened 2024-02-19
demo: fix prover

sveitser:
Currently deployment is broken and errors are hidden due to piping. Deployment seems to be working now but there's probably still something wrong.

    Light client contract address: 0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512
    Stake table initialized.
    Loading SRS from Aztec's ceremony...
    Done in 120.909ms
    Generating proving key and verification key.
    Done in 8.972s
    2024-02-19T12:47:25.758573Z  INFO hotshot_state_prover::service: Start syncing light client state.
    2024-02-19T12:47:25.758586Z  INFO hotshot_state_prover::service: Fetching the latest state signatures bundle from relay server.
    2024-02-19T12:47:25.759691Z ERROR hotshot_state_prover::service: Cannot sync the light client state: Error when communicating with the state relay server

PR #1139, opened 2024-02-22
T1092 script to deploy light client contract

### LINKED ISSUE #1092: Single way to initialize and deploy the light client contract
Currently we are using different ways for deploying the light client contract:
 * Generate the code of the light client contract with a rust binary in order to set the genesis state
 
 Use the second approach and parametrize the deployment script with the genesis state so that different test can initialize and deploy different instances of the light client contract.
 

Context: see zulip [thread](https://github.com/EspressoSystems/espresso-sequencer/pull/1041#discussion_r1487616058).

- [x] Call `diff-test mock-genesis` from the solidity deployment script in order to initialize the contract with the genesis state.
- [x] Also remove the comments to the tests added in fc9cdcbc95a99ffb72821ce85990c2999dcacc7b.
- [x] Update `launch-prover-service` as now `contracts/broadcast/LighClientDemo.s.sol `does not exist.

philippecamacho:
Closes #1092 

* Now there is one script for deploying the light client contract in "production" and one for deploying the test variant of this contract. Both contracts are kind of different as the one for testing purposes uses a different Plonk verifier key.
* Those scripts are parametrized with the number of blocks per epoch and the initial number of validators in order to avoid relying on hardcoded values in the forge script.
* The hotshot altruistic prover tests now run and rely on the `mock_ledger` crate.
* The rust binary `gen-demo-contract` has been removed.


sveitser:
Need to remove `gen-demo-contract` from `scripts/build-docker-images`, `scripts/build-docker-images-native`, `docker/prover-service.Dockerfile` and `scripts/launch-prover-service`.

philippecamacho:
@sveitser @mrain I am not sure the native demo works. The hotshot state prover looks to start fine but then there is an error communicating with the relay server.

```
> just demo-native
prover-service
   ERROR hotshot_state_prover::service: Cannot sync the light client state: Error when communicating with the state relay server: Error 404: The light client state signatures are not ready.
```

hotshot-state-prover/src/mock_ledger.rs
alxiong, line 1: I find it quite unintuitive that a `MockLedger` which I wrote only for diff testing purposes, is used anywhere else. 

It has lots of quirks that are specific to the limited requirements or workflow of the forge-based diff tests, it's not a general test stub/mock.

in general, I don't recommend moving it here.

justfile
```diff
@@ -76,8 +77,9 @@ sol-test:
...
     forge test
 
-# Develop contracts to local blockchain for development and testing
-dev-deploy url="http://localhost:8545" mnemonics="test test test test test test test test test test test junk":
+# Deploy contracts to local blockchain for development and testing
+dev-deploy url="http://localhost:8545" mnemonics="test test test test test test test test test test test junk" num_blocks_per_epoch="10" num_init_validators="5":
     forge build
-    MNEMONICS="{{mnemonics}}" forge script contracts/script/LightClientTest.s.sol \
+    MNEMONICS="{{mnemonics}}" forge script 'contracts/test/LightClientTest.s.sol' \
+    --sig "run(uint32 numBlocksPerEpoch, uint32 numInitValidators)" {{num_blocks_per_epoch}} {{num_init_validators}} \
```
alxiong, line 84: this mixed-use of justfile variable and forge script config is quite nice!

contracts/script/LightClient.s.sol
```diff
@@ -1,48 +1,51 @@
...
-            threshold
+        (admin,) = deriveRememberKey(seedPhrase, 0);
+        vm.startBroadcast(admin);
+
+        // TODO for a production deployment provide the right genesis state and value
+
+        string[] memory cmds = new string[](4);
+        cmds[0] = "diff-test";
+        cmds[1] = "mock-genesis";
+        cmds[2] = vm.toString(numBlocksPerEpoch);
```
alxiong, line 27: I think if we want to give an example, we should pass in these values?
ideally either via `.envrc` or inside `justfile`, or somehow read and parse from an artifact file?

the whole point of `gen-demo-contract.rs` was to populate these values/genesis artifacts directly embedding into the deploy script contract.

justfile
philippecamacho: Thanks!

hotshot-state-prover/src/mock_ledger.rs
philippecamacho: I moved the `MockLedger` struct into the `hotshot-state-prover` crate due to circular dependencies issues.
It is used  to have consistency between rust and solidity  values (e.g. commitments, zk proof).

contracts/script/LightClient.s.sol
philippecamacho: With the current `LightClient.s.sol` script we do not need to run `gen-demo-contract.rs` which simplifies the development workflow and is also more robust I think. Also this approach avoids having many scripts for deploying the light client contract which is the main purpose of this PR.

hotshot-state-prover/src/mock_ledger.rs
alxiong: right, but that struct is not meant to be used universally, and the circular dep is a result of violation of that assumption. 

why do we have to use that in the `hotshot-state-prover`? currently we are only reusing a tiny portion of it which is initializing the validator keypairs and stake table, correct?
that's minimum amount of code duplication, which is fine IMO. 

imo, it's an anti-pattern to move the whole `MockLedger` out of `diff-test` just to use those two test utils.

contracts/script/LightClient.s.sol
alxiong: wait, so in production, how do you imagine the workflow works? how do we compute the genesis block? where is that information being stored and how do we load that into the contract constructor call arguments?

philippecamacho: We need to decide how to handle the deployment of contracts to production environment. This is also related to how we will upgrade contracts.

The way I see it is that the same script should be used for dev/staging/production environment with some flags. Based on the flags the script will call the right rust executable that will generate the correct initialization of the contract.   I am open to other ideas.

cc @sveitser @alysiahuggins

hotshot-state-prover/src/mock_ledger.rs
philippecamacho: It seems to be that `MockLedger` is useful for writing tests beyond its original purpose. The code looks easier to read and in particular it becomes easy to ensure that the contract is initialized in a way that is compatible with the test. Having hardcoded values that need to be the same in different parts of the code is harder to maintain in my opinion.

contracts/script/LightClient.s.sol
alysiahuggins: Are we planning to deploy with a multi-sig wallet at this stage?

alxiong: > Based on the flags the script will call the rust executable that will generate the correct initialization of the contract

i see, so Chengyu's `gen-demo-contract.rs` assumes: "Rust produced init/genesis state/info --> encode into deployment script" workflow,

your code is "General deployment script --> call Rust code --> return init/genesis to deploy.sol --> encode info directly in solidity" workflow. 

I'm fine with this. Personally, I don't see clear advantage of either.

sveitser:
> @sveitser @mrain I am not sure the native demo works. The hotshot state prover looks to start fine but then there is an error communicating with the relay server.
> 
> ```
> > just demo-native
> prover-service
>    ERROR hotshot_state_prover::service: Cannot sync the light client state: Error when communicating with the state relay server: Error 404: The light client state signatures are not ready.
> ```

This is supposedly normal (see comment https://github.com/EspressoSystems/espresso-sequencer/pull/1106#issuecomment-1952822775 ). @alxiong @mrain is there a straightforward way to check if the prover service is working correctly?

sveitser:
@philippecamacho my prover service logs show this instead

image

mrain:
> > @sveitser @mrain I am not sure the native demo works. The hotshot state prover looks to start fine but then there is an error communicating with the relay server.
> > ```
> > > just demo-native
> > prover-service
> >    ERROR hotshot_state_prover::service: Cannot sync the light client state: Error when communicating with the state relay server: Error 404: The light client state signatures are not ready.
> > ```
> 
> This is supposedly normal (see comment [#1106 (comment)](https://github.com/EspressoSystems/espresso-sequencer/pull/1106#issuecomment-1952822775) ). @alxiong @mrain is there a straightforward way to check if the prover service is working correctly?

If this pops up, then some parts are working correctly. 
To test the later parts, proof generations and submitting to l1 contract, we need at lease one newly decided block to process. Maybe we could wait e.g. 5 mins to launch the prover service, and hopefully there're blocks being produced.

contracts/script/LightClient.s.sol
sveitser: I much prefer not having multiple scripts that do similar things and avoid code generation where feasible. I think if we want to decouple the creation of the genesis block and contract deployment we could save the genesis block to a JSON file, persist that JSON file and load it inside deployment scripts with https://book.getfoundry.sh/cheatcodes/parse-json. Not suggesting we do this in this PR though.

philippecamacho:
> @philippecamacho my prover service logs show this instead
> 
> image

I have the same error now. Digging into it.

contracts/script/LightClient.s.sol
alxiong: >I much prefer not having multiple scripts that do similar things and avoid code generation where feasible.

make sense. 

imo, the JSON file approach is better for production, because we will have some static artifacts to version control and refer to, which is more portable and language-agnostic
the direct calling using forge FFI is better for testing since there's no point saving some temporary json file just for tests. 

good!

alxiong:
> we need at lease one newly decided block to process. Maybe we could wait e.g. 5 mins to launch the prover service, and hopefully there're blocks being produced.

this is also my guess. since the correct contract instance did revert, which means we have configured the signer, providers properly. I think the chance of proof gen being wrong is low. it's most likely `revert OutdatedState();` in the `LightClient.sol` that's causing the error.

I think we can improve on logging by catching the [`ContractError`](https://docs.rs/ethers/latest/ethers/middleware/contract/enum.ContractError.html#variant.Revert) and logging/propagate back the revert message.
(^^ should we create another issue for this? I can help on this)

alxiong (approved):

hotshot-state-prover/src/service.rs
```diff
@@ -356,42 +366,34 @@ mod test {
...
-            amount_comm: field_to_u256(stake_table_comm.2),
-            threshold,
-        };
+        let pp = MockSystemParam::init(BLOCKS_PER_EPOCH);
+        let ledger = MockLedger::init(pp, NUM_INIT_VALIDATORS as usize);
+
+        let genesis = ledger.get_state();
+        let qc_keys = ledger.qc_keys;
+        let state_keys = ledger.state_keys;
+        let st = ledger.st;
```
sveitser, line 396: No need to change anything but I wonder why we often use names that are extremely short (`F`, `st`, `pp`, `pk`, `pi`). I understand that if you want to do many math operations the longer names make the expressions hard to read and long but I feel like in this repo we don't need to be so brief everywhere and it makes reading the code a bit more straightforward.

hotshot-state-prover/src/service.rs
```diff
@@ -478,11 +480,15 @@ mod test {
         let signer = Wallet::from(anvil.keys()[0].clone());
         let l1_wallet = Arc::new(L1Wallet::new(provider.clone(), signer));
 
-        Command::new("just")
+        let status = Command::new("just")
```
sveitser, line 483: Not an issue with this PR later but I'm not sure it's a great idea to call `just` from our rust tests. There doesn't seem to be much benefit from using `just` here it but we've created another dependency for a tool that isn't so commonplace and now needs to be installed to run `cargo test`. Any downside of using a bash script? What do others think about this?

hotshot-state-prover/src/service.rs
```diff
@@ -524,7 +530,6 @@ mod test {
 
     // This test is temporarily ignored. We are unifying the contract deployment in #1071.
```
sveitser, line 531:
```suggestion
```

scripts/launch-prover-service
```diff
@@ -4,20 +4,24 @@
...
     echo "Light client contract address not found, deploying it now."
-    gen-demo-contract --contracts-root-dir="./contracts"
     forge fmt
     forge build
-    forge script DeployLightClientDemoScript --fork-url $ESPRESSO_SEQUENCER_L1_PROVIDER --broadcast
-    ESPRESSO_SEQUENCER_LIGHTCLIENT_ADDRESS=$(cat contracts/broadcast/LightClientDemo.s.sol/*/run-latest.json | jq -r '.receipts[1].contractAddress')
+    forge script contracts/script/LightClient.s.sol:DeployLightClientContractScript  \
+     --sig "runDemo(uint32 numBlocksPerEpoch)" $NUM_BLOCKS_PER_EPOCH \
+            --fork-url $ESPRESSO_SEQUENCER_L1_PROVIDER --broadcast
+    ESPRESSO_SEQUENCER_LIGHTCLIENT_ADDRESS=$(cat contracts/broadcast/LightClient.s.sol/*/runDemo-latest.json | jq -r .receipts[1].contractAddress)
```
sveitser, line 17: We may want to get the chain ID from the L1 provider or specify it to make this a bit more robust. It would currently not work correctly if there are deployments for multiple chain IDs.

scripts/launch-prover-service
```diff
@@ -4,20 +4,24 @@
 # client contract will be deployed. This feature is intended for demo or testnet
 # purposes only.
 set -eo pipefail
+NUM_BLOCKS_PER_EPOCH=10
```
sveitser, line 7: Is there a better place to set the value of this variable? If it's interesting to change this I think there should be an env var and if not, shouldn't we hard-code it in the `runDemo` function?

sveitser (approved):
LGTM, some not so important comments. I think we should fix the demo though. @philippecamacho do you have any idea why the proof would fail to verify?

scripts/launch-prover-service
philippecamacho: Environment variable added in b182b5b.

scripts/launch-prover-service
philippecamacho: TODO added in f387a99.

contracts/broadcast/LightClient.s.sol/900/run-1709121894.json
```diff
@@ -0,0 +1,69 @@
+{
```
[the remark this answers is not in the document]
philippecamacho: Done in c92d618.

PR #1151, opened 2024-02-26
Sequencer catchup

sveitser:
- Add function to fetch forgotten leaves of the fee merkle tree.
- Add function to fetch merkle tree frontier (not hooked up yet, might change).

### TODO
- [x] Get a the right view number for querying catchup API.
- [x] Tests.
- [x] Many TODOs in code.


sequencer/src/catchup.rs
```diff
@@ -0,0 +1,78 @@
...
+                        Ok(_) => return,
+                        Err(err) => tracing::warn!("Error verifying account proof: {}", err),
+                    },
+                    Err(err) => {
+                        tracing::warn!("Error fetching account from peer: {}", err);
+                    }
+                }
+            }
+            tracing::warn!("Could not fetch account from any peer, retrying");
+            // TODO: backoff? for testing a function that doesn't loop forever is probably more convenient.
```
jbearer, line 50: IMO:
* Make backoff a parameter to set in the constructor. We can set it really short for tests that are expected to fail a few times before eventually succeeding (testing the backoff case)
* Make a separate function that returns an error, for easy unit testing. Then the main function just call that function in a loop.

sequencer/src/catchup.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    pub async fn fetch_blocks_frontier(
+        &self,
+        view: ViewNumber,
+        root: BlockMerkleCommitment,
```
jbearer, line 57: We probably should also take the tree here and `remember` instead of `verify`. Even though the frontier is kind of special, forgetting and remembering it works basically the same as any other path.

jbearer (commented):
Definitely on the right track. Let's ask the consensus team about where to get the view number

sequencer/src/state.rs
```diff
@@ -94,14 +98,17 @@ pub fn validate_proposal(
...
+        if matches!(lookup, LookupResult::NotInMemory) {
+            let view = todo!(); // where to get the view number?
+            block_on(peers.remember_account_balance(view, account, fee_merkle_tree));
+        }
+        fee_merkle_tree
+            .update_with(account, |balance| {
+                Some(balance.cloned().unwrap_or_default().add(amount))
+            })
+            .expect("update_with succeeds");
+        // TODO: can this really not fail?
```
tbro, line 111: I don't think so, b/c both `Some` or `None` cases are handled in the closure. I feel that this can be written more  succinctly, though, by *remembering* the account balance in the `Ok(NotInMemory)` case of the match on `update_with` (as you've done ~line 143).

sequencer/src/catchup.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    pub async fn fetch_accounts(
+        &self,
+        view: ViewNumber,
+        fee_merkle_tree_root: FeeMerkleCommitment,
+        accounts: impl IntoIterator<Item = FeeAccount>,
+    ) -> Vec<AccountQueryData> {
+        join_all(
```
jbearer, line 55: We have had issues in Gibraltar with the commit task firing a bunch of simultaneous requests to the query service, causing spikes in CPU usage on the commit task and the server. I don't expect we'll have too many fee accounts to fetch at one time here, but still maybe it is safer to fetch these one at a time or at least break them into fixed-size batches to fetch in parallel

sequencer/src/catchup.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    async fn fetch_account(
+        &self,
+        view: ViewNumber,
+        fee_merkle_tree_root: FeeMerkleCommitment,
+        account: FeeAccount,
+    ) -> AccountQueryData {
+        if self.clients.is_empty() {
+            panic!("No peers to fetch account from");
```
jbearer, line 70: I think it'll be easier to debug config mistakes if we panic in the constructor rather than here in what is probably a background task

sequencer/src/catchup.rs
jbearer: Actually maybe we should add a batch endpoint for fetching multiple paths at once.

sequencer/src/header.rs
```diff
@@ -244,21 +260,47 @@ impl BlockHeader<SeqTypes> for Header {
...
+            .peers
+            .fetch_accounts(
+                parent_leaf.get_view_number(),
+                parent_state.fee_merkle_tree.commitment(),
+                missing_accounts,
+            )
+            .await;
+
+        // Insert missing fee state entries
+        let mut state = parent_state.clone();
```
jbearer, line 288: Since we're cloning `state` here we could pass it into `from_info` by value and avoid a second clone (although `clone` should be fairly cheap now that stuff is `Arc`ed)

sequencer/src/catchup.rs
sveitser: Yes I will have a think about how to improve this. I think it's good to have way to run without peers. Maybe just making the state peers an `Option` is the simplest way to go.

sequencer/src/catchup.rs
sveitser: Ok. I think at this stage doing them sequentially is probably best, we can optimize later.

sequencer/src/catchup.rs
jbearer: The thing is if you run without peers and your node falls behind you're unable to vote ever again. I think we will just have to require at least one peer in order to run the node. And we can give our partners the URLs of our nodes to use for catchup.

I think ideally every node should have the URL of `f+1` other nodes, which guarantees (I think) that at least one of your peers will have the state for every view

sveitser: Yeah I agree it may be a good idea to enforce this on startup of a production node but we probably don't want to enforce it for every test deployment.

I think it would be good if

1. We can enforce running with likely correctly configured state peers.
1. We can run without real state peers for tests.
1. If we run tests we can either make it impossible accidentally to look up state from peers or panic if we do so. This avoids having many tests hang forever. Already happened to me :laughing: 

The simplest thing I can think of that achieves this is to

1. Enforce state peers aren't empty in regular constructor (as suggested).
2. Hide a test constructor behind `testing` feature flag to create empty state peer.
3. panic if we end up doing lookups with empty state peers

I do think this is kind of ugly though. Let me know if you have better idea.

If we go down the route of making a trait and a test + real implementation I think it's quite a lot of change required just to avoid a few panics in the code, so not sure it's worth it.

What I think I want to do is having a test implementation of `StatePeers` that doesn't use the network but takes as constructor input the merkle trees and then on lookup looks up value in it's own data structure and use that for tests. But again it seems a bit involved and not sure if it's worth doing right now.

jbearer: I am going to need something along those lines to get the restart test working, basically to represent an archival node without actually having state storage. I do think having a `StateCatchup` will be useful for these sorts of things. Is it maybe not so much of a change if we `Box<dyn StateCatchup>` instead of trying to make `InstanceState` generic?

sveitser:
@jbearer I tried to use box dyn trait in the last commit https://github.com/EspressoSystems/espresso-sequencer/pull/1151/commits/053d95fa1c304d0f3730125e5f9b766bda93f24a, this fails to compile with the error message below. I wonder if the actual problem is object safety due to using async in the trait and the error message is just wrong (https://github.com/rust-lang/rust/issues/119502).

I saw here [here](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#should-i-still-use-the-async_trait-macro) that using `async_trait` might help but the error message is the same.

```
    Checking sequencer v0.1.0 (/Users/lulu/r/EspressoSystems/espresso-sequencer/sequencer)
error[E0391]: cycle detected when checking effective visibilities
    |
note: ...which requires computing type of `state::<impl at sequencer/src/state.rs:254:1: 254:47>::validate_and_apply_header::{opaque#0}`...
   --> sequencer/src/state.rs:263:5
    |
263 | /     async fn validate_and_apply_header(
264 | |         &self,
265 | |         instance: &Self::Instance,
266 | |         parent_leaf: &Leaf,
267 | |         proposed_header: &Header,
268 | |     ) -> Result<Self, Self::Error> {
    | |__________________________________^
note: ...which requires computing type of opaque `state::<impl at sequencer/src/state.rs:254:1: 254:47>::validate_and_apply_header::{opaque#0}`...
   --> sequencer/src/state.rs:263:5
    |
263 | /     async fn validate_and_apply_header(
264 | |         &self,
265 | |         instance: &Self::Instance,
266 | |         parent_leaf: &Leaf,
267 | |         proposed_header: &Header,
268 | |     ) -> Result<Self, Self::Error> {
    | |__________________________________^
note: ...which requires type-checking `state::<impl at sequencer/src/state.rs:254:1: 254:47>::validate_and_apply_header`...
   --> sequencer/src/state.rs:263:5
    |
263 | /     async fn validate_and_apply_header(
264 | |         &self,
265 | |         instance: &Self::Instance,
266 | |         parent_leaf: &Leaf,
267 | |         proposed_header: &Header,
268 | |     ) -> Result<Self, Self::Error> {
    | |__________________________________^
    = note: ...which again requires checking effective visibilities, completing the cycle
note: cycle used when type-checking `header::<impl at sequencer/src/header.rs:255:1: 255:38>::new`
   --> sequencer/src/header.rs:256:5
    |
256 | /     async fn new(
257 | |         parent_state: &ValidatedState,
258 | |         instance_state: &NodeState,
259 | |         parent_leaf: &Leaf,
260 | |         payload_commitment: VidCommitment,
261 | |         metadata: <<SeqTypes as NodeType>::BlockPayload as BlockPayload>::Metadata,
262 | |     ) -> Self {
    | |_____________^
    = note: see https://rustc-dev-guide.rust-lang.org/overview.html#queries and https://rustc-dev-guide.rust-lang.org/query.html for more information

error[E0391]: cycle detected when computing type of opaque `header::<impl at sequencer/src/header.rs:255:1: 255:38>::new::{opaque#0}`
   --> sequencer/src/header.rs:256:5
    |
256 | /     async fn new(
257 | |         parent_state: &ValidatedState,
258 | |         instance_state: &NodeState,
259 | |         parent_leaf: &Leaf,
260 | |         payload_commitment: VidCommitment,
261 | |         metadata: <<SeqTypes as NodeType>::BlockPayload as BlockPayload>::Metadata,
262 | |     ) -> Self {
    | |_____________^
    |
note: ...which requires type-checking `header::<impl at sequencer/src/header.rs:255:1: 255:38>::new`...
   --> sequencer/src/header.rs:256:5
    |
256 | /     async fn new(
257 | |         parent_state: &ValidatedState,
258 | |         instance_state: &NodeState,
259 | |         parent_leaf: &Leaf,
260 | |         payload_commitment: VidCommitment,
261 | |         metadata: <<SeqTypes as NodeType>::BlockPayload as BlockPayload>::Metadata,
262 | |     ) -> Self {
    | |_____________^
    = note: ...which requires checking effective visibilities...
note: ...which requires computing type of `header::<impl at sequencer/src/header.rs:255:1: 255:38>::new::{opaque#0}`...
   --> sequencer/src/header.rs:256:5
    |
256 | /     async fn new(
257 | |         parent_state: &ValidatedState,
258 | |         instance_state: &NodeState,
259 | |         parent_leaf: &Leaf,
260 | |         payload_commitment: VidCommitment,
261 | |         metadata: <<SeqTypes as NodeType>::BlockPayload as BlockPayload>::Metadata,
262 | |     ) -> Self {
    | |_____________^
    = note: ...which again requires computing type of opaque `header::<impl at sequencer/src/header.rs:255:1: 255:38>::new::{opaque#0}`, completing the cycle
note: cycle used when comparing an impl and trait method signature, inferring any hidden `impl Trait` types in the process
   --> sequencer/src/header.rs:256:5
    |
256 | /     async fn new(
257 | |         parent_state: &ValidatedState,
258 | |         instance_state: &NodeState,
259 | |         parent_leaf: &Leaf,
260 | |         payload_commitment: VidCommitment,
261 | |         metadata: <<SeqTypes as NodeType>::BlockPayload as BlockPayload>::Metadata,
262 | |     ) -> Self {
    | |_____________^
    = note: see https://rustc-dev-guide.rust-lang.org/overview.html#queries and https://rustc-dev-guide.rust-lang.org/query.html for more information

For more information about this error, try `rustc --explain E0391`.
error: could not compile `sequencer` (lib) due to 2 previous errors
```

sequencer/src/state.rs
```diff
@@ -189,37 +262,68 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+        let missing_accounts = self.forgotten_accounts(
+            std::iter::once(proposed_header.fee_info.account)
+                .chain(l1_deposits.iter().map(|fee_info| fee_info.account)),
+        );
+
+        let view = parent_leaf.get_view_number();
+
         // Clone state to avoid mutation. Consumer can take update
         // through returned value.
         let mut validated_state = self.clone();
```
tbro, line 285: Nit: I like keeping this at the top just for emphasis. Its something that is obvious once you have familiarity w/ the code, but for the unfamiliar can clarifies somewhat how the caller will interact w/ our code.

sequencer/src/header.rs
```diff
@@ -244,21 +260,48 @@ impl BlockHeader<SeqTypes> for Header {
...
+            l1_snapshot.finalized,
+        );
+
+        // Find missing fee state entries
+        let missing_accounts = parent_state.forgotten_accounts(
+            std::iter::once(FeeAccount::from(instance_state.builder_address.address()))
+                .chain(l1_deposits.iter().map(|info| info.account())),
+        );
+
+        // Fetch missing fee state entries
```
jbearer, line 277: We also need to fetch the blocks frontier if missing

sequencer/src/state.rs
```diff
@@ -55,19 +58,69 @@ impl ValidatedState {
...
+    /// As an optimization we could try to apply updates and return the
+    /// forgotten accounts to be fetched from peers and update them later.
+    pub fn forgotten_accounts(
+        &self,
+        accounts: impl IntoIterator<Item = FeeAccount>,
+    ) -> Vec<FeeAccount> {
+        accounts
+            .into_iter()
+            .filter(|account| {
+                matches!(
```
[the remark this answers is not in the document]
sveitser: Done in 3a13ec71b52b989b63da27fbf80d9990be79b5f0

sequencer/src/catchup.rs
```diff
@@ -86,16 +86,17 @@ impl StateCatchup for MockStateCatchup {
...
+        tracing::info!("catchup: fetching frontier for view {view:?}");
+        let view = view.get_u64();
         let (_, proof) = self
             .state
             .block_merkle_tree
-            .lookup(view.get_u64())
+            .lookup(view)
             .expect_ok()
             .unwrap();
-        mt.remember(view.get_u64(), elem, proof.clone())
```
sveitser: This is currently where the tests panic. The `element` returned by lookup doesn't match the one passed in to the function.

sequencer/src/state.rs
```diff
@@ -338,6 +477,12 @@ impl FeeAmount {
     }
 }
 
+impl From<u64> for FeeAmount {
```
tbro, line 480: I think we had this before, but we removed it. Why are we putting it back again?

sveitser: I think it's currently only used here

```
error[E0277]: the trait bound `state::FeeAmount: From<{integer}>` is not satisfied
   --> sequencer/src/api.rs:920:56
    |
920 |         state.prefund_account(Default::default(), 1000.into());
    |                                                        ^^^^ the trait `From<{integer}>` is not implemented for `state::FeeAmount`
    |
    = help: the trait `From<ethers::types::U256>` is implemented for `state::FeeAmount`
    = help: for that trait implementation, expected `ethers::types::U256`, found `{integer}`
    = note: required for `{integer}` to implement `Into<state::FeeAmount>`
    ```

jbearer: I think it's just something that makes sense since `FeeAmount` is supposed to act like a number. E.g. `U256` also implements this

jbearer (approved):
@sveitser I did review your changes so if you've reviewed my changes I think we're good

sequencer/src/state.rs
tbro: Yea I don't think it hurts anything either but you can just do: `U256::from(1000).into()`.

tbro (approved):

PR #1152, opened 2024-02-27
refactor: block and namespace logic

akonring:
### tl;dr
- Removes the generic parameter `TableWord`.
- Removes the `TxTableEntry` and `TxTableEntryWord`.
- Introduces a `TableWord` struct using const generics with an implementation that mimics the "old" `TxTableEntry` but makes room for other implementations of `TableWord` (Offset, TableLen, NsId).

### Direction
This PR is a proposal for the direction of block/namespace logic:
- Table interaction should be through appropriate methods on `NamespaceTable` and `TxTable`.
Fx `Payload` should not deal with this. See comment:
https://github.com/EspressoSystems/espresso-sequencer/blob/6dfaebf134dbcb47a485bc937ee756ef69bde0a8/sequencer/src/block/payload.rs#L55-L58

- Different implementations of `TableWord` can support different words types (Offset, TableLen, NsId).    
Example Below:
```rust
pub trait Table {
    type TableLenWord: TableLenTraits;
    type NsIdWord: NsIdTraits;
    fn get_table_len(&self, offset: usize) -> TableLenWord;
    ...
} 


pub struct NameSpaceTable {
    pub(super) bytes: Vec<u8>,
}
impl Table for NameSpaceTable {
    // the word of table length is 4 bytes and is represented as u32
    type TableLenWord = TableWord<u32, 4>;
    type NsIdWord = TableWord<u64, 8>;
    ...
```


sequencer/src/block/table_word.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self(u32::from_le_bytes(bytes))
+    }
+    pub const fn byte_len() -> usize {
+        size_of::<u32>()
+    }
+
+    pub fn from_usize(val: usize) -> Self {
+        Self(
+            val.try_into()
+                .expect("usize -> TxTableEntry should succeed"),
```
philippecamacho, line 40: Replace TxTableEntry with Word?

sequencer/src/block/table_word.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self(1)
+    }
+    pub const fn to_bytes(&self) -> [u8; size_of::<u32>()] {
+        self.0.to_le_bytes()
+    }
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        Some(Self(u32::from_le_bytes(bytes.try_into().ok()?)))
+    }
+    /// Infallible constructor.
+    pub fn from_bytes_array(bytes: [u8; Self::byte_len()]) -> Self {
```
philippecamacho, line 30: Would it be equivalent to replace Self::byte_len() with SIZE?

philippecamacho:
This this PR looks like a step in the right direction. If we can achieve to implement the example shown [above](https://github.com/EspressoSystems/espresso-sequencer/pull/1152#issue-2156281100) (see "Different implementations of TableWord can support different words types (Offset, TableLen, NsId).
Example Below:") it feels to me that we would have achieved our goal of abstracting the complexity of dealing with table words.

ggutoski:
Just had a call with Anders to discuss plans.

PR #1153, opened 2024-02-27
Update hotshot

### LINKED ISSUE #1093: FTLOG don't VID `disperse` in the namespace proof endpoint
We should replace [VID `disperse`] with fetching the VID common data for this block (once the query service supports it) and using that to construct the VID scheme (once HotShot supports it), thus avoiding the need to disperse here

_Originally posted by @jbearer in https://github.com/EspressoSystems/espresso-sequencer/pull/1087#discussion_r1489576297_

jbearer:
* Fixes the state key management, using the new orchestrator feature to distribute both staking keys and state keys
* Fixes the issue with VID construction in namespace proofs by fetching the common data for the desired block and using it to construct the VID scheme

Closes #1093


nomaxg:
oh nice - so no need for https://github.com/EspressoSystems/espresso-sequencer/pull/1154?

nomaxg (approved):
LGTM. I'd recommend that we modify the VID namespace test to actually verify the proof [like this](https://github.com/EspressoSystems/espresso-sequencer/pull/1154/commits/6225de2a68a748eaade2570867d6237dc03fd290#diff-9c5c25f61c89b1c60988253125d7d03fb13becb07c9242f5966669ecfa00d373).

jbearer:
Right, this would supercede #1154 (sorry for the wasted work, I didn't know we would have this ready so soon). I will incorporate the test changes

sequencer/src/block/queryable.rs
```diff
@@ -74,7 +74,10 @@ impl QueryablePayload for Payload<TxTableEntryWord> {
             return None; // error: index out of bounds
         }
 
-        let vid = test_vid_factory(); // TODO temporary VID construction
+        // TODO temporary VID construction. We need to get the number of storage nodes from the VID
+        // common data. May need the query service to pass common into this function along with
+        // metadata.
```
ggutoski, line 79: Ugh. Shall we add a field `num_storage_nodes` to `Self::Metadata`?

sequencer/src/lib.rs
```diff
@@ -203,16 +239,23 @@ pub enum Error {
...
 async fn init_hotshot<N: network::Type>(
-    nodes_pub_keys: Vec<PubKey>,
-    known_nodes_with_stake: Vec<<PubKey as SignatureKey>::StakeTableEntry>,
     node_id: usize,
     priv_key: PrivKey,
     networks: Networks<SeqTypes, Node<N>>,
     config: HotShotConfig<PubKey, ElectionConfig>,
     metrics: &dyn Metrics,
-    instance_state: &NodeState,
+    instance_state: NodeState,
```
ggutoski, line 247: I trust there's a good reason to take ownership here.

ggutoski (approved):
VID construction LGTM.

sequencer/src/block/queryable.rs
jbearer: I don't know if it really makes sense to add it to metadata. Currently metadata is a byproduct of constructing the block, nothing to do with VID. If anything I think we could add it to the block header, currently `Header::new` takes `vid_commitment`. We could potentially add `num_storage_nodes` there. I also don't think it's unreasonable to have the query service pass in the common data when this function is called.

For now, though, I think we leave it as is until transaction proofs becomes a more pressing need.

jbearer: https://github.com/EspressoSystems/espresso-sequencer/issues/1155

sequencer/src/lib.rs
jbearer: Yeah, the type of `HotShotInitializer::from_genesis` changed so it in turn now takes ownership

PR #1169, opened 2024-02-28
chore: update hotshot 0.5.19, hotshot-query-service 0.0.10

### LINKED ISSUE #1047: Proper `VidScheme` constructor shared with hotshot

ggutoski:
close #1047 

- update hotshot to 0.5.19, use new `vid_scheme` constructor
- update hotshot-query-service to 0.0.10, use new header genesis API

# TODO

- [x] fix header genesis API usage

### DIFF
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -1,19 +1,16 @@
 use crate::block::entry::{TxTableEntry, TxTableEntryWord};
 use crate::block::payload;
 use crate::{BlockBuildingSnafu, Error, Transaction, VmId};
-use ark_bls12_381::Bls12_381;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use commit::Committable;
 use derivative::Derivative;
 use hotshot::traits::BlockPayload;
-use hotshot_types::data::VidScheme;
-use jf_primitives::{
-    pcs::{checked_fft_size, prelude::UnivariateKzgPCS, PolynomialCommitmentScheme},
-    vid::{
-        advz::payload_prover::{LargeRangeProof, SmallRangeProof},
-        payload_prover::{PayloadProver, Statement},
-        VidScheme as VidSchemeTrait,
-    },
+use hotshot_types::vid::{
+    LargeRangeProofType, SmallRangeProofType, VidCommitment, VidCommon, VidSchemeType,
+};
+use jf_primitives::vid::{
+    payload_prover::{PayloadProver, Statement},
+    VidScheme,
 };
 use num_traits::PrimInt;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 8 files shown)


sequencer/src/block/payload.rs
```diff
@@ -110,10 +107,10 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
         &self,
         ns_table: &NameSpaceTable<TxTableEntryWord>,
         ns_id: VmId,
-        vid: &VidScheme,
-        vid_common: <VidScheme as VidSchemeTrait>::Common,
+        vid: &VidSchemeType,
```
jbearer, line 110: We could consider removing this as an argument since it can be constructed from common

ggutoski: True, but VID scheme construction currently has nontrivial costs. It would be nice to avoid unnecessary VID scheme construction. OTOH if the caller doesn't already have a VID scheme then we're forcing her to make one. Ok I'll remove it.

sequencer/src/header.rs
```diff
@@ -272,14 +272,11 @@ impl BlockHeader for Header {
...
-    ) -> (
-        Self,
-        Self::Payload,
-        <Self::Payload as BlockPayload>::Metadata,
-    ) {
+        _payload_commitment: VidCommitment,
+        _metadata: <Self::Payload as BlockPayload>::Metadata,
+    ) -> Self {
         let (payload, ns_table) = Self::Payload::genesis();
         let payload_commitment = vid_commitment(&payload.encode().unwrap().collect(), 1);
```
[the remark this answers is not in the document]
ggutoski: done in https://github.com/EspressoSystems/espresso-sequencer/pull/1169/commits/231cf26113c0d1140454833eb2912c02e38976ed

sequencer/src/header.rs
```diff
@@ -303,7 +300,7 @@ impl BlockHeader for Header {
             fee_info: FeeInfo::new(instance_state.builder_address.address().into()),
```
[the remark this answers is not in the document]
ggutoski: done in https://github.com/EspressoSystems/espresso-sequencer/pull/1169/commits/231cf26113c0d1140454833eb2912c02e38976ed

sequencer/src/block/payload.rs
ggutoski: done in https://github.com/EspressoSystems/espresso-sequencer/pull/1169/commits/dac7e66b0b39ba1f35864b599d4deb868940f277

jbearer (approved):
LGTM

PR #1174, opened 2024-02-29
Data Pruning for SQL storage

imabdulbasit:
args for pruner config

Pruning is disabled by default

### DIFF
--- a/sequencer/src/persistence/sql.rs
+++ b/sequencer/src/persistence/sql.rs
@@ -1,13 +1,19 @@
-use super::{NetworkConfig, PersistenceOptions, SequencerPersistence};
-use crate::{Leaf, ValidatedState, ViewNumber};
+use std::time::Duration;
+
 use anyhow::bail;
 use async_trait::async_trait;
 use clap::Parser;
 use hotshot_query_service::data_source::{
-    storage::sql::{include_migrations, postgres::types::ToSql, Config, Query, SqlStorage},
+    storage::{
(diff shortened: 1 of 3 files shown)


sequencer/src/persistence/sql.rs
```diff
@@ -47,6 +53,20 @@ pub struct Options {
...
     #[clap(long, env = "ESPRESSO_SEQUENCER_POSTGRES_USE_TLS")]
     pub use_tls: bool,
+
+    /// This will enable the pruner and set the default pruning parameters unless provided.
+    /// Default parameters:
+    /// - pruning_threshold: 100GB
+    /// - minimum_retention: 1 day
+    /// - target_retention: 30 days
+    /// - batch_size: 1000
+    /// - max_usage: 80%
```
jbearer, line 63: Our default parameters should match the [recommended hardware](https://docs.espressosys.com/sequencer/guides/running-a-sequencer-node#hardware-requirements), based on [this policy](https://github.com/EspressoSystems/hotshot-query-service/issues/376#issuecomment-1917640481):
* pruning threshold: 3 TB
* minimum retention: 1 day
* target retention: 7 days

jbearer: I don't have a strong opinion on interval. We can run it for a bit in staging and see what works. But 1 hour seems like a reasonable starting place.

jbearer (approved):

Issue #1161, opened 2024-02-28
Hotshot state prover does not initialize BLS public keys like HotShot nodes


philippecamacho:
Currently the Hotshot state prover initialize the BLS public keys via a unique seed. However now HotShot nodes generated their own BLS key pairs.

The consequence is that now Hotshot nodes and the Hotshot state prover do not have the same stake table which result in having the snark proof be rejected by the light client contract.

**Solution:**
~~Use the orchestrator endpoint [config_after_peer_collected](https://github.com/EspressoSystems/HotShot/blob/main/crates/orchestrator/api.toml#L51) to get a HotShot config that includes the BLS keys of all nodes. We can also use [peer_pub_ready](https://github.com/EspressoSystems/HotShot/blob/main/crates/orchestrator/api.toml#L45) to wait until all BLS keys have been registered~~

Better create an environment variable with the concatenation of all BLS public keys so that they can be loaded directly from the HotShot state prover.

* [ ] Remove TODO in `hotshot_state_prover::service::sync_state`.

PR #1183, opened 2024-03-01
fix: make prover-service demo work again

mrain:
closes: #1161 

all clear

~~Stucked: `gen-demo-genesis` needs `SchnorrPrivKey` but it's in sequencer crate which hotshot-state-prover crate depends on.~~
~~Solution: remove `SchnorrPrivKey`, just use hotshot/jf types~~


sveitser:
@mrain Are you blocked on this PR?

mrain:
> @mrain Are you blocked on this PR?

Yes. Just pushed my commits. There're some new HotShot changes about `BlockPayload` that I don't know how to resolve

hotshot-state-prover/src/service.rs
```diff
@@ -58,40 +62,83 @@ pub struct StateProverConfig {
...
+                tracing::info!("Peers' keys are not ready.");
+            }
+            Err(e) => {
+                num_retries -= 1;
+                tracing::warn!("Orchestrator error {e}, retrying.");
+            }
+        }
+        sleep(Duration::from_secs(2)).await;
+    }
+    panic!("Failed initializing stake table.")
```
nomaxg, line 141: I think that this can retry indefinitely, don't see why we need a retry limit here

nomaxg (approved):
LGTM, just one comment

hotshot-state-prover/src/service.rs
mrain: changed in [43e59c7](https://github.com/EspressoSystems/espresso-sequencer/pull/1183/commits/43e59c71050e5be3330952d4b273289699775292)

Issue #1119, opened 2024-02-20
add L1 client interface to get pending deposits


jbearer:
Now that we can call async functions on the L1 client during `Header::new`, we can add a new function:
  ```rust
  async fn get_pending_deposits(prev_finalized: Option<u64>, new_finalized: u64) -> Vec<FeeInfo>;
  ```
  This function should be called from `Header:new`/`Header::from_info` whenever the new finalized block is greater than the previous finalized block, and returns a list of fee infos for each [`Deposit`](https://github.com/EspressoSystems/espresso-sequencer/blob/main/contracts/src/FeeContract.sol#L15C5-L15C57) event after the previous finalized block and up to (including) the new finalized block.

PR #1188, opened 2024-03-05
Add `get_pending_deposits` and test

tbro:
~Add `get_pending_deposits`  and test it. Call it from `Header::new`.~ Add  `get_finalized_deposits`.

### DIFF
--- a/sequencer/src/header.rs
+++ b/sequencer/src/header.rs
@@ -252,23 +252,6 @@ impl BlockHeader<SeqTypes> for Header {
         // and use `block_on` to query it.
         let l1 = instance_state.l1_client().snapshot().await;
 
-        // TODO remove this work around (aviods manipulating `from_info` signature).
-        let parent_header = parent_leaf.get_block_header();
-        // TODO cleanup and organization
-        if let Some(finalized) = l1.finalized {
-            dbg!(finalized);
-            let d = instance_state
-                .l1_client()
-                .get_finalized_deposits(
-                    parent_header.l1_finalized.map(|f| f.number),
-                    finalized.number,
-                    parent_header.fee_info.account().into(),
-                )
-                .await;
-
-            dbg!(d);
-        }
-
         Self::from_info(
             payload_commitment,
             metadata,
--- a/sequencer/src/header.rs
+++ b/sequencer/src/header.rs
@@ -259,7 +259,7 @@ impl BlockHeader<SeqTypes> for Header {
             dbg!(finalized);
             let d = instance_state
                 .l1_client()
-                .get_pending_deposits(
+                .get_finalized_deposits(
                     parent_header.l1_finalized.map(|f| f.number),
                     finalized.number,
                     parent_header.fee_info.account().into(),
--- a/sequencer/src/header.rs
+++ b/sequencer/src/header.rs
@@ -319,7 +319,10 @@ mod test_headers {
         NodeState, Payload,
     };
     use async_compatibility_layer::logging::{setup_backtrace, setup_logging};
-    use ethers::{types::RecoveryMessage, utils::Anvil};
+    use ethers::{
+        types::{Address, RecoveryMessage},
+        utils::Anvil,
+    };
     use hotshot_types::traits::block_contents::{vid_commitment, GENESIS_VID_NUM_STORAGE_NODES};
 
     #[derive(Debug, Default)]
@@ -574,7 +577,7 @@ mod test_headers {
     async fn test_validate_proposal_success() {
         let anvil = Anvil::new().block_time(1u32).spawn();
         let genesis_state = NodeState {
-            l1_client: L1Client::new(anvil.endpoint().parse().unwrap()),
+            l1_client: L1Client::new(anvil.endpoint().parse().unwrap(), Address::default()),
             ..Default::default()
         };
 
--- a/sequencer/src/header.rs
+++ b/sequencer/src/header.rs
@@ -252,6 +252,23 @@ impl BlockHeader<SeqTypes> for Header {
         // and use `block_on` to query it.
         let l1 = instance_state.l1_client().snapshot().await;
 
+        // TODO remove this work around (aviods manipulating `from_info` signature).
+        let parent_header = parent_leaf.get_block_header();
+        // TODO cleanup and organization
+        if let Some(finalized) = l1.finalized {
+            dbg!(finalized);
+            let d = instance_state
+                .l1_client()
+                .get_pending_deposits(
+                    parent_header.l1_finalized.map(|f| f.number),
+                    finalized.number,
+                    parent_header.fee_info.account().into(),
+                )
+                .await;
+
+            dbg!(d);
+        }
+
         Self::from_info(
             payload_commitment,
             metadata,
--- a/sequencer/src/l1_client.rs
+++ b/sequencer/src/l1_client.rs
@@ -146,7 +146,7 @@ impl L1Client {
             return vec![];
         }
 
-        // query for deposit events, errors are handled as "no deposits".
+        // query for deposit events, loop until successfull.
         let events =
             contract_bindings::fee_contract::FeeContract::new(address, Arc::new(&self.provider))
                 .deposit_filter()
@@ -262,13 +262,13 @@ mod test {
         for n in 1..=deposits {
             // Varied amounts are less boring.
             let amount = n as f32 / 10.0;
-            dbg!(n, amount);
             let receipt = contract
                 .deposit(wallet_address)
                 .value(parse_ether(amount).unwrap())
                 .send()
                 .await?
                 .await?;
+            // Successfull transactions have `status` of `1`.
             assert_eq!(Some(U64::from(1)), receipt.unwrap().status)
         }
 
--- a/sequencer/src/l1_client.rs
+++ b/sequencer/src/l1_client.rs
@@ -16,12 +16,13 @@
 //!   snapshot, which will cause the block builder to propose with a slightly old snapshot, but they
 //!   will still be able to propose on time.
 
+use crate::state::FeeInfo;
 use async_std::task::sleep;
 use commit::{Commitment, Committable, RawCommitmentBuilder};
 use ethers::prelude::*;
 use futures::join;
 use serde::{Deserialize, Serialize};
-use std::{cmp::Ordering, time::Duration};
+use std::{cmp::Ordering, sync::Arc, time::Duration};
 use url::Url;
 
 #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize, Hash, PartialEq, Eq)]
@@ -129,6 +130,32 @@ impl L1Client {
             }
         }
     }
+    /// Get fee info for each `Deposit` event. Returns `Vec<FeeInfo>`
+    pub async fn get_pending_deposits(
+        &self,
+        prev_finalized: Option<u64>,
+        new_finalized: u64,
+        address: Address,
+    ) -> Vec<FeeInfo> {
+        let prev = prev_finalized.unwrap_or(0);
+        if prev == new_finalized {
+            return vec![];
+        }
+        dbg!(prev, new_finalized);
+
+        let events =
+            contract_bindings::fee_contract::FeeContract::new(address, Arc::new(&self.provider))
+                .deposit_filter()
+                .from_block(prev)
+                .to_block(new_finalized)
+                .query()
+                .await;
+
+        dbg!(&events);
+        let r: Vec<FeeInfo> = events.unwrap().into_iter().map(Into::into).collect();
+        dbg!(&r);
+        r
+    }
 }
 
 async fn get_finalized_block<P: JsonRpcClient>(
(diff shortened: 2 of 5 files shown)


sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,32 @@ impl L1Client {
...
+
+        let events =
+            contract_bindings::fee_contract::FeeContract::new(address, Arc::new(&self.provider))
+                .deposit_filter()
+                .from_block(prev)
+                .to_block(new_finalized)
+                .query()
+                .await;
+
+        dbg!(&events);
```
tbro, line 154: `events` is an empty array at this point.

sequencer/src/l1_client.rs
```diff
@@ -204,4 +231,46 @@ mod test {
...
+        let provider =
+            Provider::<Http>::try_from(anvil.endpoint())?.interval(Duration::from_millis(10u64));
+        let client = SignerMiddleware::new(provider, wallet.with_chain_id(anvil.chain_id()));
+        let client = Arc::new(client);
+
+        // Initialize a contract with some deposits
+        let contract = contract_bindings::fee_contract::FeeContract::new(address, Arc::new(client));
+
+        for _ in 0..deposits {
+            contract.deposit(address).value(1).send().await?;
```
sveitser, line 258: When sending transactions need to check that they do not revert. 

For CAPE we used this helper trait

https://github.com/EspressoSystems/cape/blob/760048765297d42b19c104c160f745e9c24d6635/contracts/rust/src/assertion.rs#L59-L70

sequencer/src/l1_client.rs
```diff
@@ -204,4 +231,46 @@ mod test {
...
+        let wallet: LocalWallet = anvil.keys()[0].clone().into();
+
+        // In order to deposit we need a provider that can sign.
+        let provider =
+            Provider::<Http>::try_from(anvil.endpoint())?.interval(Duration::from_millis(10u64));
+        let client = SignerMiddleware::new(provider, wallet.with_chain_id(anvil.chain_id()));
+        let client = Arc::new(client);
+
+        // Initialize a contract with some deposits
+        let contract = contract_bindings::fee_contract::FeeContract::new(address, Arc::new(client));
```
sveitser, line 255: Currently it doesn't look like you are deploying the contract.

sveitser: The contract bindings have a `deploy` method. The `new` method is to attach to an already deployed contract.

tbro: :facepalm:

tbro: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1188/commits/340b63e0151cc1435fc387ccd6acd7ab0a7b56ed

sequencer/src/l1_client.rs
tbro: For now I added an `assert` that does the same in https://github.com/EspressoSystems/espresso-sequencer/pull/1188/commits/340b63e0151cc1435fc387ccd6acd7ab0a7b56ed.

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,34 @@ impl L1Client {
...
+    ) -> Vec<FeeInfo> {
+        let prev = prev_finalized.unwrap_or(0);
+        // When called from `Header::new` during testing both will be
+        // `0`. BUG: https://github.com/foundry-rs/foundry/issues/7245
+        // We treat this as if there were no deposits.
+        if prev == new_finalized {
+            return vec![];
+        }
+
+        // query for deposit events, errors are handled as "no deposits".
```
sveitser, line 149: Since we can't construct a valid block header if we don't have the correct deposits I think we should retry on errors instead.

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,34 @@ impl L1Client {
             }
         }
     }
+    /// Get fee info for each `Deposit` ocurring between `prev`
+    /// and `new`. Returns `Vec<FeeInfo>`
+    pub async fn get_pending_deposits(
```
sveitser, line 135: I think "pending" is a bit misleading because pending transactions usually refer to transactions that haven't been included in block so I think `get_finalized_deposits` would be a better choice of name here.

tbro: :thinking: I take "pending" to refer to sequencer state (as opposed to the l1), which makes sense to me. I don't have a strong opinion but ticket (#1119) *does* say "pending".

sequencer/src/header.rs
```diff
@@ -252,6 +252,23 @@ impl BlockHeader<SeqTypes> for Header {
         // and use `block_on` to query it.
         let l1 = instance_state.l1_client().snapshot().await;
 
+        // TODO remove this work around (aviods manipulating `from_info` signature).
+        let parent_header = parent_leaf.get_block_header();
+        // TODO cleanup and organization
+        if let Some(finalized) = l1.finalized {
+            dbg!(finalized);
```
tbro, line 259: I need to remove dbg statements.

sequencer/src/header.rs
```diff
@@ -252,6 +252,23 @@ impl BlockHeader<SeqTypes> for Header {
...
+            let d = instance_state
+                .l1_client()
+                .get_pending_deposits(
+                    parent_header.l1_finalized.map(|f| f.number),
+                    finalized.number,
+                    parent_header.fee_info.account().into(),
```
tbro, line 265: This is the wrong address.

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,34 @@ impl L1Client {
...
+        &self,
+        prev_finalized: Option<u64>,
+        new_finalized: u64,
+        address: Address,
```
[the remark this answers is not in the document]
tbro: I don't think we know what the contract address will be at this stage in development. I don't see any CLI params supplying it. Also many tests that use the `L1Client` will not have this or we will have to deploy a contract unnecessarily for all those tests... Maybe the easy solution is to make it an an `Option`...

sequencer/src/l1_client.rs
sveitser: In some ways they are pending to be credited in our own ledger but since this is a method on the L1 client I do think it's better to refer to them as "finalized deposits" because they are finalized on L1.

sequencer/src/l1_client.rs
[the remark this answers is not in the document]
jbearer: We can always just pass in `Address::default()` or something for tests, as long as we don't call any methods that use the contract in those tests it should be fine

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,34 @@ impl L1Client {
...
+    ) -> Vec<FeeInfo> {
+        let prev = prev_finalized.unwrap_or(0);
+        // When called from `Header::new` during testing both will be
+        // `0`. BUG: https://github.com/foundry-rs/foundry/issues/7245
+        // We treat this as if there were no deposits.
```
jbearer, line 144: This will also frequently happen in production, as there will be many HotShot blocks produced in between each finalized ETH block. In this case returning an empty list is the correct behavior: there are in fact no new deposits.

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,34 @@ impl L1Client {
...
+        if prev == new_finalized {
+            return vec![];
+        }
+
+        // query for deposit events, errors are handled as "no deposits".
+        let events =
+            contract_bindings::fee_contract::FeeContract::new(address, Arc::new(&self.provider))
+                .deposit_filter()
+                .from_block(prev)
```
jbearer, line 153: I think this should be from `prev + 1`, since deposits in block `prev` should already have been processed when `prev` was finalized. Unless `prev_finalized` was `None`, then we haven't processed anything so far. So we may need something like `let prev = prev_finalized.map(|prev| prev + 1).unwrap_or(0)`.

sequencer/src/l1_client.rs
tbro: fixed in d85c24bb94c26ad18d1fb202f110d67f6f90de3e

sequencer/src/l1_client.rs
tbro: fixed in 52d20ed22bfd3976bbd7b6bdfe5115c05be99379

sequencer/src/l1_client.rs
tbro: adjusted the comment in d85c24bb94c26ad18d1fb202f110d67f6f90de3e

sequencer/src/l1_client.rs
tbro: fixed in d85c24bb94c26ad18d1fb202f110d67f6f90de3e

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,43 @@ impl L1Client {
...
+        prev_finalized: Option<u64>,
+        new_finalized: u64,
+        address: Address,
+    ) -> Vec<FeeInfo> {
+        // `prev` should have allready been processed unless we
+        // haven't processed *any* blocks yet.
+        let prev = prev_finalized.map(|prev| prev + 1).unwrap_or(0);
+        // No new blocks have been finalized, therefore there are no
+        // new deposits.
+        if prev == new_finalized {
```
jbearer, line 146: I think there is still an off-by-1 issue here. Imagine the previous and new finalized blocks are both 1. But we added 1 to `prev` above so we get `2 != 1` here and fail this check when it should pass

sequencer/src/l1_client.rs
```diff
@@ -129,6 +130,43 @@ impl L1Client {
...
+            )
+            .deposit_filter()
+            .from_block(prev)
+            .to_block(new_finalized)
+            .query()
+            .await
+            {
+                Ok(events) => break events,
+                Err(e) => {
+                    tracing::warn!("Fee Event Error: {}", e);
```
jbearer, line 164: Add some delay between retries?

sequencer/src/l1_client.rs
tbro: 65c6ad67a71a644d055e95d85e6f5a3233327ab6 should address it.

sequencer/src/l1_client.rs
tbro: How embarrassing. fixed in ff623fc0c816e46adcf629ad68c7f28dc024a75b

sequencer/src/l1_client.rs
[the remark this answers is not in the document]
tbro: fixed in 82c45367204cd07cf4b22a5736c301b3068f3316

sequencer/src/l1_client.rs
```diff
@@ -136,21 +139,20 @@ impl L1Client {
...
         new_finalized: u64,
-        address: Address,
     ) -> Vec<FeeInfo> {
         // `prev` should have allready been processed unless we
         // haven't processed *any* blocks yet.
         let prev = prev_finalized.map(|prev| prev + 1).unwrap_or(0);
         // No new blocks have been finalized, therefore there are no
         // new deposits.
-        if prev == new_finalized {
+        if prev >= new_finalized {
```
jbearer, line 148: This is also problematic, I think. Take `prev_finalized = Some(1)` and `new_finalized = 2`. In this case we _should_ process new deposits (from block 2), but we won't because `prev == 2`. I think the check we want is something like `prev_finalized == Some(new_finalized)`

jbearer: Also we should probably have some tests that fail if this check is incorrect

tbro:
I believe the tests in this PR also  address [#1117](https://github.com/EspressoSystems/espresso-sequencer/issues/1117).

sequencer/src/l1_client.rs
```diff
@@ -140,15 +140,16 @@ impl L1Client {
...
         // new deposits.
-        if prev >= new_finalized {
+        if prev_finalized == Some(new_finalized) {
             return vec![];
         }
 
+        // `prev` should have allready been processed unless we
+        // haven't processed *any* blocks yet.
+        let prev = prev_finalized.map(|prev| prev + 1).unwrap_or(0);
+        dbg!(prev, new_finalized);
```
jbearer, line 152: We should replace all `dbg!` with `tracing::debug!` or something, as `dbg!` unconditionally outputs regardless of the log level which can make the logging very noisy

jbearer (approved):
Looks good, I guess next step is to integrate this into the proposal/validation logic?

sequencer/src/l1_client.rs
tbro: OH yea I need to be more careful about removing those, maybe I'll add a check to my local hooks. In any case, it's gone now.

tbro:
> Looks good, I guess next step is to integrate this into the proposal/validation logic?

Right. I think we just need to plug it in where we are currently calling `fetch_fee_receipts` here: https://github.com/EspressoSystems/espresso-sequencer/blob/5f3a222ed69ad3468d9d316f2cec2e86752a6d35/sequencer/src/header.rs#L266

It may be best at this point to build on top of #1151 to avoid doing it twice.

Issue #1190, opened 2024-03-06
Override `transaction` to get transaction without proof


akonring:
A few places in the sequencer we are iterating over transactions in a block using the `QueryablePayload` interface.
- in HotShot via transaction_commitments, to keep track of which transactions have been sequenced already
- in the query service to index newly committed transactions.  

Both of these ultimately boil down to calling `transaction_with_proof`, via the default impl of transaction:
```rust
    fn transaction(
        &self,
        meta: &Self::Metadata,
        index: &Self::TransactionIndex,
    ) -> Option<Self::Transaction> {
        Some(self.transaction_with_proof(meta, index)?.0)
    }
```
We should override transaction in the sequencer to just get the transaction, without proof, so we don't compute an expensive VID proof for no reason.

Issue #1010, opened 2024-01-20
Implement `QueryablePayload::transaction_with_proof` for multiple namespaces


ggutoski:
Next step after #991 , #931 .

- [ ] Add comment related to empty name spaces and basic tests. See https://github.com/EspressoSystems/espresso-sequencer/issues/1043#issuecomment-1913768156

philippecamacho:
Low priority as it does not block other tasks. Needed only as a feature for end users.

philippecamacho:
@jbearer do we need this feature for Cappuccino sprint 2?

PR #1191, opened 2024-03-06
add transaction to QueryablePayload impl

akonring:
closes: #1190 

Introduces `transaction` as a part of the implementation of `QueryablePayload` in the sequencer. This helps avoid computing unnecessary (expensive) proofs when querying transactions.

Note: the new method `transaction` is merely extracting appropriate parts of  `transaction_with_proof` so whatever clean-up/refactoring that needs to be done in `transaction_with_proof` probably also applies in the new `transaction`.

### DIFF
--- a/sequencer/src/block/queryable.rs
+++ b/sequencer/src/block/queryable.rs
@@ -64,6 +64,51 @@ impl QueryablePayload for Payload<TxTableEntryWord> {
         TxIterator::new(ns_table, self)
     }
 
+    fn transaction(
+        &self,
+        meta: &Self::Metadata,
+        index: &Self::TransactionIndex,
+    ) -> Option<Self::Transaction> {
+        let index_usize = index.tx_idx; // TODO fix in https://github.com/EspressoSystems/espresso-sequencer/issues/1010
+        if index_usize >= self.len(meta) {
+            return None; // error: index out of bounds
+        }
+
+        // start
+        let tx_table_range_start = if index_usize == 0 {
+            None
+        } else {
+            let range_proof_start = index_usize.checked_mul(TxTableEntry::byte_len())?;
+            Some(TxTableEntry::from_bytes(self.raw_payload.get(
+                range_proof_start..range_proof_start.checked_add(TxTableEntry::byte_len())?,
+            )?)?)
+        };
+
+        // end
+        let tx_table_range_proof_end = index_usize
+            .checked_add(2)?
+            .checked_mul(TxTableEntry::byte_len())?;
+        let tx_table_range_end = TxTableEntry::from_bytes(self.raw_payload.get(
+            tx_table_range_proof_end.checked_sub(TxTableEntry::byte_len())?
+                ..tx_table_range_proof_end,
+        )?)?;
+
+        let tx_payload_range = tx_payload_range(
+            &tx_table_range_start,
+            &tx_table_range_end,
+            &self.get_tx_table_len(),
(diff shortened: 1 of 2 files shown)


sequencer/src/block/queryable.rs
```diff
@@ -64,6 +64,51 @@ impl QueryablePayload for Payload<TxTableEntryWord> {
...
+            &tx_table_range_start,
+            &tx_table_range_end,
+            &self.get_tx_table_len(),
+            self.raw_payload.len(),
+        )?;
+        Some(
+            // TODO don't copy the tx bytes into the return value
+            // https://github.com/EspressoSystems/hotshot-query-service/issues/267
+            Transaction::new(
+                crate::VmId(0),
```
jbearer, line 106: This seems incorrect. Shouldn't we use `index.ns_idx` and look up the VM ID in the namespace table. This may also explain why we've been seeing more duplicate transactions lately, since the hash of the transaction returned here isn't going to match the hash of the submitted transaction, if it didn't originally have namespace 0.

sequencer/src/block/queryable.rs
```diff
@@ -64,6 +64,51 @@ impl QueryablePayload for Payload<TxTableEntryWord> {
...
+        if index_usize >= self.len(meta) {
+            return None; // error: index out of bounds
+        }
+
+        // start
+        let tx_table_range_start = if index_usize == 0 {
+            None
+        } else {
+            let range_proof_start = index_usize.checked_mul(TxTableEntry::byte_len())?;
+            Some(TxTableEntry::from_bytes(self.raw_payload.get(
```
jbearer, line 82: Similarly whenever we read from the `raw_payload` don't we need to offset by the position of the namespace in the payload, from the namespace table?

jbearer:
Can confirm this is a big performance improvement. Before sequencer nodes were taking >2 cores, now they're down to like half a core each, and the demo is running much smoother. I suspect the remaining CPU usage is mostly VID disperse

sequencer/src/block/queryable.rs
akonring: Ok. I thought we were looking for a quick fix to support transactions without computing the proofs but which is consistent with `transaction_with_proof`. 
The actual method is broken (issue here https://github.com/EspressoSystems/espresso-sequencer/issues/1010) but maybe we should make it higher priority to fix this.

jbearer (approved):

sequencer/src/block/queryable.rs
akonring: Right. I can start right away on getting the new transaction lookups in good shape. And then port the changes to `transaction_with_proof` later. Can continue on this PR or start another after this is merged - whatever works best(?)

jbearer: I don't really have a preference, whatever you prefer

sveitser:
@akonring can we merge this? Prefer to merge it sooner rather than later because it makes running things a lot smoother.

akonring:
> @akonring can we merge this? Prefer to merge it sooner rather than later because it makes running things a lot smoother.

Yes. Will fix the rest in another PR including issue #1010.

sequencer/src/block/queryable.rs
akonring: The namespace support has continued in this follow-up PR: https://github.com/EspressoSystems/espresso-sequencer/pull/1205

PR #1192, opened 2024-03-06
Move cargo-audit to separate action

sveitser:
- For PRs only run on dependency changes.
- Run once a day on main.
- Add audit.toml file
- Update mio and whoami to avoid CVEs.


imabdulbasit (approved):

sveitser:
@Ancient123 Could you help force merge this PR. The breaking native demo step is fixed in another PR that is also blocked because the audit failure that is fixed here.

PR #1193, opened 2024-03-06
Use port 44010 instead of 50010 for tx generator

sveitser:
Currently the logs indicate that the test fails because port 50010 is already in use on the CI.


sveitser:
Merged with https://github.com/EspressoSystems/espresso-sequencer/pull/1192

Issue #1204, opened 2024-03-07
`QueryablePayload::transaction` with support for multiple namespaces


akonring:
In #1191, the `QueryablePayload::transaction` method was introduced to limit the use of the expensive  `transaction_with_proof`. But both `transaction` and `transaction_with_proof` is broken (does not support multiple namespaces).

This issue deals with building support for multiple namespaces in `transaction`.
#1010 is the issue for namespace support in `transaction_with_proof`

PR #1205, opened 2024-03-07
transaction with support for multiple namespaces

akonring:
closes: #1204 

Currently, the `Querypayload::transaction` is broken such that it only returns correct transaction if VmId indeed is 0 and there is only a single namespace:
```rust
Transaction::new(
                crate::VmId(0),
                self.raw_payload.get(tx_payload_range.clone())?.to_vec(),
            ),
```
This PR follows up on #1191 to have `Querypayload::transaction support multiple namespaces.

### DIFF
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -592,6 +592,7 @@ mod test {
                     ns_proof_txs,
                     derived_ns
                         .tx_payloads
+                        .clone()
                         .into_iter()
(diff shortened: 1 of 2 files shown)


jbearer (approved):
In the local demo, the load generator is now receiving confirmations of the transactions it submits, which means the reported hashes must be matching the actual hashes. So I think this is working.

ggutoski (commented):
Hard to review on a phone but I see no problems.

PR #1207, opened 2024-03-07
I don't think we need ws anymore.

tbro:
I think we forgot to remove the `ws` feature from `ethers` when we updated the `L1Client`.


jbearer (approved):
Nice!

PR #1212, opened 2024-03-08
Support multiple namespaces in `QueryablePayload::transaction_with_proof`

akonring:
closes: #1010 

### Description
Currently the `QueryablePayload::transaction_with_proof` is broken as it does not support more than one namespace with namespace id: `VmId(0)`.
This PR fixes broken code such that both proving and verification can be done for transactions in all namespaces of the block.

### Other benefits
Methods like `get_tx_table_len` and `get_tx_table_len_proof` in `payload.rs` were designed when only a single tx-table existed (no namespaces). These are now removed (no other methods depended on them) and the new implementation instead makes use of the `TxTable` inside `tables.rs`.

### Considerations
A caching mechanism was designed when we only had a single tx-table (no namespaces). This could be revisited to regain some efficiency from re-using the proofs generated fx for tx-table lengths.
https://github.com/EspressoSystems/espresso-sequencer/blob/6494449d30c39ca018608d250704ee458706a91a/sequencer/src/block/payload.rs#L75-L83

### DIFF
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -756,7 +756,7 @@ mod test {
                 &block.raw_payload,
                 0..std::cmp::min(TxTableEntry::byte_len(), block.raw_payload.len()),
             )
-            .unwrap(), //block.get_tx_table_len_proof(&vid).unwrap().clone(),
+            .unwrap(),
             vid.payload_proof(&block.raw_payload, 0..3).unwrap(),
         );
 
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -9,7 +9,7 @@ use commit::Committable;
 use derivative::Derivative;
 use hotshot::traits::BlockPayload;
 use hotshot_types::vid::{
-    vid_scheme, LargeRangeProofType, SmallRangeProofType, VidCommitment, VidCommon, VidSchemeType,
+    vid_scheme, LargeRangeProofType, VidCommitment, VidCommon, VidSchemeType,
 };
 use jf_primitives::vid::{
     payload_prover::{PayloadProver, Statement},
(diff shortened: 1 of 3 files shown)


sequencer/src/block/payload.rs
```diff
@@ -789,8 +750,13 @@ mod test {
...
         let proof = queryable::gen_tx_proof_for_testing(
-            block.get_tx_table_len(),
-            block.get_tx_table_len_proof(&vid).unwrap().clone(),
+            0..block.raw_payload.len(),
+            TxTableEntry::from_usize(TxTable::get_tx_table_len(&block.raw_payload)),
+            vid.payload_proof(
+                &block.raw_payload,
+                0..std::cmp::min(TxTableEntry::byte_len(), block.raw_payload.len()),
+            )
+            .unwrap(), //block.get_tx_table_len_proof(&vid).unwrap().clone(),
```
sveitser, line 759: Is this comment left on purpose?

akonring: No, thanks for spotting this.
should be fixed: [a154708](https://github.com/EspressoSystems/espresso-sequencer/pull/1212/commits/a154708834804c74e6597a50a34d8672fca2551a)

sequencer/src/block/queryable.rs
```diff
@@ -327,26 +345,29 @@ impl TxInclusionProof {
...
-            TxTableEntry::byte_len()
+            TxTableEntry::byte_len().checked_add(self.ns_range.start)?
         };
         tx_table_range_bytes.extend(self.tx_table_range_end.to_bytes());
         let range = start
             ..index
                 .checked_add(2)?
-                .checked_mul(TxTableEntry::byte_len())?;
+                .checked_mul(TxTableEntry::byte_len())?
+                .checked_add(self.ns_range.start)?;
```
sveitser, line 370: It looks like we compute the same offsets and ranges in multiple places. Wondering if it would be worth it to add some helper methods for it and briefly comment those. I think it would make it easier to read the code, at least for people like me who aren't familiar with this part of the code.

sveitser (approved):
I'm not very familiar with the payload encoding so probably wouldn't spot an error if there was one but the changes look good to me. If you're unsure and would like a more thorough review maybe @nomaxg can have a look.

akonring:
Rebasing on `main` to incorporate #1229 and make CI happy

PR #1214, opened 2024-03-08
Fix ubuntu build

sveitser:
- Install postgres client, just
- Make it easier to test locally with docker
- Run with code in repo
- Fix contracts script name for latest foundry
- Skip SQL tests
- Less fuzz runs

### DIFF
--- a/scripts/ubuntu-install-test-no-nix-docker
+++ b/scripts/ubuntu-install-test-no-nix-docker
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# This script is to test the ubuntu installation instructions locally on
+# non-ubuntu machines by running through the installation and tests inside an
+# ubuntu docker contrainer.
+set -e
+
+# Create a temporary copy of the repository to be mounted into the container.
+REPO_COPY_DIR="$(mktemp -d)"
+
+# Remove repo copy on exit, error etc.
+trap "exit" INT TERM
+trap cleanup EXIT
+cleanup(){
+    echo "Cleaning up repo copy: $REPO_COPY_DIR, need sudo"
+    sudo rm -rf "$REPO_COPY_DIR"
+}
+
+git clone --recursive . "$REPO_COPY_DIR"
+
+# Run the installation and tests inside the container.
+docker run -it --rm \
+    -v "$REPO_COPY_DIR:/code" \
+    ubuntu:latest \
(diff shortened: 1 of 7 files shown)


sveitser:
CI run for the ubuntu install test job: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/8204651189

sveitser:
This is currently broken because foundryup downloads to XDG_CONFIG_DIR if that exists :facepalm:

sveitser:
Next try: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/8205372390/job/22442069625

scripts/ubuntu-install-test-no-nix-docker
```diff
@@ -0,0 +1,24 @@
...
+
+# Create a temporary copy of the repository to be mounted into the container.
+REPO_COPY_DIR="$(mktemp -d)"
+
+# Remove repo copy on exit, error etc.
+trap "exit" INT TERM
+trap cleanup EXIT
+cleanup(){
+    echo "Cleaning up repo copy: $REPO_COPY_DIR, need sudo"
+    sudo rm -rf "$REPO_COPY_DIR"
```
tbro, line 15: We shouldn't need `sudo` to remove a directory created by normal user. Same user should be able to delete.

scripts/ubuntu-install-test-no-nix-docker
```diff
@@ -0,0 +1,24 @@
...
+}
+
+git clone --recursive . "$REPO_COPY_DIR"
+
+# Run the installation and tests inside the container.
```
tbro, line 20: Since we are using docker anyway, instead of mounting a dir, couldn't we copy the build script into an ubuntu container and do all the git clone inside the container? This way everything is ephemeral and no need for external temp dirs.

scripts/ubuntu-install-test-no-nix-docker
sveitser: Files will be owned by root because running in docker container.

scripts/ubuntu-install-test-no-nix-docker
sveitser: It's done like this now so you can get local changes that aren't yet on github inside the container. The script is just for local debugging.

tbro (approved):

PR #1216, opened 2024-03-08
Removed unused udeps, workflow to find unused dependencies

### LINKED ISSUE #1213: Run cargo udeps on CI
Tool: https://github.com/est31/cargo-udeps

I think we can run this once a week to alert us on unused dependencies. It might also be nice to run it on PRs so we get notified in cases we no longer need a dependency.

There's a github action for it: https://github.com/aig787/cargo-udeps-action

nomaxg:
closes #1213

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -21,7 +21,7 @@ env:
 
 jobs:
   contracts:
-    timeout-minutes: 60
+    timeout-minutes: 20
     runs-on: ubuntu-latest
     steps:
       - name: Install Nix
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -21,6 +21,7 @@ env:
 
 jobs:
   contracts:
+    timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
       - name: Install Nix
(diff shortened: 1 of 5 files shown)


.github/workflows/contracts.yml
```diff
@@ -21,6 +21,7 @@ env:
 
 jobs:
   contracts:
+    timeout-minutes: 60
```
nomaxg, line 24: @mrain @alxiong is this a reasonable timeout for this job?

mrain: cc @alysiahuggins @philippecamacho do you have more inputs?

nomaxg: Actually I think this is more than reasonable, anything longer than an hour shouldn't block CI (an hour is arguably pretty generous already).

mrain: Right, I think it's more than enough

tbro: Looking at healthy tests I see they pass in ~11min, can we reduce this to ~15 or ~20?

tbro: Here you can see they consistently succeed in <= 12 min. https://github.com/EspressoSystems/espresso-sequencer/actions/workflows/contracts.yml

nomaxg: done https://github.com/EspressoSystems/espresso-sequencer/pull/1216/commits/5ea2b330911d7949b83d7d12bd2f50c9863e1209

tbro (approved):
:+1:

PR #1218, opened 2024-03-08
check spelling in lint job

tbro:
Check spelling in lint job. For now I'm just cargo installing the tool. If we decide we want this we can cache that or install the binary from github. I have also not yet added configuration to guard against any false positives.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,6 +38,6 @@ jobs:
 
       - name: Spell Good
         run: |
-          cargo install typos
+          cargo install typos-cli
           typos
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,3 +35,9 @@ jobs:
 
       - name: Check
         run: cargo clippy --workspace --all-features --all-targets -- -D warnings
+
+      - name: Spell Good
+        run: |
+          cargo install typos
+          typos
+
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,14 +30,13 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         name: Enable Rust Caching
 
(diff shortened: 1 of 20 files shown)


sveitser:
There's also a github action: https://github.com/marketplace/actions/typos-action

tbro:
> This sounds like something we should add to the flake.nix and add it to the pre-commit hooks in there. This gives much quicker feedback than having it only in the CI.

Yes I think there may be some value in having the errors in this PR so we can we what we are up against.

mrain:
> This sounds like something we should add to the flake.nix and add it to the pre-commit hooks in there. This gives much quicker feedback than having it only in the CI.

Amazing, should we propagate this action to every repo?

mrain:
> This sounds like something we should add to the flake.nix and add it to the pre-commit hooks in there. This gives much quicker feedback than having it only in the CI.

Did it for Jellyfish 🎉  https://github.com/EspressoSystems/jellyfish/pull/513/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0

tbro:
> > This sounds like something we should add to the flake.nix and add it to the pre-commit hooks in there. This gives much quicker feedback than having it only in the CI.
> 
> Did it for Jellyfish 🎉 https://github.com/EspressoSystems/jellyfish/pull/513/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0

Cool I copied your change there back over here

tbro:
> This sounds like something we should add to the flake.nix and add it to the pre-commit hooks in there. This gives much quicker feedback than having it only in the CI.

One issue with running from `pre-commit` is that the tool will descend into submodules. I don't see a way to prevent that at this time.

sveitser (approved):

PR #1224, opened 2024-03-11
Serialize bytes as base64

### LINKED ISSUE #827: Encode payload bytes as base64 for submit API
We currently use the serde-json default for `Vec<u8>` which leads to having a JSON array of numbers. We should switch to base64.

### LINKED ISSUE #1206: Serialize bytes as base 64 rather than JSON array.
There are a number of places where we have `Vec<u8>` in a serialized data structure, such as the transaction payload and VID stuff. Currently these serialize very efficiently using bincode, but with JSON they serialize as an array of numbers, which is both non-idiomatic and inefficient, compared to serializing with base64.

We should create or find a `Bytes` type which serializes appropriately in both formats (using [`is_human_readable`](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable) to distinguish) and replace uses of `Vec<u8>` with `Bytes`.

Running list of fields that need to be replaced:
- [ `Transaction::payload`](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/transaction.rs#L22)
- [`NameSpaceTable::bytes`](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/block/tables.rs#L47)
- [`Payload::raw_payload`](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/block/payload.rs#L72)

jbearer:
Closes #827
Closes #1206


sequencer/src/bytes.rs
```diff
@@ -0,0 +1,193 @@
+//! Convenient serialization for binary blobs.
```
sveitser, line 1: This would make a nice standalone crate.

jbearer: Sure, I can split it out. Any suggestions for a name?

sveitser: You mean something that isn't taken on crates.io or just in general? Uhm... base64bytes?

sequencer/src/bytes.rs
```diff
@@ -0,0 +1,193 @@
...
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    From,
+    Into,
+    CanonicalSerialize,
+    CanonicalDeserialize,
+)]
+pub struct Bytes(Vec<u8>);
```
tbro, line 45: Please don't block merge for this comment, but I wonder if going forward we can find a less generic name to avoid confusion w/ [`bytes::Bytes`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html).

jbearer:
Gonna merge this and replace the crate later, since the parts of this PR that are likely to cause conflicts if not merged soon will not be changing with the new base64-bytes crate. It will only be a simple swap of the new module that shouldn't conflict with anything

PR #1227, opened 2024-03-12
Update hotshot to 0.5.22

nomaxg:


sequencer/src/api/fs.rs
```diff
@@ -1,17 +1,14 @@
...
     data_source::{ExtensibleDataSource, FileSystemDataSource},
     node::NodeDataSource,
-    NotFoundSnafu, QueryError, QueryResult,
+    types::HeightIndexed,
+    QueryError,
 };
-use snafu::OptionExt;
 use std::{collections::BTreeMap, path::Path};
 
 #[derive(Clone, Debug, Default)]
```
jbearer, line 14: I think we can also remove `blocks_by_time` and switch from `ExtensibleDataSource` to plain `FileSystemDataSource`

jbearer (commented):
Can delete `sequencer/api/migrations/V11__timestamp_index.sql`

sequencer/src/api/fs.rs
nomaxg: I also removed `refresh_indices`

jbearer (approved):

PR #1233, opened 2024-03-13
Run audit workflow on main

sveitser:
Currently we don't run on `main`. Since the workflow is cheap I think we can add that to get alerts earlier if something requires our attention.


tbro (approved):
I think the CI failure will go away if you update this branch with main.

PR #1234, opened 2024-03-13
Move typos to separate CI workflow

sveitser:
The lint workflow is required to pass to merge. I think we shouldn't block PRs on typos, therefore moving this to a separate action.


tbro (approved):
Good Idea

PR #1235, opened 2024-03-13
Add test-node binary

sveitser:
This is a proof of concept for a sequencer test node binary. I think it would make our lives much easier for testing integrations because we will no longer have to maintain a separate docker compose project that goes out of date all the time. I expect it would also be helpful for other parties who would like to build on top of our sequencer.

Not sure if it makes sense to hide it behind the testing feature. Currently it uses some code that is only intended for tests.

To try:

    cargo run --release --features testing --bin test-node

Submit some transactions

    cargo run --release --bin submit-transactions -- http://localhost:33000

### DIFF
--- a/sequencer/src/bin/test-node.rs
+++ b/sequencer/src/bin/test-node.rs
@@ -1,0 +1,59 @@
+use async_compatibility_layer::logging::{setup_backtrace, setup_logging};
+use clap::Parser;
+use futures::StreamExt as _;
+use sequencer::{
+    api::{options, test_helpers::TestNetwork, Options},
+    catchup::mock::MockStateCatchup,
+    persistence::no_storage::NoStorage,
+    testing::TestConfig,
+    ValidatedState,
+};
+
+#[derive(Debug, Clone, Parser)]
+struct Args {
+    /// Port the API will server will listen on.
+    #[clap(
+        short,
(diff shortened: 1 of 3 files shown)


tbro:
I think it needs a README.

sequencer/src/bin/test-node.rs
```diff
@@ -0,0 +1,59 @@
...
+        .state(Default::default())
+        .submit(Default::default())
+        .query_fs(
+            Default::default(),
+            sequencer::persistence::fs::Options { path },
+        );
+    let with_state = TestNetwork::with_state(
+        api_options,
+        states,
+        [NoStorage; TestConfig::NUM_NODES],
```
nomaxg, line 51: Maybe make num_nodes an argument

sveitser:
Agree we should definitely add some documentation and also reduce the number of nodes to a minimum, as well as figure out how to handle the L1.

jbearer:
> Not sure if it makes sense to hide it behind the testing feature. Currently it uses some code that is only intended for tests.

Could add a new `test-node` feature flag, and for those things that are required in both, `#[cfg(any(feature = "test-node", feature = "testing"))]`

sveitser:
Close in favor of #1353

PR #1239, opened 2024-03-14
update hotshot tag

move47:
It accommodates changes recently introuduced to Hotshot, especially `HotshotConfig`. Moreover, It makes `hotshot-types` points to its own repo, instead of hotshot crate.

### DIFF
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -81,6 +81,10 @@ impl BlockPayload for Payload<TxTableEntryWord> {
     fn builder_commitment(&self, _metadata: &Self::Metadata) -> BuilderCommitment {
         unimplemented!("TODO builder_commitment");
     }
+
+    fn get_transactions(&self, metadata: &Self::Metadata) -> &Vec<Self::Transaction> {
+        unimplemented!("TODO get_transactions");
+    }
 }
 
 #[cfg(test)]
--- a/sequencer/src/block.rs
+++ b/sequencer/src/block.rs
@@ -82,7 +82,7 @@ impl BlockPayload for Payload<TxTableEntryWord> {
         unimplemented!("TODO builder_commitment");
     }
 
-    fn get_transactions(&self, metadata: &Self::Metadata) -> &Vec<Self::Transaction> {
+    fn get_transactions(&self, _metadata: &Self::Metadata) -> &Vec<Self::Transaction> {
         unimplemented!("TODO get_transactions");
     }
 }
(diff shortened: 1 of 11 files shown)


sequencer/src/block.rs
```diff
@@ -82,6 +82,10 @@ impl BlockPayload for Payload<TxTableEntryWord> {
     fn builder_commitment(&self, _metadata: &Self::Metadata) -> BuilderCommitment {
         unimplemented!("TODO builder_commitment");
     }
+
+    fn get_transactions(&self, _metadata: &Self::Metadata) -> &Vec<Self::Transaction> {
```
sveitser, line 86: Can't we implement this now instead of leaving it unimplemented? I think it can be the same as `transaction_commitments` above on line 77 without the `commit()`.

sequencer/src/lib.rs
```diff
@@ -140,6 +140,11 @@ impl<N: network::Type> NodeImplementation<SeqTypes> for Node<N> {
     type CommitteeNetwork = N::DAChannel;
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Delta {}
```
sveitser, line 144: We can leave this empty for now, added this issue: 

https://github.com/EspressoSystems/espresso-sequencer/issues/1242

sequencer/src/block.rs
move47: I am trying to use:
```
self.enumerate(metadata)
            .map(|(_, tx)| tx)
            .collect::<Vec<Self::Transaction>>()
            .as_ref()
```
It is throwing `can not return ref` error.

sequencer/src/lib.rs
move47: Sorry, Do you mean `pub struct Delta ()`?

move47: Moreover, the associated type also has a trait bound. So thought of this doing this way

sveitser: I mean it's fine like this but I assumed at some point we want to be more than a struct with no members. I meant to say we can do it in another PR.

move47: Oh I see, I thought you want some changes for it. Cool!

sveitser (approved):
LGTM, let's figure out how to handle the unimplemented parts without blocking the PR.

sequencer/src/block.rs
move47: So shall I keep this `unimplemented!()` for this PR, because the change will come later?

jbearer: If it doesn't break any tests it's probably fine...although it does make me wonder what this function is for if it's not being called

nyospe: I have a solution that will work. This was the reason I requested the metadata be added to the `get_transactions`. I'll take care of it later.

PR #1243, opened 2024-03-15
CI: coverage: ignore dirs

sveitser:
If I understand correctly this yaml file is only used by the github action but not something grcov understands. So we have to duplicate the ignore in the justfile and the config file here.


sveitser:
CI run here: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/8295762720

tbro (approved):
I think you are right.

Issue #999, opened 2024-01-16
Contracts upgrade deployment script


philippecamacho:
Specify and implement the scripts to upgrade the contracts, including tests to ensure the deployment is successful.
Take into account considerations introduced in #975 and other constraints such as the need for an atomic upgrade across all the contracts.

- Automatically Rollback if post-deployment tests fail

Potential steps
- create deployment script that allows for deployments with a multisig wallet
- add tests that must be satisfied in order for the deployment to be successful
- perform upgrade deployment with forked version of mainnet
- consider a deployment that executes in one transaction and fails if the test cases aren't satisfied

alysiahuggins:
I was stuck on deploying & upgrading the light client contract for a bit - I contacted openzeppelin defender support and this issue (https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/issues/26) created by the team. In the end using the `internal` visibility for library functions worked - adding this info here for documentation purposes.

alysiahuggins:
I'm reducing the scope until we get more support from OpenZeppelin.

I had written these potential steps on the ticket:
```
1. create deployment script that allows for deployments with a multisig wallet
2. add tests that must be satisfied in order for the deployment to be successful
3. perform upgrade deployment with forked version of mainnet
4. consider a deployment that executes in one transaction and fails if the test cases aren't satisfied
```

Step 1 is done 
- by using openzeppelin defender, we get some upgradable validation checks e.g.
- `upgrade safety`(e.g. the contract does not have a constructor, an initialize function exists etc https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#what-does-it-mean-for-a-contract-to-be-upgrade-safe)
- `upgrade compatibility` (e.g. compatible storage layout https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#what-does-it-mean-for-an-implementation-to-be-compatible)

Steps 2 & 3 
- have not been implemented because I'm not clear on how to test with a forked version of mainnet whilst using openzeppelin defender. This talks about using forked networks https://docs.openzeppelin.com/defender/v2/guide/forked-network but it does not support upgrades. which is when the tests are mostly needed.


Step 4 relies on step 2/3 working

[image.png](https://espresso.zulipchat.com/user_uploads/40450/qnt0PYntrYCSoK9ObbXFAPpn/image.png)
I asked OpenZeppelin support about the best way to test locally with Defender and his response was to ask in a forum.


[...]

PR #1248, opened 2024-03-15
999 contracts upgrade deployment script

alysiahuggins:
uses openzeppelin defender + gnosis safe multisig wallets to deploy the light client contract and the fee contract
changes were made to both contract to make it upgrade safe. e.g. no immutable variables

deployment steps
- set up the prerequisites as written in contracts/script/README.md
- run foundry deployment command 
- open the open zeppelin ui and navigate to the network where the deployment occurs and this provides you with a link to the transaction to be approved by the wallets via the Safe UI
- the signers approve the transactions and one signer executes the transactions once the signature threshold is met. Normally two transactions are needed  - the deployment of the implementation contract as well as the proxy contract.


.gitmodules
```diff
@@ -13,15 +13,15 @@
...
 [submodule "contracts/lib/foundry-devops"]
 	path = contracts/lib/foundry-devops
 	url = https://github.com/chainaccelorg/foundry-devops
+[submodule "contracts/lib/openzeppelin-foundry-upgrades"]
+	path = contracts/lib/openzeppelin-foundry-upgrades
+	url = https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades
+[submodule "contracts/lib/openzeppelin-contracts-upgradeable"]
+	path = contracts/lib/openzeppelin-contracts-upgradeable
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
+[submodule "contracts/lib/forge-std"]
```
sveitser, line 25: This looks like a duplicate the module in line 1.

contracts/script/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+Steps:
+
+1. Run the Deployment Command This command requires you to go to OpenZeppelin Defender's UI to see the transaction.
+   Click that transaction which opens up the Safe UI where your signers for that Safe multi-sig wallet can confirm the
+   transaction. The two transactions to be confirmed are: (i) deployment of implementation contract (ii) deployment of
+   proxy contract
+
+```bash
+export FOUNDRY_PROFILE=defender && \
```
sveitser, line 38: Is this required for `forge build`?

If not, 

```
env FOUNDRY_PROFILE=defender forge script contracts/script/FeeContractWithDefender.s.sol:FeeContractDefenderDeployScript --ffi --rpc-url https://ethereum-sepolia.publicnode.com
```
Would only set the env var for the command, so there's no need to restore it at the end.

contracts/script/README.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+The profile, `profile.default` is used by default. Setting the `FOUNDRY_PROFILE` variable overrides the `foundry.toml`
+settings.
+
+## Prerequisites
+
+1. Create a multisig wallet using [Safe](https://app.safe.global/welcome/accounts) on the network you'd like to deploy
+   to.
+2. In OpenZeppelin Defender, create an Approval Process that requires the multisig wallet you created above.
```
sveitser, line 21: I get this message when using the address from step 1 in defender.


image

contracts/script/README.md
```diff
@@ -0,0 +1,142 @@
...
+3. In OpenZeppelin Defender, create a deployment environment and use the approval process created in Step 2. Be sure to
+   copy the Defender secret and key, that is shown at the end of this step, into the .env file.
+
+## Deployments
+
+### Deploying the Fee Contract
+
+Steps:
+
+1. Run the Deployment Command This command requires you to go to OpenZeppelin Defender's UI to see the transaction.
```
sveitser, line 32: Where in the UI should I see this transaction?

I got this as output from `forge script`:

```
[⠊] Compiling...
[⠆] Compiling 92 files with 0.8.23
[⠒] Solc 0.8.23 finished in 17.23s
Compiler run successful with warnings:
Warning (2018): Function state mutability can be restricted to view
   --> contracts/lib/openzeppelin-foundry-upgrades/src/internal/Utils.sol:133:5:
    |
133 |     function getOutDir() internal returns (string memory) {
    |     ^ (Relevant source part starts here and spans across multiple lines).

Warning (2018): Function state mutability can be restricted to pure
  --> contracts/test/PlonkVerifier.t.sol:29:5:
   |
29 |     function sanitizeScalarField(uint256 a) public view returns (uint256) {
   |     ^ (Relevant source part starts here and spans across multiple lines).

2024-03-18T14:16:10.827858Z  WARN foundry_evm_core::fork::cache: Failed to read cache file err=Os { code: 2, kind: NotFound, message: "No such file or directory" } path="/home/lulu/.foundry/cache/rpc/sepolia/5511676"
```

contracts/src/LightClient.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: MIT
```
sveitser, line 1: Does this need to be changed? I think we haven't yet decided on what license to use for our contracts.

.env
```diff
@@ -84,3 +84,9 @@ ESPRESSO_SEQUENCER_DEPLOY_LIGHTCLIENT_CONTRACT=yes
 # Load generator
 ESPRESSO_SUBMIT_TRANSACTIONS_DELAY=1s
 ESPRESSO_SUBMIT_TRANSACTIONS_PORT=44010
+
+# Openzeppelin Defender Deployment Profile
+DEFENDER_KEY="3vKMu7Dwajrg8GMV3B4qVZpspdocj1Pt"
+DEFENDER_SECRET="4TJWVRbw6kWGqR3bTrAQQTBcBQiNduvZvEqUv9o7fAChWZW5pcgqNkkj1RgKjg1b"
```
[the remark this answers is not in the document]
sveitser: I think it would be better to leave empty placeholders in the file instead.

contracts/test/FeeContract.t.sol
```diff
@@ -17,160 +17,6 @@ import { FeeContract } from "../src/FeeContract.sol";
 import { DeployFeeContractScript } from "../script/Fee.s.sol";
 
 /// @title FeeContract Test
-contract FeeContractTest is Test {
```
sveitser, line 20: I wonder why this was deleted.

contracts/script/FeeContractWithDefender.s.sol
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+contract FeeContractDefenderUpgradeScript is Script {
+    function run() public returns (string memory proposalId, string memory proposalUrl) {
+        bytes32 contractSalt = bytes32(abi.encodePacked(vm.envString("FEE_CONTRACT_SALT")));
+        Options memory opts;
+        opts.defender.useDefenderDeploy = true;
+        opts.defender.salt = contractSalt;
+        address proxyAddress = 0x61B4C96475B99A6ce01AfF0da7910605D048c125;
```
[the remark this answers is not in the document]
alysiahuggins: I manually added it for this step and this will change for our actual deployment - when using openzeppelin defender I can't put a parameter in the run() function.

contracts/script/README.md
alysiahuggins: Both profiles (default and defender) work for forge build but generating the bindings fail for the defender profile. The defender profile sets build info to true and the existence of the build info in the generated json for the contracts causes the binding generation to fail. (so this is what inspired creating a separate foundry profile for running defender scripts)

contracts/script/README.md
alysiahuggins: Yes so at this point, you'd go into openzepelin defender (OPD) and you'd see the transaction waiting to be signed by the wallets in the multisig 'safe'. I just checked OPD to see if there is a transaction waiting but I don't. I'll DM you

contracts/src/LightClient.sol
alysiahuggins: Ah yes, defender didn't allow unlicensed contracts so that's something we have to decide to be able to move forward with defender.

contracts/test/FeeContract.t.sol
alysiahuggins: I thought that it was redundant as there were the same tests in the `FeeContractUpgradabilityTest`

.gitmodules
alysiahuggins: Ooh let me remove that - i had to update forge-std in the process

contracts/script/README.md
sveitser: Is there an open issue about this on the foundry github repo? If not we should open one.

contracts/script/README.md
sveitser: We noticed that my initial attempt wasn't properly indexed but it worked when I tried again a few hours later.

philippecamacho:
Is the plan still to have a single script that upgrades both light client and fee contract? Or shall this be done in another PR?

contracts/test/FeeContract.t.sol
philippecamacho: Some tests were not redundant. They have been readded in bf55348.

contracts/script/README.md
sveitser: So the issue is that `build_info = true` creates an extra file in the `out` folder that forge bind tries to generate bindings from.

sveitser: Upstream issue:
- https://github.com/foundry-rs/foundry/issues/7443

sveitser: @alysiahuggins https://github.com/foundry-rs/foundry/pull/7444 to fix this issue has been merged. We can temporarily switch to using nightly foundry until this fix lands in the monthly release in order to avoid having to have two foundry profiles.

In about 15 hours [foundry.nix](https://github.com/shazow/foundry.nix) should be updated and at that point we can also use a foundry version with this fix in flake.nix by changing

https://github.com/EspressoSystems/espresso-sequencer/blob/29edcb1cf2ebfa4c54a1773cec07844d2d1fb5e6/flake.nix#L26-L27

to 

```
  inputs.foundry.url = "github:shazow/foundry.nix/main";
```

The CI is already using nightly so shouldn't need to update any of the workflows.

sveitser:
> Is the plan still to have a single script that upgrades both light client and fee contract? Or shall this be done in another PR?

@philippecamacho I think it would be better to have a bash script for the deployment instead of having it (only) in the readme so there are less manual steps involved in doing a deployment. The script is also less likely to go out of date compared to the readme.

alysiahuggins:
> Is the plan still to have a single script that upgrades both light client and fee contract? Or shall this be done in another PR?

is it that you want it deployed in the same transaction or the same script?

philippecamacho:
> > Is the plan still to have a single script that upgrades both light client and fee contract? Or shall this be done in another PR?
> 
> is it that you want it deployed in the same transaction or the same script?

Yes, that is what I had in mind when writing ticket #999 ("consider a deployment that executes in one transaction and fails if the test cases aren't satisfied").  I am not sure  how forge deployment scripts work in practice though. Maybe if we use a single `deploy` function for both contracts we get atomicity for free?

philippecamacho:
> > Is the plan still to have a single script that upgrades both light client and fee contract? Or shall this be done in another PR?
> 
> @philippecamacho I think it would be better to have a script for the deployment instead of having it (only) in the readme so there are less manual steps involved in doing a deployment. The script is also less likely to go out of date compared to the readme.

I agree.

sveitser:
I think there's not straightforward way to batch transactions but we could look into https://github.com/ind-igo/forge-safe for example. It's quite a big change for this PR though so if we do it let's do it in another PR.

sveitser:
So I made some changes locally to use nightly foundry and remove the need for a second foundry profile however there's the additional issue that setting `build_info = true` disables the cache so we probably don't want to turn it on by default anyway. 

It's also not ideal that running `forge script` twice compiles everything twice during deployment which is a consequence of this. This would be avoided if we made script that combines the scripts to deploy the light client and fee contract if that works.

Unfortunately there also doesn't seem to be a way to run `forge build --build-info` once and then not have `forge script` recompile.

philippecamacho:
> I think there's not straightforward way to batch transactions but we could look into https://github.com/ind-igo/forge-safe for example. It's quite a big change for this PR though so if we do it let's do it in another PR.

I agree, if this is too complicated let us do it in another PR or even reevaluate if we want to do this in this sprint. In particular let us sync again with people in charge of the broader task of making HotShot upgradeable.

contracts/test/FeeContract.t.sol
alysiahuggins: thanks @philippecamacho!!

blowme8888 (commented):
Lol

contracts/src/LightClient.sol
alysiahuggins: This is the error that Openzeppelin Defender returns:
```Error: License type UNLICENSED is not a valid SPDX license identifier for block explorer verification. Specify a valid license type.```

The rationale from threads like these https://forum.openzeppelin.com/t/solidity-0-6-8-introduces-spdx-license-identifiers/2859 suggest that the inclusion of valid license idenfiers (https://spdx.org/licenses/) is needed for verification of contracts on websites like Etherscan.

File: .gitmodules (hunk 1 above)
Comment (alysiahuggins) [reply .gitmodules]: https://github.com/EspressoSystems/espresso-sequencer/commit/8dc1578d530709c54c1ba2c345e5156b63f8db91

File: foundry.toml [hunk 9]
```diff
@@ -17,6 +17,32 @@ solc = "0.8.23"
...
+# Version should match the solc installed via flake, otherwise the contract
+# artifacts may differ slightly.
+solc = "0.8.23"
+# The bytecode hash is no longer deterministic and prevents us from generating
+# deterministic bindings. The "none" setting prevents solc from appending the
+# IPFS hash to the bytecode.
+bytecode_hash = "none"
+ffi = true
+# needed for openzeppelin defender
+build_info = true
```
sveitser, line 42: So my suggestion is to add `extra_output` and `fs_permissions` to the default profile and then run `forge script` with `--build-info` when we do a defender deployment. We can also keep the `defender` profile. Note that it should not be necessary to duplicate the entire `default` profile because all other profiles inherit values from the default profile.

contracts/script/FeeContractWithDefender.s.sol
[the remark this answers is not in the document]
alysiahuggins: the proxy Address is now retrieved from the output logs. see https://github.com/EspressoSystems/espresso-sequencer/commit/a4d119f2de16927fc7fc64fcd2d381d561983f16 and 
https://github.com/EspressoSystems/espresso-sequencer/commit/e458801a694c20792712fc799516aa91d5669557

foundry.toml
alysiahuggins: I've removed duplications in the defender profile: https://github.com/EspressoSystems/espresso-sequencer/commit/d17dac1a2bffb30144de9584a30cf39bcd2668ea. 

Using --build-info true in the script command works but defender also requires the files to be built in `out` instead of `contracts/out` which is the current setting in the default profile. 

Do you have a preference for where the build files go? the `gen-bindings` currently uses contracts/out

alysiahuggins:
> > I think there's not straightforward way to batch transactions but we could look into https://github.com/ind-igo/forge-safe for example. It's quite a big change for this PR though so if we do it let's do it in another PR.
> 
> I agree, if this is too complicated let us do it in another PR or even reevaluate if we want to do this in this sprint. For this purpose we can sync again with the people in charge of the broader task of making HotShot upgradeable.

Issue created here https://github.com/EspressoSystems/espresso-sequencer/issues/1287

alysiahuggins:
an issue for creating a local testing environment for openzeppelin defender https://github.com/EspressoSystems/espresso-sequencer/issues/1251

blowme8888:
Thanks

philippecamacho (approved):
LGTM

PR #1249, opened 2024-03-16
[DRAFT DRAFT] Builder Executable

move47:


sequencer/src/context.rs
```diff
@@ -67,6 +74,7 @@ impl<N: network::Type, Ver: StaticVersionType + 'static> SequencerContext<N, Ver
         state_relay_server: Option<Url>,
         metrics: &dyn Metrics,
         node_id: u64,
+        events_streaming_server: Option<Url>,
```
jbearer, line 77: Nit: if this is just the URL to listen on, we usually would just make it a port (`u16`) and then do something like `app.serve(format!("0.0.0.0:{port}"))`. Any particular reason it needs to be a full `Url` (this allows the caller to pass in weird things that don't make sense for a server)

sequencer/src/context.rs
```diff
@@ -111,7 +124,14 @@ impl<N: network::Type, Ver: StaticVersionType + 'static> SequencerContext<N, Ver
             state_signer = state_signer.with_relay_server(url);
         }
 
-        Ok(Self::new(handle, persistence, node_id, state_signer))
+        Ok(Self::new(
+            handle,
+            persistence,
+            node_id,
+            state_signer,
+            Some(event_streamer),
```
jbearer, line 132: Why use `Option` if this is unconditionally `Some`?

sequencer/src/context.rs
```diff
@@ -132,8 +155,21 @@ impl<N: network::Type, Ver: StaticVersionType + 'static> SequencerContext<N, Ver
...
+                persistence,
+                ctx.state_signer.clone(),
+                event_streamer.clone(),
+            ),
         );
+        // spwan the task for running the event streaming api
+        if let Some(event_streamer) = event_streamer {
+            ctx.spawn(
+                "event streaming api",
+                run_hotshot_event_streaming_api(events_streaming_server.unwrap(), event_streamer),
```
jbearer, line 169: From the implementation of `init` it looks like `event_streamer` could be `Some` even when `events_streaming_server` is `None`. Is it possible for this `unwrap` to panic?

sequencer/src/context.rs
```diff
@@ -241,5 +278,28 @@ async fn handle_events<Ver: StaticVersionType>(
...
+        Version01,
+    >(&EventStreamingApiOptioins::default())
+    .expect("Failed to define hotshot eventsAPI");
+
+    let mut app = App::<_, EventStreamApiError, Version01>::with_state(source);
+
+    app.register_module("hotshot_events", hotshot_events_api)
+        .expect("Failed to register hotshot events API");
+
+    spawn(app.serve(url, STATIC_VER_0_1));
```
jbearer, line 304: We should be using `ctx.spawn` here so we don't leak the task

jbearer (commented):
Why did we make the events streaming service an entirely separate server, rather than just making it an additional sub-module of the main query API? That might require a bit less setup and would be more convenient for clients who could then use the same URL for both.

See how other API modules are hooked up in `api/options.rs`

nyospe:
> Why did we make the events streaming service an entirely separate server?

It's hooked up to the internal consensus, rather than the outcome on Decide. It is completely unusable with nodes that aren't part of the DA committee, has zero need for caching, and must be real-time (e.g. no benefit to being able to fetch results from peers).

jbearer:
We already have a bunch of modules under `api/options` that can be independently selected at runtime. They don't all hook into the query service storage and fetching, e.g. the `submit` module only requires a `SystemContextHandle`. And they all run under the same `App`

builder/src/lib.rs
```diff
@@ -0,0 +1,1317 @@
...
+    pub hotshot_handle: Consensus<N>,
+
+    /// Index of this sequencer node
+    pub node_index: u64,
+
+    /// Context for generating state signatures.
+    state_signer: Arc<StateSigner<Ver>>,
+
+    /// An orchestrator to wait for before starting consensus.
+    pub wait_for_orchestrator: Option<Arc<OrchestratorClient>>,
```
nyospe, line 75: Is this for the permissioned instance? Because I don't think we can do this for the permissionless...

move47: Yes, it is for permissioned instance. A permission node would do this `init_node()` call. It is similar to `init_node` in sequencer but without persistence.

builder/src/non_permissioned.rs
```diff
@@ -0,0 +1,377 @@
...
+        // Start Client for the event streaming api
+        let client = Client::<EventStreamApiError, Version01>::new(hotshot_events_api_url.clone());
+
+        assert!(client.connect(Some(Duration::from_secs(60))).await);
+
+        tracing::info!("Builder client connected to the hotshot events api");
+
+        // client subscrive to hotshot events
+        let subscribed_events = client
+            .socket("hotshot_events/events")
```
nyospe, line 145: Minor modification: surf-disco default for streaming is json; for performance reasons, add this line:

```rust
            .socket("hotshot_events/events")
            .header(ACCEPT, "application/octet-stream")
            .subscribe::<BuilderEvent<SeqTypes>>()
```

sequencer/src/context.rs
nyospe: Let's fix this one.

builder/src/non_permissioned.rs
move47: Done!

sequencer/src/context.rs
move47: d1ba612c74d0131e4b06907bd3b5dd3b96aad345

sequencer/src/context.rs
move47: d1ba612c74d0131e4b06907bd3b5dd3b96aad345

sequencer/src/context.rs
move47: d1ba612c74d0131e4b06907bd3b5dd3b96aad345

sequencer/src/context.rs
move47: d1ba612c74d0131e4b06907bd3b5dd3b96aad345

.vscode/settings.json.example
nyospe, line 1: Why did you remove this?

Cargo.toml
```diff
@@ -43,13 +44,17 @@ ethers = { version = "2.0", features = ["solc"] }
 futures = "0.3"
 hotshot = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.28" }
 hotshot-orchestrator = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.28" }
-hotshot-query-service = { git = "https://github.com/EspressoSystems/hotshot-query-service", tag = "0.1.0" }
+#hotshot-query-service = { git = "https://github.com/EspressoSystems/hotshot-query-service", tag = "0.1.0" }
```
nyospe, line 47: to be fixed before making non-draft... including removing all commented out lines.

builder/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "builder"
+description = "A standalone builder service"
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+hotshot-builder-core = { git = "https://github.com/EspressoSystems/hs-builder-core", branch = "hg/test_builder_executable" }
```
nyospe, line 9: And the one below absolutely should be removed.

builder/src/bin/permissioned_builder.rs
nyospe, line 1: Let's rename these to permissioned-builder.rs and permissionless-builder.rs

builder/src/bin/permissioned_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+        builder_pub_key,
+        private_staking_key,
+        bootstrapped_view,
+        opt.channel_capacity,
+        sequencer_version,
+    )
+    .await?;
+
+    // Start doing consensus.
+    ctx.start_consensus().await;
```
nyospe, line 233: 👍  I believe this order is correct.

builder/src/lib.rs
```diff
@@ -0,0 +1,732 @@
...
+    app.register_module("block_info", builder_api)
+        .expect("Failed to register the builder API");
+
+    app.register_module("txn_submit", private_mempool_api)
+        .expect("Failed to register the private mempool API");
+
+    async_spawn(app.serve(url, STATIC_VER_0_1));
+}
+
+#[cfg(test)]
```
nyospe, line 102: A lot of this feels like it belongs in non_permissioned.rs or permissioned.rs

It's okay, the non-test contents of lib.rs fall within the reasonable terseness, but it makes the file look huge on first glance.

sequencer/src/api/options.rs
```diff
@@ -169,6 +181,9 @@ impl Options {
                 App::<_, Error, Ver>::with_state(RwLock::new(super::State::from(&context)));
 
             self.init_hotshot_modules(&mut app)?;
+
+            self.init_hotshot_event_streaming_module(&mut app)?;
```
nyospe, line 185: Is there a check on the options I missed?

sequencer/src/api.rs
```diff
@@ -32,9 +35,20 @@ impl<N: network::Type, Ver: StaticVersionType + 'static> From<&SequencerContext<
...
     }
 }
+use futures::stream::BoxStream;
+
+#[async_trait]
+impl<N: network::Type, Ver: StaticVersionType> EventsSource<SeqTypes> for State<N, Ver> {
+    type EventStream = BoxStream<'static, Arc<BuilderEvent<SeqTypes>>>;
+
+    async fn get_event_stream(&self) -> Self::EventStream {
+        self.events_streamer.write().await.get_event_stream().await
```
jbearer, line 49: This seems like it could/should be `read()` instead of `write()`

sequencer/src/api/options.rs
```diff
@@ -151,7 +161,9 @@ impl Options {
             let status_api = status::define_api(&Default::default(), bind_version)?;
             app.register_module("status", status_api)?;
 
+            // Initialise only here because Extensivle Data Source doesn't impl the EventsSource trait.
```
jbearer, line 164: This means we silently ignore the hotshot events module when other modules are present. Is this what the PR in hotshot-query-service is intended to fix?

jbearer (commented):
Think we need to add a new optional module in `sequencer/options.rs` so be passed on the command line to enable the `HotShotEvents` API option in `main.rs`. See how other optional modules work e.g. `query`

.vscode/settings.json.example
nyospe: ... rephrasing. Please put this back.

sequencer/src/api/options.rs
```diff
@@ -220,7 +235,6 @@ impl Options {
         app.register_module("availability", endpoints::availability(bind_version)?)?;
         app.register_module("node", endpoints::node(bind_version)?)?;
 
-        self.init_hotshot_modules(&mut app)?;
```
nyospe, line 223: ????

sequencer/src/api/options.rs
```diff
@@ -151,7 +161,9 @@ impl Options {
             let status_api = status::define_api(&Default::default(), bind_version)?;
             app.register_module("status", status_api)?;
 
+            // Initialise only here because Extensivle Data Source doesn't impl the EventsSource trait.
             self.init_hotshot_modules(&mut app)?;
+
```
nyospe, line 166: Why no invocation here?

sequencer/src/api/options.rs
```diff
@@ -182,12 +205,32 @@ impl Options {
 
             self.init_hotshot_modules(&mut app)?;
 
-            self.init_hotshot_event_streaming_module(&mut app)?;
-
             context.spawn(
                 "API server",
                 app.serve(format!("0.0.0.0:{}", self.http.port), bind_version),
             );
+            // Start the event streaming API server if it is enabled.
```
jbearer, line 211: Can we make this a function to avoid copypasta?

sequencer/src/api/options.rs
```diff
@@ -168,6 +168,29 @@ impl Options {
                 "API server",
                 app.serve(format!("0.0.0.0:{}", self.http.port), bind_version),
             );
+
+            // It runs to different port and app because State and Extensible Data source needs to support required
```
jbearer, line 171: Do we not initialize this in the previous case or am I missing it (`if let Some(query_opt) = self.query.take() {`)?

jbearer (commented):
Ok, I'm satisfied with this handling of the API configuration (modulo a couple comments)

sequencer/src/options.rs
```diff
@@ -364,4 +372,5 @@ pub struct Modules {
     pub status: Option<api::options::Status>,
     pub state: Option<api::options::State>,
     pub catchup: Option<api::options::Catchup>,
+    pub hotshot_events: Option<api::options::HotshotEvents>,
```
jbearer, line 375: Need to handle this in `main.rs`

sequencer/src/context.rs
jbearer: This is still `Option` and seems like it doesn't need to be

jbearer (approved):
`sequencer/` stuff looks good

nyospe (approved):
I think that's everything...

PR #1252, opened 2024-03-18
Check deposits inserted when building header

sveitser:


sequencer/src/header.rs
```diff
@@ -600,6 +585,56 @@ mod test_headers {
...
+    #[test]
+    fn test_new_header_deposits_many() {
+        TestCase {
+            l1_deposits: [
+                (Address::default(), 1),
+                (Address::default(), 2),
+                (Address::random(), 3),
+            ]
+            .iter()
+            .map(|(address, amount)| FeeInfo::new(*address, *amount))
```
tbro, line 606: I feel like this could become a common pattern. Does it warrant implementing `From<(Address, u64) for FeeInfo`?

sveitser: Was also wondering. It might. I think we can add it later if we really need it.

PR #1261, opened 2024-03-19
feat: new method `Payload::namespace`

### LINKED ISSUE #1171: Add `Payload::namespace` interface
Sometimes it is useful to get a namespace without any proof. For example we may have already verified the entire block payload and just want to extract a certain namespace. Without generating the proof, we may be able to have a faster implementation, and a leaner interface (for example maybe there is no need to take VID common as a parameter)

ggutoski:
close #1171 

Minimalist implementation with basic correctness test.

### DIFF
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -96,14 +96,17 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
         0..self.ns_table.len()
     }
 
-    /// Returns the flat bytes for namespace `ns_id`.
-    pub fn namespace(&self, ns_id: NamespaceId) -> Option<&[u8]> {
+    /// Returns the list of txs for namespace `ns_id`.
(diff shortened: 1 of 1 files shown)


sequencer/src/block/payload.rs
```diff
@@ -96,6 +96,16 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
         0..self.ns_table.len()
     }
 
+    /// Returns the flat bytes for namespace `ns_id`.
+    pub fn namespace(&self, ns_id: NamespaceId) -> Option<&[u8]> {
```
jbearer, line 100: I feel like this should return `Vec<Transaction>` and do the parsing for you

ggutoski: done in 77c3e717cb0bcb2f80e6c25d3aa7158e64c07f4c

jbearer (approved):
Nice!

PR #1268, opened 2024-03-20
Add VBS and Merklized State support

imabdulbasit:
This PR:
- adds VBS support (https://github.com/EspressoSystems/espresso-sequencer/pull/1189) 
- Merklized State storage and queries (https://github.com/EspressoSystems/hotshot-query-service/pull/473)


sequencer/api/migrations/V14__state_tables.sql
```diff
@@ -0,0 +1,37 @@
...
+  CONSTRAINT fee_merkle_tree_pk PRIMARY KEY (pos, created);
+
+CREATE INDEX fee_merkle_tree_path ON fee_merkle_tree USING GIST (pos);
+
+CREATE TABLE block_merkle_tree (
+  pos LTREE NOT NULL, 
+  created BIGINT NOT NULL, 
+  hash_id INT NOT NULL REFERENCES hash (id), 
+  children INT[], 
+  children_bitvec BIT(3),
```
jbearer, line 27: Why 3? Isn't this tree binary?

sequencer/src/api.rs
```diff
@@ -25,7 +25,9 @@ struct State<N: network::Type, Ver: StaticVersionType> {
     handle: SystemContextHandle<SeqTypes, Node<N>>,
 }
 
-impl<N: network::Type, Ver: StaticVersionType> From<&SequencerContext<N, Ver>> for State<N, Ver> {
+impl<N: network::Type, Ver: StaticVersionType + 'static> From<&SequencerContext<N, Ver>>
```
jbearer, line 28: Not a big deal, but I feel like `StaticVersionType` might as well require `'static` in the trait definition. This would need to be a change in the vbs crate so we don't need to fix it right now, but cc @nyospe

sequencer/src/api/options.rs
```diff
@@ -31,6 +38,7 @@ pub struct Options {
     pub submit: Option<Submit>,
     pub status: Option<Status>,
     pub state: Option<State>,
+    pub merklized_state: Option<MerklizedState>,
```
jbearer, line 41: IMO: rename this to `state` and rename  `state` to `catchup` which is a better name for it anyways

sequencer/src/api/sql.rs
```diff
@@ -17,6 +31,80 @@ impl SequencerDataSource for DataSource {
...
+            blocks_delta,
+            fees_delta,
+        } = delta.as_ref();
+
+        // Insert block merkle tree nodes
+        for delta in blocks_delta {
+            let (_, proof) = block_merkle_tree
+                .lookup(delta)
+                .expect_ok()
+                .context("Index not found in block merkle tree")?;
```
jbearer, line 76: I don't think deltas should ever be missing, but still I don't think we should panic if they are. Better to return an error, some state will be missing but we can recover from that once we implement state catchup.

sequencer/src/api/sql.rs
```diff
@@ -17,6 +31,80 @@ impl SequencerDataSource for DataSource {
...
+            traversal_path,
+            block_number,
+        )
+        .await
+        .context("failed to insert merkle nodes! ")
+    }
+}
+
+#[async_trait]
+impl<D: SequencerDataSource + Send + Sync> UpdateStateStorage<SeqTypes, D> for ValidatedState {
```
jbearer, line 53: This implementation has nothing specific to SQL so I think it's confusing to be in this file. Might make more sense where `SequencerDataSource` is defined

sequencer/src/api/update.rs
```diff
@@ -47,3 +58,68 @@ where
 
     Ok(())
 }
+
+#[async_trait]
+impl<N, D, Ver: StaticVersionType> UpdateStateStorage<SeqTypes, StorageState<N, D, Ver>>
```
jbearer, line 63: This is defined twice?

sequencer/src/lib.rs
```diff
@@ -133,33 +138,67 @@ pub type PrivKey = <PubKey as SignatureKey>::PrivateKey;
...
     type QuorumNetwork = N::QuorumChannel;
     type CommitteeNetwork = N::DAChannel;
-    type Storage = SeqDAStorage;
+    type Storage = ToBeReplacedStorage<SeqTypes>;
 }
 
-#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
-pub struct Delta {}
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct Delta {
```
jbearer, line 198: This should probably be defined near `ValidatedState` since it's closely related and involved in validation

sequencer/src/lib.rs
```diff
@@ -133,33 +138,67 @@ pub type PrivKey = <PubKey as SignatureKey>::PrivateKey;
...
+    blocks_delta: HashSet<u64>,
```
jbearer, line 199: Technically there can only be one block delta per block. Now that I think about it, we don't even necessarily need to record this explicitly. I believe we can figure out which block changed when it's time to update the state, it's just `block_height - 1`.

sequencer/src/lib.rs
```diff
@@ -461,6 +500,9 @@ pub mod testing {
                 da_staked_committee_size: num_nodes,
                 da_non_staked_committee_size: 0,
                 my_own_validator_config: Default::default(),
+                // ???
+                view_sync_timeout: Duration::from_secs(1),
+                data_request_delay: Duration::from_secs(1),
```
jbearer, line 505: For testing purposes I think this is fine 👍

sequencer/src/state.rs
```diff
@@ -271,6 +292,9 @@ fn validate_and_apply_header(
         tracing::warn!("Invalid Builder: {}", e);
         return Err(BlockError::InvalidBlockHeader);
     };
+
+    delta.fees_delta.insert(proposed_header.fee_info.account);
```
jbearer, line 296: Nit: for deposits in `validate_and_apply_proposal`, we update `fees_delta` in the callee, but here for `validate_and_charge_builder` it's left to the caller. Doesn't really matter but I think it's less confusing if we are consistent here.

sequencer/api/migrations/V14__state_tables.sql
imabdulbasit: uses [LightWeightSHA3MerkleTree](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/prelude/type.LightWeightSHA3MerkleTree.html) which  Arity = U3 (

sequencer/src/api/sql.rs
imabdulbasit: it never panics as the update function in query service just logs the errors

sequencer/src/api/update.rs
imabdulbasit: Oh yeah this should be a wrapper impl which uses the inner data source

sequencer/api/migrations/V14__state_tables.sql
jbearer: lmao why did we do that

sequencer/src/api/sql.rs
jbearer: Oh sorry, I got confused by `expect_ok()` (because `expect()` would be a panic. Disregrad

nyospe:
We might as well do the commit directly from this branch to main, rather than through this.

nyospe (approved):
We could do this... or we could just merge your branch to main.

imabdulbasit (commented):
can we merge this PR and add Storage in a separate PR? @jbearer @sveitser

sequencer/src/api/endpoints.rs
```diff
@@ -189,13 +202,15 @@ where
     Ok(api)
 }
 
-pub(super) fn state<S>() -> anyhow::Result<Api<S, Error>>
+pub(super) fn state<S, Ver: StaticVersionType + 'static>(
```
jbearer, line 205: Rename to catchup?

jbearer (commented):
We need to add a new optional module for catchup (we already had one for state, which now applies to merklized state) in `sequencer/src/options.rs`, and check it in `main.rs` to enable the `catchup` API when appropriate

jbearer:
> can we merge this PR and add [Storage](https://hotshot.docs.espressosys.com/hotshot_types/traits/storage/trait.Storage.html) in a separate PR? @jbearer @sveitser

In principle I'm fine with this but we shouldn't break the build/demo in the meantime. If you can find a quick way to amke things generally work without implementing Storage I'm fine with that.

nyospe:
closes https://github.com/EspressoSystems/versioned-binary-serialization/issues/8

sequencer/src/api/options.rs
```diff
@@ -202,26 +296,30 @@ impl Options {
...
             app.register_module("submit", submit_api)?;
         }
 
         // Initialize state API.
-        if self.state.is_some() {
+        if self.catchup.is_some() {
             tracing::info!("initializing state API");
-            let state_api = endpoints::state()?;
+            let state_api = endpoints::state(bind_version)?;
             app.register_module("state", state_api)?;
```
jbearer, line 319: Rename to `catchup`?

sequencer/src/options.rs
```diff
@@ -277,6 +277,7 @@ module!("query", api::options::Query, requires: "http");
 module!("submit", api::options::Submit, requires: "http");
 module!("status", api::options::Status, requires: "http");
 module!("state", api::options::State, requires: "http");
```
jbearer, line 279: Maybe add "sql-storage" to the requires list, since this module is ignored if we have file system storage

jbearer (approved):

PR #1269, opened 2024-03-20
Fix Docker demo

jbearer:
* Fix typo in deploy image path
* Make deploy script runnable outside of the source tree


jbearer:
Need to think about this, since the bindings are checked into version control we have been able to build without compiling the contracts first, but this breaks that. Maybe it is better to read the bytecode object at runtime and just make sure it gets copied into the Docker image for the deploy script

sveitser (approved):

sequencer/src/bin/deploy.rs
```diff
@@ -295,11 +295,15 @@ async fn deploy_light_client_contract<M: Middleware + 'static>(
...
         .await?;
 
-    // Link with LightClient's bytecode artifacts
-    let bytecode_path = Path::new(env!("CARGO_MANIFEST_DIR"))
-        .join("../contracts/out/LightClient.sol/LightClient.json");
-    let json: Value = serde_json::from_reader(BufReader::new(File::open(bytecode_path)?))?;
-
+    // Link with LightClient's bytecode artifacts. We include the unlinked bytecode for the contract
+    // in this binary so that the contract artifacts do not have to be distributed with the binary.
+    // This should be fine because if the bindings we are importing are up to date, so should be the
```
sveitser, line 300: I think currently we don't always have the contract artifacts when we run `cargo test` etc. because everything that cargo test needs is in the bindings. If we want this to work we either need to check in the artifacts or always compile the contracts before we run tests.

sveitser: Didn't see your comment.

jbearer: Yeah, I ended up checking in the artifacts (just the one that is needed for the build, for now). My other idea, loading the artifacts at runtime and making sure they got copied into the docker image, ended up being a bigger headache because there's like 5 different places where we build docker images (we should fix that eventually)

PR #1283, opened 2024-03-27
Fix docker demo

nomaxg:


docker-compose.yaml
```diff
@@ -147,8 +147,7 @@ services:
     image: ghcr.io/espressosystems/espresso-sequencer/sequencer:main
     ports:
       - "$ESPRESSO_SEQUENCER1_API_PORT:$ESPRESSO_SEQUENCER_API_PORT"
-    # Run the availability API only, using postgres for persistence
-    command: sequencer -- storage-sql -- http -- query -- state
+    command: sequencer -- http -- query -- catchup -- status
```
jbearer, line 150: Why did we get rid of SQL here?

jbearer (changes requested):

docker-compose.yaml
nomaxg: switched back

jbearer (approved):

PR #1284, opened 2024-03-27
Workflow that fails if process compose is edited but docker compose isn't

nomaxg:


jbearer (approved):
Maybe have it print something like "If this was intentional, you may ignore this failure". Since in some cases you might change something in the native demo that doesn't affect the docker demo

PR #1291, opened 2024-04-01
Implement storage for fs and sql persistence

imabdulbasit:


sequencer/src/api/sql.rs
```diff
@@ -62,7 +65,7 @@ mod impl_testable_data_source {
     #[async_trait]
     impl TestableSequencerDataSource for DataSource {
         type Storage = TmpDb;
-        type Persistence = SqlStorage;
+        type Persistence = Arc<RwLock<SqlStorage>>;
```
jbearer, line 68: Curious why we needed this change? I thought the data source would always be `RwLock`ed at a higher level

sequencer/src/persistence/fs.rs
```diff
@@ -53,6 +64,113 @@ impl Persistence {
...
+            .open(file_path)?;
+
+        let proposal_bytes = bincode::serialize(&proposal).context("serialize proposal")?;
+
+        file.write_all(&proposal_bytes)?;
+
+        Ok(())
+    }
+    async fn record_action(&self, view: ViewNumber, action: HotShotAction) -> anyhow::Result<()> {
+        if let HotShotAction::Vote = action {
```
jbearer, line 128: Ah, so the `voted_view` stuff was from before we realized that there were other attributable actions. We want to store the latest view in which we have taken _any_ attributable action, so we should just store the view number here regardless of the type of action, and change `load_voted_view` to `load_latest_acted_view` or something

sequencer/src/persistence/fs.rs
```diff
@@ -53,6 +64,113 @@ impl Persistence {
...
+            .create(true)
+            .open(file_path)?;
+
+        let proposal_bytes = bincode::serialize(&proposal).context("serialize proposal")?;
+
+        file.write_all(&proposal_bytes)?;
+
+        Ok(())
+    }
+    async fn record_action(&self, view: ViewNumber, action: HotShotAction) -> anyhow::Result<()> {
```
jbearer, line 127: I don't love this taking `&self`. The update is non-atomic (first check that the view is newer than the old one, then overwrite the file), so we were relying on the fact that there was only one task in the node writing to this file at a time, which in turn was enforced by `&mut self`. Without that, we may need a `Mutex` protecting this functionality somehow.

jbearer: I'm beginning to wonder if we should implement the HotShot `Storage` trait on `RwLock<SequencerPersistence>` rather than directly on `SequencerPersistence`. This would fix the issue of `SequencerPersistence` requiring `Clone` and the awkwardness of setting `Storage = Arc<RwLock<Storage>>` in the testable data source

sequencer/src/persistence/fs.rs
```diff
@@ -53,6 +64,113 @@ impl Persistence {
...
+                    return Ok(());
+                }
+            }
+            fs::write(self.voted_view_path(), view.get_u64().to_le_bytes())?;
+        }
+
+        Ok(())
+    }
+
+    async fn update_high_qc(&self, high_qc: QuorumCertificate<SeqTypes>) -> anyhow::Result<()> {
```
jbearer, line 140: Same here

sequencer/src/persistence/sql.rs
```diff
@@ -186,7 +196,8 @@ impl PersistenceOptions for Options {
     type Persistence = Persistence;
 
     async fn create(self) -> anyhow::Result<Persistence> {
-        SqlStorage::connect(self.try_into()?).await
+        let storage = SqlStorage::connect(self.try_into()?).await?;
+        Ok(Arc::new(RwLock::new(storage)))
```
jbearer, line 200: Like I think this is gonna result in a nested `RwLock` pair when we `RwLock` it again to share it into the web server state

sequencer/src/persistence.rs
```diff
@@ -220,20 +252,209 @@ mod persistence_tests {
...
+            QuorumCertificate::genesis()
+        );
+
+        // store qcs with different view number
+        qc.view_number += 1;
+        storage.update_high_qc(qc.clone()).await.unwrap();
+
+        qc.view_number += 1;
+        storage.update_high_qc(qc.clone()).await.unwrap();
+
```
jbearer, line 314: Check that `update_high_qc` with a lower view number doesn't update anything

sequencer/src/persistence/fs.rs
```diff
@@ -147,6 +177,169 @@ impl SequencerPersistence for Persistence {
...
+        let dir_path = self.da_dir_path();
+
+        if !dir_path.exists() {
+            return Ok(None);
+        }
+
+        let entries = fs::read_dir(dir_path)?;
+        let view = view.get_u64();
+
+        for entry in entries {
```
jbearer, line 208: Since we know the view number, just figure out the file name and open it directly, instead of searching

jbearer (approved):

PR #1294, opened 2024-04-01
[Sequencer] New network integration

### LINKED ISSUE #1299: Add new CDN and Libp2p implementations
The updated HotShot allows us to use both our new CDN and Libp2p in a combined fashion. This ticket is to integrate those into the sequencer.

rob-maron:
Closes #1299

# This PR:

TL;DR: Updates the sequencer to use the combined network (Libp2p and Push CDN).

Specifically:
- Removes code related to the original CDN
- Adds binaries, environment variables, and containers for the new CDN
- Adds environment variables and setup for the Libp2p implementation
- Has some miscellaneous HotShot-update-related items
- Removes Cap'n'Proto dependencies


sequencer/src/lib.rs
```diff
@@ -380,20 +397,46 @@ pub async fn init_node<Ver: StaticVersionType + 'static>(
...
+    let cdn_network = PushCdnNetwork::new(
+        network_params.cdn_endpoint,
+        vec!["Global".into(), "DA".into()],
+        KeyPair {
+            public_key: WrappedSignatureKey(my_config.public_key),
+            private_key: my_config.private_key.clone(),
+        },
+    )
+    .await
+    .with_context(|| "Failed to create CDN network")?;
```
rob-maron, line 410: I have in-progress CDN changes to make the connection be lazily initialized, but they aren't in yet.

sequencer/src/bin/cdn-broker.rs
```diff
@@ -0,0 +1,99 @@
...
+use sha2::Digest;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+/// The main component of the push CDN.
+struct Args {
+    /// The discovery client endpoint (including scheme) to connect to.
+    /// With the local discovery feature, this is a file path.
+    /// With the remote (redis) discovery feature, this is a redis URL (e.g. `redis://127.0.0.1:6789`).
+    #[arg(short, long)]
```
sveitser, line 22: Shouldn't we add env vars to enable configuration of the new binaries?

rob-maron: Good idea, yes, will do

sequencer/src/bin/cdn-broker.rs
```diff
@@ -0,0 +1,123 @@
...
+        long,
+        default_value = "0.0.0.0:1738",
+        env = "ESPRESSO_CDN_BROKER_PUBLIC_BIND_ADDRESS"
+    )]
+    public_bind_address: String,
+
+    /// The user-facing address to advertise
+    #[arg(
+        long,
+        default_value = "local_ip:1738",
```
sveitser, line 56: Took a brief look at the push cdn code. So `local_ip` is a magic value that will be resolved to a local IP? What will it choose if there are multiple network interfaces? Would we ever not want this to be set to `local_ip`?

rob-maron: Yessir, downstream it is resolved to whatever `local_ip::local_ip_address()` resolves to. I made it like this so it's easier to specify it if it's unknown and not set (e.g. in AWS).

If there are multiple networking interfaces, my thought is to use `0.0.0.0` to bind to all of them.  For example, we would want it (`public_advertise_address` specifically) to be `0.0.0.0` on a fully public deployment that does not use load balancers

rob-maron:
cc @imabdulbasit for builder-related changes

docker/cdn-broker.Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apt-get update \
+    &&  apt-get install -y curl libcurl4 wait-for-it tini \
+    &&  rm -rf /var/lib/apt/lists/*
+
+COPY target/$TARGETARCH/release/cdn-broker /bin/cdn-broker
+RUN chmod +x /bin/cdn-broker
+
+ENV RUST_LOG="info"
+
```
jbearer, line 13: Helthcheck

docker-compose.yaml
```diff
@@ -118,15 +129,16 @@ services:
     command: sequencer -- http -- query -- catchup -- status -- submit
     environment:
       - ESPRESSO_SEQUENCER_ORCHESTRATOR_URL
-      - ESPRESSO_SEQUENCER_DA_SERVER_URL
-      - ESPRESSO_SEQUENCER_CONSENSUS_SERVER_URL
+      - ESPRESSO_SEQUENCER_CDN_ENDPOINT
```
jbearer, line 132: Needs to depend on CDN

sequencer/src/bin/orchestrator.rs
```diff
@@ -190,12 +191,33 @@ async fn main() {
...
     };
+
+    // The Libp2p configuration
+    let libp2p_config = Libp2pConfig {
+        bootstrap_nodes: Vec::new(),
+        node_index: 0,
+        bootstrap_mesh_n_high: 4,
+        bootstrap_mesh_n_low: 4,
+        bootstrap_mesh_outbound_min: 2,
+        bootstrap_mesh_n: 4,
```
jbearer, line 202: envvrs

docker-compose.yaml
```diff
@@ -54,6 +54,8 @@ services:
 
   # The CDN system's main entry point; where users contact first.
   marshal-0:
+    environment:
+      RUST_LOG: info
```
jbearer, line 58: Could just do `RUST_LOG` as this is set in `.env`

docker-compose.yaml
```diff
@@ -154,6 +160,12 @@ services:
         condition: service_healthy
       state-relay-server:
         condition: service_healthy
+      broker-0:
+        condition: service_started
```
jbearer, line 164: Why not `service_healthy`?

sequencer/src/bin/cdn-broker.rs
```diff
@@ -80,15 +79,13 @@ struct Args {
...
-#[cfg_attr(async_executor_impl = "async-std", async_std::main)]
+#[async_std::main]
 async fn main() -> Result<()> {
     // Parse command line arguments
     let args = Args::parse();
 
     // Initialize tracing
-    setup_logging();
-    setup_backtrace();
+    tracing_subscriber::fmt::init();
```
jbearer, line 88: why?

rob-maron: It wasn't printing any logs to `stdout` until I changed this. I can investigate it further if needed

jbearer: not a big deal. Just weird because I think we use the `setup_logging()` thing in other binaries. We can keep it this way though

docker-compose.yaml
```diff
@@ -55,18 +55,28 @@ services:
...
     image: ghcr.io/espressosystems/espresso-sequencer/cdn-marshal:main
     command:
       - -d
       - redis://:changemeplease!!@keydb:6379
       - --bind-port 
       - $ESPRESSO_CDN_SERVER_PORT
+      - --metrics-enabled
+    healthcheck:
+      # Curl the metrics endpoint
+      test: "curl localhost:9090"
```
jbearer, line 68: Use the env var rather than hard-coded port

docker-compose.yaml
```diff
@@ -55,18 +55,28 @@ services:
...
-      RUST_LOG: info
+      - RUST_LOG
     image: ghcr.io/espressosystems/espresso-sequencer/cdn-marshal:main
     command:
       - -d
       - redis://:changemeplease!!@keydb:6379
       - --bind-port 
       - $ESPRESSO_CDN_SERVER_PORT
+      - --metrics-enabled
+    healthcheck:
```
jbearer, line 66: Better to put this healthcheck in the Dockerfile itself so we can still use it even with different orchestration

rob-maron: Are we also moving it out of here? My worry is that if we want to change the port there, we would have to rebuild the Dockerfile, so in the compose would not respect the environment variable. 

We could also change the compose to build the images instead of pulling them and then pass in the environment variable there.

jbearer: You can reference environment variables inside the Dockerfile in the healthcheck command, and if they are overridden at runtime the command will respect that

jbearer (approved):

PR #1296, opened 2024-04-02
add test for merklized state api

imabdulbasit:


nomaxg (commented):
Could we also verify the proof here just as a sanity check?

imabdulbasit:
> Could we also verify the proof here just as a sanity check? We actually found some subtle bugs like this during the VID integration.

We only get proof for the last inserted block for the BlockMerkleTree, so the test is not very reliable to compare proofs here. If I shut down the consensus, the event handler is also dropped, so the proof in the database for the last inserted block is not inserted most of the time.

jbearer (approved):

PR #1307, opened 2024-04-08
1175 make the light client contract permissioned

alysiahuggins:
Added two variables:
- `permissionedProver` which stores the address of the prover that is able to update the latest finalized state on the light client contract.
- `permissionedProverEnabled` a flag that determines when the contract is operating in permissionedProver mode

Added two methods:
- `setPermissionedProver(address)` - Sets the permissioned prover and turns on permissioned prover mode
- `disablePermissionedProver()` - turns off permissioned prover mode

Updated one method:
- `newFinalizedState(...)` - can only be executed by the `permissionedProver` when the contract is in permissioned prover mode (permissionedProverEnabled == true) otherwise, anyone can call it


hotshot-state-prover/src/service.rs
```diff
@@ -666,10 +685,14 @@ mod test {
 
         let anvil = Anvil::new().spawn();
         let (_wallet, contract) = deploy_contract_for_test(&anvil).await?;
+        print!("{:?}", _wallet.address());
         let mut config = StateProverConfig::default();
         config.update_l1_info(&anvil, contract.address());
         // sanity check on `config`
 
+        //set the approved prover
+        // let _ = contract.update_approved_prover(_wallet.clone().address()).send().await?;
```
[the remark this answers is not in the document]
alysiahuggins: ohh let me remove that comment, I update the approved prover in the `deploy_contract_for_test` function. https://github.com/EspressoSystems/espresso-sequencer/blob/c611cbffa7f89816567ccbc2527909349cd8e69b/hotshot-state-prover/src/service.rs#L624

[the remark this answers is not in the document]
alysiahuggins: stale comment removed https://github.com/EspressoSystems/espresso-sequencer/commit/6383e21e07537395af52200a76deab701ebb6a5b

contracts/test/mocks/DeployLightClientTestScript.sol
sveitser, line 1: Why is this file in the `mocks` directory?

flake.nix
```diff
@@ -162,7 +162,7 @@
               description = "Spell checking";
               entry = "typos";
               pass_filenames = true;
-              excludes = [ "contract-bindings/artifacts" ];
+              excludes = [ "contract-bindings" ];
```
sveitser, line 165: The excludes should all go into `.typos.toml` and I think can be deleted here.

justfile
```diff
@@ -86,10 +86,12 @@ sol-test:
...
 # Deploy contracts to local blockchain for development and testing
-dev-deploy url="http://localhost:8545" mnemonics="test test test test test test test test test test test junk" num_blocks_per_epoch="10" num_init_validators="5":
-    MNEMONICS="{{mnemonics}}" forge script contracts/test/LightClientTest.s.sol:DeployLightClientTestScript \
-    --sig "run(uint32 numBlocksPerEpoch, uint32 numInitValidators)" {{num_blocks_per_epoch}} {{num_init_validators}} \
-    --fork-url {{url}} --broadcast
+dev-deploy url="" mnemonics="" num_blocks_per_epoch="10" num_init_validators="5" admin="0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266" private_key="":
+    MNEMONICS="{{mnemonics}}" forge script contracts/test/mocks/DeployLightClientTestScript.sol:DeployLightClientTestScript \
+    --sig "run(uint32 numBlocksPerEpoch, uint64 numInitValidators, address owner)" {{num_blocks_per_epoch}} {{num_init_validators}} {{admin}} \
+    --fork-url {{url}} --broadcast \
+    --private-key {{private_key}}
```
sveitser, line 93: Does this work if no private key is specified?

To make optional arguments we can use `*args` like on line 11 and pass them to `forge script`.

```
down *args:
    docker compose down {{args}}
 ```

File: contracts/src/LightClient.sol [hunk 4]
```diff
@@ -107,6 +110,10 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     error InvalidProof();
     /// @notice Wrong stake table used, should match `finalizedState`
     error WrongStakeTableUsed();
+    /// @notice Invalid address
+    error InvalidAddress();
+    /// @notice Only an approved prover can perform this actionx
```
sveitser, line 115:
```suggestion
    /// @notice Only an approved prover can perform this action
```

justfile
alysiahuggins: It didn't without the private key for me, I got this error:
`No associated wallet for addresses: {0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266}. Unlocked wallets: []`

and used this solution https://github.com/Cyfrin/foundry-full-course-f23/discussions/866#discussioncomment-7000638

contracts/test/mocks/DeployLightClientTestScript.sol
alysiahuggins: Mostly because it was used to deploy the mock light client, i can put it in the parent folder, test, if that's preferred.

flake.nix
alysiahuggins: thanks for identifying this dupe, updated https://github.com/EspressoSystems/espresso-sequencer/commit/241141ba5e41883171da720b3e4741cb0aeec3b4

justfile
sveitser: Sorry I think I wasn't explicit enough.

I think currently the way this just command is written if one runs it without specifying a private key the command line will turn into `forge script ... --private-key ""` which I think is not a valid `forge script` command because  I would expect that `--private-key` expects a private key to follow and the empty string isn't a valid private key.

That's why I suggested putting `*args` there instead so that `dev-deploy` can accept additional arguments and hand them to `forge script` via `{{args}}`.

sveitser, line 93:
```suggestion
dev-deploy url="" mnemonics="" num_blocks_per_epoch="10" num_init_validators="5" admin="0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266" *args:
    MNEMONICS="{{mnemonics}}" forge script contracts/test/mocks/DeployLightClientTestScript.sol:DeployLightClientTestScript \
    --sig "run(uint32 numBlocksPerEpoch, uint64 numInitValidators, address owner)" {{num_blocks_per_epoch}} {{num_init_validators}} {{admin}} \
    --fork-url {{url}} --broadcast \
    {{args}}
```

alysiahuggins: Ok thank you for clarifying. I spoke to @alxiong about this too and we agreed to create a separate ticket to change the deployment implementation so we'll leave this as is for this ticket.

alysiahuggins:
A ticket has been created to cover the work required to call the newly added methods
`setPermissionedProverMode(bool)`
`updatePermissionedProver(address)` 
via Defender and a Multisig wallet https://github.com/EspressoSystems/espresso-sequencer/issues/1328

contracts/test/mocks/DeployLightClientTestScript.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/ffa2bcfc908135609f99cd22292ce92d7b9c2145, https://github.com/EspressoSystems/espresso-sequencer/commit/6f2b61f91a51e0619f6f237208482c7e60540c2e

contracts/src/LightClient.sol
```diff
@@ -37,6 +37,15 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     /// @notice upgrade event when the proxy updates the implementation it's pointing to
     event Upgrade(address implementation);
 
+    /// @notice the approved prover was updated
```
sveitser, line 40: Update the doc string to use "permissioned"?

contracts/src/LightClient.sol
```diff
@@ -292,4 +322,26 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+            revert InvalidAddress();
+        }
+        permissionedProver = _prover;
+        emit PermissionedProverUpdated(_prover);
+    }
+
+    /// @notice Set the permissioned prover mode
+    /// @dev when this is set to true, only the `permissionedProver` is able to execute the
+    /// `newFinalizedState` method
+    function setPermissionedProverMode(bool mode) public onlyOwner {
```
sveitser, line 339: Is it useful to allow setting the mode to the mode that it's already in? I feel like we don't need to emit an enabled/disabled event in case the mode wasn't changed.

contracts/test/DeployLightClientTestScript.s.sol
```diff
@@ -0,0 +1,74 @@
...
+import { LightClientTest as LCTest } from "./mocks/LightClientTest.sol";
+import { LightClient as LC } from "../src/LightClient.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+
+contract DeployLightClientTestScript is Script {
+    function run(uint32 numBlocksPerEpoch, uint64 numInitValidators, address owner)
+        external
+        returns (address payable proxyAddress, address admin, LC.LightClientState memory)
+    {
+        // TODO for a production deployment provide the right genesis state and value
```
sveitser, line 15: Do we still need this todo? If yes, would it be better to have a github issue since this file isn't something that would be used in production.

contracts/test/DeployLightClientTestScript.s.sol
```diff
@@ -0,0 +1,74 @@
...
+    /// @notice deploys the impl, proxy & initializes the impl
+    /// @return proxyAddress The address of the proxy
+    /// @return admin The address of the admin
+
+    function deployContract(
+        LC.LightClientState memory state,
+        uint32 numBlocksPerEpoch,
+        address owner
+    ) public returns (address payable proxyAddress, address admin, LC.LightClientState memory) {
+        // string memory seedPhrase = vm.envString("MNEMONIC");
```
sveitser, line 52: Wondering why these lines are commented out, but still here.

contracts/test/DeployLightClientTestScript.s.sol
```diff
@@ -0,0 +1,74 @@
...
+        bytes memory result = vm.ffi(cmds);
+        LC.LightClientState memory state = abi.decode(result, (LC.LightClientState));
+
+        return deployContract(state, numBlocksPerEpoch, owner);
+    }
+
+    /// @notice deploys the impl, proxy & initializes the impl
+    /// @return proxyAddress The address of the proxy
+    /// @return admin The address of the admin
+
```
sveitser, line 46: Also returns the light client state.

contracts/test/LightClient.t.sol
```diff
@@ -112,6 +142,124 @@ contract LightClient_constructor_Test is LightClientCommonTest {
...
+        //assert that the permissionedProver has not been set and thus the zero address
+        assert(newLc.permissionedProver() == address(0));
+
+        //any prover can call the newFinalizedState method as the contract is not in permissioned
+        // prover mode
+        vm.prank(makeAddr("randomUser"));
+        newLc.newFinalizedState(newState, newProof);
+    }
+
+    function test_RevertWhen_NoPermissionedProverSetInPermissionedMode() external {
```
sveitser, line 215: Maybe this has been discussed and I missed it but I wonder what the use case is for being able to enable permissioned prover mode without having a permissioned prover set. We could also make this impossible at the contract level.

contracts/test/LightClient.t.sol
```diff
@@ -112,6 +142,124 @@ contract LightClient_constructor_Test is LightClientCommonTest {
...
+        //set permissioned prover mode to true but do not set the permissionedProver address so that
+        // it's still address(0)
+        vm.expectEmit(true, true, true, true);
+        emit LC.PermissionedProverModeEnabled();
+        vm.prank(admin);
+        newLc.setPermissionedProverMode(true);
+
+        assert(newLc.permissionedProver() == address(0));
+
+        //any prover can call the newFinalizedState method as the contract is not in permissioned
```
sveitser, line 231: Update comment

contracts/src/LightClient.sol
```diff
@@ -107,6 +122,13 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     error InvalidProof();
     /// @notice Wrong stake table used, should match `finalizedState`
     error WrongStakeTableUsed();
+    /// @notice Invalid address
+    error InvalidAddress();
```
sveitser, line 126:
```suggestion
    error InvalidPermissionedProverAddress();
```
I feel like the name is a bit too non-descriptive now.

hotshot-state-prover/src/service.rs
```diff
@@ -666,6 +691,7 @@ mod test {
 
         let anvil = Anvil::new().spawn();
         let (_wallet, contract) = deploy_contract_for_test(&anvil).await?;
+        print!("{:?}", _wallet.address());
```
sveitser, line 694: Remove?

contracts/src/LightClient.sol
alysiahuggins: > I feel like the name is a bit too non-descriptive now.

I made it more general because using the zero address in most cases would not be valid. In the event that the code functionality expands and a check for a non-zero address needs to happen - the same error type can be used. Do you still prefer a very specific error message?

contracts/test/LightClient.t.sol
sveitser: In other words we could have a function to disable the permissioned mode and one to enable it that takes an address as an argument that can't be the zero address. Then it's no longer possible to "disable" the lightlclient contract. Currently it seems it's possible to set permissioned prover mode without a permissioned prover which in fact disables sending proofs. If having a disabled state is desired we should make it explict, if it's not desired I think it's best to make it less likely. It's of course still possible to set an address that doesn't exist.

cc @alysiahuggins @philippecamacho

contracts/src/LightClient.sol
sveitser: I do prefer it to be explicit but it's not exactly strong preference. I just think it makes debugging easier for anyone who ever sees the error. Leave it up to you.

contracts/test/LightClient.t.sol
alysiahuggins: i agree that you should not be able to update the permissioned prover mode in an instance with the prover address is not specified. I'll update the function.

justfile
alxiong: yes, there's already a ticket: https://github.com/EspressoSystems/espresso-sequencer/issues/1177

i'll notify you guys when I have a PR.

contracts/src/LightClient.sol
```diff
@@ -292,4 +322,26 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        }
+        permissionedProver = _prover;
+        emit PermissionedProverUpdated(_prover);
+    }
+
+    /// @notice Set the permissioned prover mode
+    /// @dev when this is set to true, only the `permissionedProver` is able to execute the
+    /// `newFinalizedState` method
+    function setPermissionedProverMode(bool mode) public onlyOwner {
+        permissionedProverMode = mode;
```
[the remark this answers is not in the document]
sveitser: There are some comments further up the page on changing the API.

hotshot-state-prover/src/service.rs
```diff
@@ -580,14 +580,23 @@ mod test {
     ) -> Result<(Arc<L1Wallet>, LightClient<L1Wallet>)> {
         let provider = Provider::<Http>::try_from(anvil.endpoint())?;
         let signer = Wallet::from(anvil.keys()[0].clone());
-        let l1_wallet = Arc::new(L1Wallet::new(provider.clone(), signer));
+        let client =
+            SignerMiddleware::new(provider.clone(), signer.with_chain_id(anvil.chain_id()));
+        let l1_wallet = Arc::new(client);
+        let wallet_address = format!("{:?}", l1_wallet.clone().address());
```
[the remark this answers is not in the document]
alysiahuggins: I recall that when i used that, it truncated the address

hotshot-state-prover/src/service.rs
```diff
@@ -580,14 +580,23 @@ mod test {
...
+
+        // deploy the light client contract via a proxy
         let output = Command::new("just")
             .arg("dev-deploy")
             .arg(anvil.endpoint())
             .arg(TEST_MNEMONIC)
             .arg(BLOCKS_PER_EPOCH.to_string())
             .arg(NUM_INIT_VALIDATORS.to_string())
+            .arg(wallet_address.clone())
+            .arg(private_key)
```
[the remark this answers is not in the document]
alysiahuggins: it complained about the account not being unlocked. e.g. No associated wallet for addresses: {0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266}. Unlocked wallets: []

contracts/src/LightClient.sol
```diff
@@ -292,4 +322,26 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
             )
         );
     }
+
+    /// @notice Update the address of the permissioned prover
+    /// @dev this address is only considered when the `permissionedProverMode` is set to true
+    function updatePermissionedProver(address _prover) public onlyOwner {
```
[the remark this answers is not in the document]
alysiahuggins: yes true, https://github.com/EspressoSystems/espresso-sequencer/commit/906339c6b20ef92630d2cee141e6ab9a408ae72d

contracts/test/LightClient.t.sol
```diff
@@ -260,16 +442,19 @@ contract LightClient_newFinalizedState_Test is LightClientCommonTest {
...
         // outdated view num
         vm.expectRevert(LC.OutdatedState.selector);
+        vm.prank(permissionedProver);
         lc.newFinalizedState(newState, proof);
 
         // outdated block height
         state.viewNum = genesis.viewNum;
         state.blockHeight = numBlockSkipped + 1;
         vm.expectRevert(LC.OutdatedState.selector);
+        vm.prank(permissionedProver);
```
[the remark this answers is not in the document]
alysiahuggins: yes that's cleaner, thank you! https://github.com/EspressoSystems/espresso-sequencer/commit/906339c6b20ef92630d2cee141e6ab9a408ae72d#diff-08d922b027356f5ca532486fcea6a356b593fadcc63440be9268b59f2503d5d8

hotshot-state-prover/src/service.rs
alysiahuggins: done

contracts/test/LightClient.t.sol
alysiahuggins: in this case the test does indeed check to see whether the contract is in permissioned mode and it's not

contracts/test/LightClient.t.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/c6ff57f7d67653b9bc5bb3bf69bda7f5609f779e

contracts/test/DeployLightClientTestScript.s.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/c6ff57f7d67653b9bc5bb3bf69bda7f5609f779e

contracts/test/DeployLightClientTestScript.s.sol
alysiahuggins: removed

contracts/src/LightClient.sol
alysiahuggins: yes that's true! addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/c6ff57f7d67653b9bc5bb3bf69bda7f5609f779e

contracts/src/LightClient.sol
alysiahuggins: done

hotshot-state-prover/src/service.rs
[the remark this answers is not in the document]
alysiahuggins: it complained about the account not being unlocked. e.g. No associated wallet for addresses: {0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266}. Unlocked wallets: []

hotshot-state-prover/src/service.rs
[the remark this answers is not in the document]
alysiahuggins: I recall that when i used that, it truncated the address, which made it invalid input for the method that calls it

contracts/src/LightClient.sol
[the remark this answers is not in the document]
alysiahuggins: @alxiong these are the comments @sveitser is referring to https://github.com/EspressoSystems/espresso-sequencer/pull/1307#discussion_r1568690412

contracts/test/DeployLightClientTestScript.s.sol
alysiahuggins: Yes we do, issue created here https://github.com/EspressoSystems/espresso-sequencer/issues/1346

contracts/test/LightClient.t.sol
```diff
@@ -112,6 +142,124 @@ contract LightClient_constructor_Test is LightClientCommonTest {
...
+        //set permissioned flag to false
+        vm.expectEmit(true, true, true, true);
+        emit LC.PermissionedProverModeDisabled();
+        vm.prank(admin);
+        lc.setPermissionedProverMode(false);
+
+        //any prover can call the newFinalizedState method as the contract is not in permissioned
+        // prover mode
+        vm.prank(makeAddr("randomUser"));
+        lc.newFinalizedState(newState, newProof);
```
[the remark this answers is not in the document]
alysiahuggins: good point - added those emits alongside other left out ones here https://github.com/EspressoSystems/espresso-sequencer/commit/c060c87c55a2dc6a13d91e1ee61b9304c9ef20f4

contracts/test/LightClient.t.sol
sveitser: I think setting the prover to zero address if permissioned mode is disabled is a good idea. I think the `update` function is fine, for the other one, I think we should split it in two to get rid of the boolean parameter and make it more obvious to the operator what's they are doing. E. g.

1. `enablePermissionedProverMode(address)`
2. `disablePermissionedProverMode()`

What do you think?

alysiahuggins: Yea that's clearer should reduce complexity for users, will do!

alysiahuggins: refactored here https://github.com/EspressoSystems/espresso-sequencer/commit/830ab2f0a835ba46a897cb429b68f5ed0bb57c98

contracts/src/LightClient.sol
```diff
@@ -292,4 +327,54 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        }
+    }
+
+    /// @notice set the permissionedProverMode to true and set the permissionedProver to the
+    /// non-zero address provided
+    /// @dev if it was already enabled (permissioneProverMode == true), then revert with
+    /// NoChangeRequired
+    function enablePermissionedProverMode(address prover) public onlyOwner {
+        if (!permissionedProverMode) {
+            if (prover == address(0) && permissionedProver == address(0)) {
```
philippecamacho, line 356: Why is the condition `prover == address(0)` not enough?

hotshot-state-prover/src/service.rs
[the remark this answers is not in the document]
sveitser: Yeah the default formatter gives something like `0x12...34` but `{:?}` and `{:#x}` give the full content with a hex prefix. I think `{:#x}` is a bit more correct to use when one doesn't just want to log the address but we currently use both in our code. There is also `ToHexExt::encode_hex_with_prefix()` which allows one to write `address().encode_hex_with_prefix()` but this is even longer than using `format!` so 🤷

contracts/src/LightClient.sol
sveitser: The two functions are now almost the same it's just that the `enable` function reverts if the permissioned prover mode is already enabled. I wonder if this is an important feature. My feeling is that it isn't and we can provide a single function to set the prover and enable permissioned prover mode (if it was previously disabled). cc @philippecamacho @alysiahuggins @alxiong

hotshot-state-prover/src/service.rs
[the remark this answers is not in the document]
sveitser: As mentioned, this will be removed soon anyway.

contracts/src/LightClient.sol
```diff
@@ -189,6 +216,14 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         LightClientState memory newState,
         IPlonkVerifier.PlonkProof memory proof
     ) external {
+        //revert if we're in permissionedProver mode and the permissioned prover has not been set
+        if (permissionedProverMode == true && msg.sender != permissionedProver) {
```
sveitser, line 220: Since we no longer allow permissioned prover mode to be true if there is no permissioned prover set we don't need to check if permissioned prover mode is enabled and checking `msg.sender` should be enough, shouldn't it?

alysiahuggins: very true, good catch and should hopefully save on a bit of gas fees

contracts/src/LightClient.sol
alysiahuggins: @philippecamacho , you're right, the first condition is enough. 

As far as reverting if the permissionedProver is already set - there's another function `updatePermissionedProver` which allows the admin to update the permissioned prover. So are you suggesting that we get rid of the `updatePermissionedProver` function? If so, I agree, this isn't a function that should be called that many times anyway, so having multiple functions to set the prover with different logging doesn't seem necessary.

alysiahuggins: consolidated the update and enable prover functions into one `setPermissionedProver` https://github.com/EspressoSystems/espresso-sequencer/commit/f521dac663d5563820273529913ebe88b86fa915

contracts/src/LightClient.sol
```diff
@@ -37,6 +37,12 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     /// @notice upgrade event when the proxy updates the implementation it's pointing to
     event Upgrade(address implementation);
 
+    /// @notice an permissioned prover is needed to interact `newFinalizedState`
```
philippecamacho, line 40:
```suggestion
    /// @notice a permissioned prover is needed to interact `newFinalizedState`
```

philippecamacho (approved):
LGTM

PR #1308, opened 2024-04-08
Generate only necessary bindings.

### LINKED ISSUE #1068: improve handling of forge bindings
**Goal**:  decide if bindings should be committed and if so, decide which folders it should be created for and write a `forge bind` command which handles it appropriately. 

**Issue**: Using this command `forge bind --contracts ./contracts/src --crate-name contract-bindings --bindings-path contract-bindings --overwrite --force`, I still get bindings generated for other folders within the `./contracts` folder. e.g. the `./contracts/demo` folder

**Context**:
what we definitely don't want:

test/ (esp test/mocks/ which look like normal contract)
scripts/
lib/ (esp forge's std library)

what we strongly prefer NOT to have:

demo code

philippecamacho:
Closes #1068.


alxiong (approved):
holy cow, finally!
the `-30k` diff makes me happy 👍

sveitser (approved):

zerosnacks:
Hey @philippecamacho, saw your mention in https://github.com/foundry-rs/foundry/issues/4375

We've also recently released Alloy `0.1`: https://www.paradigm.xyz/2024/06/alloy-release

As part of that we've also updated `forge bind` to use Alloy: https://github.com/foundry-rs/foundry/pull/7919

Would be great to get feedback on this by using `forge bind` with the `--alloy` flag

PR #1314, opened 2024-04-10
Update h2: 0.3.24 -> 0.3.26

sveitser:
https://rustsec.org/advisories/RUSTSEC-2024-0332.html


sveitser:
The problem is that `commit` is renamed to `committable` and we should use the crates.io version once dependencies have also switched.

PR #1316, opened 2024-04-10
 merklized state catchup

imabdulbasit:
This PR:
- updates sequencer to latest hotshot version
- adds support for merklized state catchup


builder/Cargo.toml
```diff
@@ -14,14 +14,14 @@ async-std = { workspace = true }
 async-trait = { workspace = true }
 clap = { workspace = true }
 cld = { workspace = true }
-commit = { git = "https://github.com/EspressoSystems/commit.git" }
+committable = "0.2"
```
jbearer, line 17: Can we make all of these workspace dependencies?

sequencer/src/api/options.rs
```diff
@@ -307,19 +303,27 @@ impl Options {
...
         if self.hotshot_events.is_some() {
             self.init_and_spawn_hotshot_event_streaming_module(&mut context, bind_version)?;
         }
 
+        let ds = sql::DataSource::create(mod_opt, provider(query_opt.peers, bind_version), false)
+            .await?;
+
+        context.spawn(
+            "merklized state storage updater",
+            update_state_storage(Arc::new(RwLock::new(ds)), context.node_state()),
```
jbearer, line 324: Maybe `RwLock` not necessary?

sequencer/src/bin/nasty-client.rs
```diff
@@ -772,8 +773,8 @@ async fn serve(port: u16, metrics: PrometheusMetrics) {
         PATH = ["/metrics"]
         METHOD = "METRICS"
     };
-    let mut app = App::<_, ServerError, _>::with_state(RwLock::new(metrics));
-    app.module::<ServerError>("status", api)
+    let mut app = App::<_, ServerError>::with_state(RwLock::new(metrics));
+    app.module::<ServerError, StaticVersion<0, 1>>("status", api)
```
jbearer, line 777:
```suggestion
    app.module::<ServerError, SequencerVersion>("status", api)
```

sequencer/src/block/payload.rs
```diff
@@ -123,7 +123,7 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
         ns_id: NamespaceId,
         vid_common: VidCommon,
     ) -> Option<NamespaceProof> {
-        if self.raw_payload.len() != VidSchemeType::get_payload_byte_len(&vid_common) {
+        if self.raw_payload.len() as u32 != VidSchemeType::get_payload_byte_len(&vid_common) {
```
jbearer, line 126:
```suggestion
        if self.raw_payload.len().try_into().ok()? != VidSchemeType::get_payload_byte_len(&vid_common) {
```

sequencer/src/catchup.rs
```diff
@@ -138,8 +147,11 @@ impl<Ver: StaticVersionType> StateCatchup for StatePeers<Ver> {
                             continue;
                         };
                         match mt.remember(view.get_u64(), *elem, &frontier) {
-                            Ok(_) => return,
-                            Err(err) => tracing::warn!("Error verifying block proof: {}", err),
+                            Ok(_) => return Ok(()),
+                            Err(err) => {
+                                tracing::warn!("Error verifying block proof: {}", err);
+                                bail!(err);
```
jbearer, line 153: Don't bail here, just keep retrying

sequencer/src/header.rs
```diff
@@ -279,7 +279,8 @@ impl BlockHeader<SeqTypes> for Header {
                     parent_state.fee_merkle_tree.commitment(),
                     missing_accounts,
                 )
-                .await;
+                .await
+                .unwrap();
```
jbearer, line 283: Comment explaining that this unwrap will never panic

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
     Ok(())
 }
 
+#[derive(Debug)]
+struct SqlStateCatchup<'a> {
+    db: RwLockReadGuard<'a, SqlDataSource<SeqTypes, Provider>>,
```
jbearer, line 326: Maybe don't need `RwLock`? `&'a mut`?

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+
+        if bh == 0 {
+            return Ok(());
+        }
+
+        let proof = self
+            .db
+            .get_path(Snapshot::<SeqTypes, BlockMerkleTree, 3>::Index(bh), bh - 1)
+            .await
+            .unwrap();
```
jbearer, line 393: `?` instead of `unwrap`

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+
+    let storage_read = storage.read().await;
+    let last_height = storage_read.get_last_state_height().await?;
+
+    let mut leaves = storage_read.subscribe_leaves(last_height).await;
+
+    while let Some(leaf) = leaves.next().await {
+        let leaf = leaf.leaf();
+        let header = leaf.get_block_header();
+        let validated_state = ValidatedState::from_header(header);
```
jbearer, line 418: This is the parent state we're applying onto so we need to create it from the parent header

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+
+        let catchup = SqlStateCatchup {
+            db: storage.read().await,
+            block_height: leaf.get_height(),
+        };
+
+        let (state, delta) = validated_state
+            .apply_header(&instance, leaf, header, catchup)
```
jbearer, line 426: `leaf` should be parent leaf

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+        } = state;
+
+        let Delta { fees_delta } = delta;
+
+        // Insert fee merkle tree nodes
+        for delta in fees_delta {
+            let (_, proof) = fee_merkle_tree
+                .universal_lookup(delta)
+                .expect_ok()
+                .context("Index not found in fee merkle tree")?;
```
jbearer, line 442: Retry

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+            .store_state::<BlockMerkleTree, 3>(proof.proof, path, block_number)
+            .await
+            .context("failed to insert merkle nodes for block merkle tree")?;
+
+        storage
+            .write()
+            .await
+            .set_last_state_height(block_number as usize)
+            .await?;
+    }
```
jbearer, line 482: `commit`

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+                    block_info.number,
+                )
+                .await
+        } else {
+            vec![]
+        };
+
+        let mut delta = Delta::default();
+
+        catchup
```
jbearer, line 518: Only call catchup if we need to (if `self.need_to_fetch_blocks_mt_frontier`)

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+            .remember_blocks_merkle_tree(
+                parent_leaf.get_view_number(),
+                &mut validated_state.block_merkle_tree,
+            )
+            .await?;
+
+        for FeeInfo { account, .. } in l1_deposits {
```
jbearer, line 525: Only call catchup for missing accounts (`self.forgotten_accounts`). Also collect all missing accounts into a `Vec` and only call `fetch_accounts` once

sequencer/src/state.rs
```diff
@@ -308,6 +321,228 @@ fn validate_and_apply_header(
...
+                .await?;
+
+            let proof = &account_proof.first().unwrap().proof;
+
+            proof
+                .remember(&mut validated_state.fee_merkle_tree)
+                .unwrap();
+            delta.fees_delta.insert(account);
+        }
+
```
jbearer, line 541: Missing stuff from `validate_and_apply_header`, e.g. proposer fee and block merkle tree

sequencer/src/state.rs
```diff
@@ -364,7 +599,8 @@ impl HotShotState<SeqTypes> for ValidatedState {
                 .peers
                 .as_ref()
                 .remember_blocks_merkle_tree(view, &mut validated_state.block_merkle_tree)
-                .await;
+                .await
+                .unwrap();
```
jbearer, line 603: Comment explaining why `unwrap` won't panic

sequencer/src/state.rs
```diff
@@ -402,83 +406,110 @@ impl<'a> StateCatchup for SqlStateCatchup<'a> {
...
+                parent_leaf.clone(),
+                leaf.clone(),
+            )
             .await
-            .set_last_state_height(block_number as usize)
-            .await?;
+            .is_err()
+            {
+                tracing::error!(
+                    "failed to update state storage for leaf {:?}. retrying..",
```
jbearer, line 504: We should include the error in this log. Also probably a small delay before retrying would be good

sequencer/src/state.rs
```diff
@@ -402,83 +406,110 @@ impl<'a> StateCatchup for SqlStateCatchup<'a> {
 }
 
 pub async fn update_state_storage(
-    storage: Arc<RwLock<SqlDataSource<SeqTypes, Provider>>>,
+    storage: &mut SqlDataSource<SeqTypes, Provider>,
     instance: NodeState,
+    parent_leaf: LeafQueryData<SeqTypes>,
+    proposed_leaf: LeafQueryData<SeqTypes>,
```
jbearer: I think we don't need to take ownership of these things, `&` would do fine. This could save us from unnecessary cloning in `update_state_storage_loop`

sequencer/src/state.rs
```diff
@@ -512,17 +543,71 @@ impl ValidatedState {
...
+        if self.need_to_fetch_blocks_mt_frontier() {
+            tracing::warn!("fetching block frontier from peers");
+            instance
+                .peers
+                .as_ref()
+                .remember_blocks_merkle_tree(view, &mut validated_state.block_merkle_tree)
+                .await
+                .unwrap();
+
+            // validate proposal is descendent of parent by appending to parent
```
jbearer, line 565: This comment is now outdated, we're not validating anything here, we're just appending.

jbearer: Also we need to do this outside of this `if` statement, whether or not we had to fetch the frontier from peers is irrelevant, we need to insert the new block regardless

sequencer/src/state.rs
```diff
@@ -512,17 +543,71 @@ impl ValidatedState {
...
+        );
+
+        let view = parent_leaf.get_view_number();
+
+        // Ensure merkle tree has frontier
+        if self.need_to_fetch_blocks_mt_frontier() {
+            tracing::warn!("fetching block frontier from peers");
+            instance
+                .peers
+                .as_ref()
```
jbearer, line 560: Wow just realized we need to replace all these `instance.peers` with `catchup`. Then we will pass in `instance.peers` as `catchup` _only_ in the case where we actually want to use peer nodes for catchup (ie when we're being called from HotShot

jbearer: _Or_, perhaps even better, we could get rid of the `catchup` argument, use `instance.peers`, and then _set_ `node_state.peers` to be the SQL catchup in `update_state_storage`

sequencer/src/state.rs
```diff
@@ -512,17 +543,71 @@ impl ValidatedState {
...
+                    .proof
+                    .remember(&mut validated_state.fee_merkle_tree)
+                    .expect("proof previously verified");
+            }
+        }
 
         let mut delta = Delta::default();
 
+        // Unwrapping here is okay as we retry until we get the proof or until the task is canceled.
         catchup
```
jbearer, line 602: This `remember_blocks_merkle_tree` is already done above

sequencer/src/state.rs
```diff
@@ -512,17 +543,71 @@ impl ValidatedState {
...
                 parent_leaf.get_view_number(),
                 &mut validated_state.block_merkle_tree,
             )
-            .await?;
+            .await
+            .expect("block merkle tree proof member");
 
-        for FeeInfo { account, .. } in l1_deposits {
+        for FeeInfo { account, amount } in l1_deposits {
             let account_proof = catchup
```
jbearer, line 611: Already done above

jbearer (changes requested):
Have some more comments on the storage update and `apply_header`. Also, once we're happy with `apply_header`, let's not forget to update `validate_and_apply_header` to just call `apply_header` and do whatever extra validation is needed, rather than having the apply logic duplicated and spread out over many different functions

sequencer/src/state.rs
```diff
@@ -407,22 +402,22 @@ impl<'a> StateCatchup for SqlStateCatchup<'a> {
...
-    parent_leaf: LeafQueryData<SeqTypes>,
-    proposed_leaf: LeafQueryData<SeqTypes>,
+    instance: &NodeState,
+    parent_leaf: &LeafQueryData<SeqTypes>,
+    proposed_leaf: &LeafQueryData<SeqTypes>,
 ) -> anyhow::Result<()> {
     let proposed_leaf = proposed_leaf.leaf();
     let parent_leaf = parent_leaf.leaf();
-    let header = proposed_leaf.get_block_header();
+    let header = parent_leaf.get_block_header();
```
jbearer, line 411: Ah, but `header` is still the new header we're applying, so it has to come from the `proposed_leaf`. In other words `header` to apply and `validated_state` come from different leaves

sequencer/src/state.rs
```diff
@@ -543,6 +534,35 @@ impl ValidatedState {
         } else {
             vec![]
         };
+
+        // pushing a block into merkle tree shouldn't fail
+        validated_state
+            .block_merkle_tree
+            .push(parent_leaf.get_block_header().commit())
+            .unwrap();
```
jbearer, line 542: This needs to happen after the `remember_blocks_merkle_tree`; that `remember` is the thing that ensures this won't fail.

In general the flow here can be divided into three separate parts:
* Fetch any inputs we need from external sources (currently, just the deposits from L1)
* Fetch any missing data from our own state (missing accounts, block merkle tree frontier if needed)
* Apply state changes (push new block, update fee state based on deposits, charge builder fee). This step cannot fail assuming the previous steps succeeded.

May be a good idea to divide these into three separate functions; then we can use the type system to enforce that e.g. the third step is synchronous (not `async`) and infallible (doesn't return `Result`)

sequencer/src/state.rs
```diff
@@ -543,6 +534,35 @@ impl ValidatedState {
...
+
+        let mut delta = Delta::default();
+        for FeeInfo { account, amount } in l1_deposits.iter() {
+            validated_state
+                .fee_merkle_tree
+                .update_with(account, |balance| {
```
jbearer, line 548: Same here, needs to happen after remember

sequencer/src/state.rs
```diff
@@ -555,18 +575,12 @@ impl ValidatedState {
         // Ensure merkle tree has frontier
         if self.need_to_fetch_blocks_mt_frontier() {
             tracing::warn!("fetching block frontier from peers");
-            instance
-                .peers
-                .as_ref()
+
+            // Unwrapping here is okay as we retry until we get the proof or until the task is canceled.
+            catchup
```
jbearer, line 580: I guess technically this is not the case anymore, since this same function is also called with the Postgres catchup, in which we might actually return an `Err` (e.g. if we lose connection to the database). What we should probably do is just propagate the error here  (this function returns `Result` anyway) and then this unwrap will happen in `validate_and_apply_header` in the `ValidatedState` trait, because at _that_ point we are only dealing with the infallible `StatePeers` implementation of catchup.

Alternatively, we could in turn loop `validate_and_apply_header` forever, because why not: we get rid of a potential panic and HotShot will eventually cancel the task anyways

sequencer/src/state.rs
```diff
@@ -608,10 +524,69 @@ impl ValidatedState {
...
+        if charge_fee(
+            &mut validated_state.fee_merkle_tree,
+            proposed_header.fee_info,
+        )
+        .is_err()
+        {
+            bail!("Insufficient funds")
+        };
+
+        delta.fees_delta.insert(proposed_header.fee_info.account);
```
jbearer, line 540: `charge_fee` should do this

sequencer/src/state.rs
```diff
@@ -632,83 +607,17 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
-                    .expect("proof previously verified");
-            }
-        }
+        validate_builder(proposed_header).unwrap();
 
-        let mut delta = Delta::default();
+        let (validated_state, delta) = self
+            .apply_header(instance, parent_leaf, proposed_header)
+            .await
+            .unwrap();
```
jbearer, line 618: Here is where that "this unwrap will never fail" comment should finally go

sequencer/src/state.rs
```diff
@@ -526,19 +522,18 @@ impl ValidatedState {
 
         let mut delta = Delta::default();
 
-        apply_proposal(&mut validated_state, &mut delta, parent_leaf, l1_deposits);
-        // validate_builder(&mut self.fee_merkle_tree, &mut delta, proposed_header)?;
+        apply_proposal(&validated_state, &mut delta, parent_leaf, l1_deposits);
```
jbearer, line 525: We need to do something with the return value here! Currently we continue using `validated_state` which is unmodified by this function

sequencer/src/state.rs
```diff
@@ -565,11 +560,12 @@ pub async fn get_l1_deposits(
 }
 
 pub fn apply_proposal(
```
jbearer, line 562: Maybe throw a `#[must_use]` on this function

jbearer (approved):
Looks good, once tests are working

PR #1318, opened 2024-04-11
chain config, add ChainConfig struct

### LINKED ISSUE #1319: Add ChainConfig struct with chain ID, base fee, maximum payload size

sveitser:
Close #1319 

### TODO
- [x] Fix builder crate.
- [x] Add env vars for genesis chain config.
- [x] Add tests that shows that blocks fail to validate if chain config is changed.
- [x] Change `PartialEq` so that equality means complete equality instead of having the same commit.


sequencer/src/header.rs
```diff
@@ -21,14 +21,61 @@ use hotshot_types::{
...
     vid::VidCommitment,
 };
+use itertools::Either;
 use jf_primitives::merkle_tree::prelude::*;
 
 use serde::{Deserialize, Serialize};
 use time::OffsetDateTime;
 
+#[derive(Clone, Debug, Copy, Deserialize, Serialize, Eq)]
+pub struct ResolvableChainConfig {
```
sveitser, line 31: Should probably use a different name, since it's currently not resolvable.

sequencer/src/chain_config.rs
```diff
@@ -0,0 +1,66 @@
...
+            chain_id: chain_id.into(),
+            max_block_size,
+            base_fee: base_fee.into(),
+        }
+    }
+}
+
+impl Committable for ChainConfig {
+    fn tag() -> String {
+        "ChainConfig".to_string()
```
tbro, line 56: The convention elsewhere appears to be "CHAIN_CONFIG".

sequencer/src/state.rs
```diff
@@ -332,6 +332,12 @@ impl HotShotState<SeqTypes> for ValidatedState {
         // through returned value.
         let mut validated_state = self.clone();
 
+        // validate `ChainConfig`
+        if proposed_header.chain_config != instance.chain_config.into() {
```
sveitser, line 336: I think we should add a test that hits this. Or do we have it and I'm not seeing it?

sequencer/src/chain_config.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+impl ChainId {
+    // TODO: this duplicate code, consider a macro for U256 newtypes
+    pub(crate) fn to_fixed_bytes(self) -> [u8; 32] {
+        let mut bytes = [0u8; core::mem::size_of::<U256>()];
+        self.0.to_little_endian(&mut bytes);
+        bytes
+    }
```
ImJeremyHe, line 22: self.0.to_fixed_bytes()

ImJeremyHe: I do think we can add a macro in `commit` to avoid writing these code every time and being non-consistent

data/header.json
```diff
@@ -13,5 +13,6 @@
     },
     "block_merkle_tree_root":"MERKLE_COMM~yB4_Aqa35_PoskgTpcCR1oVLh6BUdLHIs7erHKWi-usUAAAAAAAAAAEAAAAAAAAAJg",
     "fee_merkle_tree_root":"MERKLE_COMM~VJ9z239aP9GZDrHp3VxwPd_0l28Hc5KEAB1pFeCIxhYgAAAAAAAAAAIAAAAAAAAAdA",
-    "fee_info":{"account":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266","amount":"0x0"}
+    "fee_info":{"account":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266","amount":"0x0"},
+    "chain_config":{"chain_config":{"Left":{"chain_id":"0x8a19","max_block_size":10240,"base_fee":"0x0"}}}
```
ImJeremyHe, line 17: redudant "chain_config" here

sveitser: Sorry @ImJeremyHe what is `untag`?

sequencer/src/chain_config.rs
sveitser: I added a macro in the sequencer crate: https://github.com/EspressoSystems/espresso-sequencer/pull/1318/commits/037958322a8ee46b444ffd01ce8cb6b5cc8f238a, I'm not sure what exactly you would like to have a macro for in the commit crate.

data/header.json
ImJeremyHe: Sorry, I mean we could customize the `serde` using this https://serde.rs/enum-representations.html#untagged

tbro (approved):

sequencer/src/header.rs
```diff
@@ -21,14 +21,61 @@ use hotshot_types::{
...
+    chain_config: Either<ChainConfig, Commitment<ChainConfig>>,
+}
+
+impl ResolvableChainConfig {
```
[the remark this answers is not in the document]
sveitser: No, that seems to be an oversight.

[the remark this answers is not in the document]
sveitser: @tbro feel free to change this. I won't get around to it today.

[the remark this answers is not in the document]
tbro: 5376435f47f8f4aa4f7074d119d33bc9b75fd84d

sveitser:
@tbro I think the changes look good. I can't approve on github because I opened the PR.

tbro (approved):

sequencer/src/header.rs
```diff
@@ -203,6 +210,7 @@ impl Header {
         let fee_merkle_tree_root = state.fee_merkle_tree.commitment();
 
         let header = Self {
+            chain_config: chain_config.into(),
```
jbearer, line 213: Should this be `chain_config.commit().into()`? Otherwise we're including the full chain config in every header

sequencer/src/state.rs
```diff
@@ -139,13 +140,25 @@ impl ValidatedState {
...
     delta: &mut Delta,
     parent_leaf: &Leaf,
     proposal: &Header,
     receipts: Vec<FeeInfo>,
 ) -> anyhow::Result<()> {
     let parent_header = parent_leaf.get_block_header();
+
+    // validate `ChainConfig`
+    anyhow::ensure!(
+        proposal.chain_config == expected_chain_config.into(),
```
jbearer, line 154: Due to the way `PartialEq` works on `ResolvableChainConfig`, I think this will only pass if `proposal.chain_config` is exactly `expected_chain_config` (the full thing). I think this should be something like `proposal.chain_config.commit() == expected_chain_config.commit()`

sequencer/src/header.rs
tbro: we have that defined on the `Header` as `chain_config: ResolvableChainConfig`. should that be changed to `Commitment<ChainConfig>`?

tbro: My understanding of the docs is that the Header field can be either the full `ChainConfig` or the `Commitment`. As it is not that `into()` will resolve to `ResolvableChainConfig` which holds an `Either` wrapping those two things.

jbearer: Ah sorry, maybe I wasn't clear in the docs. It can indeed be either, but we mainly only want the full thing in the genesis. After that it's better to include just a commitment to keep the headers small (this will benefit more as we add more fields to it)

PR #1326, opened 2024-04-12
Don't block prover API during proof generation

### LINKED ISSUE #1325: Prover service does not respond to API requests when generating proof
This may make the healthcheck requests time out leading to the load balancer believing the service is unhealthy. The service should be modified to be able to respond to healthchecks while it's generating proofs.

### LINKED ISSUE #1352: "UpgradeBoxScript: Added upgradeBox function to upgrade proxy contract"
This commit message describes the addition of the upgradeBox function to the UpgradeBoxScript contract, which is used to upgrade a proxy contract to a new implementation. This function is essential for managing contract upgrades and ensuring that the system remains up-to-date with the latest features and bug fixes. By adding this function, the contract is now capable of performing upgrades in a more controlled and efficient manner, reducing the risk of errors and improving the overall user experience.

sveitser:
Close #1325


sveitser:
@jbearer Is this a reasonable solution? I did some manual testing and it does seem to unblock the healthcheck during proof generation.

scripts/smoke-test-demo
```diff
@@ -59,8 +59,8 @@ done
 
 echo "Checking if builder is healthy at http://localhost:$ESPRESSO_BUILDER_SERVER_PORT"
 curl -s --fail http://localhost:$ESPRESSO_BUILDER_SERVER_PORT > /dev/null
```
sveitser, line 61: Need to add retries to these checks.

mrain (approved):
LGTM. Pending CI fix

PR #1327, opened 2024-04-12
Fix Demo

rob-maron:
Fixes the Docker demo by changing the healthcheck endpoints in the images.

Also changes it so that Libp2p addresses are supplied as strings and then later parsed as socket addresses. 

This lets us resolve a particular host to bind or advertise to. For example, we want to support binding to `localhost` and advertising as a load balancer.


sveitser (approved):

sveitser:
@rob-maron just requires `cargo fmt` I think. The contracts test is also failing due to that.

rob-maron:
> @rob-maron just requires `cargo fmt` I think. The contracts CI job is also failing due to that.

Thanks Mathis, good catch

builder/src/bin/permissioned-builder.rs
```diff
@@ -210,10 +210,24 @@ async fn main() -> anyhow::Result<()> {
     // get from the private key
     let builder_pub_key = BLSPubKey::from_private(&private_staking_key);
 
+    // Parse supplied Libp2p addresses to their socket form
+    let libp2p_advertise_address = opt
+        .libp2p_advertise_address
+        .to_socket_addrs()?
+        .find(|x| x.is_ipv4())
```
sveitser, line 217: Could you add a comment on why we only allow ipv4 addresses?

sveitser: Thank you!

PR #1329, opened 2024-04-14
chore: update ark-srs dep

alxiong:
Pointing `ark-srs` to `0.2.0` on crates-io instead of to git repo, after updates in https://github.com/alxiong/ark-srs/pull/7. This is synced with https://github.com/EspressoSystems/HotShot/pull/2854

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -203,6 +203,7 @@
             nixpkgs-fmt
             entr
             process-compose
+            gh
             # `postgresql` defaults to an older version (15), so we select the latest version (16)
             # explicitly.
             postgresql_16
@@ -228,6 +229,9 @@
             # with rustup installations.
             export CARGO_HOME=$HOME/.cargo-nix
             export PATH="$PWD/$CARGO_TARGET_DIR/release:$PATH"
+            export ROOT_DIR=$(dirname "$(realpath ./flake.nix)")
+            export AZTEC_SRS_PATH="$ROOT_DIR/data/aztec20/kzg10-aztec20-srs-1048584.bin"
+            ./scripts/download_srs_aztec.sh
           '' + self.checks.${system}.pre-commit-check.shellHook;
           RUST_SRC_PATH = "${stableToolchain}/lib/rustlib/src/rust/library";
           FOUNDRY_SOLC = "${solc}/bin/solc";
--- a/scripts/download_srs_aztec.sh
+++ b/scripts/download_srs_aztec.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -f "$AZTEC_SRS_PATH" ]; then
+    echo "SRS file $AZTEC_SRS_PATH exists"
+else
+    echo "SRS file $AZTEC_SRS_PATH does not exist, downloading ..."
+    gh release download --repo alxiong/ark-srs v0.2.0 -p "$(basename $AZTEC_SRS_PATH)" -O "$AZTEC_SRS_PATH" --skip-existing
+fi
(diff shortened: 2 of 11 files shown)


scripts/download_srs_aztec.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -f "$AZTEC_SRS_PATH" ]; then
+    echo "SRS file $AZTEC_SRS_PATH exists"
+else
+    echo "SRS file $AZTEC_SRS_PATH does not exist, downloading ..."
+    gh release download --repo alxiong/ark-srs v0.2.0 -p "$(basename $AZTEC_SRS_PATH)" -O "$AZTEC_SRS_PATH" --skip-existing
```
sveitser, line 9: Why not download with wget? E. g.

```
wget https://github.com/alxiong/ark-srs/releases/download/v0.2.0/kzg10-aztec20-srs-1048584.bin
```

alxiong:
The "security audit" ci failure is due to "commit" crate renamed to committable, I tried to upgrade but just got more errors mostly due to HotShot version we relied on is "main head pinning" this crate thus fragile in upgrade and I don't find a quick fix.

We probably need a separate PR to upgrade the hotshot dep together. So just ignore that CI failure for now.

scripts/download_srs_aztec.sh
alxiong: I tried with curl -L and wget and many other combinations of option flags, but while running them works directly in terminal, running them in bash script doesn't.

And the gh CLI is the simplest solution i can find.

sveitser: Weird. It seems to work for me to run this wget command in bash script. It's true curl seems to be troublesome though.

sveitser:
The update for committable is in https://github.com/EspressoSystems/espresso-sequencer/pull/1316

scripts/download_srs_aztec.sh
alxiong: Oh that's interesting, can you double check the sha256sum of the downloaded against https://github.com/alxiong/ark-srs/blob/main/src/constants.rs#L11?

If wget works for you, then I'll try again tmr.

alxiong: you are right, wget can be done, [63c40c3](https://github.com/EspressoSystems/espresso-sequencer/pull/1329/commits/63c40c3069ac3a2d00aa435e0e5b36466e79fc15)
(previously I was testing locally outside the nix-shell, not sure it wasn't work then, nvm)

alxiong:
any more comments? @sveitser 
shall we merge this soon?

flake.nix
```diff
@@ -228,6 +228,9 @@
             # with rustup installations.
             export CARGO_HOME=$HOME/.cargo-nix
             export PATH="$PWD/$CARGO_TARGET_DIR/release:$PATH"
+            export ROOT_DIR=$(dirname "$(realpath ./flake.nix)")
```
sveitser, line 231: Isn't this the same as $PWD?

.github/workflows/contracts.yml
```diff
@@ -18,6 +18,7 @@ concurrency:
 
 env:
   RUST_LOG: info,libp2p=off
+  GH_TOKEN: ${{ github.token }}
```
sveitser, line 21: Since we're using wget now can we remove this?

sveitser (approved):
LGTM, just 2 questions.

.github/workflows/contracts.yml
alxiong: agree, removed in [3cc95e0](https://github.com/EspressoSystems/espresso-sequencer/pull/1329/commits/3cc95e028d73534fa20beee5bb0970e62b6a5e91)

PR #1330, opened 2024-04-15
[DRAFT] Builder ECDSA signatures

sveitser:
Right now just to discuss if this goes in the right direction


sequencer/src/eth_signature_key.rs
```diff
@@ -0,0 +1,198 @@
...
+use ethers::{
+    signers::{
+        coins_bip39::{English, Mnemonic},
+        LocalWallet, WalletError,
+    },
+    types::{Address, Signature},
+    utils::public_key_to_address,
+};
+use hotshot_types::traits::signature_key::BuilderSignatureKey;
+use k256::ecdsa::{SigningKey, VerifyingKey};
```
alxiong, line 11: i believe `ethers` re-exported this so we don't have to declare a separate dep, and use `ethers::core::k256::ecdsa::SigningKey` instead.

alxiong: same comment for [`coins_bip39`](https://docs.rs/ethers/latest/ethers/signers/coins_bip39/index.html)

sveitser: I don't think they export everything we need. For example the serde feature for `VerifyingKey`.

alxiong: really? I thought they literally just rexport the entire `k256` crate.

sveitser: They don't enable the `serde` and `pem` features on it which are required to get serialization so if we use that we have to add those traits ourselves. I prefer to add a dependency in a toml file than trait impls for serde.

For coins_bip39 I have to check if I can use that instead of coins_bip32.

alxiong: i see, thanks for the clarification!

PR #1333, opened 2024-04-16
Feature flag Libp2p

rob-maron:
Adds a feature flag `libp2p` that enables Libp2p. Without it, only the CDN is used.

### DIFF
--- a/sequencer/src/lib.rs
+++ b/sequencer/src/lib.rs
@@ -95,16 +95,16 @@ pub mod network {
     }
 
     #[derive(Clone, Copy, Default)]
-    pub struct Networks;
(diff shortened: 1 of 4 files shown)


sequencer/src/lib.rs
```diff
@@ -93,13 +95,20 @@ pub mod network {
     }
 
     #[derive(Clone, Copy, Default)]
-    pub struct Combined;
+    pub struct Networks;
```
jbearer, line 98: I think this name is too vague, what about like `Production` to distinguish it from `Memory` which is used only in tests?

jbearer (approved):

PR #1334, opened 2024-04-16
[Push CDN] Dockerfiles - change `ENTRYPOINT` to `CMD`

rob-maron:


Ancient123:
This is going to break the docker compose file (and cloud deploy) unless we update the commands on those as well.

rob-maron:
> This is going to break the docker compose file (and cloud deploy) unless we update the commands on those as well.

Good point, fixed

Ancient123 (approved):

PR #1337, opened 2024-04-17
Use new transcript logic in solidity Plonk Verifier.

### LINKED ISSUE #1336: Update transcript logic in the Plonk verifier
Jellyfish 0.4.3 contains the new transcript logic. However the current solidity code does not support it and thus many tests fail.
Create a PR off https://github.com/EspressoSystems/espresso-sequencer/pull/1316

philippecamacho:
Closes #1336 

@alxiong I added some TODOs that I will handle when merge the changes related to gas optimization.


imabdulbasit (approved):

contracts/src/libraries/Transcript.sol
```diff
@@ -60,15 +60,19 @@ library Transcript {
...
-        return BN254.fromLeBytesModOrder(BytesLib.slice(abi.encodePacked(h1, h2), 0, 48));
+        // TODO Philippe create function in BN254.sol
+        // TODO Optimize
+        bytes memory beBytes = abi.encodePacked(hash);
+        uint256 ret = 0;
+        for (uint256 i = 0; i < beBytes.length; i++) {
+            ret = mulmod(ret, 256, BN254.R_MOD);
+            ret = addmod(ret, uint256(uint8(beBytes[i])), BN254.R_MOD);
+        }
+        return ret;
```
alxiong, line 75: personally I find this to be gas inefficient. since we know for sure that `beBytes` will be 32 bytes, (first of all, i don't know why we convert it to a generic `bytes` from `bytes32`)

is the following wrong?:
```solidity
uint256 ret = uint256(hash) % BN254.R_MOD;
```

philippecamacho: Thanks, you are right. I fixed it in @imabdulbasit 's branch: 9cb190cc
With this change verifying the proof now costs 600k instead of 757k!!! Very exciting when this change is merged with the other optimizations :)

alxiong: great to hear!

PR #1338, opened 2024-04-18
chore: remove dev-deploy and refactor deploy.rs

### LINKED ISSUE #1177: Remove calls to `just` in rust tests
Not an issue with this PR but I'm not sure it's a great idea to call `just` from our rust tests. There doesn't seem to be much benefit from using `just` here it but we've created another dependency for a tool that isn't so commonplace and now needs to be installed to run `cargo test`. Any downside of using a bash script? What do others think about this?

_Originally posted by @sveitser in https://github.com/EspressoSystems/espresso-sequencer/pull/1139#discussion_r1505785291_

alxiong:
Close #1177 

Major changes include:
- refactor `bin/deploy.rs` to separate out the reusable code to a `utils/deployer.rs` module
  - meanwhile fix the different deployment logic for LightClientMock and LightClient as I reviewed in https://github.com/EspressoSystems/espresso-sequencer/pull/1322
- remove `just dev-deploy` command and replace deployment logic inside rust to directly use `sequencer_utils::deployer`'s APIs.


sveitser (approved):
LGTM, but let's wait for an ark-srs update that works with hotshot.

PR #1340, opened 2024-04-18
Use tini entrypoint to enable exit with SIGTERM

sveitser:


ImJeremyHe (approved):

docker/cdn-broker.Dockerfile
```diff
@@ -5,11 +5,12 @@ ARG TARGETARCH
...
 
 COPY target/$TARGETARCH/release/cdn-broker /bin/cdn-broker
 RUN chmod +x /bin/cdn-broker
 
 ENV RUST_LOG="info"
 
 HEALTHCHECK --interval=1s --timeout=1s --retries=100 CMD curl --fail http://localhost:${ESPRESSO_CDN_SERVER_METRICS_PORT}/metrics || exit 1
-CMD ["cdn-broker"]
\ No newline at end of file
+CMD ["cdn-broker"]
```
ImJeremyHe, line 16: it looks like we need a linter configed in nix

ImJeremyHe (approved):

docker/cdn-broker.Dockerfile
sveitser: We can add this: https://github.com/marketplace/actions/linelint

PR #1347, opened 2024-04-19
Patch for patch for branch...

nyospe:


sequencer/src/header.rs
```diff
@@ -79,6 +79,7 @@ pub struct Header {
     pub l1_finalized: Option<L1BlockInfo>,
 
     pub payload_commitment: VidCommitment,
+    pub builder_commitment: BuilderCommitment,
```
jbearer, line 82: Hmm why is this a field of header?

nyospe: Because it needs to be propagated with the quorum proposal to make it possible to successfully confirm identity between a DAProposal and a QuorumProposal, and if it isn't committed to, a bad actor running the hotshot-event-service could forge and manipulate (rival) builders.

nyospe: Putting it in the header means everyone sees the same identity, and the identity conforms to the same block as the quorum proposal, so lying to a builder shouldn't be possible.

jbearer: But isn't it derivable from other fields in the header?

nyospe: Nope. Attempting to derive it from other fields in the header was a BUGBUGBUG

nyospe: It is, explicitly, a commitment to the block payload that is cryptographically secure, but much much much faster and lower load than the vid_commitment

jbearer: Ohhh I see this is the _payload_ builder commitment. Then I think returning this from `Header::builder_commitment` is wrong, we need to hash in the fee amount and metadata so we end up with a commitment signed by the builder that binds the desired payload to a desired fee

nyospe: Uh... do we currently have those available in the `BlockPayload`? Because this needs to be computed from the `DAProposal`, then checked against what is sent in the `QuorumProposal`... its current purpose is correlating the two between events emitted by `HotShot` and disseminated by the `hotshot-event-service` subscription stream. If it provides value by binding the fee, that's not unreasonable, but it is an additional property that the original use case doesn't care about.

nyospe: Oh!!! This isn't the hash that the builder signs when it sends over available blocks for the proposer. This is what _every_ builder consumes, not what the _selected_ builder creates.

nyospe: But, yes, they could be made synonymous, and it wouldn't be harmful.

jbearer: Discussed offline: `Header::builder_commitment` is actually something entirely different from the fee commitment/signature. I will update the fee signing/validation logic to use a different function to get the signed message and then `builder_commitment` can be a simple getter for the builder commitment *to the payload*, as it is in this PR

nyospe: And I've got a pin in a minor issue to rename and thoroughly document the `builder_payload_commitment`...

jbearer (approved):

PR #1353, opened 2024-04-22
Add dev-node binary

ImJeremyHe:


sequencer/src/bin/test-node.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// Account index in the L1 wallet generated by MNEMONIC to use when deploying the contracts.
+    #[clap(
+        long,
+        name = "ACCOUNT_INDEX",
+        env = "ESPRESSO_DEPLOYER_ACCOUNT_INDEX",
+        default_value = "0"
+    )]
+    account_index: u32,
+    /// Port to run the builder server on.
+    #[clap(short, long, env = "ESPRESSO_BUILDER_SERVER_PORT")]
```
sveitser, line 49: I think we can put some defaults here for convenience. Like 3 consecutive ports that aren't by default used by any other tools we use. For example, maybe

API: 8770
Builder: 8771
Commit: 8772

ImJeremyHe: [b042b8e1](https://github.com/EspressoSystems/espresso-sequencer/pull/1353/commits/b042b8e1af4adf715117ab8858d589d951fd5018)

.github/workflows/build.yml
```diff
@@ -72,6 +72,7 @@ jobs:
             target/release/keygen
             target/release/permissionless-builder
             target/release/nasty-client
+            target/release/test-node
```
sveitser, line 75: I think since we eventually want to distribute this we should call the binary something that identifies the binary more easily out of the context of this repo. I don't have a great idea but something like `espresso-test-node` is probably better than just `test-node`.

sequencer/src/bin/test-node.rs
```diff
@@ -0,0 +1,366 @@
...
+    deployer::{deploy, Contract, Contracts},
+    AnvilOptions,
+};
+use tide_disco::{error::ServerError, Api};
+use url::Url;
+use vbs::version::StaticVersionType;
+
+#[derive(Clone, Debug, Parser)]
+struct Args {
+    /// A JSON-RPC endpoint for the L1 to deploy to.
```
sveitser, line 26: Describe what happens if this argument is not passed.

sequencer/src/bin/test-node.rs
```diff
@@ -0,0 +1,366 @@
...
+        let builder_port = pick_unused_port().unwrap();
+        let commitment_task_port = pick_unused_port().unwrap();
+        let api_port = pick_unused_port().unwrap();
+        let postgres_port = pick_unused_port().unwrap();
+
+        let mut db = Command::new("docker")
+            .arg("compose")
+            .arg("up")
+            .arg("-d")
+            .arg("sequencer-db")
```
sveitser, line 238: We also need to add this to the binary don't we? I think ideally we don't use docker-compose for it but call docker directly so we don't need to be in this repo for it to work.

ImJeremyHe: This indeed makes it simpler to launch the test-node binary in sequencer repo. However, since developers would likely prefer using docker image directly, hardcoding the docker commands into the binary seems somewhat inappropriate.

sveitser: In https://github.com/EspressoSystems/espresso-sequencer/pull/1353/commits/1624cd465ebc03a0038f681f1b392a89344d07c9 we have a self contained docker image now that starts postgres, anvil and the dev node.

.github/workflows/build.yml
sveitser: renamed to `espresso-dev-node`

sequencer/Cargo.toml
```diff
@@ -9,10 +9,16 @@ edition = "2021"
 testing = ["hotshot-testing"]
 libp2p = []
 
+[[bin]]
+name = "espresso-dev-node"
+required-features = ["testing"]
```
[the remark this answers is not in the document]
sveitser: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    tracing::info!("starting hotshot commitment task");
+    run_hotshot_commitment_task::<es_version::SequencerVersion>(&commitment_task_options).await;
+
+    Ok(())
+}
+
+// In the test node binary, for now we don't need to run the builder. We just hardcode the builder address
+// and expose the needed endpoint.
+fn start_builder_server<Ver: StaticVersionType + 'static>(
```
imabdulbasit, line 154: TestNetwork already runs the simple builder so we don't need to run it separately.

ImJeremyHe:
@imabdulbasit Thank you!

sequencer/src/bin/deploy.rs
```diff
@@ -104,95 +85,26 @@ struct Options {
...
-        contracts
-            .deploy_tx(
-                Contract::FeeContractProxy,
-                ERC1967Proxy::deploy(l1.clone(), (fee_contract_address, data))?,
-            )
-            .await?;
-    }
+    let contracts = Contracts::from(opt.contracts);
+
+    let genesis = light_client_genesis(&opt.orchestrator_url, opt.stake_table_capacity).await?;
```
imabdulbasit, line 96: native demo is failing because this binary is waiting for pubkeys from orchestrator
`Peers' keys are not ready, retrying.`

sveitser: @ImJeremyHe @imabdulbasit I think the deployment of the sequencer-contracts (fee contract) must work without needing the orchestrator. I think that's why the deployment is split into two tasks (sequencer / prover contracts).

Currently in this PR the orchestrator needs the sequencers to provide the peer configs and the deploy-sequencer-contracts needs to orchestrator and the sequencers need the deploy-sequencer-contract so there's a circular dependency.

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -0,0 +1,332 @@
...
+        setup_logging();
+        setup_backtrace();
+        let anvil = AnvilOptions::default().spawn().await;
+        // let commitment_task_port = pick_unused_port().unwrap();
+        let commitment_task_port = 10002;
+        let api_port = pick_unused_port().unwrap();
+        let postgres_port = pick_unused_port().unwrap();
+
+        let mut db = Command::new("docker")
+            .arg("compose")
```
sveitser, line 201: Let's not introduce a dependency on docker compose and call docker directly instead.

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -0,0 +1,332 @@
...
+            commitment_task_port
+        );
+        println!("commitment task url: {}", commitment_task_url);
+
+        let sequencer_get_header_url =
+            format!("http://localhost:{}/availability/header/3", api_port);
+        println!("sequencer url: {}", sequencer_get_header_url);
+
+        // Waiting for the test node running completely
+        async_sleep(Duration::from_secs(100)).await;
```
sveitser, line 247: Can we replace this with something that waits for endpoints to be available instead of a sleep? The sleep might fail on slower hosts and waste time otherwise.

imabdulbasit: Yeah I was thinking to call availability api

sequencer/src/bin/espresso-dev-node.rs
imabdulbasit: This will be replaced by TmpDb from query service

sequencer/src/bin/espresso-dev-node.rs
sveitser: Yeah or we could re-try to submit until that works.

sequencer/src/bin/espresso-dev-node.rs
sveitser: Done in https://github.com/EspressoSystems/espresso-sequencer/pull/1353/commits/76dec3068f2d29c1a582a5693a3a1eae4e26cc06

sequencer/src/bin/espresso-dev-node.rs
imabdulbasit: adjusted in b47398eb9a4970b61364f2f281681c6840a92131

sequencer/src/bin/deploy.rs
imabdulbasit: Changed the `genesis` parameter to be a `BoxFuture` that is awaited only for `LightClient` contract.  We can also split it into separate functions for each contract.

tbro:
> Since the dev-node requires the `testing` feature, how about we add a separate docker build step on the CI where we `cargo build --release --features testing --bin espresso-dev-node `.


I think https://github.com/EspressoSystems/espresso-sequencer/pull/1353/commits/805aee159035696d502e3464798f277e3eea781a accomplishes this.

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -0,0 +1,326 @@
...
+    )]
+    account_index: u32,
+    /// Port that the HTTP API will use.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_API_PORT")]
+    sequencer_api_port: u16,
+    /// If provided, the service will run a basic HTTP server on the given port.
+    ///
+    /// The server provides healthcheck and version endpoints.
+    #[clap(short, long, env = "ESPRESSO_COMMITMENT_TASK_PORT")]
+    commitment_task_port: u16,
```
ImJeremyHe, line 56: I think we still need to add the `builder_port` back here.

imabdulbasit (commented):
TestNetwork runs the simple builder, so the start function needs to be changed to take this port.

sequencer/src/bin/espresso-dev-node.rs
tbro: I understand the need for this, but its going to require some modification to `HotShot`. There is a lot of work here so I think its best to merge. I've created an issue so we can deal w/ this in a follow up: https://github.com/EspressoSystems/espresso-sequencer/issues/1522

sequencer/src/bin/deploy.rs
imabdulbasit: 76dec3068f2d29c1a582a5693a3a1eae4e26cc06

sveitser (approved):

PR #1354, opened 2024-04-22
Restartability

jbearer:


jbearer:
check-edits failure is because I added the `state` API to the native demo, but it was already present in the docker demo

PR #1357, opened 2024-04-22
CDN Whitelist

rob-maron:
Allows for whitelisting of orchestrator-derived public keys for the CDN.


jbearer (approved):

rob-maron:
@Ancient123 This will add another service that only needs to run once on startup. It ensures that only whitelisted users can access the CDN. If it fails it will exit with code != 0

Issue #1341, opened 2024-04-18
Review public API of rust code


sveitser:
We typically weren't super careful when deciding on visibility so should take a moment to review and not expose more than desired.

PR #1360, opened 2024-04-22
Make API initiliazation lazy; healthchecks faster to start

jbearer:
This change makes the initalization of the API state, which depends on conesensus initialization, lazy. This allows us to start the API server before conesensus is running, which means we get healthchecks immediately after the sequencer node starts (since healthchecks do not depend on any state). Meanwhile, endpoints that depend on consensus state can simply await the lazy state each time they are called. Normally, this will yield the consensus state immediately, but during initialization, this causes stateful endpoints to block until the consensus state is ready.


sequencer/src/api.rs
```diff
@@ -21,44 +28,119 @@ mod update;
...
+            event_streamer: ctx.get_event_streamer(),
+            node_state: ctx.node_state(),
             handle: ctx.consensus().clone(),
         }
     }
 }
+
+#[derive(Derivative)]
+#[derivative(Clone(bound = ""), Debug(bound = ""))]
+struct State<N: network::Type, P: SequencerPersistence, Ver: StaticVersionType> {
```
sveitser, line 59: Should we call it `ApiState`, since we have quite a few `...State` types now?

sequencer/src/context.rs
```diff
@@ -299,3 +294,47 @@ async fn handle_events<Ver: StaticVersionType>(
         }
     }
 }
+
+#[derive(Debug, Default)]
+pub struct TaskList(Vec<(String, JoinHandle<()>)>);
```
sveitser, line 299: I recently created #1341 because I think we expose a lot more than we would want to. If we all agree on that we should probably at the very least not unnecessarily expose new types.

I think we could also say we don't make any guarantees in terms of public API stability because at the moment this is first and foremost an application crate. In that case we should probably put this in the readme though.

sveitser (approved):
LGTM just some questions/suggestions.

sequencer/src/context.rs
jbearer: I can make this one `pub(crate)`, and add a disclaimer to the README

PR #1367, opened 2024-04-23
CI: fix contracts fuzz tests

sveitser:
- Update flake.lock to get a foundry bugfix that fixes inline config:

  https://github.com/foundry-rs/foundry/pull/7431

- Only set `RAYON_NUM_THREADS` in prover. Foundry also uses it to set how many tests run in parallel. In addition it sets parallelism for `diff-test` which is run by foundry tests.


sveitser:
Ran the CI like it runs on `main` [here](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/8797286381/job/24141832386). Seems to take much less than 30 minutes again now.

alysiahuggins (approved):

PR #1369, opened 2024-04-23
builder changes/update hotshot

move47:
- Introduces two new env variables and removes some unused env variables. 
- Updates to hotshot 0.5.41 and other relevant dependencies.

Edit: Fulfilled. ~~REQUEST TO REVIEWERS: I am not aware of our env variables and where we run `permission less-builder` binary with env variables, I request reviewer to make push those relevant changes before we merge this. CI MIGHT FAIL BECAUSE OF THIS.~~


builder/src/bin/permissionless-builder.rs
```diff
@@ -110,6 +86,25 @@ pub struct NonPermissionedBuilderOptions {
     /// BUILDER CHANNEL CAPACITY
     #[clap(short, long, env = "ESPRESSO_BUILDER_CHANNEL_CAPACITY")]
     pub channel_capacity: NonZeroUsize,
+
+    /// The amount of time a builder can wait before timing out a request to the API.
+    #[clap(
+        short,
+        long,
+        env = "WEBSERVER_RESPONSE_TIMEOUT_DURATION",
```
sveitser, line 94: Please prefix with `ESPRESSO_BUILDER`

sveitser:
To answer the questions the env vars are set in the `.env` file and if you want to pass it to docker containers also in `docker-compose.yaml` rarely also in the `process-compose.yaml` file because for the native demo the env vars in `.env` are all available to the binaries.

Here it looks like you have workable default values then you don't need to add anything to any other files.

builder/src/bin/permissioned-builder.rs
```diff
@@ -133,24 +133,43 @@ pub struct PermissionedBuilderOptions {
     pub state_peers: Vec<Url>,
 
     /// Port to run the builder server on.
-    #[clap(short, long, env = "BUILDER_SERVER_PORT")]
+    #[clap(short, long, env = "ESPRESSO_BUILDER_SERVER_PORT")]
     pub port: u16,
 
     /// Port to run the builder server on.
-    #[clap(short, long, env = "BUILDER_ADDRESS")]
+    #[clap(short, long, env = "ESPRESSO_BUILDER_ADDRESS")]
```
sveitser, line 140: Is this variable used? The address is obtained from the private key derived from the mnemonic.

builder/src/bin/permissionless-builder.rs
```diff
@@ -110,6 +86,25 @@ pub struct NonPermissionedBuilderOptions {
     /// BUILDER CHANNEL CAPACITY
     #[clap(short, long, env = "ESPRESSO_BUILDER_CHANNEL_CAPACITY")]
     pub channel_capacity: NonZeroUsize,
+
+    /// The amount of time a builder can wait before timing out a request to the API.
+    #[clap(
+        short,
+        long,
+        env = "ESPRESSO_WEBSERVER_RESPONSE_TIMEOUT_DURATION",
```
jbearer, line 94:
```suggestion
        env = "ESPRESSO_BUILDER_WEBSERVER_RESPONSE_TIMEOUT_DURATION",
```

builder/src/bin/permissioned-builder.rs
sveitser: Actually nevermind, I thought we will remote the permissioned builder anyway, so no need to fix the code I think.

jbearer (approved):

PR #1370, opened 2024-04-23
Only set RUSTFLAGS for coverage

sveitser:
Currently the coverage workflow is broken because we set RUSTFLAGS globally that only work with nightly.


sveitser:
CI run: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/8802942220

Issue #1321, opened 2024-04-11
Clean up builder code


sveitser:
Currently the builders require arguments that are only required because the re-use code for that was written to set up consensus nodes. The code should be de-duplicated and unnecessary parts removed.

PR #1372, opened 2024-04-23
Remove unused parameters from permissionless builder

jbearer:


move47 (approved):
Thanks for this. It closes #1321 also :). I tried removing some in #1369

PR #1373, opened 2024-04-23
Submit transactions to builder private mempool

jbearer:
Add an option in `submit-transactions` to submit to a builder. Add a second transaction generator task in the demos to submit to the builder. This both serves as a test of the builder API and allows us to directly compare the latencies of transactions submitted via the public and private mempools.


philippecamacho (approved):
LGTM

PR #1380, opened 2024-04-24
update hotshot

jparr721:
This PR updates HotShot to the latest version which includes, among other things, significant changes to the interfaces for encoding and handling of bytes (essentially removes all the copying that existed before). It also includes a slew of updates for the trait updates Artemii made. Lastly, it implements `EncodeBytes` for the NamespaeTable.

The principal change, however, was a bugfix within `HotShot` which fixed an issue where `instance_state` would get passed to sequencer via a lock on the `consensus` state object. As a result, if sequencer got stuck in a loop while it held the lock, it'd lock consensus from writing and neither would proceed, even when messages that would ordinarily break the loop arrived.

### DIFF
--- a/sequencer/src/block/tables.rs
+++ b/sequencer/src/block/tables.rs
@@ -48,8 +48,7 @@ pub struct NameSpaceTable<TableWord: TableWordTraits> {
 
 impl<TableWord: TableWordTraits> EncodeBytes for NameSpaceTable<TableWord> {
     fn encode(&self) -> std::sync::Arc<[u8]> {
-        let boxed_slice: Box<[u8]> = self.bytes.clone().into_boxed_slice();
-        Arc::from(boxed_slice)
+        Arc::from(self.bytes.clone())
     }
 }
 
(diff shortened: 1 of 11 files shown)


sequencer/src/block/tables.rs
```diff
@@ -44,6 +46,13 @@ pub struct NameSpaceTable<TableWord: TableWordTraits> {
     pub(super) phantom: PhantomData<TableWord>,
 }
 
+impl<TableWord: TableWordTraits> EncodeBytes for NameSpaceTable<TableWord> {
+    fn encode(&self) -> std::sync::Arc<[u8]> {
+        let boxed_slice: Box<[u8]> = self.bytes.clone().into_boxed_slice();
+        Arc::from(boxed_slice)
```
jbearer, line 52: Can't just do `Arc::from(self.bytes.clone())`?

jbearer (approved):

sequencer/src/block/tables.rs
jparr721: Ah crap sorry I meant to swap this.

PR #1382, opened 2024-04-24
Enforce max_block_size

tbro:
Omit transactions that exceed max_block_size from payload. Use mocked `max_block_size` for now. 

Depends on:  https://github.com/EspressoSystems/HotShot/pull/3096
Related to: https://github.com/EspressoSystems/hotshot-builder-core/pull/122

### DIFF
--- a/sequencer/src/block/payload.rs
+++ b/sequencer/src/block/payload.rs
@@ -18,7 +18,6 @@ use serde::{Deserialize, Serialize};
 use snafu::OptionExt;
 use std::default::Default;
 use std::mem::size_of;
-use std::ops::AddAssign;
 use std::{collections::HashMap, fmt::Display};
 use trait_set::trait_set;
 
@@ -166,12 +165,17 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
         for tx in txs.into_iter() {
             block_size += (tx.payload().len() + size_of::<TxTableEntry>()) as u64;
 
-            // block_size is updated when we encounter new namespaces
-            Payload::<TableWord>::update_namespace_with_tx(&mut namespaces, tx, &mut block_size);
+            // block_size is updated when we encounter a new namespace
+            if namespaces.get(&tx.namespace()).is_none() {
+                block_size += size_of::<TxTableEntry>() as u64;
+            }
 
-            if block_size >= chain_config.max_block_size() {
+            if block_size > chain_config.max_block_size() {
+                dbg!(block_size);
                 break;
             }
+
+            Payload::<TableWord>::update_namespace_with_tx(&mut namespaces, tx);
         }
 
         structured_payload.generate_raw_payload(namespaces)?;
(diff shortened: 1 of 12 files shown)


sveitser:
As this is a distinct / new feature for of our block building logic we should add a unittest that fails if we don't enforce the max block size.

sveitser:
@tbro anything blocking more progress here?

sequencer/src/block/payload.rs
```diff
@@ -155,13 +152,19 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
...
+        chain_config: &ChainConfig,
     ) -> Result<Self, Error> {
         let mut namespaces: HashMap<NamespaceId, NamespaceInfo> = Default::default();
         let mut structured_payload = Self {
             raw_payload: vec![],
             ns_table: NameSpaceTable::default(),
         };
+        let mut block_size = 0u64;
         for tx in txs.into_iter() {
+            block_size += tx.payload().len() as u64;
```
jbearer, line 164: This is not quite right as the payload contains other metadata, like a transaction table for each namespace, that could cause it to exceed the max block size even if the sum of transaction payloads doesn't.

Not sure of the best solution, maybe @ggutoski has some ideas. One idea could be to add a `len()` function to `NamespaceInfo`, which would return something like `TxTableEntry::byte_len() + tx_table.len() + tx_bodies.len()` (the first term is for the one word used to encode the length of the tx table). Then we would basically loop until `namespaces.values().map(NamespaceInfo::len).sum() > chain_config.max_block_size()`.

sequencer/src/header.rs
```diff
@@ -355,8 +368,7 @@ impl BlockHeader<SeqTypes> for Header {
             validated_state,
             instance_state.chain_config,
         )
-        // TODO we should be able to return an error from `Header::new`
-        .unwrap_or_else(|err| panic!("invalid proposal: {err:#}"))
+        .map_err(|e| InvalidBlockHeader::new(e.to_string()))
```
jbearer, line 371: For formatting anyhow errors, prefer `format!("{e:#}")` instead of `e.to_string()`. The former will print the whole chain of errors/contexts while the latter only prints the outermost

sequencer/src/block/payload.rs
ggutoski: I'm currently rewriting everything in `block`. I can definitely accommodate this. For this PR I suggest you do as little as possible to get it working.

It should be possible to keep a running count of the block byte length for each new transaction *that includes tx table sizes*, then just stop when we hit the limit.

(Note that a single large tx could cause a lot of wasted block space if it happens to be the one that triggers `break`. We might need to do something more intelligent in the future.)

Happy to hop on a call to nail this down.

ggutoski: Now that I think of it, maybe this is easy. Just do
```rust
block_size += tx.payload.len() + [size of a tx table entry];
```

ggutoski: that's a tiny amount of data, but it could in principle push you over the limit. You could account for it on-the-fly by doing `block_size += size_of::<TxTableEntry>()` each time you encounter a *new* namespace. Unfortunately that's wrapped up in `update_namespace_with_tx`, so you might need to hack that method to get what you want. I suggest you don't worry about ugly code for this PR though because it will all be replaced soon.

tbro: I'm still going over the tests to make sure they are still valid, but I *think* 61dd04f5b9a2db478af802ac29ac8344e8824ac6 gets us what we need. Note that I removed a `TODO` that seemed relevant.

ggutoski: i like it!

sequencer/src/block/payload.rs
```diff
@@ -159,13 +161,17 @@ impl<TableWord: TableWordTraits> Payload<TableWord> {
...
+
         let mut block_size = 0u64;
         for tx in txs.into_iter() {
-            block_size += tx.payload().len() as u64;
+            block_size += (tx.payload().len() + size_of::<TxTableEntry>()) as u64;
+
+            // block_size is updated when we encounter new namespaces
+            Payload::<TableWord>::update_namespace_with_tx(&mut namespaces, tx, &mut block_size);
+
             if block_size >= chain_config.max_block_size() {
```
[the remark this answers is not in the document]
tbro: A yes. hummm its a complicated bit of spaghetti to try and reason about.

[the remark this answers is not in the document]
tbro: The only thing I can think of is just to check the hashmap and see if the namespace of the current tx is present or not. Its redundant b/c `update_namespace_with_tx` does that check again, but if @ggutoski is going to give all this an overhaul anyway maybe it doesn't matter.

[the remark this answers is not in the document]
tbro: fixed here :crossed_fingers: b849d445058bb070e52b9229ea3c16c0a2bcbcc7

PR #1385, opened 2024-04-25
Move default ports out of private/dynamic port range

### LINKED ISSUE #1194: Move default ports out of private/dynamic port range
We should use ports in the user port range and avoid conflicts with ports of applications we use. For example we could replace the `500xy` with `440xy`.
      
      Port numbers are assigned in various ways, based on three ranges: System
      Ports (0-1023), User Ports (1024-49151), and the Dynamic and/or Private
      Ports (49152-65535); the different uses of these ranges are described in
      [RFC6335]. According to Section 8.1.2 of [RFC6335], System Ports are
      assigned by the "IETF Review" or "IESG Approval" procedures described in
      [RFC8126]. User Ports are assigned by IANA using the "IETF Review" process,
      the "IESG Approval" process, or the "Expert Review" process, as per
      [RFC6335]. Dynamic Ports are not assigned.

See https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/CI.3A.20native.20demo.20issues

TropicalDog17:
Closes #1194


tbro (approved):
Thank you for your contribution @TropicalDog17

PR #1390, opened 2024-04-26
Enforce base fee

jbearer:
* Add logic in `validate_proposal` to enforce that the provided fee is at least the base fee
* Enable base fees in the demo, enforcing in the sequencers and configuring the builder to pay the base fee
* Deploy bridge contract before starting sequencers, read deposits from here
* Add a tool to work with the bridge, use this to fund the builder instead of prefunded accounts

The new demo configuration should work on a real network like Sepolia or Ethereum, and demonstrates the startup order/dependencies.

Draft because I think this should go in after the latest builder and HotShot updates, but I don't know how to do those updates.

# Testing

A good end-to-end test is to run `demo-native`, convince yourself that it's working as you normally would (look at block height, submit-transactions logs, etc) and then watch the balance of the builder account and burner account:
* `cargo run --release --all-features --bin bridge -- balance -e http://localhost:44001 -a 0x23618e81E3f5cdF7f54C3d65f7FBc0aBf5B21E8f`
* `cargo run --release --all-features --bin bridge -- balance -e http://localhost:44001 -a 0x0000000000000000000000000000000000000000`

The former should be steadily decreasing while the latter is steadily increasing, as the builder pays fees


move47 (commented):
/builder changes looks good.

move47:
I think this can go first and later we can update with hotshot 0.5.45(needs to release tag) which will have simplify builder interaction, and update builder along with that.

jbearer:
We need to wait until @Ancient123 is ready for the new contracts deployment order

docker-compose.yaml
```diff
@@ -29,6 +43,26 @@ services:
         condition: service_healthy
       orchestrator:
         condition: service_healthy
+      # Make sure this doesn't start until the other contracts have been deployed, since we use the same mnemonic.
+      deploy-sequencer-contracts:
+        condition: service_completed_successfully
+
+  fund-builder:
+    image: ghcr.io/espressosystems/espresso-sequencer/bridge:main
```
sveitser, line 51: Changes for the new docker image are missing in this PR.

sequencer/src/bin/bridge.rs
```diff
@@ -0,0 +1,343 @@
...
+type EspressoClient = surf_disco::Client<ClientError, SequencerVersion>;
+
+type FeeMerkleProof = MerkleProof<FeeAmount, FeeAccount, Sha3Node, { FeeMerkleTree::ARITY }>;
+
+/// Command-line utility for working with the Espresso bridge.
+#[derive(Debug, Parser)]
+enum Command {
+    Deposit(Deposit),
+    Balance(Balance),
+    L1Balance(L1Balance),
```
sveitser, line 34: Since most of the arguments are the same I wonder why not put the command specific arguments into a `subcommand` and the rest into a top level struct.

docker-compose.yaml
jbearer: oops, thanks

sequencer/src/bin/bridge.rs
jbearer: I don't think any of the arguments are exactly the same across all three commands unfortunately. `mnemonic` and `account_index` are shared by all of them, but they are required for `deposit` and optional for the balance commands (can just specify a public address instead).

I checked and `cast` seems to work the same way (only top-level options are `-h` and `-v`, everything else is under a sub-command), probably for similar reasons.

docker-compose.yaml
jbearer: Added in [97fdcf1](https://github.com/EspressoSystems/espresso-sequencer/pull/1390/commits/97fdcf175d02e15a6a42f40e32d1cc4f69c9adae)

sequencer/src/bin/bridge.rs
```diff
@@ -0,0 +1,343 @@
...
+/// Deposit ETH from the L1 into Espresso.
+#[derive(Debug, Parser)]
+struct Deposit {
+    /// L1 JSON-RPC provider.
+    #[clap(short, long, env = "L1_PROVIDER")]
+    rpc_url: Url,
+
+    /// Espresso query service provider.
+    ///
+    /// This must point to an Espresso node running the node, availability, and Merklized state
```
tbro, line 46: I think the second use of "node" here is ambiguous.

jbearer: Maybe I can put `/node` to make it clearer?

scripts/smoke-test-demo
```diff
@@ -29,17 +29,31 @@ function wait_for() {
 	echo "Endpoint $what @ $url is Ok, after $elapsed seconds"
 }
 
+# usage: get_balance <address> <block>
+function get_balance() {
+	(
+		unset MNEMONIC
+		cargo run --release --all-features --bin bridge -- balance -e $SEQUENCER_API -a $1 -b $2
```
sveitser, line 36: Not a huge deal but I think it's kind of nice that this script is currently portable and doesn't need a rust toolchain. Maybe it would be useful to expose reading the bridge APIs via the explorer then we could query that instead. If we do add the `cargo run` here we should probably add rust-cache to the `test-demo` github action workflow.

sequencer/src/chain_config.rs
```diff
@@ -1,58 +1,73 @@
...
+    /// tested, this may be made mandatory.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_FEE_CONTRACT_PROXY_ADDRESS")]
+    pub fee_contract: Option<Address>,
+    /// Account that receives sequencing fees.
+    #[clap(
+        long,
+        env = "ESPRESSO_SEQUENCER_FEE_BURN_ACCOUNT",
+        default_value = "0x0000000000000000000000000000000000000000"
+    )]
+    pub fee_burn_account: FeeAccount,
```
sveitser, line 58: Do we need a configurable burn account or could we just use burn to the address if the fee_contract is None? If this is too implicit then suggest we at least make it impossible to have both a burn address and the fee contract address set.

scripts/smoke-test-demo
jbearer: What do you mean exactly by reading the bridge APIs via the explorer? 

Another approach could be to use `bridge` as an executable for the native demo (since the native demo already relies on cargo targets being added to `$PATH`) and use the `bridge` docker for the docker demo, which would also have the nice benefit of testing that the new docker is built correctly. I could add a CLI flag to this script to choose the bridge executable

sequencer/src/chain_config.rs
jbearer: Oh, the intent was that this parameter is orthogonal to the fee contract being set: when fees are paid in Espresso, we always burn to this address, and then, whenever we deploy a fee contract, it is responsible for figuring out how to handle "withdrawals" from the burn address...ie what to do retroactively with burned tokens.

So it's important that this is set if the fee contract is not so we don't lose tokens forever before we've deployed the contract, but it is also meaningful if the contract is set. I can add comments to clarify

jbearer: See [2ac26c0](https://github.com/EspressoSystems/espresso-sequencer/pull/1390/commits/2ac26c031bb6c55f818660fe482c8ecbc8232093)

sequencer/src/bin/bridge.rs
jbearer: See [2ac26c0](https://github.com/EspressoSystems/espresso-sequencer/pull/1390/commits/2ac26c031bb6c55f818660fe482c8ecbc8232093)

tbro: Maybe. Thinking about it... it isn't really ambiguous, b/c it is referring to the `NodeApi`. But it still feels confusing b/c "node" is such a generic word. I suggest we capitalize "Node" and "Availability", but also it is a *nit* so not worth blocking over.

sequencer/src/state.rs
```diff
@@ -157,17 +172,21 @@ impl ValidatedState {
...
-            format!("missing account state for {account}")
-        );
-        // Fail if there was an error during `update_with`, otherwise succeed.
+
+        // Fail if there was an error during `persistent_update_with` (e.g. insufficient balance).
         if let Some(err) = err {
-            Err(err)
-        } else {
-            Ok(())
+            return Err(err);
```
tbro, line 178: Not really an issue w/ this PR, but I wonder if `persistent_update_with` would take a closure that returns a `Result` instead of an `Option`. I *think* we could then bubble the error up to the caller instead of mutating `err`. This code *appears* to demonstrate that we do need to handle more than just the `Some` and `None` cases. I might be confused though.

jbearer: Yeah I think that would be cleaner. This pattern is basically just implementing that in a messier way. @mrain maybe you can clean this up in jf-merkle-tree whenever you find yourself with some spare time?

scripts/smoke-test-demo
sveitser: I mean if we had a HTTP API that allowed to query the bridge balances either run by the hotshot node or via the block explorer that would be useful here and I think it would be convenient with anyone interacting with our sequencer too but it's also somewhat out of scope.

We could also call `bridge` if it's in the PATH and otherwise try the docker container then we don't need a flag.

sequencer/src/chain_config.rs
sveitser: Thanks for the explanations. I think I was really confused by the term "burn" and hadn't read the implementation in detail yet. In Ethereum burn usually means making the funds inaccessible by sending it to an address for which no-one has the private keys (or somewhat more involved actually making them vanish). Can we call it fee recipient account?

sequencer/src/l1_client.rs
```diff
@@ -164,6 +163,7 @@ impl L1Client {
     /// and `new`. Returns `Vec<FeeInfo>`
     pub async fn get_finalized_deposits(
         &self,
+        fee_contract_address: Address,
```
sveitser, line 166: Could we pass this to the L1Client on construction instead or is that inconvenient?

sequencer/src/state.rs
```diff
@@ -1227,8 +1230,11 @@ mod test {
         let vid_common = vid_scheme(1).disperse(payload).unwrap().common;
 
         let state = ValidatedState::default();
-        let instance =
-            NodeState::mock().with_chain_config(ChainConfig::new(0, MAX_BLOCK_SIZE as u64, 0));
+        let instance = NodeState::mock().with_chain_config(ChainConfig {
+            max_block_size: MAX_BLOCK_SIZE as u64,
+            base_fee: 0.into(),
```
sveitser, line 1287: 0 is the already default

sequencer/src/state.rs
```diff
@@ -1202,4 +1271,92 @@ mod test {
...
+        let instance = NodeState::mock().with_chain_config(ChainConfig {
+            base_fee: 1000.into(), // High base fee
+            max_block_size,
+            ..Default::default()
+        });
+        let parent = Leaf::genesis(&instance);
+        let header = parent.get_block_header();
+
+        // Validation fails because the genesis fee (0) is too low.
+        let err = validate_proposal(&state, instance.chain_config, &parent, header, &vid_common)
```
sveitser, line 1318: Inspect the error to ensure that it failed because of the base fee being too low?

sequencer/src/state.rs
```diff
@@ -1202,4 +1271,92 @@ mod test {
...
+
+        tracing::info!("test insufficient balance");
+        state.burn_fee(fee_info, dst).unwrap_err();
+        assert_eq!(state.balance(src), Some(0.into()));
+        assert_eq!(state.balance(dst), Some(amt));
+
+        tracing::info!("test src not in memory");
+        let mut state = new_state();
+        state.fee_merkle_tree.forget(src).expect_ok().unwrap();
+        state.burn_fee(fee_info, dst).unwrap_err();
```
sveitser, line 1354: Here as well I think if we have tests that are trying to hit a specific type of validation error we should make sure that it did hit the right type of validation error. Otherwise it's easy for tests to no longer fail for the right reasons as the code base evolves. Conversely I think it's preferable to have well defined error types instead of using `anyhow` so I think we should move away from using ad-hoc anyhow errors for all our validation code. If there's potential consensus on this let's open an issue and redo all the validation errors in a separate PR.

sequencer/src/state.rs
```diff
@@ -202,15 +221,24 @@ pub fn validate_proposal(
         )
     );
 
+    // validate block size and fee
+    let block_size = VidSchemeType::get_payload_byte_len(vid_common) as u64;
     anyhow::ensure!(
-        (VidSchemeType::get_payload_byte_len(vid_common) as u64)
-            < expected_chain_config.max_block_size(),
+        block_size < expected_chain_config.max_block_size,
```
sveitser, line 227: This was added some time ago but wouldn't it be more intuitive to make the range inclusive so that a block of `max_block_size` is still allowed? If yes, let's add an issue.

sequencer/src/state.rs
```diff
@@ -90,6 +94,14 @@ impl ValidatedState {
         self.fee_merkle_tree.update(account, amount).unwrap();
     }
 
+    pub fn balance(&mut self, account: FeeAccount) -> Option<FeeAmount> {
```
sveitser, line 97: This function is only used in tests (unless my rust-analyzer is tripping). I'm not sure returning `None` is a good idea for the `NotInMemory` case, but if we make it test code only then I think it's not a problem.

sveitser (changes requested):
The request for change is because I think we should avoid the term "burn" unless we do light things on fire.

The other comments are just suggestions / nitpicky stuff.

sequencer/src/l1_client.rs
jbearer: I was trying to isolate state into the `ChainConfig` struct. Eventually, the validation logic will need to keep track of two different chain configs and switch between them when we reach an upgrade. This makes it easy to just pass the new fee address into the L1 client, rather than having to mutate or create a new L1 client.

scripts/smoke-test-demo
jbearer: > I mean if we had a HTTP API that allowed to query the bridge balances either run by the hotshot node or via the block explorer that would be useful here and I think it would be convenient with anyone interacting with our sequencer too but it's also somewhat out of scope.

I agree, but yeah, I think adding this to the block explorer is a lot of work. We do have an HTTP API (that is how the `bridge` tool gets its information) but it outputs a merkle path and the balance is base64 encoded which makes it a bit annoying to work with, hence the `bridge` tool.

sequencer/src/state.rs
jbearer: Yeah, I think let's do this in a separate PR

jbearer: https://github.com/EspressoSystems/espresso-sequencer/issues/1463

sequencer/src/state.rs
sveitser: I guess it's fine to show that we really need `0` in this test.

sequencer/src/chain_config.rs
jbearer: Done [4d34e25](https://github.com/EspressoSystems/espresso-sequencer/pull/1390/commits/4d34e25bd278d5cd707c3a18f444c84829ecaa51)

scripts/smoke-test-demo
jbearer: [61c3086](https://github.com/EspressoSystems/espresso-sequencer/pull/1390/commits/61c3086869a5f30324df84cc1b6b9b646fb57041)

sequencer/src/state.rs
jbearer: https://github.com/EspressoSystems/espresso-sequencer/issues/1464

sequencer/src/state.rs
jbearer: I know we don't have a stable public Rust API yet but I feel like a function like this should be part of it. Could easily be useful outside tests for clients and apps built on top of this. What do you think such a public function should do in the `NotInMemory` case? Should it return `Result` instead of `Option`?

sveitser (approved):

PR #1402, opened 2024-04-30
Print deployer address from deploy script

jbearer:


Ancient123 (approved):

sequencer/src/bin/deploy.rs
```diff
@@ -109,6 +109,15 @@ async fn main() -> anyhow::Result<()> {
     let owner = wallet.address();
     let l1 = Arc::new(SignerMiddleware::new(provider, wallet));
 
+    // As a sanity check, check that the deployer address has some balance of ETH it can use to pay
+    // gas.
+    let balance = l1.get_balance(owner, None).await?;
+    ensure!(
+        balance > 0.into(),
```
tbro, line 116: Could we add configuration item where we pass in a minimum expected cost? Maybe not a priority at this time, but this seems like there is still a lot of room for failure.

jbearer: Yeah, I just don't know that we would ever use this feature because it's hard to know the minimum cost without simulating the whole deployment. This is mostly just to fail fast if we are using some random address that has never been funded

PR #1409, opened 2024-05-01
Set ESPRESSO_SEQUENCER_IS_DA=true in docker demo

sveitser:
- In order for the `is_da` variable to be set to true the env var cannot be empty.
- Log the non-sensitive sequencer options on startup instead of only the modules. This should make it easier to spot configuration errors.
- Log with info level instead of warn since they aren't indications of anything potentially being wrong.


sequencer/src/main.rs
```diff
@@ -18,10 +18,10 @@ async fn main() -> anyhow::Result<()> {
     setup_logging();
     setup_backtrace();
 
-    tracing::warn!("sequencer starting up");
```
sveitser, line 21: @jbearer just saw you recently changed this. Feels sort of wrong to log with warn or error level if there's no cause for alarm. Wondering why this was changed.

rob-maron:
This is my fault, I incorrectly assumed it was pulling from the ".env", but I had removed it from there

rob-maron (approved):

sequencer/src/main.rs
```diff
@@ -18,10 +18,10 @@ async fn main() -> anyhow::Result<()> {
...
+    tracing::info!("sequencer starting up");
     let opt = Options::parse();
+    tracing::info!("options: {:?}", opt);
```
jbearer, line 23: For this one I think `warn!` does make sense, as we would almost always want this to show up in the logs even if we are trying to hide other INFO level messages.

jbearer (approved):
Nice!

sequencer/src/main.rs
sveitser: Ok I'll change it back

sveitser: https://github.com/EspressoSystems/espresso-sequencer/pull/1409/commits/69a8b5fd1619faa1d41802f07de8eb753ddc1634

Issue #1401, opened 2024-04-30
Add CI job to check compilation works with tokio


sveitser:
I think running `cargo check` is enough here.

```
env RUSTFLAGS='--cfg async_executor_impl="tokio" --cfg async_channel_impl="tokio"' cargo check
```

Not a high priority now because we'll make it less easy to mess up the RUSTFLAGS.

sveitser:
Comment from @jbearer: ... if we want to run with tokio, we need to replace all our async_std::task_spawn with async_spawn from the compatibility layer, and similarly with all of our other explicit async_std references.

sveitser:
We are running with tokio now

PR #1414, opened 2024-05-02
Set working RUSTFLAGS in .cargo/config.toml

### LINKED ISSUE #845: Build fails => see screenshot attached
Screenshot 2023-12-13 at 02 35 35

I'm struggling to setup a HotShot node. Ideally I'd like to run the espresso-sequencer without smart contracts and just test the submission & query functionality exposed by the API.  Is someone available to quickly walk me through these steps? 

Help would be much appreciated. Thanks in advance!

sveitser:
Closes #845

### This PR:
Avoids developers running into compilation errors due to RUSTFLAGS.

The RUSTFLAGS in `.cargo/config.toml` will be used for all rustc invocations unless the RUSTFLAGS env var is set. In that case the env var will override the value in the config file.

Removes setting of `--cfg hotshot_example` (except for on the CI). This does lead to some warnings when compiling the tests with `--release` but requires less configuration. I think the main problem here is that we need to run tests with `--release` and we should instead fix that by for example enabling optimization of dependencies and leave the warning there until we fix it.

### This PR does not:
Fix issues with compilation with tokio. Tracked in #1401

### Key places to review:
It would be good if a vscode user could test if the changes to the settings are okay. cc @ImJeremyHe

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,0 @@
-[net]
-git-fetch-with-cli = true
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[net]
+git-fetch-with-cli = true
+
+[build]
+rustflags =  ['--cfg', 'async_executor_impl="async-std"', '--cfg', 'async_channel_impl="async-std"']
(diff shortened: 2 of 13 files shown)


sveitser:
So somehow the static nix shell does not use the rustflags set in `.cargo/config.toml` :upside_down_face:

sveitser:
This no longer needed, see #1856

PR #1417, opened 2024-05-02
Prevent possible race condition in state persistence/resetartability

jbearer:
### This PR:

Prevents a race condition where a node is shut down after saving a decided leaf but before saving the corresponding validated state. If this happens to all nodes, then there is no node able to serve catchup queries, and the system stalls.

We want to keep the asynchronous state catchup, because it is extremely simple and resilient to all kinds of failures in the underlying event processing, and because it's auxiliary in a sense: as long as we have the block data, we have all the information we need to reconstruct the state at any time, and so it makes no sense to make the state computation blocking.

In light of this, one potential solution would just make the state loading at startup wait/retry until the state reconstruction loop has reconstructed the desired state. However, a better, more generalized solution is to incorporate the state storage into the normal catchup flow, which already has a retry loop. This removes the need to load entire state snapshots into memory at all, which is very nice: it is a step towards having the in-memory `ValidatedState` being just a cache for the full persisted state. This also further unifies the restart procedure with the normal catchup procedure. In other words, it is as if a node that has restarted has merely been offline for some amount of time.

This PR has a lot of changes, but a fair bit of it is just moving things around and implementing some boilerplate traits for wrapper types. The primary semantic changes of this PR:
1. Catchup is able to fetch missing state both from peers and from the local database (if supported)
    * Most importantly, this means we don't rely on having the entire state saved the moment we start up anymore
    * This can also make catchup more efficient: previously, archival nodes that had all the information they needed locally would still reach out to other nodes during catchup
2. Serving catchup queries for other nodes can also utilize persisted state, an important bit of symmetry with (1)
3. We no longer even try to load a validated state snapshot on startup. Technically, we could still do this, and it might prevent the catchup flow from being triggered in some cases. However, it is lighter weight to load only the state we need on demand, plus by forcing everything through the same code path of catchup we can have more confidence that this code path is correct, and that we are not relying on having the full state loaded at startup more than we should be.
4. Catchup API queries no longer return the decided state if no view number is specified. This was not being used anyways.

### Key places to review:
* Fetching state for catchup is implemented in `api/sql.rs`
* New catchup client logic in `catchup.rs`
  - Pulled the body of each retry loop out into its own function, which makes it easier to write catchup wrappers/adaptors
  - Added a new catchup implementation for `Vec<T> where T: StateCatchup`, which lets us try to catchup from multiple sources (e.g. peers and local database)
  - Moved `SqlStateCatchup` here from `state.rs`
* New catchup server in `catchup.toml`, `api/endpoints.rs`

### Things tested

Removed a loop in `test_restart` after stopping consensus but before shutting down the nodes which waited for merklized state to be persisted. This loop was intended to subvert exactly the race condition described above. Previously, it was needed for the test to pass consistently, but now the test passes without it, which is good confirmation that the change is working as intended. See [6cacced](https://github.com/EspressoSystems/espresso-sequencer/pull/1417/commits/6cacceda6c69c5c93bfaa7b459f1f27857531d4f).

### DIFF
--- a/builder/src/lib.rs
+++ b/builder/src/lib.rs
@@ -392,6 +392,7 @@ pub mod testing {
             };
 
             let node_state = NodeState::new(
+                i as u64,
                 ChainConfig::default(),
                 L1Client::new(self.anvil.endpoint().parse().unwrap(), Address::default()),
                 MockStateCatchup::default(),
@@ -529,6 +530,7 @@ pub mod testing {
         ) -> Self {
             // setup the instance state
             let node_state = NodeState::new(
+                u64::MAX,
                 ChainConfig::default(),
                 L1Client::new(
                     hotshot_test_config.get_anvil().endpoint().parse().unwrap(),
@@ -590,6 +592,7 @@ pub mod testing {
         ) -> Self {
             // setup the instance state
             let node_state = NodeState::new(
+                node_id,
                 ChainConfig::default(),
                 L1Client::new(
                     hotshot_test_config.get_anvil().endpoint().parse().unwrap(),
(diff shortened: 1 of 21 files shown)


imabdulbasit (approved):
LGTM!

PR #1421, opened 2024-05-03
Keep incremental state in memory during Merklized state update loop

jbearer:
This means we don't hammer the database for the parent state with every single update.


jbearer:
idk why the docker demo is running on this branch, maybe because I named it `release/*` so we could test in staging. Should be safe to ignore that as it's only meant to work on `main`

nyospe (approved):
Took a minute to work out everything, but makes sense, and a nice optimization.

PR #1434, opened 2024-05-07
Added configuration setting for builder: initial node_count

nyospe:
Added configuration setting for builder: initial node_count, so that bootstrap doesn't use a default (1)


jbearer (approved):
LGTM

PR #1435, opened 2024-05-07
Make builder aware of maximum block size

jbearer:


nyospe (approved):
Looks fine for this part. The way we detect what transactions are included means no changes should be needed for hotshot-builder-core...

PR #1440, opened 2024-05-08
More demo fixes

jbearer:
* Move all query nodes to postgres storage. The file system storage is causing query nodes to fall behind in consensus. Not sure why, but it's not worth spending time on now as we don't plan to run query nodes with file system storage in production
* Upgrade query service to set a default timeout of 500ms instead of 0 when fetching missing resources
* Add a script to nicely format process-compose logs


tbro (approved):
LGTM

PR #1441, opened 2024-05-08
Make L1 block number configurable at genesis

jbearer:


tbro (approved):
Looks good. My only doubt is if there may be some danger in allocating a very large vector of chunks. But I don't currently have any good ideas of how to avoid it, and if it were a problem could be seen as a future optimization.

PR #1443, opened 2024-05-09
Define a genesis config file

jbearer:
This aims to separate onchain data, which must be the same for all nodes (chain ID, base fee, etc) from node-specific configuration. Node configuration continues to be done via environment variables, whereas onchain data is defined in a configuration file which we can distribute to all node operators when launching a network. This should greatly simplify coordination and reduce the frequency of errors due to misconfigured nodes.

For officially supported networks (demo, staging, testnet, and eventually mainnet) the genesis file is distributed via the sequencer docker image itself, which makes the deployment process far less error prone -- the only thing we need to get right is to ensure all node operators are using the same Docker tag. It is also a bit easier to toggle between networks now, as you can switch over all the onchain data just by changing `ESPRESSO_SEQUENCER_GENESIS_FILE` to a different bundled genesis file. You still have to change the other env vars to connect to the right orchestrator, CDN, etc though.

# Key places to review

* Changes in `genesis.rs` and `options.rs` are the meat of the change
* Serialization changes (to allow flexible deserialization from string or integer) in `utils/src/ser.rs`, `sequencer/src/state.rs`, and `sequencer/src/chain_config.rs`. In particular check the new tests, that are intended to prove this is backwards compatible with Cappuccino serialization
* Changes to `docker-compose.yaml` and `process-compose.yaml`


data/genesis.toml
```diff
@@ -0,0 +1,8 @@
+[chain_config]
```
sveitser, line 1: not JSON :raised_hands:

sequencer/src/chain_config.rs
```diff
@@ -50,9 +70,36 @@ impl ChainConfig {
             base_fee: base_fee.into(),
         }
     }
+
+    pub fn from_toml(toml: &toml::Value) -> anyhow::Result<Self> {
```
sveitser, line 74: Couldn't we use toml with serde as described here: https://docs.rs/toml/latest/toml/#deserialization-and-serialization to reduce the amount of code we need?

jbearer: We could. I wanted to allow both integers and strings with units for sizes, and ended up writing some custom code for that, and also for the `U256` stuff because the default parsing for `U256` only accepts hex strings (event without a `0x` prefix and only decimal digits, it's interpreted as hex!).

But thinking about it since we already have custom wrappers for everything (`FeeAmount`, `FeeAccount`, `ChainId`) I could probably just override the deserialization for those types specifically. I can try this when I get some time.

jbearer:
Changed base to #1390, since this was touching a lot of the same code. We'll have to merge this later anyways, since it requires a substantial reworking of the deploy config.

sequencer/src/chain_config.rs
jbearer: Done in [b193fd5](https://github.com/EspressoSystems/espresso-sequencer/pull/1443/commits/b193fd530cb0504ab7430c935f4f1a3ed2388362)

.env
```diff
@@ -30,10 +30,9 @@ ESPRESSO_SEQUENCER1_API_PORT=24001
 ESPRESSO_SEQUENCER2_API_PORT=24002
 ESPRESSO_SEQUENCER3_API_PORT=24003
 ESPRESSO_SEQUENCER4_API_PORT=24004
-ESPRESSO_SEQUENCER_MAX_BLOCK_SIZE=1mb
-ESPRESSO_SEQUENCER_BASE_FEE=1
 ESPRESSO_SEQUENCER_URL=http://sequencer0:${ESPRESSO_SEQUENCER_API_PORT}
 ESPRESSO_SEQUENCER_STORAGE_PATH=/store/sequencer
+ESPRESSO_SEQUENCER_GENESIS_FILE=/genesis
```
sveitser, line 35:
```suggestion
ESPRESSO_SEQUENCER_GENESIS_FILE=/genesis.toml
```
Not important, just looks strange to me without extension.

sequencer/src/state.rs
```diff
@@ -964,6 +965,56 @@ impl FromStr for FeeAmount {
...
+
+    fn from_binary(b: Self::Binary) -> anyhow::Result<Self> {
+        Ok(Self(b))
+    }
+
+    fn from_integer(i: Self::Integer) -> anyhow::Result<Self> {
+        Ok(i.into())
+    }
+
+    fn from_string(s: String) -> anyhow::Result<Self> {
```
sveitser, line 980: Not sure if it's better there is a function for this in ethers. 

https://github.com/gakonst/ethers-rs/blob/51fe937f6515689b17a3a83b74a05984ad3a7f11/ethers-core/src/utils/mod.rs#L245

sveitser: I think one difference is that it allows to specify fractional values, but checks that there aren't any fractional wei. So could be nice to use it to get some errors handled and the extra functionality.

jbearer: [1b771e5](https://github.com/EspressoSystems/espresso-sequencer/pull/1443/commits/1b771e5f791d2b97481a2e45ff2929e79fee8f92)

Ancient123 (approved):

data/genesis/cappuccino.toml
```diff
@@ -0,0 +1,11 @@
...
+capacity = 200
+
+[chain_config]
+chain_id = 0
+base_fee = '0 wei'
+max_block_size = '30mb'
+fee_recipient = '0x0000000000000000000000000000000000000000'
+
+[header]
+timestamp = "1970-01-01T00:00:00Z"
```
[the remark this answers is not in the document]
jbearer: yep!

PR #1444, opened 2024-05-10
chore: update Jellyfish

mrain:
No issue related.

### This PR:

- Update Jellyfish w.r.t the recent crates restructure.
- Bump Push-CDN from 0.3.2 to 0.3.6.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,17 +403,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ark-pallas"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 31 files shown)


mrain:
Jellyfish related things are all good now.

Handed over to @rob-maron for the Push-CDN related updates here: https://github.com/EspressoSystems/espresso-sequencer/pull/1444/files#diff-f9c63954d0f37c31e704df9bbab44af369313500172400b76c21f4cbe01ed390

tbro (approved):
Looks like there are not any functional changes. Just renaming imports and adjusting return values to use anyhow. So I think if tests are passing it is good enough.

PR #1456, opened 2024-05-13
Demo fixes

jbearer:
* Avoid Linux local port range (>= 32768)
* Some configuration fixes in the Docker demo


ImJeremyHe (commented):
I am reading something about choosing port.
And I found this one:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=30000

We should also avoid using `30000` according to this document.

sveitser:
I think using a part that another software uses isn't such a big problem as long as we don't run that software. The problem is with dynamic ports because on each run there's a small probability that the port is used, which leads to non-determinism. If we would happen to choose a port that is always used by software running on the github runner it would fail every time and we would know.

sveitser (approved):

sveitser (approved):

PR #1461, opened 2024-05-15
fix: correct threshold for light client state update

mrain:
No issue related.

### This PR:

- Let n be the total stake. We only need n/3+1 signatures (at least 1 signature from the honest party) to update the light client contract. 
- Correct the comment/documentation for `use_mock_contract` option.


sveitser:
Is it feasible to add a regression test for this?

mrain:
> Is it feasible to add a regression test for this?

Any suggestion for this? I don't know where to put.

sveitser:
Also to be clear, I don't intend to delay merging of this PR. It's just a suggestion to reduce tech-debt and future problems.

alxiong:
I agree with @sveitser, single responsibility principle when possible.

mrain:
> Not sure. It's kind of a "magic", it's not that obvious that the correct computation is `x / 3 + 1` to the point where we even implemented it wrongly at first. So it would be better if this was computed only in one place instead of four places (or maybe two places, with the second one being a test to verify that the first computation is correct, so that if someone changed the computation it would break that test).
> 
> Would it make sense to add a `quorum_threshold` method to the stake table instead, or do you think it doesn't belong there?

Good idea! Just implemented this.

hotshot-state-prover/src/service.rs
```diff
@@ -79,6 +79,12 @@ pub struct StateProverConfig {
     pub stake_table_capacity: usize,
 }
 
+#[inline]
+/// A helper function to compute the quorum threshold given a total amount of stake.
+pub fn compute_quorum_threshold(total_stake: U256) -> U256 {
```
jbearer, line 84: I think we should call this `one_honest_threshold`, not quorum

mrain: Right: [0a6b8b1](https://github.com/EspressoSystems/espresso-sequencer/pull/1461/commits/0a6b8b1d97294c7b3625836a6c33bec03003b261)

jbearer (approved):

PR #1465, opened 2024-05-15
Make text metrics parseable

jbearer:
Uses the new labels functionality of the metrics trait to make parseable the version information and public key. Also adds a test that these metrics are correct and available before the orchestrator starts.


philippecamacho (approved):
LGTM

PR #1467, opened 2024-05-15
remove `genesis`

tbro:
Depends on https://github.com/EspressoSystems/HotShot/pull/3164


imabdulbasit:
I think this can be closed

PR #1468, opened 2024-05-16
Fix prover in local demo

jbearer:
`ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS` was mistakenly set to the address of the light client implementation contract instead of the proxy.

This also adds a check to `smoke-test-demo` that the light client contract is being successfully updated.


philippecamacho (approved):
The native demo works correctly on my machine now.

PR #1472, opened 2024-05-16
Add skeleton types types for auctions

### LINKED ISSUE #1471: Layout requirements for handling auctions
The sequencer application should define what interfaces and what additional information it needs from consensus to validate all the requirements for the marketplace.

sveitser:
Closes #1471 

### TODO
- [ ] Sketch out validation to know if we need new information from consensus.

### This PR:
This is a very early sketch based on [the notion document](https://www.notion.so/espressosys/MVP-0-Marketplace-Sprint-Planning-f6aab8d007954b96a1a6045d69786993) for discussion and experimentation.

### This PR does not:
Implement anything.

### DIFF
--- a/sequencer/src/auction.rs
+++ b/sequencer/src/auction.rs
@@ -8,7 +8,7 @@ struct Auction {
     /// current phase.
     phase: AuctionPhase,
     /// A structure to find which Phase should be the current one.
-    // This could probably be an enum if we we structure.
+    // This could probably be an enum with some restructuring.
     possible_phases: [AuctionPhase; 3],
 }
 
--- a/sequencer/src/auction.rs
+++ b/sequencer/src/auction.rs
@@ -1,4 +1,7 @@
-use crate::{state::FeeAmount, NamespaceId, Transaction};
+use crate::{
+    state::{FeeAccount, FeeAmount},
+    NamespaceId, Transaction,
+};
 use hotshot_types::data::ViewNumber;
 use std::num::NonZeroU64;
 
@@ -72,6 +75,7 @@ struct Signature;
 struct BidTx {
     auction: AuctionId,
     amount: FeeAmount,
+    account: FeeAccount,
     namespace: NamespaceId,
     nonce: Nonce,
     signature: Signature,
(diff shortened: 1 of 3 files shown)


sequencer/src/auction.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+    fn send_txn(&self, txn: Transaction) {
+        unimplemented!();
+    }
+}
+
+// - needs to be configured in genesis block
+// - needs to be updatable
+/// Configuration for the auction system
+struct AuctionConfig {
```
tbro, line 38: I can see in a very generic way why we need this information, but I'm having trouble reasoning about how we will actually use it. I feel like we will need to receive an event announcing the start of a phase. And from there we can use this to calculate the view at which that phase will end. Are there indeterminate spans between phases? if end of one phase == start of next then we would only need the start of the first phase on the chain, which could just be a constant (or come from config).

sequencer/src/auction.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.phase = self
+            .possible_phases
+            .into_iter()
+            .find(|phase| (view > phase.start && view < phase.end))
+            .unwrap();
+    }
+    fn recv_bid(&self, view: ViewNumber, bid: BidTx) {
+        self.update(view);
+        match self.phase.kind {
+            AuctionPhaseKind::Bid => self.send_txn(bid.as_txn()),
```
tbro, line 26: Rereading, I think the actions that will actually occur are validation and persistence.

sequencer/src/auction.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Uniquely identifies one auction phase for a specific auction
+#[derive(Clone, Copy)]
+struct AuctionPhaseId(AuctionId, AuctionPhaseKind);
+
+/// Describes one auction phase for a specific auction
+#[derive(Clone, Copy)]
+struct AuctionPhase {
+    id: AuctionPhaseId,
+    kind: AuctionPhaseKind,
+    start: ViewNumber,
```
tbro, line 57: Reading over some discussion it looks like timing will be determined by block instead of view, so these might need to be updated to be `BlockHeight`.

tbro: This has since been clarfied

PR #1473, opened 2024-05-16
Wait for L1 genesis block to be finalized, not just available

jbearer:


sequencer/src/l1_client.rs
```diff
@@ -121,26 +121,32 @@ impl L1Client {
...
-                Ok(height) if height > number.into() => break,
-                Ok(height) => {
-                    tracing::warn!(number, %height, "waiting for L1 block");
-                }
-                Err(err) => {
-                    tracing::error!(%err, "failed to get L1 block height");
-                }
+        // Wait for the block to finalize.
+        let finalized = loop {
+            let Some(block) = self.get_finalized_block().await else {
```
tbro, line 131: I think using our `get_finalized_block` proxy here is causing us to do extra work. It calls external `get_finalize_block` which internally calls `provider.get_block` and pulls the block_number and hash off of the block. So for this PR my suggestion is to just use `self.provider.get_block` here to reduce the number of HTTP requests. Long term I think it points to a need to refactor L1Client to better fit it different use cases.

sequencer/src/l1_client.rs
```diff
@@ -121,26 +121,32 @@ impl L1Client {
...
             sleep(interval).await;
+            continue;
+        };
+
+        if finalized.number == number {
+            return finalized;
         }
 
-        // Get the block, retrying until we succeed.
+        // The finalized block may have skipped over the block of interest. In this case, our block
```
tbro, line 148: I'm not 100% sure if we have adopted a standard on this, but comments are consistently rolling over to the next line. I *feel* they should be 80 chars, but maybe we don't care about this :shrug:

tbro (changes requested):

sequencer/src/l1_client.rs
jbearer: I think Rust code tends to use a 100 character line length. At least that's what rustfmt does for code lines by default

tbro: Ok my mistake

sequencer/src/l1_client.rs
tbro: If we have the block in the first loop, can we not avoid the second loop entirely? This would avoid however many requests are made in the second loop. Which would probably just be 1 given that we've already shown that the block is present.

tbro: Ah no now I understand we first need to find out what the finalized block is. I was misreading.

tbro (approved):

PR #1481, opened 2024-05-17
add endpoint for hotshot config and env variables

imabdulbasit:
Closes https://github.com/EspressoSystems/espresso-sequencer/issues/1413

### This PR:

adds endpoints for
- hotshot config
- env vars


sequencer/src/api/data_source.rs
```diff
@@ -134,6 +142,71 @@ pub(crate) trait CatchupDataSource {
...
+
+        state.serialize_field("public_key", &self.0.public_key)?;
+        state.serialize_field("stake_value", &self.0.stake_value)?;
+        state.serialize_field("is_da", &self.0.is_da)?;
+
+        state.end()
+    }
+}
+
+impl Serialize for MyHotShotConfig {
```
imabdulbasit, line 169: can't think of a better way to skip field without adding it in hotshot

sveitser: I think ideally sensitive data would have a Display implementation that prints a redacted version of it.

sveitser: Or I guess that wouldn't even help here. Does it even make sense to use `Serialize` here? I think serialization usually implies that deserialization is the inverse which won't work anyway if we hide some fields.

sequencer/src/api/endpoints.rs
```diff
@@ -286,3 +289,35 @@ where
...
+pub(super) fn config<S, Ver: StaticVersionType + 'static>(_: Ver) -> Result<Api<S, Error, Ver>>
+where
+    S: 'static + Send + Sync + ReadState,
+    S::State: Send + Sync + HotShotConfigDataSource,
+{
+    let toml = toml::from_str::<toml::Value>(include_str!("../../api/config.toml"))?;
+    let mut api = Api::<S, Error, Ver>::new(toml)?;
+
+    let env_vars: Vec<(String, String)> = env::vars()
+        .filter(|(key, _)| key.starts_with("ESPRESSO_"))
```
sveitser, line 302: This feels a bit dangerous too because it's easy to include something private.

sequencer/src/api/data_source.rs
imabdulbasit: shouldn't we skip serialization for these fields in hotshot types?

imabdulbasit: but then we load the config from the storage when starting up the node

jbearer: Yeah, I think the solution to *that* problem is storing the config encrypted in the db/file system. But IMO that's much less critical than exposing private keys in plaintext *in API endpoints*, so having a solution specifically for the latter seems good for now.

FWIW I think this approach is fine. And the serialization here is bidirectional, because the "hiding" of the private keys happens on creation of the `MyHotShotConfig`: we can successfully round trip a `MyHotShotConfig`, hidden fields and all, we just can't get back a `HotShotConfig` from it, which is fine.

jbearer: My only concerns:
* by specifically selecting which fields to hide, it seems very easy to accidentally miss something and leak secret data over a public API. While it is going to be messier, I think it would be safer to take the opposite approach and specifically select the fields we want to include. If we do that, we might as well just make it a proper struct with all the fields we want from the `HotShotConfig`, instead of a `json::Value` wrapper
* We should call it `PublicHotShotConfig` or something, `My` doesn't mean anything

sequencer/src/api/endpoints.rs
jbearer: Agree, I'm afraid here we also need a hard-coded list of the variables we want to pull :(

sequencer/src/api.rs
```diff
@@ -239,6 +239,18 @@ impl<N: network::Type, Ver: StaticVersionType + 'static, P: SequencerPersistence
     }
 }
 
+impl<
```
jbearer, line 242: Whenever we are running, we have the config in memory. So we could consider putting the config in `SequencerContext` and implementing this also for non-storage nodes, the same way we pull other things out of the context like the state signer and event streamer

jbearer (changes requested):
Thank you so much for doing this! I have wanted this for so long lol

sequencer/src/api.rs
imabdulbasit: SystemContextHandle already has that field so I think I can get it from there too

sequencer/src/api/endpoints.rs
imabdulbasit: I was thinking of adding a function to crate::Options and its modules for public env vars but then each bin also has options, and we might add some binaries in the future. How about a file containing all the variables for this endpoint?

jbearer: I like it

sequencer/src/api/data_source.rs
```diff
@@ -141,36 +145,109 @@ pub(crate) trait CatchupDataSource {
...
-        let validator_config = value
-            .get_mut("my_own_validator_config")
-            .context("my_own_validator_config not found")?;
-
-        let private_key = validator_config
-            .get_mut("private_key")
-            .context("private_key not found")?;
-        *private_key = "******".into();
+#[derive(Debug, Serialize)]
+pub struct PublicValidatorConfig {
```
jbearer, line 149: We should have a comment explaining what this is/why it exists

sequencer/src/api/data_source.rs
```diff
@@ -141,36 +145,109 @@ pub(crate) trait CatchupDataSource {
...
+    pub public_key: PubKey,
+    pub stake_value: u64,
+    pub is_da: bool,
+    pub private_key: &'static str,
+    pub state_key_pair: &'static str,
```
jbearer, line 154: We can at least display the public state key (`state_key_pair.ver_key()`) in plaintext

sequencer/src/api/data_source.rs
```diff
@@ -141,36 +145,109 @@ pub(crate) trait CatchupDataSource {
...
 }
 
-impl Serialize for MyHotShotConfig {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        self.0.serialize(serializer)
+#[derive(Debug, Serialize)]
+pub struct PublicHotShotConfig {
```
jbearer, line 178: Comment

sequencer/src/api/data_source.rs
```diff
@@ -141,36 +145,109 @@ pub(crate) trait CatchupDataSource {
...
+    pub builder_timeout: Duration,
+    pub data_request_delay: Duration,
+    pub builder_url: Url,
+}
+
+impl From<HotShotConfig<PubKey>> for PublicHotShotConfig {
+    fn from(v: HotShotConfig<PubKey>) -> Self {
+        // Destructure all fields from HotShotConfig to return an error
+        // if new fields are added to HotShotConfig. This makes sure that we handle
+        // all fields appropriately and do not miss any updates.
```
jbearer, line 205: 👏 👏

sequencer/src/api/options.rs
```diff
@@ -215,6 +215,10 @@ impl Options {
                 )?;
             }
 
+            if self.config.is_some() {
+                app.register_module("config", endpoints::config(bind_version)?)?;
+            }
```
jbearer, line 220: Could this be moved into `init_hotshot_modules`?

public-env-vars.toml
```diff
@@ -0,0 +1,110 @@
...
+    "ESPRESSO_CDN_BROKER_METRICS_BIND_ENDPOINT",
+    "ESPRESSO_CDN_BROKER_CA_CERT_PATH",
+    "ESPRESSO_CDN_MARSHAL_DISCOVERY_ENDPOINT",
+    "ESPRESSO_CDN_MARSHAL_BIND_PORT",
+    "ESPRESSO_CDN_MARSHAL_METRICS_BIND_ENDPOINT",
+    "ESPRESSO_CDN_MARSHAL_CA_CERT_PATH",
+    "ESPRESSO_CDN_WHITELIST_DISCOVERY_ENDPOINT",
+    "ESPRESSO_SEQUENCER_URL",
+    "ESPRESSO_SEQUENCER_HOTSHOT_ADDRESS",
+    "ESPRESSO_SEQUENCER_HOTSHOT_ACCOUNT_INDEX",
```
jbearer, line 54: I don't think we should expose these. Technically it's useless without the mnemonic, but it is one of the inputs that feeds into the private key derivation so feels a little too close to a secret

public-env-vars.toml
```diff
@@ -0,0 +1,110 @@
...
+
+
+variables = [
+    "ESPRESSO_SEQUENCER_ORCHESTRATOR_URL",
+    "ESPRESSO_SEQUENCER_CDN_ENDPOINT",
+    "ESPRESSO_SEQUENCER_LIBP2P_BIND_ADDRESS",
+    "ESPRESSO_SEQUENCER_LIBP2P_ADVERTISE_ADDRESS",
+    "ESPRESSO_STATE_RELAY_SERVER_URL",
+    "ESPRESSO_SEQUENCER_PREFUNDED_BUILDER_ACCOUNTS",
+    "ESPRESSO_SEQUENCER_L1_PROVIDER",
```
jbearer, line 14: We should probably hide this as the L1 RPC URL will often have an API key in it

sequencer/api/config.toml
```diff
@@ -6,4 +6,4 @@ DOC = "Get the Hotshot configuration for the current node."
 [route.env]
 PATH = ["/env"]
 METHOD = "GET"
-DOC = "Get all ESPRESSO environment variables set for the current node."
\ No newline at end of file
+DOC = "Get all ESPRESSO environment variables defined in public-env-vars.toml file for the current node"
```
jbearer, line 9: I think the documentation is fine the way it was before. This will be exposed publicly in the auto-generated docs page, and users of this API won't necessarily know what `public-env-vars.toml` is

public-env-vars.toml
```diff
@@ -0,0 +1,110 @@
+# TOML file for the `/env` route of Config endpoint
```
jbearer, line 1: Maybe put this file in `sequencer/api` next to the other API toml specs?

public-env-vars.toml
imabdulbasit: remove all these? or only the L1?

jbearer: All the L1_PROVIDER URLs

public-env-vars.toml
jbearer: Sorry, I meant to hide any account index variables. I think you deleted the 4 lines shown here but I only meant to comment on 54 (github just shows the rest for context)

imabdulbasit: np I'll add those back. Is the ACCOUNT_INDEX different from this one?

public-env-vars.toml
imabdulbasit: does the ESPRESSO_PROVIDER variable needs to be removed as well?

PR #1482, opened 2024-05-17
Use the correct validated state on restart

jbearer:


tbro (approved):
Looks like it conforms to the discussion I saw in chat.

PR #1486, opened 2024-05-20
Add/improve tests for serialization and commitment compatibility

jbearer:
No issue, this is necessary now that we are supporting a deployed network (Cappuccino). We need to avoid breaking changes or at least be conscious of when we are making them.

### This PR:
* Adds round-trip serialization check to reference data structure tests
* Prints out a valid serialization if compat tests fail, so that we can more easily update the test vectors when we change data structures
* Generates test vectors from Rust objects instead of hard-coding JSON
* Adds reference tests for ChainConfig, FeeInfo
* Adds a new test checking serialization compatibility for HotShot messages

### This PR does not:

Catch all possible instances of compatibility breaking changes. The checks in here are based on serialization of specific instantiations of various data structures. Compatibility breakages that only affect variants/structures that are not specifically tested will not trigger a failure. Short term, a combination of these tests and running in staging with multiple different versions at the same time should be sufficient to catch the most common and severe breakages. Long term, we need a deterministic schema for all serialized data structures so we can symbolically check every possible variant for breaking changes.

### Things tested

These tests pass on the release tag used for Cappuccino (`20240517`). Thus keeping these tests passing without changes should indicate compatibility with the code that is currently deployed. Branch with these tests cherry-picked on top of `20240517`: https://github.com/EspressoSystems/espresso-sequencer/tree/jb/backport-ser-test


sequencer/src/message_compat_tests.rs
```diff
@@ -0,0 +1,243 @@
...
+        .map(MessageKind::Consensus)
+        .chain(data_messages.into_iter().map(MessageKind::Data))
+        .map(|kind| Message { kind, sender })
+        .collect::<Vec<Message<SeqTypes>>>();
+
+    // Ensure the current serialization implementation generates the same JSON as the committed
+    // reference.
+    let actual = serde_json::to_value(&messages).unwrap();
+    let expected: Value = serde_json::from_str(include_str!("../../data/messages.json")).unwrap();
+    assert_eq!(
```
sveitser, line 222: I think we can make the this a bit more convenient for our futures selves. For example we could use https://github.com/rust-pretty-assertions/rust-pretty-assertions, with that if we compare the json strings instead of the `Value`s the output looks pretty nice and indicates where the difference is.

However the output is still pretty big, so we could also write the new JSON out to a file next to `messages.json` which makes it easier to manually diff it and update the file in the repo if desired.

sequencer/src/message_compat_tests.rs
```diff
@@ -0,0 +1,243 @@
...
+    let messages = seq_messages
+        .map(MessageKind::Consensus)
+        .chain(data_messages.into_iter().map(MessageKind::Data))
+        .map(|kind| Message { kind, sender })
+        .collect::<Vec<Message<SeqTypes>>>();
+
+    // Ensure the current serialization implementation generates the same JSON as the committed
+    // reference.
+    let actual = serde_json::to_value(&messages).unwrap();
+    let expected: Value = serde_json::from_str(include_str!("../../data/messages.json")).unwrap();
```
sveitser, line 221: Is there any benefit to including the json file in the test binary here?

jbearer: Yeah if we load it at runtime then we have to run the binary from the root of the repo. If you e.g. run `cargo test` from the `sequencer` sub-directory you would get a confusing failue

sequencer/src/message_compat_tests.rs
jbearer: Love it

sequencer/src/message_compat_tests.rs
jbearer: Actually I guess if I use `CARGO_MANIFEST_DIR` then the only requirement is you run the tests on the same machine where they were built...which seems reasonable for tests. I can do that to make the binary smaller/faster to build

jbearer: [111fe29](https://github.com/EspressoSystems/espresso-sequencer/pull/1486/commits/111fe2947c02b705bd7675713c3997bfe1499e09)

sequencer/src/message_compat_tests.rs
jbearer: [111fe29](https://github.com/EspressoSystems/espresso-sequencer/pull/1486/commits/111fe2947c02b705bd7675713c3997bfe1499e09)

sequencer/src/reference_tests.rs
```diff
@@ -0,0 +1,245 @@
...
+    setup_logging();
+    setup_backtrace();
+
+    // Load the expected serialization from the repo.
+    let data_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("../data");
+    let expected_bytes = std::fs::read(data_dir.join(format!("{name}.json"))).unwrap();
+    let expected: Value = serde_json::from_slice(&expected_bytes).unwrap();
+
+    // Check that the reference object matches the expected serialized form.
+    let actual = serde_json::to_value(&reference).unwrap();
```
ss-es, line 137: Is there a reason to use `serde_json` rather than `vbs::Serializer::<hotshot_types::constants::Version01>::serialize` that HotShot uses?

It makes the checked in file human-readable, but if I understand correctly, I think only checking roundtrip json serialization would not catch a lot of the more subtle errors (e.g. reordering but not renaming enum variants)

jbearer: [4a30c28](https://github.com/EspressoSystems/espresso-sequencer/pull/1486/commits/4a30c28e6e506da21c61aee8d4ee95869db7e18d)

ss-es (approved):
Looks good to me!

PR #1490, opened 2024-05-21
Add Libp2p peerID derivation to public key binary

rob-maron:
No linked issue.

Adds the ability to derive a Libp2p peerID (public key) to the public key binary. This will make an upgrade to Libp2p easier as the keys need to be known to bootstrap into the network


sveitser (approved):
LGTM but I wonder if we should merge this functionality into the keygen binary instead of having two utilities for handling keys.

rob-maron:
> LGTM but I wonder if we should merge this functionality into the keygen binary instead of having two utilities for handling keys.

Yeah, I think this is worth it. I can do a follow-up PR. We can probably have it do both based on subcommands like `gen` or `print_public_key` or something

PR #1493, opened 2024-05-22
DRAFT PR: 1470 escape hatch add sliding windows to light client contract

### LINKED ISSUE #1470: Escape hatch: add sliding window(s) to light client contract
We need to store a 7 day window of historical state roots so that optimistic rollups' dispute handling contracts can access it during disputes.

alysiahuggins:
Closes #1470

### This PR:
- lagOverEscapeHatchThreshold(l1BlockHeight, delayThreshold)
- getHotShotCommitment(hotShotBlockHeight)
- getL1BlockUpdatesCount()
- getHotShotBlockCommitmentsCount()

### This PR does not:
- Purge the newly added arrays once there is 10 days of data, another issue has been created
- The edgecase: the user submits a blocknumber > block.number and the last light client update block - block.number > threshold

### Key places to review:
- lagOverEscapeHatchThreshold function
- getHotShotCommitment function

### How to test this PR:  
forge test --match-contract LightClient_StateUpdatesTest -vvv --ffi
forge test --match-contract LightClient_HotShotCommUpdatesTest -vvv --ffi

### Things tested 
- `lagOverEscapeHatchThreshold` happy paths and edge cases
- `getHotShotCommitment` functionality
- `stateUpdateBlockNumbers` is updated when the `newFinalizedState` function is successfully called
- `hotShotCommitments` is updated when the `newFinalizedState` function is successfully called


contracts/src/LightClient.sol
```diff
@@ -341,4 +367,77 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        if (newThreshold < 1 || newThreshold > type(uint32).max) {
+            revert InvalidDelayThreshold();
+        }
+        delayThreshold = newThreshold;
+        emit NewDelayThreshold(newThreshold);
+    }
+
+    /// @notice checks the L1BlockUpdates array and the delayThreshold to determine if hotshot was
+    /// down at a specified L1 block number
+    function isHotshotAlive(uint256 l1BlockNumber) public view returns (bool) {
```
ImJeremyHe, line 387: isHotShotLive?

contracts/src/LightClient.sol
```diff
@@ -84,6 +87,12 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     /// @notice a flag that indicates when a permissioned provrer is needed
     bool public permissionedProverEnabled;
 
+    /// @notice an array to store the L1 Block Heights where the finalizedState was updated
+    uint256[] public l1BlockUpdates;
+
+    /// @notice the threshold used to determine if hotshot is down
+    uint32 public delayThreshold;
```
sveitser, line 94: @alysiahuggins @nomaxg @ImJeremyHe @philippecamacho Since rollups may have different thresholds, wouldn't it make more sense to not have this variable in the light client contract? I think this contract should only store the sliding window and add a "getter" function that takes delay threshold as parameter so the rollup can call `isHotShotLive(l1BlockNum, delayThreshold)` with a delay threshold of their choice.

ImJeremyHe: @sveitser Yeah that makes sense. I think there are actually 2 questions here:
- Is the hotshot live?
- When should the rollups use the escape hatch?

For the 1st question, I think whether hotshot is live or not should merely determined by us. We are supposed to select this `delayThreshold` according to the properties of HotShot.
For the 2nd question, rollups can have their own choice. But I guess it would also be reasonable that assuming they would use the escape hatch as long as the hotshot is down

alysiahuggins: I agree with @ImJeremyHe's take

sveitser: What's the advantage of having the variable in the contract? I don't think there is any. We want rollups to be able to set the thresholds themselves. Having it in the contract only makes it inflexible (same for all rollups) and we have to make the decision what the value should be. We probably can't make a good decision on the value either because this is not a feature that we are going to use.

Maybe we should use a different name than `isHotShotLive`, but I don't think it matters that much here. We could call it `shouldTriggerEscapeHatch` for example.

cc @nomaxg @jbearer

contracts/src/LightClient.sol
```diff
@@ -341,4 +367,77 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        // The block being requested is in the future so return false
+        // if(l1BlockNumber>block.number){
+        //     return true;
+        // }
+
+        // If there is only one block recorded then use the block ledger's current block as the
+        // 'nextBlock'
+        if (l1BlockUpdates.length == 1) {
+            if (
+                block.number - l1BlockUpdates[0] >= delayThreshold
```
nomaxg, line 397: Why are both these checks necessary? Isn't it sufficient to check that `l1BlockNumber - l1BlockUpdates[0] >= delayThreshold`

contracts/src/LightClient.sol
```diff
@@ -341,4 +367,77 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+            prevBlock = l1BlockUpdates[i];
+            nextBlock = l1BlockUpdates[i + 1];
+
+            //if we're at the last item in the array, the nextBlock is the block ledger's block
+            // number
+            if (i == l1BlockUpdates.length - 1) {
+                nextBlock = block.number;
+            }
+
+            if (prevBlock <= l1BlockNumber && nextBlock >= l1BlockNumber) {
```
nomaxg, line 422: I think you can condense the first [check](https://github.com/EspressoSystems/espresso-sequencer/pull/1493/files#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292R395) with this loop in the following way:
1. Scan from end of the block updates array until you find the first block less than l1BlockNumber at index i. (edge case: no such block is found, this means that you are querying for a block too early in history, throw an error). 
2. prevBlock = l1BlockUpdates[i]
3. nextBlock = l1BlockUpdates[i+1] OR block.number if i+1 is out of bounds. 
4. If `l1BlockNumber - prevBlock >= delayThreshold`, return false

nomaxg: Actually, we don't even need the nextBlock, since all we really need is to calculate the distance between l1BlockNumber and prevBlock, so this should work:
``` solidity
function isHotshotAlive(uint256 l1BlockNumber) public view returns (bool) {
        uint256 prevBlock;
        bool prevBlockFound;

         for (uint256 i = l1BlockUpdates.length - 1; !prevBlockFound;) {
            if (i == 0) {
                break;
            }
            else if (l1BlockUpdates[i] <= l1BlockNumber) {
                prevBlockFound = true;
                prevBlock = l1BlockUpdates[i];
                break;
            }
        }

        // If no snapshot is found, we don't have enough history stored to tell whether HotShot was down.
        if (!prevBlockFound) {
            revert InsufficientSnapshotHistory();
        }

        return l1BlockNumber - prevBlock <= delayThreshold;
    }
```

contracts/src/LightClient.sol
```diff
@@ -341,4 +380,103 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+
+        // Edgecase 2: There have only been two HotShot updates so far
+        // we only start checking if HotShot is live after it has received atleast two updates
+        if (updatesCount < 2) {
+            return true;
+        }
+
+        for (uint256 i = 0; i < updatesCount; i++) {
+            uint256 blockStart = l1BlockUpdates[i];
+            uint256 blockEnd = (i < updatesCount - 1) ? l1BlockUpdates[i + 1] : block.number;
```
sveitser, line 415: Maybe this code would be easier to read if the iteration only went until `i < updatesCount - 1` and the last case was handled after the loop?

alysiahuggins:
I actually handled it that way before but the linter complained that the
function was too complex. I’ve since changed the function a bit so I can
try that again

contracts/src/LightClient.sol
alysiahuggins: This is a much cleaner way of thinking about it and implementing it! Thanks @nomaxg

contracts/src/LightClient.sol
alysiahuggins: resolving this as we decided to remove it from the contract and allow the user to specific the delayThreshold in their request

contracts/src/LightClient.sol
alysiahuggins: resolving this as the implementation has since changed

contracts/src/LightClient.sol
alysiahuggins: resolving as the implementation has since changed

contracts/src/LightClient.sol
```diff
@@ -341,4 +381,75 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
     }
+
+    /// @notice checks the stateUpdateBlockNumbers array and the delayThreshold to determine if
+    /// hotshot was
+    /// down at a specified L1 block number
+    /// @param l1BlockNumber This L1 block number used to reference a point in time when this light
+    /// client contract was expected to be updated in a given threshold
+    /// @param delayThreshold The delay threshold used to determined if this Light Client contract
+    /// was updated in the desired time (measured in blocks)
+    function wasL1Updated(uint256 l1BlockNumber, uint256 delayThreshold)
```
sveitser, line 392: I propose to invert the meaning of this function because I think it makes it clearer what it is for and easier to explain the logic. Currently this function is queried to turn on the escape hatch if the function returns false. With the change the function is called to turn on the escape hatch if the function returns true.

```solidity
    /// @notice checks that no more than threshold blocks passed since the last update before
    /// l1BlockNumber
    /// @param blockNumber The L1 block number
    /// @param threshold The number of blocks updates to this contract is allowed to lag behind
    function lagOverThreshold(uint256 blockNumber, uint256 threshold) public view returns (bool) {
        uint256 latestUpdateBefore;
        bool updateFound;
        // Iterate over updates in revers order to find the latest update before the given block
        // number.
        uint256 index = stateUpdateBlockNumbers.length - 1;
        while (!updateFound) {
            if (stateUpdateBlockNumbers[index] <= blockNumber) {
                updateFound = true;
                latestUpdateBefore = stateUpdateBlockNumbers[index];
            }
            if (index == 0) {
                break;
            }
            index--;
        }

        // If no snapshot is found, we don't have enough history stored to tell whether HotShot was
        // down.
        if (!updateFound) {
            revert InsufficientSnapshotHistory();
        }

        return blockNumber - latestUpdateBefore > threshold;
    }
```

thoughts?

cc @alysiahuggins @ImJeremyHe @nomaxg

alysiahuggins: i like this too actually - It matches with the mindset of the person who's calling the function and wants to determine if they need to do some action in event of a lag

nomaxg: make sense to me

contracts/src/LightClient.sol
```diff
@@ -84,14 +84,21 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+
     // === Data Structure ===
     //
     /// @notice The finalized HotShot state (as the digest of the entire HotShot state)
     /// @param viewNum The latest view number of the finalized HotShot chain
     /// @param blockHeight The block height of the latest finalized block
     /// @param blockCommRoot The merkle root of historical block commitments (BN254::ScalarField)
     /// @param feeLedgerComm The commitment to the fee ledger state (type: BN254::ScalarField)
-    /// @param stakeTableBlsKeyComm The commitment to the BlsVerKey column of the stake table
+    /// @param stakeTableBlsKeyComm The commitment to the BlsVerKey column of the stake tableg
```
nomaxg, line 101: typo

contracts/src/LightClient.sol
```diff
@@ -341,4 +381,78 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+
+        uint256 i = updatesCount - 1;
+        while (!prevUpdateFound) {
+            if (stateUpdateBlockNumbers[i] <= blockNumber) {
+                prevUpdateFound = true;
+                prevBlock = stateUpdateBlockNumbers[i];
+            }
+
+            // We don't consider the lag time for the first two updates
+            if (i < 2) {
```
nomaxg, line 413: Why don't we consider the lag time for the first two updates?

contracts/src/LightClient.sol
```diff
@@ -341,4 +381,78 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        returns (BN254.ScalarField hotShotCommitment)
+    {
+        uint256 commitmentsHeight = hotShotCommitments.length;
+        if (
+            hotShotCommitments[0].blockHeight > hotShotBlockHeight
+                || hotShotBlockHeight > hotShotCommitments[commitmentsHeight - 1].blockHeight
+        ) {
+            revert InvalidHotShotBlockForCommitmentCheck();
+        }
+        for (uint256 i = 1; i < commitmentsHeight; i++) {
```
nomaxg, line 447: Why start at index 1? Is there a genesis commitment edge case?

hotshot-state-prover/src/service.rs
```diff
@@ -645,11 +646,13 @@ mod test {
     async fn test_read_contract_state() -> Result<()> {
         setup_logging();
         setup_backtrace();
-
+        println!("deploy contract");
```
nomaxg, line 649: change to log?

contracts/src/LightClient.sol
```diff
@@ -341,4 +381,78 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+    }
+
+    /// @notice get the number of L1 block updates
+    function getStateUpdateBlockNumbersCount() public view returns (uint256) {
+        return stateUpdateBlockNumbers.length;
+    }
+
+    /// @notice get the HotShot commitment at the specified block height
+    /// @param hotShotBlockHeight hotShotBlockHeight
+    function getHotShotCommitment(uint256 hotShotBlockHeight)
```
nomaxg, line 435: @ImJeremyHe now that we have this function, we no longer need to worry about the L1 proof height and I think we can remove from the header.

nomaxg (approved):
LGTM just a couple of nits and questions

contracts/src/LightClient.sol
alysiahuggins: at one point we discussed and said that maybe there may be a lag between when the light client client contract was first deployed & instantiated, and the 1st update from a prover. So then in that instance we only consider lag times between update 2 onwards. what do you think? paging @sveitser @ImJeremyHe to hear their thoughts too

hotshot-state-prover/src/service.rs
alysiahuggins: ahh this was during debugging at one point, will remove

contracts/src/LightClient.sol
```diff
@@ -251,6 +276,22 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
 
         // upon successful verification, update the latest finalized state
         states[finalizedState] = newState;
+
+        /**
+         * TODO purge elements from the stateUpdateBlockNumbers array after a decided number of
```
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/issues/1541

contracts/src/LightClient.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/12b6bf256ec4136d6a2d62edaf40809818d7e6d8#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292

contracts/src/LightClient.sol
alysiahuggins: yes but has been addressed since our convo https://github.com/EspressoSystems/espresso-sequencer/commit/12b6bf256ec4136d6a2d62edaf40809818d7e6d8#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292

hotshot-state-prover/src/service.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/12b6bf256ec4136d6a2d62edaf40809818d7e6d8#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292

contracts/src/LightClient.sol
```diff
@@ -341,4 +381,78 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+    function lagOverEscapeHatchThreshold(uint256 blockNumber, uint256 threshold)
+        public
+        view
+        returns (bool)
+    {
+        uint256 updatesCount = stateUpdateBlockNumbers.length;
+
+        // Handling Edge Cases
+        // Edgecase 1: The block is in the future or in the past before HotShot was live
+        if (blockNumber > block.number || updatesCount < 3) {
```
ImJeremyHe, line 389: If the block is in the future, doen't that mean hotshot has already down?

alysiahuggins: hmm, if the current ethereum block is 10 and the user requested for block 12, i'd say that we can't say whether our light client contract is missing an update / hotshot is down because we don't have enough data points to be able to assert that.

alysiahuggins:
An issue was created to handling purging of the newly added arrays, `hotShotCommitments` and `stateUpdateBlockNumbers`
https://github.com/EspressoSystems/espresso-sequencer/issues/1541

contracts/src/LightClient.sol
sveitser: IMO it's not meaningful to ask if the light client contract lagged behind at an Ethereum block number in the future so I think it's correct to throw an error here.

contracts/src/LightClient.sol
sveitser: I think this makes sense. I think the idea is that we don't expect rollups to be deployed before the prover has sent its first update to the contract. I think maybe the comment is misleading but the implementation is correct?

If a rollup has threshold 10 and we have the following history (first column is the Ethereum block number)

0 genesis
10 first LC update
12 rollup deployed
25 second LC update

Say we are now at some block after 25 and check if the lag is over the threshold at block 21.

```
blockNumber <- 21
updatesCount <- 3
i <- 2
stateUpdateBlockNumbers[2] is 25 
25 <= 21 // false
2 < 2 // false
i <- 1
stateUpdateBlockNumbers[1] is 10
10 <= 21 // true
```
So we do find the previous update.

nomaxg: I see, makes sense to me!

nomaxg:
LGTM

PR #1498, opened 2024-05-23
Avoid always using binaries in target/release

sveitser:
If one previously ran `cargo build` without nix and then switches to nix, the demo script uses the old binaries and things don't work.

### DIFF
--- a/scripts/demo-native
+++ b/scripts/demo-native
@@ -10,8 +10,6 @@ if [ -z "${IN_NIX_SHELL-}" ]; then
     export "PATH=${TARGET_DIR}/release:$PATH"
 fi
 
-export PATH=./target/release:$PATH
-
 ESPRESSO_BASE_STORAGE_PATH=$(mktemp -d -t espresso-XXXXXXXX)
 export ESPRESSO_BASE_STORAGE_PATH
 echo "Using sequencer storage path: $ESPRESSO_BASE_STORAGE_PATH"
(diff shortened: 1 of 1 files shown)


scripts/demo-native
```diff
@@ -1,6 +1,15 @@
...
+# If we aren't in a nix shell (where this handled in flake.nix) add the target
+# directory to the path so that the binaries are found by process-compose.
+if [ -z "${IN_NIX_SHELL-}" ]; then
+    REPO_ROOT="$(dirname "$(dirname "$(readlink -fm "$0")")")"
+    # Default to CARGO_TARGET_DIR if set, otherwise use the default target directory.
+    TARGET_DIR="${CARGO_TARGET_DIR:-${REPO_ROOT}/target}"
+    export "PATH=${TARGET_DIR}/release:$PATH"
+fi
+
 export PATH=./target/release:$PATH
```
jbearer, line 13: Delete this, right?

scripts/demo-native
```diff
@@ -1,6 +1,15 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# If we aren't in a nix shell (where this handled in flake.nix) add the target
+# directory to the path so that the binaries are found by process-compose.
+if [ -z "${IN_NIX_SHELL-}" ]; then
```
jbearer, line 6: Where does this get set?

sveitser: By nix

scripts/demo-native
sveitser: yup, already done

jbearer (approved):

PR #1499, opened 2024-05-23
chore: rewrite block payload

### LINKED ISSUE #1076: Follow-up refactoring of `Payload`
The Payload struct has been redesigned (see #1012) however there are some pending improvements including:

- [ ]  Get rid of type `TxTableEntry` See https://github.com/EspressoSystems/espresso-sequencer/blob/dbf68c9a12ec04a0b2b5be1967f7e9b651071efb/sequencer/src/block2/entry.rs#L10
- [ ] This type should be replaced with `TableWord` with the appropriate trait bounds which will likely involve the use of const generics.
  - [ ] Don't allow arithmetic directly on table entries: `zero()`, `one()`, `checked_add_mut()`
- [ ]  Change BlockPayload::Encode to avoid copying the payload.
- [ ]  Add unit tests for `NameSpaceTable`
- [ ]  Add unit tests for `TxTable`
- [ ]  Handle errors
- [ ] See if the TxTable / NameSpaceTable can store only a reference to the payload

ggutoski:
Closes #1076 

### This PR:
- Complete rewrite of everything in `sequencer/src/block`. 
- Maintain serialization compatibility with current main branch, though it's janky. Search `serialization compatibility` in comment TODOs.

### This PR does not:
- Only minimal tests for basic correctness in `block/tests.rs`. Deep testing for arbitrary block payloads is deferred to a future PR.

### Key places to review:
- `TODO` in code comments. I've checked them all prior to posting this PR. Many are flags to request changes to upstream code such as `TODO(QueryablePayload)`. Some are requests for feedback from reviewers.
- Anything *outside* module `block`. Especially `NamespaceProofQueryData` struct in `endpoints.rs` as discussed in [zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/namespace.20query.20endpoint.20for.20non-existent.20ID/near/441305372), plus knock-on changes in `api.rs`, `nasty-client.rs`.
- Proposal to move `NamespaceId` struct into `ns_table.rs`: https://github.com/EspressoSystems/espresso-sequencer/pull/1499#issuecomment-2134065090, plus a laundry list of complaints: https://github.com/EspressoSystems/espresso-sequencer/blob/ab276e55676e746873b1311775bbb9c771e3966b/sequencer/src/transaction.rs#L9-L28
- ~~Should we allow duplicate namespace IDs in the namespace table? See `ns_table.rs` (search "duplicate"), especially `NsIter` struct. See also [discussion in zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/enforce.20uniqueness.20of.20namespace.20IDs.20at.20consensus.3F)~~
  - Current plan is to allow duplicate namespace IDs for now, but forbid it in the future.
- FYI detailed spec in rustdoc in `ns_table.rs` and `types.rs`. (Not that you need to read it!)

# Top-down tour of code

## Module: `block`

`block.rs`: Exposes an API to the rest of this workspace. Currently restricted only to
https://github.com/EspressoSystems/espresso-sequencer/blob/46bbe9a7d09745b8a932c1582f422f336d97742c/sequencer/src/block.rs#L6
I expect in the future we might also need to expose `TxProof` and possibly other items. The goal is to expose as little as possible outside `block` module.

`block/full_payload.rs`, `block/namespace_payload.rs`: An attempt to impose a hierarchy of isolation/abstraction within `block` so that it's not just a flat wall of types. (It's difficult to define a natural hierarchy; but this presentation seems not bad to me.)

- `full_payload`: Any code that needs to know the structure of a multi-namespace payload. Contains the all-important `Payload` struct (which impls `BlockPayload`, `QueryablePayload` traits) plus `NsTable`, `NsProof`, etc.
- `namespace_payload`: Self-contained implementation of an individual namespace. `NsPayload` plus `TxProof`, etc.
 
As usual, these modules expose as little as possible. Lots of use of `pub(in crate::block)` to restrict visibility within `block` module.

## Isolation of low-level binary format

Low-level details of binary format is isolated as much as possible. All reads from a block payload must use the following 2 modules:

- `block/full_payload/ns_table.rs`. 3 constants define byte lengths of objects that appear in a namespace table:
https://github.com/EspressoSystems/espresso-sequencer/blob/46bbe9a7d09745b8a932c1582f422f336d97742c/sequencer/src/block/full_payload/ns_table.rs#L20-L23
- `block/namespace_payload/types.rs` (Open to suggestions for a better name.) Similar to the namespace table, 2 constants define byte lengths of objects that appear in a transaction table:
https://github.com/EspressoSystems/espresso-sequencer/blob/46bbe9a7d09745b8a932c1582f422f336d97742c/sequencer/src/block/namespace_payload/types.rs#L12-L14

## Extensive use of newtypes to prevent bugs

There are many different items floating around: payload byte length, namespace byte length, number of transactions in a namespace, index of a transaction, index of a namespace, byte range for a namespace, byte range for a transaction, etc.

It's easy to accidentally confuse these quantities. Example: accidental use of *payload* byte length when you intend to use *namespace* byte length. I've made extensive use of newtypes to prevent nasty bugs that might arise from such accidental misuse. 

### Example: construct a `TxIter`

- `TxIter::new` has an arg of type `NumTxs`: a newtype wrapper for `usize` that indicates the number of transactions in a namespace.
- To get a `NumTxs` call `NumTxs::new`, which has args of type `NumTxsUnchecked` and `NsPayloadByteLen`.
- To get a `NumTxsUnchecked` call `NsPayload::read_num_txs`. This method extracts the declared number of transactions from the transaction table. A transaction table might contain corrupted data, so this quantity might cause a panic if used directly without checking. That's why you cannot construct a `TxIter` from a `NumTxsUnchecked` -> you must get a `NumTxs`.
- Similarly, you also need a `NsPayloadByteLen` to construct a `NumTxs`. You get that from `NsPayload::byte_len` or `NsPayloadRange::byte_len`. This newtype prevents you from accidentally using the *payload* byte len instead of the *namespace* byte len.
- Normally I would not bother to decompose this process into so many low-level steps because it adds complexity to the codebase. But `TxProof` demands many low-level pieces. Given that these pieces must exist for `TxProof`, it makes sense to create them, then build higher-level helpers from these low-level tools as needed.
    - Example: `NsPayload::read_num_txs` is actually a convenience helper that bundles several lower-level steps. 

## Modules `full_payload`, `namespace_payload`

Module `full_payload` is relatively simple because `NsProof` is not very demanding because the correctness of the namespace table is trusted as part of the block header.

Module `namespace_payload` is complex because `TxProof` is very demanding, so let's talk about the contents of `namespace_payload`:

- `ns_payload.rs`: Struct `NsPayload` for payload bytes for an individual namespace. No surprises here.
- `ns_payload_range.rs`: Newtype `NsPayloadRange` for `Range<usize>` that indicates a subslice of a `Payload` for an individual namespace. Lots of important metadata can be obtained from only this range with no need to have a full `NsPayload`. Indeed, in some contexts such as `TxProof::verify` a full `NsPayload` is not available so the only option is to rely on `NsPayloadRange`.
- `types.rs`: All low-level binary details of a namespace are isolated in this file. Most of the newtypes mentioned above depend on these low-level binary details, so most of those newtypes are also defined here.
- `iter.rs`: A public-facing iterator `Iter` over all transactions in a block payload. Under the hood it's a simple Cartesian product `NsIter` x `TxIter`.
    - `NsIter` is defined in `ns_table.rs` and can be accessed publicly via `NsTable::iter`.
    - `TxIter` is defined in `types.rs` and can be accessed via `NsPayload::iter`. But `TxIter` is not publicly accessible outside `block`.

## Serialization for fixed-byte-length integers in `uint_bytes.rs`

The magic constants defined in `ns_table.rs`, `types.rs` indicate the byte length of various items in the payload.

Example: `NUM_NSS_BYTE_LEN = 4` indicates that a namespace table shall use 4 bytes to encode the number of namespaces in the block. Perhaps in the future we'll decide that 3 bytes is sufficient to describe this quantity, as we expect the number of namespaces in a single block will never exceed 16 million. In that case, you can change `NUM_NSS_BYTE_LEN` from `4` to `3` and everything should work perfectly.

In order to make the code generic over these constants I had to write my own serialization functions in `uint_bytes.rs`. In pseudocode:
```rust
fn usize_to_bytes<const N: usize>(n: usize) -> [u8; N];
fn usize_from_bytes<const N: usize>(bytes: &[u8]) -> usize;
```
Most uses of this serialization are for a `usize` quantity, but `NamespaceId` serialization is `u64`. The only way to make these functions generic over primitive integer types `usize`, `u64`, etc is to put them in a macro `uint_bytes_impl`.

This file includes other helper functions, extensive rustdoc, and detailed tests. There's also a separate macro `bytes_serde_impl` for custom `serde` derivation to/from a fixed-length bytes array `[u8; N]`.

### Limitations

This code should work perfectly if all the magic constants are at most `8`. Anything beyond `8` exceeds the byte length of `usize`, `u64` and so you can expect build failure and/or startup panic if you try it.

If you want to accommodate sizes larger than `8` then you'll need to change the implementation of all affected types to use something other than `usize`, `u64`. I do not foresee any demand for sizes larger than 8 bytes: typical use is to serialize an integer quantity, and we do not expect any such quantity to exceed 64 bits.

A possible exception: `NamespaceId` is an identifier, not a quantity. In the future we might use 32 bytes for `NamespaceId` to allow a namespace ID to be a hash output. In that case `NamespaceId` will need modification and can no longer use the functions in `uint_bytes.rs`.


contracts/lib/forge-std
sveitser, line 1: Is this an intentional change?

ggutoski: Hell no. I don't recall doing this. Good catch!

sveitser:
I have only one failure locally:

    failures:
        message_compat_tests::test_message_compat

This is the diff

```diff
diff data/messages-actual.json data/messages.json 
11c11
<                 "builder_commitment": "BUILDER_COMMITMENT~_FlsTMQYXvjU_M1TpJFyrGe9BGPU9Di-qALdKJb2hegZ",
---
>                 "builder_commitment": "BUILDER_COMMITMENT~tEvs0rxqOiMCvfe2R0omNNaphSlUiEDrb2q0IZpRcgA_",
32c32,34
<                 "ns_table": "AAAAAA==",
---
>                 "ns_table": {
>                   "bytes": "AAAAAA=="
>                 },
320c322,324
<               "metadata": "AQAAAAEAAAALAAAA",
---
>               "metadata": {
>                 "bytes": "AQAAAAEAAAALAAAA"
>               },
```

ggutoski:
> I have only one failure locally:

Yep I just found that one, too. Took me a while to dig it out of the logs. I'll fix that one shortly. Thank you!! ❤️

contracts/lib/forge-std
ggutoski: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/e6903db4493a4f34626ab8fdfccaa4160fbd06a8

sequencer/src/block/namespace_payload/ns_payload.rs
```diff
@@ -0,0 +1,133 @@
+//! The only thing [`NsPayload2`] does is naively read from its payload given a
```
philippecamacho, line 1: This comment does not look up to date.

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -0,0 +1,229 @@
+//! Types related to a namespace table.
+//!
+//! All code that needs to know the binary format of a namespace table is
+//! restricted to this file.
```
philippecamacho, line 4: Would be nice to add more comments following this helpful document: https://www.notion.so/espressosys/Block-payload-format-with-namespace-transaction-tables-81425c5bd3ba495bafc3198ce287c487

philippecamacho:
> A possible exception: NamespaceId is an identifier, not a quantity. In the future we might use 32 bytes for NamespaceId to allow a namespace ID to be a hash output. In that case NamespaceId will need modification and can no longer use the functions in uint_bytes.rs.

I understand that while currently the set of validators is permissioned, the set of rollups is not. Can we still use `u64` for `NamespaceId`?

sequencer/src/block/full_payload/payload.rs
```diff
@@ -0,0 +1,233 @@
...
+use jf_vid::VidScheme;
+use serde::{Deserialize, Serialize};
+use sha2::Digest;
+use std::{collections::HashMap, fmt::Display, sync::Arc};
+
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct Payload {
+    // Concatenated payload bytes for each namespace
+    #[serde(with = "base64_bytes")]
+    payload: Vec<u8>,
```
philippecamacho, line 23: Would be nice to add some comments following this helpful document: https://www.notion.so/espressosys/Block-payload-format-with-namespace-transaction-tables-81425c5bd3ba495bafc3198ce287c487

sequencer/src/block/full_payload/payload.rs
```diff
@@ -0,0 +1,233 @@
...
+        let mut ns_builders = HashMap::<NamespaceId, NsPayloadBuilder>::new();
+        for tx in transactions.into_iter() {
+            // accounting for block byte length limit
+            block_byte_len += tx.payload().len() + NsPayloadBuilder::tx_overhead_byte_len();
+            if !ns_builders.contains_key(&tx.namespace()) {
+                // each new namespace adds overhead
+                block_byte_len += NsTableBuilder::ns_overhead_byte_len()
+                    + NsPayloadBuilder::fixed_overhead_byte_len();
+            }
+            if block_byte_len > max_block_byte_len {
```
philippecamacho, line 92: Why not throw an error if this happens?

ggutoski: Previously we did not do anything (post log or return error) in this event. See #1382. If we throw an error here then we are breaking previous behaviour.

I put a `warn!` log because the user probably does not intend to trigger max block size, but strictly speaking it's not an "error".

Current enforcement of max block size is very primitive. I imagine in the future builders might want more fine-grained info about how each transaction contributes to block size. (Perhaps `from_transactions` might return more info instead of silently truncating the block!)

Anyway, I suggest any improvement here is a matter for a future PR. I suspect we do not wish to allocate resources to this task at the moment.

ggutoski:
Commit https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/b593a96adb42ea41bf4321fd5945078c7fefc64b reverts hotfix PR #1504, which is no longer necessary. But I don't know where to put a test.

Note that API/serialization for `NamespaceId` accepts any `u64`. Thus, the sequencer in its current form will crash if the constant `NS_ID_BYTE_LEN` is anything different from `8`.

This PR does not touch `NamespaceId` but perhaps it should. Unfortunately, a proper fix requires that `NamespaceId` serialization code has visibility of `NS_ID_BYTE_LEN`, which I'd very much like to keep private to `ns_table.rs`.

One solution is to move `NamespaceId` struct into `ns_table.rs`. I like this solution because `NamespaceId` depends on the low-level binary details of a payload, much like similar new structs such as `NsIndex`, `TxIndex`, `NumTxs`, etc.

I'll let this idea sit overnight and gather opinions. If I don't hear anything by tomorrow then I'll move `NamespaceId` to `ns_table.rs` as part of this PR.

sequencer/src/block/full_payload/payload.rs
philippecamacho: Sounds good, thanks for the explanation.

ggutoski:
> I understand that while currently the set of validators is permissioned, the set of rollups is not. Can we still use `u64` for `NamespaceId`?

I don't understand the concern here. What could a malicious rollup do to exploit a 8-byte namespace ID?

philippecamacho:
> > I understand that while currently the set of validators is permissioned, the set of rollups is not. Can we still use `u64` for `NamespaceId`?
> 
> I don't understand the concern here. What could a malicious rollup do to exploit a 8-byte namespace ID?

This is not a security concern, rather I wonder how the short ID will be assigned to rollups in practice given that any rollup can join?

sequencer/src/block/full_payload/ns_proof.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// convention. But that would require a change to our RPC endpoint API at
+    /// [`endpoints`](crate::api::endpoints), which is a hassle.
+    pub fn new(payload: &Payload, ns_id: NamespaceId, common: &VidCommon) -> Option<NsProof> {
+        let payload_byte_len = payload.byte_len();
+        if !payload_byte_len.is_consistent(common) {
+            return None; // error: vid_common inconsistent with self
+        }
+        let Some(ns_index) = payload.ns_table().find_ns_id(&ns_id) else {
+            // ns_id does not exist
+            return Some(NsProof {
```
philippecamacho, line 49: Should not we have some cryptographic evidence that the payload is empty for namespace `ns_id`?

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        // TODO hack to deserialize `NamespaceId` from `NS_ID_BYTE_LEN` bytes
+        NamespaceId::from(u64_from_bytes::<NS_ID_BYTE_LEN>(
+            &self.0[start..start + NS_ID_BYTE_LEN],
+        ))
+    }
+
+    /// Does the `index`th entry exist in the namespace table?
+    pub fn in_bounds(&self, index: &NsIndex) -> bool {
+        // The number of entries in the namespace table, including all duplicate
```
philippecamacho, line 58: Can't the replicas ensure the namespaces are not duplicated?

philippecamacho:
A general question: do we need to do something special in order to handle possible changes regarding how namespaces, payloads will be serialized in the future, or is this problem already handled by a general solution we have? cc @nyospe @jbearer

sequencer/src/block/full_payload/ns_table.rs
ggutoski: They could, and perhaps they will. This design was originally conceived on the assumption that replicas would not enforce anything, and a namespace table could have arbitrary bytes. I'd like to maintain that property as long as it is reasonable to do so, just in case we need it again in the future.

philippecamacho: I see. Thank you for the clarification.

sequencer/src/block/full_payload/payload.rs
ggutoski: addressed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/5c9fe51fa9f95bb392d5c8446166068663ef0ac1. Added a detailed spec in rustdoc

sequencer/src/block/full_payload/ns_table.rs
ggutoski: addressed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/5c9fe51fa9f95bb392d5c8446166068663ef0ac1. Added a detailed spec in rustdoc. Also added namespace payload spec in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/311a7f3353cc24194b401e81df7b03f4893b8a22

sequencer/src/block/namespace_payload/ns_payload.rs
ggutoski: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/5c9fe51fa9f95bb392d5c8446166068663ef0ac1

ggutoski:
> > > I understand that while currently the set of validators is permissioned, the set of rollups is not. Can we still use `u64` for `NamespaceId`?
> > 
> > 
> > I don't understand the concern here. What could a malicious rollup do to exploit a 8-byte namespace ID?
> 
> This is not a security concern, rather I wonder how the short ID will be assigned to rollups in practice given that any rollup can join?

I believe the concern is not that we'll run out of 64-bit chain IDs. Perhaps you're concerned that different rollups might accidentally collide at a single namespace ID? If rollups choose their namespace IDs at random then the probability of collision is acceptably small. If instead rollups choose their namespace IDs maliciously then we have the same problem whether the chain ID is 8 bytes or 32 bytes.

My understanding is that precedent for EVM chains is that each chain simply chooses its own chain ID so as not to conflict with any other known chain. I suspect all parties are incentivized to avoid a chain ID conflict, and so we never hear of any problem caused by it. There's even a public list of EMV chain IDs: https://chainlist.org/ It seems that EVM chain IDs are also 64 bits.

I expect a similar dynamic to hold for namespaces in our setting. tldr; I don't think it's a problem.

sequencer/src/block/full_payload/ns_proof.rs
ggutoski: Here there is significant benefit if we take advantage of the fact that every node sees and comes to consensus on the namespace table: non-existence of a namespace ID can be verified directly by searching the namespace table.

Now that you mention it... it is kinda silly for `NsProof` to have a `existence` field, and pretend to "prove" non-existence of a namespace ID to a "verifier" who will just search the namespace table.

Perhaps I should tweak `NsProof` so that its constructor refuses to make a proof for a non-existent namespace ID. I think I'll do that next time I'm at my desk.

philippecamacho: Oh I see, good to know as it is relevant for the design of the zk rollups circuits.

philippecamacho:
> > > > I understand that while currently the set of validators is permissioned, the set of rollups is not. Can we still use `u64` for `NamespaceId`?
> > > 
> > > 
> > > I don't understand the concern here. What could a malicious rollup do to exploit a 8-byte namespace ID?
> > 
> > 
> > This is not a security concern, rather I wonder how the short ID will be assigned to rollups in practice given that any rollup can join?
> 
> I believe the concern is not that we'll run out of 64-bit chain IDs. Perhaps you're concerned that different rollups might accidentally collide at a single namespace ID? If rollups choose their namespace IDs at random then the probability of collision is acceptably small. If instead rollups choose their namespace IDs maliciously then we have the same problem whether the chain ID is 8 bytes or 32 bytes.
> 
> My understanding is that precedent for EVM chains is that each chain simply chooses its own chain ID so as not to conflict with any other known chain. I suspect all parties are incentivized to avoid a chain ID conflict, and so we never hear of any problem caused by it. There's even a public list of EMV chain IDs: https://chainlist.org/ It seems that EVM chain IDs are also 64 bits.
> 
> I expect a similar dynamic to hold for namespaces in our setting. tldr; I don't think it's a problem.

Concern is probably too strong of a word :). I just wonder if we should create a task for letting the rollups define their namespace ID (something similar to chainlist.org you mentioned) or if this is already covered for Decaf.

sequencer/src/block/full_payload/ns_proof.rs
ggutoski: > Perhaps I should tweak `NsProof` so that its constructor refuses to make a proof for a non-existent namespace ID. I think I'll do that next time I'm at my desk.

done in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/f907be50155e5ff42fafc5840e4bd96752ebedd8

sequencer/src/block/namespace_payload/iter.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = Index;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let Some(ns_index) = self.ns_iter.peek() else {
+                return None; // ns_iter consumed
```
akonring, line 59: nit: other places you've used `break None`.

sequencer/src/block/namespace_payload/ns_payload.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// Iterator over all transactions in this namespace.
+    pub fn iter(&self) -> TxIter {
+        self.iter_from_num_txs(&self.read_num_txs())
+    }
+
+    /// Return all transactions in this namespace. The namespace ID for each
+    /// returned [`Transaction`] is set to `ns_id`.
+    pub fn export_all_txs(&self, ns_id: &NamespaceId) -> Vec<Transaction> {
```
akonring, line 48: nit: (just a comment) providing `ns_id` seems redundant in many cases (NsPayload should already be aware of the corresponding NsId). Extending NsPayload with a NsId (making NsPayload id-aware) would fix this but likely messes up the desired abstraction.

akonring (commented):
Looks good.
The type-driven approach seems to take the code in a less error-prone direction 👍

sequencer/src/bin/nasty-client.rs
```diff
@@ -940,11 +936,13 @@ impl ResourceManager<BlockQueryData<SeqTypes>> {
...
             })
             .await?;
-        if header.ns_table.is_empty() {
+        let num_namespaces = header.ns_table.iter().count();
+        if num_namespaces == 0 {
             tracing::info!("not fetching namespace because block {block} is empty");
             return Ok(());
         }
-        let ns = header.ns_table.get_table_entry(index).0;
+        let ns_index = header.ns_table.iter().nth(index % num_namespaces).unwrap();
```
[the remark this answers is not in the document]
ggutoski: Actually, I think this *fixes* the nasty-client. 😁  The purpose is to select a random entry from the ns_table.

Old version: a random `index` is usually huge, causing `get_table_entry` to return a bunch of zeros. Not a very good test!

New version: I want to be careful that the random entry is selected only among entries that do not duplicate a previous entry's namespace ID. I'm pleased with this solution: the new namespace iterator already does this, so I could just use `nth`.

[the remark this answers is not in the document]
ggutoski: Indeed, the type system prevents anyone from making a `NsIndex` that improperly points to a duplicate entry in the namespace table. (Unless you deserialize from hacked bytes.) So really this is the *only* way for nasty-client to get the `NsIndex` it needs to proceed. 🙂

philippecamacho (approved):
This is a significant improvement compared to the previous version of the code.

sequencer/src/block/full_payload/payload.rs
```diff
@@ -0,0 +1,261 @@
...
+
+        // add each tx to its namespace
+        let mut ns_builders = HashMap::<NamespaceId, NsPayloadBuilder>::new();
+        for tx in transactions.into_iter() {
+            // accounting for block byte length limit
+            block_byte_len += tx.payload().len() + NsPayloadBuilder::tx_overhead_byte_len();
+            if !ns_builders.contains_key(&tx.namespace()) {
+                // each new namespace adds overhead
+                block_byte_len += NsTableBuilder::ns_overhead_byte_len()
+                    + NsPayloadBuilder::fixed_overhead_byte_len();
```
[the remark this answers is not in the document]
ggutoski: Yep. Each new namespace requires:
1. A new entry in the namespace table: `NsTableBuilder::ns_overhead_byte_len()`
2. A new transaction table inside the namespace payload: `NsPayloadBuilder::fixed_overhead_byte_len()`

These names are pretty uninformative. 😕 Unfortunately the task to enforce max payload size is a recent addition to the codebase and so it's awkward to add it to the current design. What you see here is not much different from what we had before. Suggestions welcome.

sequencer/src/block/namespace_payload/ns_payload.rs
ggutoski: I wrestled with this question, and I even have a id-aware implementation in a draft branch: https://github.com/EspressoSystems/espresso-sequencer/blob/5feebba5a351ab8152a72f5a8478a63f82da362f/sequencer/src/block2/ns_payload.rs#L53-L58

TBH I can't quite remember why I decided against it. Your suggestion that it "messes up the desired abstraction" sounds familiar though. Some quick notes:
- Some computation would be repeated or unnecessary, especially in `TxProof`.
- If `NsPayload` has any other fields beyond `[u8]` then it breaks `NsPayloadOwned` due to limitations in Rust, which would force me to write a lot of boilerplate code for delegation, etc.

I agree it's not ideal to have a `ns_id` arg here. I'm willing to reconsider this question, but I also want to be mindful not to spend time chasing rabbits down holes. 😉 🐰 🕳️

sequencer/src/block/namespace_payload/iter.rs
ggutoski: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/a1269faf9d827acf444390f46c71fa52a45ad532

ggutoski:
> A general question: do we need to do something special in order to handle possible changes regarding how namespaces, payloads will be serialized in the future, or is this problem already handled by a general solution we have? cc @nyospe @jbearer

@philippecamacho I don't have a complete answer to this question but I can tell you that we have tests that catch any breaking changes to serialization in `reference_tests.rs`, `message_compat_tests.rs`. (Note that the present PR makes breaking changes, so I've updated these tests accordingly.)

sequencer/src/block/full_payload/payload.rs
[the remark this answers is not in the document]
akonring: Makes sense. I just misinterpreted this part.

ggutoski:
On further discussion with @jbearer , I need to investigate whether it's feasible to maintain serialization compatibility. Putting this PR back into draft mode for now.

sequencer/src/block/full_payload/ns_proof.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// running non-Rust code, in which case the client is unable to perform
+    /// this post-processing himself.
+    pub fn verify(
+        &self,
+        ns_table: &NsTable,
+        commit: &VidCommitment,
+        common: &VidCommon,
+    ) -> Option<(Vec<Transaction>, NamespaceId)> {
+        VidSchemeType::is_consistent(commit, common).ok()?;
+        if !ns_table.in_bounds(&self.ns_index) {
```
jbearer, line 90: Nit: it seems like we often check `in_bounds` and then read from the index with `ns_range` or whatever. Maybe it would be cleaner to have `ns_range` do the bounds checking and return an `Option`? And if there are places where we already know the index is in range and want to skip the extra check for (dubious) efficiency, we could have `ns_range_unchecked`

sequencer/src/block/full_payload/payload.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[async_trait]
+impl BlockPayload<SeqTypes> for Payload {
+    type Error = crate::Error;
+    type Transaction = Transaction;
+    type Instance = NodeState;
+    type Metadata = NsTable;
+    type ValidatedState = ValidatedState;
+
+    // TODO(BlockPayload): return type should not include `Self::Metadata`
```
jbearer, line 145: I think for TODOs referring to other repos, it is better to create issues in the appropriate repos and assign them to the PL

sequencer/src/transaction.rs
```diff
@@ -6,6 +6,26 @@ use hotshot_types::traits::block_contents::Transaction as HotShotTransaction;
 use jf_merkle_tree::namespaced_merkle_tree::{Namespace, Namespaced};
 use serde::{Deserialize, Serialize};
 
+/// TODO [`NamespaceId`] has historical debt to repay:
+/// - It must fit into 4 bytes in order to maintain serialization compatibility
+///   for [`crate::block::NsTable`], yet it currently occupies a `u64`. Thus,
+///   any code that creates a [`NamespaceId`] must ensure that it fits into 4
+///   bytes, which is bug-prone. We should either (i) allow it to occupy 8
+///   bytes, breaking serialization compatibility, or (ii) define it as a 4-byte
+///   struct such as `u32`.
```
jbearer, line 15: So it seems like we're going with the 4-byte restriction for now. We can lift this and switch to 8 bytes as soon as we have a solution for upgrading data structures.

In the meantime, though, why not just change this to `u32` right now? This will prevent bugs and errors. As is, I think we at least have to update the deserialization to fail if the input doesn't fit in a `u32`, or add back the check in the submit endpoint to enforce that nobody is submitting transactions with namespaces that don't fit. Otherwise we will have transactions in the mempool that cause payload construction to fail, and the builder is not smart enough to filter those out.

jbearer (commented):
This is unbelievably well thought out. I applaud you.

Most of my comments are very minor and you should feel free to ignore them. The only really important one is the one on `NamespaceId`.

sequencer/src/transaction.rs
ggutoski: > why not just change this to `u32` right now?

I strongly prefer this solution. I didn't put it in this PR because I wasn't sure whether it would introduce a serialization break. (And because this PR is already way too big.) I don't even know whether `reference_tests.rs` would catch a change in `Transaction` serialization.

But now that I have this feedback I'll give it a shot.

jbearer: Ah, shoot I think you're right actually. When we serialize a transaction by itself, outside of a payload, this might be a break. For JSON it definitely won't matter. For bincode we might get away with it if using a variable length integer encoding. I'm not sure if that's turned on by default, so it's worth trying.

If we can't make the `u32` switch work, I think we should override *de*serialization so it parses a u64 but returns an error if it exceeds `u32::MAX`.

ggutoski: Confirmed: changing it to `u32` breaks reference tests. 😕 

> If we can't make the `u32` switch work, I think we should override _de_serialization so it parses a u64 but returns an error if it exceeds `u32::MAX`.

Could do. But that won't stop anyone from using `NamespaceId::from` on an arbitrary `u64`, which is what we do in `endpoints.rs`. How would you feel about removing `From` from the `derive` list for `NamespaceId`? (There's a bunch of other cruft there I'd like to remove, too. See the rest of this code comment block.)

sequencer/src/api/endpoints.rs
```diff
@@ -178,16 +168,6 @@ where
...
 
-            // Transactions with namespaces that do not fit in the u32
-            // cannot be included in the block.
-            // TODO: This issue will be addressed in the next release.
-            if tx.namespace() > NamespaceId::from(u32::MAX as u64) {
-                return Err(Error::Custom {
-                    message: "Transaction namespace > u32::MAX".to_string(),
-                    status: StatusCode::BadRequest,
-                });
-            }
```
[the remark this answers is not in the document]
ggutoski: Yikes! I removed this code when I switched namespace ID to 8 bytes in the namespace table, but then forgot to restore it when I reverted it back to 4 bytes for serialization compatibility. Thanks for catching it! I'll restore this code as needed pending resolution of https://github.com/EspressoSystems/espresso-sequencer/pull/1499#discussion_r1631308091

sequencer/src/bin/nasty-client.rs
```diff
@@ -940,11 +936,13 @@ impl ResourceManager<BlockQueryData<SeqTypes>> {
                     .context(format!("fetching header {block}"))
             })
             .await?;
-        if header.ns_table.is_empty() {
+        let num_namespaces = header.ns_table.iter().count();
```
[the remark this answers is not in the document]
ggutoski: Can't use `ExactSizeIterator` here because the only way to compute the length is by consuming the iterator because we currently allow duplicate namespace IDs in the namespace table and so we don't know which entries to ignore unless we scan the whole thing. I could add a `len()` method but it would just wrap `iter().count()`.

I think I'll leave it as-is for now. In the future we intend to enforce uniqueness of namespace IDs at the consensus level. When that happens we can deduce the number of entries in the namespace table in constant time and so a `len()` method would make more sense.

sequencer/src/bin/nasty-client.rs
```diff
@@ -940,11 +936,13 @@ impl ResourceManager<BlockQueryData<SeqTypes>> {
...
+        if num_namespaces == 0 {
             tracing::info!("not fetching namespace because block {block} is empty");
```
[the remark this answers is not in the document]
ggutoski: If `num_namespaces == 0` then it's impossible even to create a `NsIndex` needed to retrieve a `NamespaceId` for the query, thanks to my militant use of newtypes.

In order to execute the query I'd need to guess a namespace ID. The only thing I could test with such a query is that my guessed ID caused the endpoint to return an error, but I expect that outcome anyway even for a non-empty block because the ID was guessed.

Current `main` branch does the same thing: just return `Ok(())` immediately if the namespace table is empty. So I'm going to leave this as-is.

sequencer/src/block/full_payload/ns_proof.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct NsProof {
+    ns_index: NsIndex,
+    ns_payload: NsPayloadOwned,
+    ns_proof: Option<LargeRangeProofType>, // `None` if ns_payload is empty
+}
+
+impl NsProof {
+    /// Returns the payload bytes for namespace `ns_id`, along with a proof of
+    /// correctness for those bytes. Returns `None` if `ns_id` is not in the
+    /// namespace table, or on error.
```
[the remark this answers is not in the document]
ggutoski: This doc was out of date. Fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/65b6f62cfedb4a47544a652b788ad8fd04e20966

This method actually takes a `NsIndex` arg instead of a `NamespaceId` arg. An out-of-bounds index is viewed as an error, so this method actually returns `None` only on error. (We could change error handling to use `Result` here but I'm not aware that we've settled on a good error handling pattern so for now I'm continuing to live in denial and return an `Option` instead.)

sequencer/src/block/full_payload/ns_proof.rs
ggutoski: Fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/176f0015e466ee3baf2a7ef473b617fbe003784b

sequencer/src/transaction.rs
jbearer: Ah, I didn't realize we were using `NamespaceId::from`, I thought we were just deserializing the whole transaction in the endpoint.

Anyway, yeah, how about we replace `From<u64>` with `From<u32>` and that will place the typechecking/range checking burden on whoever is trying to create a namespace ID?

jbearer: Oh, we have two endpoints that deal with namespace IDs: the `submit` endpoint just deserializes a whole transaction (which recursively calls deserialize on the namespace ID). The namespace proof endpoint uses `NamespaceId::from` directly. So I think if we fix both deserialize and `From`, we will be safe

ggutoski: `serde` deserialize fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/fed11618f98fdca3a1f738fc4ab5de74875db0d3

`From<u32>` fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/26a6b299ce260a210e6fa3253bbd50fb57f3023d

I prefer to drop `From<u32>` and force the caller to rely solely on `serde`. But at first glance it seems that `serde` is invoked only by `body_auto` in [RequestParams in tide_disco::request - Rust](https://tide-disco.docs.espressosys.com/tide_disco/request/struct.RequestParams.html#method.body_auto), which seems to want to parse the entire request body. By contrast, if you want only an individual param then you need to use `param`, which doesn't seem to use `serde` (?) and instead converts directly to fundamental types such as `u64`, which you must then deserialize into a `NamespaceId`. Instead of fighting with it I just implemented both solutions as you suggest.

Wow this is ugly. New issue https://github.com/EspressoSystems/espresso-sequencer/issues/1574

sequencer/src/api/endpoints.rs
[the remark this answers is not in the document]
ggutoski: This code shall remain removed. It's no longer needed because `serde` now enforces this condition as per https://github.com/EspressoSystems/espresso-sequencer/pull/1499#discussion_r1633684374

sequencer/src/block/full_payload/payload.rs
ggutoski: Posted new issues and linked to them in https://github.com/EspressoSystems/espresso-sequencer/pull/1499/commits/79308305b6b9d49015e44c685bb3289d3feeba4f

New issues spawned from this PR:

# This repo

- #1577 
  - #1574
  - #1575 
  - #1576 

# Hotshot repo

- https://github.com/EspressoSystems/HotShot/issues/3298
- https://github.com/EspressoSystems/HotShot/issues/3300

# hotshot-query-service repo

- https://github.com/EspressoSystems/hotshot-query-service/issues/639

jbearer (approved):
Think we're good here 👍

PR #1505, opened 2024-05-24
[Libp2p upgrade] Load bootstrap info from genesis file

rob-maron:
Closes https://github.com/EspressoSystems/HotShot/issues/3198 and https://github.com/EspressoSystems/HotShot/issues/3197

### This PR:

- Breaks apart `network.rs` into separate files (`cdn.rs` and `libp2p.rs`)
  - Adds a function for loading in genesis network information from the file 
  - Updates the genesis files to add a comment with example values that work for the demo
  - Adds a function for updating the config file with the new values. For both persistently-loaded and orchestrator-loaded configurations, the genesis network file will replace what's in there.
  - Calls the function after loading the config in 

- Waits for either network to be ready on a combined run 

### This PR does not:

Change anything about connections when not using the `libp2p` flag.

### Key places to review:

- `sequencer/lib.rs`
- `network/mod.rs`
- `libp2p.rs`

### How to test this PR:
- Enable the `libp2p` feature in the sequencer and the builder and try supplying the file in the demo.

### Tests run

- [x] Libp2p network does not initialize if peers in supplied file are unreachable
- [x] Libp2p network works when correct example values are supplied
- [x] Binary returns if the file was supplied but malformed
- [x] Made sure it overrides values from the orchestrator


sequencer/src/network/cdn.rs
```diff
@@ -0,0 +1,104 @@
+/// Trait implementations for the CDN
```
rob-maron, line 1: No changes here, broke this out of `network.rs`

sequencer/src/network/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+/// The genesis network configuration. Overrides what is received from the orchestrator
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct GenesisNetworkConfig {
+    /// An optional list of bootstrap nodes (multiaddress and peer ID) to connect to
+    pub bootstrap_nodes: Option<Vec<BootstrapNode>>,
+}
+
+impl GenesisNetworkConfig {
+    /// Load the genesis network config from a file
+    pub fn load_from_file(path: String) -> Result<GenesisNetworkConfig> {
```
jbearer, line 22: Is this used anywhere, since we normally just get the network config as a subset of the genesis file?

rob-maron: Nope, forgot to remove this when I switched things over. Thanks for catching it

jbearer (approved):
Looks good. How does this work in the demo since we don't supply any bootstrap addresses? Isn't it just not connected via libp2p? Is setting that up a next step?

rob-maron:
> Looks good. How does this work in the demo since we don't supply any bootstrap addresses? Isn't it just not connected via libp2p? Is setting that up a next step?

Enabling the libp2p feature flag for the demo and not supplying bootstrap addresses will just have them use the ones from the orchestrator.

I tested seeding it with addresses from a real run of the demo (currently commented out) and it works fine. I also made sure it was using the bootstrap nodes by testing what happens if you supply an empty vector (libp2p doesn't work)

PR #1506, opened 2024-05-27
JSON serialize numbers as decimals

sveitser:
Adds two functions to be used with `(de)serialize_with`. This avoids
having to define a macro, the downside is that it makes it possible to
specify `serialize_with` without `deserialize_with`.

- Changes the serialization of ChainID and FeeAmount.
- Updates reference types. Apparently `chain_config.json` was not sorted
  before.
- Add test to check if a roundtrip works.
- Add test well as a test to check that bincode serialization is not
  affected by the customization.

Close https://github.com/EspressoSystems/espresso-sequencer/issues/1496
Close https://github.com/EspressoSystems/espresso-sequencer/issues/1497


ggutoski (approved):
LGTM but you might want approval from others, too.

PR #1524, opened 2024-05-29
Specifiy fetch requests rate limit for SQL

imabdulbasit:
This PR updates the hotshot-query-service to tag 0.1.26, allowing configuration of the maximum number of concurrent fetch requests from peers.

https://github.com/EspressoSystems/hotshot-query-service/pull/599


sequencer/src/persistence/sql.rs
```diff
@@ -90,6 +90,10 @@ pub struct Options {
 
     #[clap(long, env = "ESPRESSO_SEQUENCER_STORE_UNDECIDED_STATE", hide = true)]
     pub(crate) store_undecided_state: bool,
+
+    // Specifies the maximum number of concurrent fetch requests allowed from peers.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_FETCH_RATE_LIMIT", hide = true)]
```
jbearer, line 95: Let's not hide this, this is a useful thing that we very much want other operators to use

jbearer (approved):

sequencer/src/persistence/sql.rs
imabdulbasit: oops I copied the above field :D

PR #1525, opened 2024-05-30
Add backoff to state catchup retry loops

jbearer:


sequencer/src/catchup.rs
```diff
@@ -59,15 +63,21 @@ pub trait StateCatchup: Send + Sync + std::fmt::Debug {
...
                     Err(err) => {
-                        tracing::warn!(%account, "Could not fetch account, retrying: {err:#}");
-                        async_std::task::sleep(self.retry_interval()).await;
+                        tracing::warn!(%account, ?delay, "Could not fetch account, retrying: {err:#}");
+                        sleep(delay).await;
+
+                        // Try a few times with a short delay, on the off chance that the problem
+                        // resolves quickly. Back off until we eventually reach the maximum delay,
+                        // which should be pretty long.
+                        delay = min(delay * BACKOFF_FACTOR, MAX_RETRY_DELAY);
```
imabdulbasit, line 80: Should we also add some jitter?

imabdulbasit: Or maybe the fetch call can take different time each time it is called

jbearer: sure

jbearer: [071e234](https://github.com/EspressoSystems/espresso-sequencer/pull/1525/commits/071e234ec7f0ce1a4e16b3307e428432dea51bdb)

imabdulbasit (approved):

PR #1534, opened 2024-05-31
Serialize L1 block numbers as hex strings

### LINKED ISSUE #1507: Follow up work on JSON number serialization
Discussion on [zulip](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/Serializing.20amounts.20as.20decimals).

- [x] Make it backwards compatible so that we can still de-serialize the old hex strings.
- [ ] Use hex strings for L1 data but decimal strings for hotshot data.

sveitser:
- Add newtype for L1 block number.
- Implement serialization in backwards compatible manner. Although there is currently no test for this backwards compatibility.
- Don't ignore "actual" files created from reference tests.
- Remove "actual" files for successful tests, if they exist.

After implementing this and looking at the JSON file I'm now of the opinion that we should throw the compatibility with L1 away. It looks rather messy and I don't see what we really gain from it. I now think we should just serialize everything in the way we think is best, most pretty or whatever.

Close #1507


sveitser:
So what I'm saying is we should keep the new L1 block number type but actually serialize L1 block numbers as JSON number like our height, L1 timestamp like our timestamp. I think it's more important that our header serialization is internally consistent instead of L1 derived data being consistent with L1.

PR #1535, opened 2024-05-31
[WEEKLY RELEASE] HotShot 0.5.56

jparr721:
Closes #<ISSUE_NUMBER>


imabdulbasit (approved):

jbearer:
It looks like `test_restart` is timing out, and it appears to be because `SystemContextHandle::shut_down` is not returning. Try running `cargo test --release --all-features --all-targets -- test_restart --nocapture` with this change:
```diff
diff --git a/sequencer/src/context.rs b/sequencer/src/context.rs
index 781f0b15..e4b6b5b8 100644
--- a/sequencer/src/context.rs
+++ b/sequencer/src/context.rs
@@ -248,6 +248,7 @@ impl<N: network::Type, P: SequencerPersistence, Ver: StaticVersionType + 'static
     pub async fn shut_down(&mut self) {
         tracing::info!("shutting down SequencerContext");
         self.handle.write().await.shut_down().await;
+        tracing::info!("shut down consensus");
         self.tasks.shut_down().await;
     }
 ```

You will see `INFO sequencer::context: shutting down SequencerContext`, which gets logged just before we call `shut_down()`, but not `shut down consensus`, which happens just after

**Comment by jbearer** [comment]
There's a simple fix for this. We just need to make sure the read lock taken by the test is dropped before we call a method that gets a write lock:
```diff
diff --git a/sequencer/src/api.rs b/sequencer/src/api.rs
index 1557f30f..609b63ea 100644
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -1182,9 +1182,11 @@ mod test {
         let decided_view = chain.last().unwrap().leaf().view_number();
 
         // Get the most recent state, for catchup.
-        let consensus = network.server.consensus();
-        let consensus_reader = consensus.read().await;
-        let state = consensus_reader.decided_state().await;
+        let state = {
+            let consensus = network.server.consensus();
+            let consensus_reader = consensus.read().await;
+            consensus_reader.decided_state().await
+        };
         tracing::info!(?decided_view, ?state, "consensus state");
 
         // Fully shut down the API servers.
```

However, I think this is indicative of an API problem with the introduction of the `RwLock` around the consensus handle. It is dangerous/easily misused for the `SequencerContext` to have both a method that returns the lock itself, for the caller to use how they want, as well as methods that implicitly take a lock internally. I think we should remove the `consensus` method and instead add specific methods for whatever we need to do with consensus, that guarantee a lock is never held outside of a `SequencerContext` method.

WDYT @jparr721 ?

jbearer (approved):

jbearer (changes requested):
This looks good but I'm requesting changes just to prevent it from being merged until the next round of cappuccino upgrades is complete. Will lift the hold later today

imabdulbasit:
> This looks good but I'm requesting changes just to prevent it from being merged until the next round of cappuccino upgrades is complete. Will lift the hold later today

restart test still fails sometimes

jbearer (approved):
Good to merge whenever now, we have tagged a release for cappuccino

PR #1543, opened 2024-06-03
Add state validation error types

### LINKED ISSUE #1463: Validation should use `thiserror` instead of `anyhow`
We often care, at least in tests, exactly why validation failed, so we can be sure we are covering all the edge cases. Anyhow makes it difficult and brittle to tell why validation failed when it failed. We should instead use a specific error enum so we can check that our tests are hitting the correct cases.

- [x] Change all state validation to use specific error types
- [x] Update `test_validation_max_block_size` to check the error type
- [x] Update `test_validation_base_fee` to check the error type
- [x] Update `test_charge_fee` to check the error type
- [x] Update `test_validate_proposal_error_cases` to check error cases using enum variants rather than string comparisions
- [x] Look for other tests which ought to be updated

tbro:
Add error types for state validation functions and update tests to assert these types.

Closes #1463

### DIFF
--- a/sequencer/src/state.rs
+++ b/sequencer/src/state.rs
@@ -41,8 +41,8 @@ use jf_merkle_tree::{
     prelude::{LightWeightSHA3MerkleTree, MerkleProof, Sha3Digest, Sha3Node},
     universal_merkle_tree::UniversalMerkleTree,
     AppendableMerkleTreeScheme, ForgetableMerkleTreeScheme, ForgetableUniversalMerkleTreeScheme,
-    LookupResult, MerkleCommitment, MerkleTreeScheme, PersistentUniversalMerkleTreeScheme,
-    ToTraversalPath, UniversalMerkleTreeScheme,
+    LookupResult, MerkleCommitment, MerkleTreeError, MerkleTreeScheme,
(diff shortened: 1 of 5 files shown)


sequencer/src/state.rs
```diff
@@ -210,112 +218,167 @@ impl ValidatedState {
     }
 }
 
+/// Possible proposal validation failures
+#[derive(Error, Debug, Eq, PartialEq)]
+pub enum ProposalValidationError {
+    #[error("Invalid ChainConfig: expected={expected}, proposal={proposal}")]
+    InvalidChainConfig { expected: String, proposal: String },
```
jbearer, line 225: How come we use strings here instead of just putting the `ChainConfig`?

jbearer (approved):
Looks good

sequencer/src/state.rs
tbro: Using `String` is just an easy way to avoid[large enum varient](https://rust-lang.github.io/rust-clippy/master/index.html#/large_enum_variant) clippy warning. Since we are only using these error types in tests and for our own comprehension, this easy solution seems good enough to me.

PR #1544, opened 2024-06-04
fix: use espresso-bridge instead of bridge for fund-builder

imabdulbasit:


docker-compose.yaml
```diff
@@ -50,7 +50,7 @@ services:
 
   fund-builder:
     image: ghcr.io/espressosystems/espresso-sequencer/bridge:main
-    command: bridge deposit
+    command: espresso-bridge deposit
```
sveitser, line 53: Thanks. The command is still be "bridge" in the docker image. I kept it like this so we don't have to update it everywhere. But yeah it's confusing. Maybe we should use `ENTRYPOINT` instead for our docker images so we don't have to specify the command. What do you think @Ancient123 ?

sveitser, line 53:
```suggestion
    command: bridge deposit
```

sveitser:
Actually let's update the bridge image and call the executable `espresso-bridge` there too and also symlink `/bin/bridge` to it so that it's consistent within this repo but also backwards compatible. @imabdulbasit

jbearer (approved):

PR #1557, opened 2024-06-05
test: reference_tests use nontrivial payload, ns_table

### LINKED ISSUE #1556: Reference tests use nontrivial payload, namespace table
Current reference tests use only a trivial payload (zero transactions) and namespace table (zero entries). Thus, the reference tests will fail to detect an accidental break of compatibility.

Example: If I change the byte length of a namespace ID in the namespace table then this change will go undetected in the reference test because the reference namespace table has zero entries.

ggutoski:
Closes #1556


jbearer (approved):
LGTM

PR #1562, opened 2024-06-06
Update light client contract to return the height alongside the block commitment

nomaxg:
### This PR:
Our Arbitrum integration needs to know the hotshot height of the block commitment to fetch the merkle proof.


contracts/src/LightClient.sol
```diff
@@ -426,19 +426,19 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
         }
         for (uint256 i = 0; i < commitmentsHeight; i++) {
             if (hotShotCommitments[i].blockHeight >= hotShotBlockHeight) {
-                return hotShotCommitments[i].blockCommRoot;
+                return hotShotCommitments[i];
             }
         }
 
-        return hotShotCommitments[commitmentsHeight - 1].blockCommRoot;
+        return hotShotCommitments[commitmentsHeight - 1];
```
jbearer, line 441: Not related to this change but isn't this unreachable? Here we would be returning something that we already visited in the loop, thus `!(hotShotCommitments[commitmentsHeight - 1].block_height >= hotShotBlockHeight)`, thus we would have reverted on line 433. I feel like it's misleading to have an unreachable `return` here and this should probably be a `revert` instead

jbearer (approved):
Change looks good, my comment pertains to something I noticed about the existing logic of the contract

contracts/src/LightClient.sol
nomaxg: yep, that's right, I believe this is unreachable @alysiahuggins when you revisit this can we simplify this logic?

alysiahuggins: I put it mainly because the compiler complained about not having a default return, would you prefer a revert instead? @nomaxg

PR #1564, opened 2024-06-07
use mold in nix environments

tbro:
Install mold into nix environment and configure rust to use it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,7 @@
     let
       # Mold linker is faster, but is not supported on MacOS.
       moldLinker = !pkgs.stdenv.isDarwin;
-
+      rustDeps = [];
       # node=error: disable noisy anvil output
       RUST_LOG = "info,libp2p=off,isahc=error,surf=error,node=error";
       RUST_BACKTRACE = 1;
--- a/flake.nix
+++ b/flake.nix
@@ -50,7 +50,7 @@
       RUST_LOG = "info,libp2p=off,isahc=error,surf=error,node=error";
       RUST_BACKTRACE = 1;
       ASYNC_FLAGS = " --cfg async_executor_impl=\"async-std\" --cfg async_channel_impl=\"async-std\" ";
-      RUSTFLAGS = "${ASYNC_FLAGS} --cfg hotshot_example";
+      RUSTFLAGS = "${ASYNC_FLAGS} --cfg hotshot_example -C link-arg=-fuse-ld=/usr/local/bin/mold";
       RUSTDOCFLAGS = ASYNC_FLAGS;
       # Use a distinct target dir for builds from within nix shells.
       CARGO_TARGET_DIR = "target/nix";
@@ -182,6 +182,7 @@
             exec ${pkgs.nixFlakes}/bin/nix --experimental-features "nix-command flakes" "$@"
           '';
           solc = pkgs.solc-bin.latest;
+          stdenv = pkgs.stdenvAdapters.useMoldLinker pkgs.clangStdenv;
         in
         mkShell (rustEnvVars // {
           buildInputs = [
--- a/flake.nix
+++ b/flake.nix
@@ -46,11 +46,20 @@
     }:
     flake-utils.lib.eachDefaultSystem (system:
     let
+      # Mold linker is faster, but is not supported on MacOS.
+      moldLinker = !pkgs.stdenv.isDarwin;
+
       # node=error: disable noisy anvil output
       RUST_LOG = "info,libp2p=off,isahc=error,surf=error,node=error";
       RUST_BACKTRACE = 1;
       ASYNC_FLAGS = " --cfg async_executor_impl=\"async-std\" --cfg async_channel_impl=\"async-std\" ";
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -182,16 +191,23 @@
             exec ${pkgs.nixFlakes}/bin/nix --experimental-features "nix-command flakes" "$@"
           '';
           solc = pkgs.solc-bin.latest;
-        in
-        mkShell (rustEnvVars // {
-          buildInputs = [
-            # Rust dependencies
+          stdenv =
```
sveitser, line 194: Is `stdenv` meant to be passed to `mkShell`?

I get compilation errors on NixOS currently (and also if i pass `stdenv` to `mkShell`).

```
  = note: gcc: error: unrecognized command-line option '-fuse-ld=/nix/store/b7hb6xpazdnidp0s87fain2im6lk4f4r-mold-2.30.0/bin/mold'
 ```
So it looks like the devShell still brings in gcc and that's used for compilation.

I can compile with this change

```diff
diff --git a/flake.nix b/flake.nix
index 66d29e7a..7e0872a5 100644
--- a/flake.nix
+++ b/flake.nix
@@ -205,7 +205,8 @@
             jq
           ] ++ lib.optionals moldLinker [ mold ];
         in
-        mkShell (rustEnvVars // {
+        stdenv.mkDerivation (rustEnvVars // {
+          name = "shell";
           buildInputs = [
             stableToolchain
```

However linking still seems fairly slow. What kind of speedup do you get?

tbro: > Is `stdenv` meant to be passed to `mkShell`?
> 
> I get compilation errors on NixOS currently (and also if i pass `stdenv` to `mkShell`).
> 
> ```
>   = note: gcc: error: unrecognized command-line option '-fuse-ld=/nix/store/b7hb6xpazdnidp0s87fain2im6lk4f4r-mold-2.30.0/bin/mold'
> ```

Sorry my nix skills are very thin, so I'm probably doing something wrong. The `gcc` error is the same that caused me to [add the env var](6d071ed7daa7). I'm unsure what configuration is causing `gcc` to be selected in some cases and not others. Possibly the issue is related to the gcc *version*. 
 
> So it looks like the devShell still brings in gcc and that's used for compilation.
> 
> I can compile with this change
> 
> ```diff
> diff --git a/flake.nix b/flake.nix
> index 66d29e7a..7e0872a5 100644
> --- a/flake.nix
> +++ b/flake.nix
> @@ -205,7 +205,8 @@
>              jq
>            ] ++ lib.optionals moldLinker [ mold ];
>          in
> -        mkShell (rustEnvVars // {
> +        stdenv.mkDerivation (rustEnvVars // {
> +          name = "shell";
>            buildInputs = [
>              stableToolchain
> ```
> 
> However linking still seems fairly slow. What kind of speedup do you get?

It is noticeably faster for me. It isn't very straightforward to get real figures, because the biggest speedups are on incremental changes and modifying `RUSTFLAGS` will trigger a recompile of many things.

sveitser: What I think works quite well and is somewhat realistic for development is to run a full build. Then `touch sequencer/src/lib.rs` and then time the full build again.

sveitser:
```bash
#!/usr/bin/env bash

function build() {
    echo "building..."
    cargo clippy --all-features
}

build
touch sequencer/src/lib.rs
echo "timing re-build ..."
time build
```

I tried with this script and for both the new shell and the old shell I get 4-5 seconds `real` time.

sveitser: I think the linker is actually not invoked when running cargo clippy. If I make this change

```
diff --git a/flake.nix b/flake.nix
index 66d29e7a..5015ff00 100644
--- a/flake.nix
+++ b/flake.nix
@@ -55,7 +55,7 @@
       ASYNC_FLAGS = " --cfg async_executor_impl=\"async-std\" --cfg async_channel_impl=\"async-std\" ";
       LINKER_FLAGS =
         if moldLinker then
-          "-C link-arg=-fuse-ld=${pkgs.mold}/bin/mold"
+          "-C link-arg=-fuse-ld=${pkgs.mold}/bin/mold-blah"
         else
           ""
       ;
```

things still work.

sveitser: Also the shell still doesn't seem to be using `mold` for linking in my case 🤔

tbro: maybe you can confirm like so:

```bash
readelf -p .comment target/debug/sequencer

String dump of section '.comment':
  [     0]  **mold** 2.31.0 (20fa8d56f5e0c47d1f4bbf7b829c12d3f43298e1; compatible with GNU ld)
  [    4f]  rustc version 1.78.0 (9b00956e5 2024-04-29)
  [    7b]  GCC: (Debian 12.2.0-14) 12.2.0
```

tbro: > I think the linker is actually not invoked when running cargo clippy. If I make this change

Cargo clippy evokes `rustc` w/ linker arguments... I think some linking would have to go on since you are compiling a new binary... but I'm really not clear on the internals.

tbro: Maybe if you want to get a better idea if it mold is useful for you, maybe start by removing nix from the equation. Then we can figure out how to fix my terrible nix code :/.

PR #1566, opened 2024-06-07
[WEEKLY RELEASE] HotShot 0.5.57

jparr721:
Closes #<ISSUE_NUMBER>


sequencer/src/persistence/fs.rs
```diff
@@ -408,6 +412,86 @@ impl SequencerPersistence for Persistence {
             },
         )
     }
+    async fn append_quorum_proposal(
+        &mut self,
+        proposal: &Proposal<SeqTypes, QuorumProposal<SeqTypes>>,
+    ) -> anyhow::Result<()> {
+        let view_number = proposal.data.view_number().u64();
```
jparr721, line 419: Thoughts @jbearer?

sequencer/src/persistence/fs.rs
```diff
@@ -408,6 +412,86 @@ impl SequencerPersistence for Persistence {
...
+        let dir_path = self.proposals_dir_path();
+
+        fs::create_dir_all(dir_path.clone()).context("failed to create proposals dir")?;
+
+        let file_path = dir_path.join(view_number.to_string()).with_extension("txt");
+        self.replace(
+            &file_path,
+            |_| {
+                // Don't overwrite the file, we want to append since we can get multiple
+                // saved quorum proposals.
```
jbearer, line 429: Not for the same view right? I think not overwriting the file is fine, but it's not going to cause us to append, it's just going to keep what was already there

sequencer/src/persistence/fs.rs
```diff
@@ -408,6 +412,86 @@ impl SequencerPersistence for Persistence {
...
+    async fn append_quorum_proposal(
+        &mut self,
+        proposal: &Proposal<SeqTypes, QuorumProposal<SeqTypes>>,
+    ) -> anyhow::Result<()> {
+        let view_number = proposal.data.view_number().u64();
+        let dir_path = self.proposals_dir_path();
+
+        fs::create_dir_all(dir_path.clone()).context("failed to create proposals dir")?;
+
+        let file_path = dir_path.join(view_number.to_string()).with_extension("txt");
```
jbearer, line 424: Do we ever prune old ones?

sequencer/src/persistence/sql.rs
```diff
@@ -452,6 +455,30 @@ impl SequencerPersistence for Persistence {
             .transpose()
     }
 
+    async fn load_quorum_proposals(
+        &self,
+    ) -> anyhow::Result<Option<BTreeMap<ViewNumber, Proposal<SeqTypes, QuorumProposal<SeqTypes>>>>>
+    {
+        let rows = self
+            .db
+            .query_static("SELECT * FROM quorum_proposals")
```
jbearer, line 464: We need a migration to create this table

jbearer (changes requested):

sequencer/src/persistence/fs.rs
jbearer: If it is correct to delete anything equal or less than the decided view (I don't know if that's what we want, but if it is) it should be to just add this to the `delete_files` calls in `collect_garbage`. And likewise the `DELETE` statements in `collect_garbage` in `persistence/sql.rs`

sequencer/src/persistence/fs.rs
jparr721: Sorry, this was an old comment. It'll write a new entry per view. I can fix this.

sequencer/src/persistence/fs.rs
jparr721: Yeah I think that's the case, I can update that.

sequencer/src/persistence/fs.rs
jbearer: Looks reasonable

sequencer/src/persistence/sql.rs
```diff
@@ -452,6 +458,30 @@ impl SequencerPersistence for Persistence {
...
+            .await?;
+
+        Ok(Some(BTreeMap::from_iter(
+            rows.map(|row| {
+                let row = row?;
+                let view: i64 = row.get("view");
+                let view_number: ViewNumber = ViewNumber::new(view.try_into()?);
+                let bytes: Vec<u8> = row.get("proposal");
```
jbearer, line 475: Should be `data` now I think

jbearer (approved):
Looks good

PR #1578, opened 2024-06-10
Update the CDN

rob-maron:
- Improves throughput ~4x for larger (1000 node) runs by reducing contention
- Better logging
- Moves allocation pool (limiter for the the amount of total outstanding bytes not processed) to be definable during runtime
- Adds TCP+TLS support. On my local machine it's much faster (10x??) than QUIC for the scenario I tested, we may want to use it in the future. But it may be from lack of GSO support on MacOS 
- Adds latency buckets for tail latency of message received -> sent to everyone
  - Adds 30s sliding window tail latency metric

Unrelated changes:
- `Topics` didn't like being updated out of band with HotShot, so I pulled them and and added static assertions that they are equal

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -674,7 +674,7 @@ dependencies = [
  "async-std",
  "async-trait",
  "color-eyre",
- "console-subscriber",
+ "console-subscriber 0.2.0",
  "flume 0.11.0",
  "futures",
  "tokio",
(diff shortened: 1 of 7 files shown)


jbearer (approved):

rob-maron:
Closed in #1530

Issue #1328, opened 2024-04-12
update prover permissions on the light client contract via openzeppelin defender and safe wallets


alysiahuggins:
Context: 
Defender Admin is used so that you can manage your smart contracts using your multisig walletes and timelocks. Read more here (https://docs.openzeppelin.com/defender/v1/admin). There are two additional functions that have been added to the LightClient contract https://github.com/EspressoSystems/espresso-sequencer/issues/1175 that can only be called by the admin (a multisig)

Objective: 
Create a workflow via the defender-admin-client (https://www.npmjs.com/package/@openzeppelin/defender-admin-client) to create proposals for calling the following functions on the LightClient contract via our multisig wallet.
- setPermissionedProverMode(bool)
- updatePermissionedProver(address)

Theres's a preference to use pnpm over yarn/npm to manage the node packages.

PR #1585, opened 2024-06-11
combined the prover proposal scripts

### LINKED ISSUE #1502: De-duplicate typescript contract operation scripts
Combine `setProverProposal.ts` and `disableProverProposal.ts` into one script with 2 subcommands.

              De-duplicate typescript contract operation scripts

_Originally posted by @sveitser in https://github.com/EspressoSystems/espresso-sequencer/pull/1454#discussion_r1613393853_

alysiahuggins:
Closes #1502 

### This PR:
There was previously a separate script for setting the prover and disabling the prover on the light client contract. This combines the two into one script. 

### Key places to review:
The command line arguments required for each proposal and the fact that the setProver depends on the env variable, APPROVED_PROVER_ADDRESS for setting the prover. This hasn't changed from the previous implementation but worth mentioning. 

 ### How to test this PR:
See the read me - it was tested this way in #1328 as well. 

### Things tested 
Manually tested the disableProver and setProver workflow. 

See PR https://github.com/EspressoSystems/espresso-sequencer/pull/1454


contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -28,17 +32,24 @@ async function main() {
...
-    await proposeSetProverTransaction(
-      safeSdk,
-      safeService,
-      await orchestratorSigner.getAddress(),
-      safeAddress,
-      permissionedProverAddress,
-    );
+    const orchestratorSignerAddress = await orchestratorSigner.getAddress();
+
+    if (command == SET_PROVER_CMD) {
```
Ayiga, line 37: `===` is the prefered comparison operator in a JavaScript environment between JavaScript primitives.  Otherwise loose or truthy equality might get through, in this case it likely doesn't matter since we're comparing strings.  But I think, in general, it's better to err on the side of caution.

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -48,6 +59,21 @@ async function main() {
...
+  const args = process.argv.slice(2); // Remove the first two args (node command and script name)
+  let command: string;
+  if (args.length === 0) {
+    console.log("No arguments provided.");
+    throw new Error(`No arguments provided, either ${SET_PROVER_CMD} or ${DISABLE_PROVER_CMD}`);
+  } else {
+    command = args[0];
+    if (command != SET_PROVER_CMD && command != DISABLE_PROVER_CMD) {
+      throw new Error(`Unrecognized argument ${command} provided, either ${SET_PROVER_CMD} or ${DISABLE_PROVER_CMD}`);
+    }
```
Ayiga, line 72: t seems odd to perform the comarison checks on the commands here, since we're already going to be comparing them in `main`.  We could just catch in unmatched commands in `main` and catch the error there instead.

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -48,6 +59,21 @@ async function main() {
   }
 }
 
+function processCommandLineArguments(): string {
+  const args = process.argv.slice(2); // Remove the first two args (node command and script name)
+  let command: string;
+  if (args.length === 0) {
+    console.log("No arguments provided.");
```
Ayiga, line 66: The language here might want to specify that no "command" was found instead of "arguments"

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -48,6 +59,21 @@ async function main() {
   }
 }
 
+function processCommandLineArguments(): string {
+  const args = process.argv.slice(2); // Remove the first two args (node command and script name)
```
Ayiga, line 63: Do we want to possibly consider command line flags for this script?

Will this always be invoked with this form:

```sh
ts-node <script-name> <command>
```

Or might it be invoked in another way, like a node-binary that could be invoked with `npx` or `tsx`?

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -102,6 +128,57 @@ function createPermissionedProverTxData(proverAddress: string): string {
...
+  safeAddress: string,
+) {
+  // Prepare the transaction data to disable permissioned prover mode
+  let data = createDisablePermissionedProverTxData();
+
+  const contractAddress = getEnvVar("LIGHT_CLIENT_CONTRACT_ADDRESS");
+  isValidEthereumAddress(contractAddress);
+
+  // Create the Safe Transaction Object
+  const safeTransaction = await createSafeTransaction(safeSDK, contractAddress, data, "0");
```
Ayiga, line 151: Due to the signature of createSafeTransaction, this doesn't indicate it's return type in intellisense.  By removing the explicit return type in `createSafeTransaction`, you can get some type inference here, that could prevent future errors due to the type checking alone.  Othwerise, this allows `createSafeTransaction` to change it's implementation with no errors indicated in the editor, or when the typescript gets interpretted.

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
```diff
@@ -28,17 +32,24 @@ async function main() {
...
-      await orchestratorSigner.getAddress(),
-      safeAddress,
-      permissionedProverAddress,
-    );
+    const orchestratorSignerAddress = await orchestratorSigner.getAddress();
+
+    if (command == SET_PROVER_CMD) {
+      console.log(`${command}`);
+      const permissionedProverAddress = getEnvVar("APPROVED_PROVER_ADDRESS");
+      isValidEthereumAddress(permissionedProverAddress);
```
Ayiga, line 40: I think the name `isValidEthereumAddress` doesn't really convey that this function throws.  By the name alone, it seems like this would just return a `boolean`, and it looks like this `boolean` isn't inspected at all which seems odd, and makes the function call seem pointless.

However, upon inspecting the function, it seems that this function will return `true` or throw an `Error`.  So it seems like this is really a validation function for the address to ensure that it is a correctly formatted address.  I think returning `true` in this case is unecessary as nothing is even done with the result.  I this this function should probably be named more appropriately to reflect that this is validating.

There's also room for improvement with the error message returned here, to provide the context of which address is throwing the error.

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
Ayiga: It looks like `SafeTransaction` isn't exposed directly in the `@safe-global/protocol-kit` package, which is saddening.  You **could** reach into the package to retrieve it via 
```typescript
import SafeTransaction from "@safe-global/protocol-kit/dist/src/utils/transactions/SafeTransaction";
```

But this seems rather rude.  A better solution might be to declare a local forward declaration of the type via the resolved return type of `safeSDK.createTransaction`:


```typescript
type SafeTransaction = Awaited<ReturnType<Safe['createTransaction']>>;
```

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: oh good point, addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/b9c694be95c90c23b91c79271db861cc9991c0cb

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: the idea for processing them here is to make sure that the user executed the script with an appropriate command before the doing other work such as checking for other environment configs and setting up the safeSDK

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: ah i feel you, updated here https://github.com/EspressoSystems/espresso-sequencer/commit/ac3bd820121b9d8b1f0a9d25bf2ef60f7353755a

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: Good point, i've updated it here https://github.com/EspressoSystems/espresso-sequencer/commit/418c16ff56e0a0ad830c4fd0f47bf2c5c941a9c6

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: This script is going to be run rarely, but can also be ran with npx. What is your preference?

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
alysiahuggins: Yea I initially considered creating a local type definition when I realized that it wasn't exposed but thought that it wasn't the best approach. I love this approach, thank you. implemented here https://github.com/EspressoSystems/espresso-sequencer/commit/158ed1bcf3d3ee310bbd185f2d6c90690c13472d

contracts/script/multisigTransactionProposals/safeSDK/modifyProverModeProposal.ts
Ayiga: Ah.  I just wanted to make sure that the calculation of the args here would always be correct.  Just to make sure that this is indeed where the args will start.

Ayiga (approved):

PR #1588, opened 2024-06-11
test: reference test for Payload

### LINKED ISSUE #1587: Reference tests for `Payload` struct
`reference_tests.rs` is missing a test for `Payload` struct. PR #1499 introduced a breaking change to serialization for this struct that was not detected in `reference_tests.rs`. Create a test that passes against `main` branch immediately prior to #1499 . Then modify the current tip of `main` to pass that test.

ggutoski:
Closes #1587 

### This PR:
A new reference test for `Payload` was created against pre- #1499 `main` branch in 6d0a2bfb5626d53c445d6ecdf5334bad93b4a4ad . This PR cherry-picks that new test at current tip of `main`, post- #1499, in order to ensure that #1499 did not introduce breaking changes to `Payload`.


sequencer/src/block/full_payload/payload.rs
```diff
@@ -281,3 +281,20 @@ impl hotshot_types::traits::block_contents::TestableBlock<SeqTypes> for Payload
         self.len(&self.ns_table) as u64
     }
 }
+
+// `Committable` used only in `reference_tests.rs`
+#[cfg(any(test, feature = "testing"))]
+impl committable::Committable for Payload {
```
jbearer, line 287: I don't like adding this implementation solely to test that the implementation is backwards compatible. Can we just pull out the parts of `reference_test` that deal with serialization into a separate function and not do the commitment test for payload?

jbearer (changes requested):

sequencer/src/block/full_payload/payload.rs
ggutoski: Done in https://github.com/EspressoSystems/espresso-sequencer/pull/1588/commits/321d1772c6cf24db309e1f87614206f61fd13d08 .

Deserialization tests also rely on `Committable`, so I had to tweak them to instead rely on `Eq + Debug`. This works because everything we're testing impls `Eq` and `Debug`.

If we don't wish to impose these trait bounds on our test types then we could instead simulate the `Committable`-based test without the `Committable` trait bound via serialize + hash. But I did not do that.

jbearer (approved):

PR #1600, opened 2024-06-12
[Prover] Sequencer node stake table initialization

rob-maron:
Allows the prover to pull in the stake table from sequencer nodes running with `--` `config`. 

This makes it so that a transient failure of both the prover and orchestrator does not cause the prover to lose the stake table permanently. This would cause it to stop working entirely

### DIFF
--- a/hotshot-state-prover/src/service.rs
+++ b/hotshot-state-prover/src/service.rs
@@ -1,7 +1,7 @@
 //! A light client prover service
 
 use crate::snark::{generate_state_update_proof, Proof, ProvingKey};
-use anyhow::anyhow;
+use anyhow::{anyhow, Context, Result};
 use async_std::{
     io,
     sync::Arc,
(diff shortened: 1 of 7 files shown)


hotshot-state-prover/src/service.rs
```diff
@@ -103,62 +102,70 @@ pub fn init_stake_table(
     Ok(st)
 }
 
-async fn init_stake_table_from_orchestrator(
-    orchestrator_url: &Url,
+#[derive(Debug, Deserialize)]
+pub struct PublicHotShotConfig {
```
jbearer, line 106: Nit: why not just import this from sequencer?

mrain (approved):
LGTM

jbearer (approved):
Nice

hotshot-state-prover/src/service.rs
rob-maron: I tried that but cargo told me there was a cyclic dependency problem somewhere. I also considered moving it to `utils` but it didn't seem in the util spirit. 

I can keep pushing if we want though, maybe it wouldn't be too bad

mrain: Deploying the light client contract will invoke `light_client_genesis` fn from `hotshot-state-prover` repo. Thus a cyclic dep.

PR #1607, opened 2024-06-14
feat: enforce namespace table validity at consensus

### LINKED ISSUE #1605: Consensus changes for namespace table
See [Proposal: consensus changes for namespace table](https://www.notion.so/espressosys/Proposal-consensus-changes-for-namespace-table-ca1dddb0edf549d286f368389efe182e?d=8fe5a35ba462481bb62c8681be6ab0b7#65637f77d34b4562a4b26b5dd1072611).

Summary excerpted from the above doc:
1. Disallow duplicate namespace IDs
2. Sort by namespace ID
3. Monotonically increasing offsets
4. No partial entries

Code to enforce these new consensus rules goes here: https://github.com/EspressoSystems/espresso-sequencer/blob/1951d7a7e579f5e2f852375d3d8ac4d44e3d6d2a/sequencer/src/state.rs#L257

ggutoski:
Closes #1605 

### This PR:
- New method `NsTable::validate` to check whether the bytes of a namespace table meet the desired conditions.
- Exact conditions specified in docs here: https://github.com/EspressoSystems/espresso-sequencer/blob/058ea4bd9619590c697074b847a1f598460335da/sequencer/src/block/full_payload/ns_table.rs#L210-L218
- Modify fn `validate_proposal` from `state.rs` to call `proposal.ns_table.validate()?;` at the end.
- New unit tests in new source file `sequencer/src/block/full_payload/ns_table/test.rs`
- Simplify `NsIter` to be like `TxIter` (ie. just a `Range` wrapper) now that we can assume there are no duplicate namespace IDs.
- Fix test in `persistence.rs`  

### This PR does not:

Test `submit_test_with_query_module` is hard to debug. If something bad happens then the test just runs forever with log messages like
```
2024-06-14T23:51:50.4143933Z   2024-06-14T23:51:50.414097Z  WARN hotshot_task_impls::transactions: Couldn't get a block: no available blocks
2024-06-14T23:51:50.4145539Z     at /home/runner/.cargo/git/checkouts/hotshot-0ac703037ea89b98/828f718/crates/task-impls/src/transactions.rs:290
2024-06-14T23:51:50.4146966Z     in hotshot_task_impls::transactions::wait_for_block with id: 1, view: 1408
2024-06-14T23:51:50.4148270Z     in hotshot_task_impls::transactions::Transaction task with id: 1, view: 1407
2024-06-14T23:51:50.4148937Z
2024-06-14T23:51:50.4402287Z ##[error]The action 'Test' has timed out after 40 minutes.
``` 

### Key places to review:
- `NsTable::validate`
- Serde for `NsTable` needed a workaround to call `validate` during deserialization. Search `382589140` for relevant code.
- All changes to `sequencer/src/state.rs`
- Reduce the traits supported by `NamespaceId`: no more `Copy`, `Default`, `Namespace`. Also remove `Namespaced` from `Transaction`.
- `test_append_and_collect_garbage` is `persistence.rs` was malformed. This PR caused it to fail. Fixed the test in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/5ffd2e2241ec6b3a1167ccf3e90a2ab4ec3efb62. Fixed test persists only an empty (genesis) block--is this ok or do you want it to test a nontrivial block? (Previously the block it tested was malformed, which triggered the new namespace table validation code of the present PR.)


tbro (commented):
I haven't quite made it though the code yet, but here are a few nits in the meantime.

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -116,28 +118,64 @@ const NS_ID_BYTE_LEN: usize = 4;
...
     /// Search the namespace table for the ns_index belonging to `ns_id`.
     pub fn find_ns_id(&self, ns_id: &NamespaceId) -> Option<NsIndex> {
         self.iter()
             .find(|index| self.read_ns_id_unchecked(index) == *ns_id)
     }
 
+    /// Number of entries in the namespace table.
+    ///
+    /// Returns the maximum number of entries that could fit in the namespace
+    /// table, ignoring what's declared in the table header.
```
jbearer, line 168: This seems more like a `capacity` function than `len` then?

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -166,17 +204,58 @@ impl NsTable {
...
+    /// 2. All namespace IDs and offsets must increase monotonically. Offsets
+    ///    must be nonzero.
+    /// 3. Header consistent with byte length (obsolete after
+    ///    <https://github.com/EspressoSystems/espresso-sequencer/issues/1604>)
+    pub fn validate(&self) -> Result<(), NsTableValidationError> {
+        use NsTableValidationError::*;
+
+        // Byte length for a table with `x` entries must be exactly
+        // `x * NsTableBuilder::entry_byte_len() + NsTableBuilder::header_byte_len()`
+        if self.bytes.len() % NsTableBuilder::entry_byte_len() != NsTableBuilder::header_byte_len()
```
jbearer, line 223: Nit: this would fail if `header_byte_len` is a multiple of `entry_byte_len`

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -237,6 +317,16 @@ impl Committable for NsTable {
     }
 }
 
+/// Return type for [`NsTable::validate`].
+///
+/// TODO wtf is our error handling pattern?
```
jbearer, line 322: The new convention, that we are aiming to stick with, is to use `thiserror` when we want to distinguish different types of errors in code (even in tests) and use `anyhow` otherwise. So this should be good

sequencer/src/persistence.rs
```diff
@@ -451,18 +451,13 @@ mod persistence_tests {
...
 
-        let mut seed = [0u8; 32];
-        let mut rng = rand_chacha::ChaChaRng::from_entropy();
-        rng.fill_bytes(&mut seed);
-
-        let tx = Transaction::random(&mut rng);
-        let tx_hash = Sha256::digest(tx.payload()).to_vec();
-        let block_payload_signature =
-            BLSPubKey::sign(&privkey, &tx_hash).expect("Failed to sign tx hash");
+        // TODO `leaf` is the genesis (empty) block; is that ok?
```
jbearer, line 454: Yes, this should be fine. The only thing that really matters for this test is the view number, since that's the key we use for GC

sequencer/src/transaction.rs
```diff
@@ -21,21 +20,14 @@ use serde::{de::Error, Deserialize, Deserializer, Serialize};
...
 ///   it makes the most sense to put serialization. See
 ///   <https://github.com/EspressoSystems/espresso-sequencer/pull/1499#issuecomment-2134065090>
-/// - It impls [`Namespace`] from [`jf_merkle_tree`], but this seems unneeded
-///   now that we're not using jellyfish's namespace merkle tree.
-/// - We derive lots of things that perhaps we shouldn't: `Into`, `From`,
-///   `Default`, `Ord`. Perhaps derivations for [`NamespaceId`] should match
-///   that of [`Transaction`].
 #[derive(
     Clone,
-    Copy,
```
jbearer, line 31: Why shouldn't this be copyable?

jbearer (approved):

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -116,28 +118,64 @@ const NS_ID_BYTE_LEN: usize = 4;
...
+    where
+        D: Deserializer<'de>,
+    {
+        let unchecked = NsTable::deserialize(deserializer)?;
+        unchecked.validate().map_err(de::Error::custom)?;
+        Ok(unchecked)
+    }
+}
+
+// Boilerplate to check invariants on deserialization
```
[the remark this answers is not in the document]
ggutoski: Clarified in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/043dd5f00c759016da87af05f48fd0a27d223202

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -116,28 +118,64 @@ const NS_ID_BYTE_LEN: usize = 4;
...
-/// #[derive(Clone, Debug, Default, Deserialize, Eq, Hash, PartialEq, Serialize)]
+/// #[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 /// #[serde(transparent)]
 /// pub struct NsTable(#[serde(with = "base64_bytes")] Vec<u8>);
 /// ```
 /// but we need to maintain serialization compatibility.
 /// <https://github.com/EspressoSystems/espresso-sequencer/issues/1575>
-#[derive(Clone, Debug, Default, Deserialize, Eq, Hash, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+#[serde(remote = "Self")] // workaround https://github.com/serde-rs/serde/issues/1220#issuecomment-382589140
```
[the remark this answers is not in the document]
ggutoski: Reworded in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/043dd5f00c759016da87af05f48fd0a27d223202.

Perhaps "workaround" isn't the right word. It is an ugly hack with a lot of boilerplate to achieve something basic.

sequencer/src/block/full_payload/ns_table.rs
ggutoski: Fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/1807a243f96a5ddcf45f8276da07beb1e7a2eb83.

That is not a nit; it's a latent bug. Maybe one day I'll post a PR where you don't find a bug in review. 🤦‍♂️

ggutoski: Fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/e1ef35f86a979648bae18329e22fa24533b2f567.

That's not a nit; it's a latent bug. Perhaps one day I'll post a PR where you don't find a bug in review. 🤦‍♂️

sequencer/src/persistence.rs
ggutoski: comment removed in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/5afc03f0c9c8b3c95cc12d49d1513270bf194275

sequencer/src/transaction.rs
ggutoski: `Copy` restored in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/b01f5717cd8f1fde6805a3c2131666d7a619212c.

This is a matter of philosophy. I like my types to promise as little as possible. `Copy` suggests that cloning this type will always be a fast bitwise copy. That's probably true, but what's the benefit of committing to that now?

For a different perspective, [official Rust doc](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy) says "Generally speaking, if your type *can* implement `Copy`, it should" and then immediately proceeds to a warning about making promises you can't keep. 🤷‍♂️

sequencer/src/block/full_payload/ns_table.rs
ggutoski: Thanks. Comment removed in https://github.com/EspressoSystems/espresso-sequencer/pull/1607/commits/71583bde78b58123adbadd55fca83c1847a094c4.

PR #1610, opened 2024-06-17
Test framework for restartability

jbearer:
Set up some Rust automation for tests that spin up a sequencer network and restart various combinations of nodes, checking that we recover liveness. Instantiate the framework with several combinations of nodes as outlined in https://www.notion.so/espressosys/Persistence-catchup-and-restartability-cf4ddb79df2e41a993e60e3beaa28992.

As expected, the tests where we restart >f nodes do not pass yet, and are ignored. The others pass locally.

There are many things left to test here, including:
* Testing with a valid libp2p setup
* Testing with _only_ libp2p and no CDN
* Checking integrity of the DA/query service during and after restart But this is a pretty good starting point.

I considered doing this with something more dynamic like Bash or Python scripting, leaning on our existing docker-compose or process-compose infrastructure to spin up a network. I avoided this for a few reasons:
* process-compose is annoying to script and in particular has limited capabilities for shutting down and starting up processes
* both docker-compose and process-compose make it hard to dynamically choose the network topology
* once the basic test infrastructure is out of the way, Rust is far easier to work with for writing new checks and assertions. For example, checking for progress is way easier when we can plug directly into the HotShot event stream, vs subscribing to some stream via HTTP and parsing responses with jq

Closes #<ISSUE_NUMBER>


jbearer:
Alright, I've found a potential deadlock that might explain the CI timeouts.

During the process of initializing a node, we ([sometimes](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/files#diff-835e175467c9abeffff8cd340e292c8289baac4102027459d1d0a9935a3c5fa7R755-R758)) [call out](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/files#diff-5fa8c38c7b8802cca20b09a82671e6a1452d67b9c1b94165c5b58c0d78265a4dR404-R406) [to the API data source](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/files#diff-835e175467c9abeffff8cd340e292c8289baac4102027459d1d0a9935a3c5fa7R761-R770). Basically, this happens when there were unprocessed leaves from before the shutdown, that we now try to process. The event consumer then [tries to get an exclusive lock](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/files#diff-09ac4501918a6027b5399406fc941a4f295afd5b7bb657fdfd73e621774a0550R40) to update the data source.

Unfortunately, there are [certain API endpoints](https://github.com/EspressoSystems/espresso-sequencer/blob/a61d0fde20eea25155a297c9d4b20fd40a043c52/sequencer/src/api/endpoints.rs#L222-L275) that also lock the data source (for reading) which then [block on initialization completing](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/files#diff-9c5c25f61c89b1c60988253125d7d03fb13becb07c9242f5966669ecfa00d373R108) so they can read from the sequencer context.

Thus, the potential deadlock is:

* Node starts API server
* Another node sends a catchup request (for chain config, frontier, or fee account)
* Request handler locks API state and then awaits future for context initialization
* Context initialization gets to the point where it has to process old leaves and blocks on a write lock on API state

I believe a clean solution to this is to separate the back-processing of old persisted leaves from the loading of consensus state, and move it instead to the beginning of the asynchronous task which processes future leaves. This means that node initialization will no longer block on the data source lock. This will also make consensus state loading read-only, which the fact that it wasn't was a code smell anyways

jbearer:
Fix for the above deadlock in [7bda75e](https://github.com/EspressoSystems/espresso-sequencer/pull/1610/commits/7bda75ede8820039fcaf54feefe1c62232799544), and tests are now passing!

tbro (approved):
There is a lot here, but I've done as thorough of a review as I am able. I think the extensive test coverage should make up the difference. I'm approving. I've made a couple of minor comments, but I leave it up to you if should continue discussing them in another venue.

> I considered doing this with something more dynamic like Bash or Python scripting

I'm glad you didn't do that.

jbearer:
Superseded by #1985

PR #1626, opened 2024-06-19
Make `Header.fee_info` iterable

tbro:
Wrap `Vec<FeeInfo>` in a new type so we can implement things like `Committable` over it. Update code where possible to deal with the structure change. Further changes needed in hotshot (and builder?).

### This PR:
Allows us to have multiple fees per block in order to implement Decaf version of Marketplace. It is an early stage to take comment on general direction.

### This PR does not:
Currently compile. Some changes in the traits defined in HotShot will need to be updated before this can be shipped.

### Key places to review:
  * New type `FeeInfos` is added to `state.rs`, and most code updates occur there. 
  * BlockHeader is updated to use `FeeInfos` in `headers.rs`

### DIFF
--- a/sequencer/src/header.rs
+++ b/sequencer/src/header.rs
@@ -4,7 +4,7 @@ use crate::{
     eth_signature_key::BuilderSignature,
     genesis::UpgradeType,
     l1_client::L1Snapshot,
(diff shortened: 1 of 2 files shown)


sveitser:
Superseded by #1659

PR #1633, opened 2024-06-20
Implement BidTx

tbro:
### This PR:
This PR itroduces `FullNetworkTx` by way of a single enum variant, `BidTx`. To meet upcoming goals we only need `BidTx` but having it nested in long term structure may be helpful. There is a tiny bit of test coverage to ensure a moderate degree of santity.

This is a DRAFT and exists for collaboration and to gather feedback.

### Key places to review:
 Most of the juicy bits are in `auction.rs`. `Header` in `header.rs` has been given some useful methods (mocks for now) and validation logic in `state.rs` has been updated to execute the transaction. None of the logic is complete or refined, but it more or less covers the *breadth* of requirements for the first iteration of `BidTx`. 

### How to test this PR: 
Tests are currently testing, you can run `cargo test` if you like.

### Things tested
There is currently a test for verifying signature of a mock `BidTx`.

### DIFF
--- a/sequencer/src/auction.rs
+++ b/sequencer/src/auction.rs
@@ -1,11 +1,10 @@
 use crate::{
-    eth_signature_key::EthKeyPair,
+    eth_signature_key::{EthKeyPair, SigningError},
     state::{FeeAccount, FeeAmount},
     NamespaceId, ValidatedState,
 };
 use committable::{Commitment, Committable};
 use ethers::types::Signature;
-use hotshot::types::SignatureKey;
 use hotshot_types::{data::ViewNumber, traits::signature_key::BuilderSignatureKey};
 use serde::{Deserialize, Serialize};
 use std::{
@@ -78,7 +77,6 @@ pub enum FullNetworkTx {
 
 impl FullNetworkTx {
     pub fn execute(&self, state: &ValidatedState) -> Result<(), (ExecutionError, FullNetworkTx)> {
-        dbg!("execute");
         match self {
             Self::Bid(bid) => bid.execute(state),
         }
@@ -141,6 +139,13 @@ impl Committable for BidTxBody {
                 &bincode::serialize(&self.bundle.as_slice()).unwrap(),
             );
         comm.finalize()
+    }
+}
+
+impl BidTxBody {
+    /// Sign Tx
+    fn sign(&self, key: &EthKeyPair) -> Result<Signature, SigningError> {
+        FeeAccount::sign_builder_message(key, self.commit().as_ref())
     }
 }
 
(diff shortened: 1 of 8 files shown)


sequencer/src/auction.rs
```diff
@@ -0,0 +1,304 @@
...
+    Ok(T),
+    Err(E, K),
+}
+
+// enum ExecutionResult {
+//     Bid(String),
+// }
+
+/// A solution to one auction of sequencing rights for namespaces
+struct AuctionResultTx {
```
jparr721, line 243: Is this a single result, so an HTTP call to the solver to get the auction results for this view would be `Vec<AuctionResultTx>`?

jparr721: Also, where would the builder URL be stored in this case?

tbro: I'm resolving this as it was discussed in zulip that this is a mistake left over from pre-JIT version.

sequencer/src/auction.rs
```diff
@@ -0,0 +1,312 @@
...
+    fn commit(&self) -> Commitment<Self> {
+        let comm = committable::RawCommitmentBuilder::new(&Self::tag())
+            .fixed_size_field("account", &self.account.to_fixed_bytes())
+            .fixed_size_field("gas_price", &self.gas_price.to_fixed_bytes())
+            .fixed_size_field("bid_amount", &self.bid_amount.to_fixed_bytes())
+            .var_size_field("url", self.url.as_str().as_ref())
+            .u64_field("slot", self.slot.0)
+            .var_size_field(
+                "bundle",
+                // TODO what is the correct way to serialize `Vec<NamespaceId>`
```
tbro, line 134: @ggutoski do you know the answer?

ggutoski: This code uses bincode via serde. I see no problem with this. Not sure why you need `as_slice()` here though.

tbro: I don't! thanks for the feedback!

elliedavidson (changes requested):

sequencer/src/auction.rs
```diff
@@ -0,0 +1,231 @@
...
+    // Failed Tx in the future. Maybe we just want its name.
+    pub fn execute(
+        &self,
+        state: &mut ValidatedState,
+    ) -> Result<(), (ExecutionError, FullNetworkTx)> {
+        self.verify()
+            .map_err(|e| (e, FullNetworkTx::Bid(self.clone())))?;
+
+        // TODO review when this actually occurs in JIT auction
+        // charge the bid
```
tbro, line 161: This should be correct for JIT assuming that only winning bids are being sent to the sequencer, which is my reading of the diagram: https://www.notion.so/espressosys/Yet-Another-Marketplace-Spec-WIP-c09873314c8241e4b4e6463594d44bb6#7a316e557df44de7a834f82182d2b7fc

tbro: Per [discussion](https://espresso.zulipchat.com/#narrow/stream/440749-Marketplace/topic/BidTx.20Type/near/448843471), It appears that this is correct.

tbro:
Closing superseded by #1696

PR #1637, opened 2024-06-24
versioning

imabdulbasit:
This PR:
- moves types from the sequencer crate
- Moved `SeqTypes` to the `types` crate.
- implements all the hotshot traits for the `SeqTypes` 
- moves all the helper functions, types and some traits needed for hotshot trait implementations or types
- creates a versioned Header type in v0/header.rs and the serialization/deserialization

things to review:
- v0/header.rs
- Serialization and test_versioned_header_serialization() in  v0/impls/header.rs 
- check that all types and implementations are organized clearly within their respective modules.


types/src/v0/header.rs
```diff
@@ -1,20 +1,166 @@
...
+                }
+            }
+
+            fn visit_map<V>(self, mut map: V) -> Result<Header, V::Error>
+            where
+                V: MapAccess<'de>,
+            {
+                // retrieving the first entry from the map, which could either be a
+                // `chain_config` (for v0.1 headers) or a `version` (for `VersionedHeader` structs).
+                let Some((_, chain_config_or_version)) =
```
imabdulbasit, line 134: I am assuming that first field is always chain_config or version but that might not be the case.

types/src/v0/header.rs
```diff
@@ -1,20 +1,166 @@
...
+        D: Deserializer<'de>,
+    {
+        struct HeaderVisitor;
+
+        #[derive(Deserialize)]
+        #[serde(field_identifier, rename_all = "lowercase")]
+        enum StructFields {
+            Version,
+            Fields,
+        }
```
jbearer, line 89: neat

types/src/v0/header.rs
```diff
@@ -1,20 +1,166 @@
...
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("Header")
+            }
+
+            fn visit_seq<V>(self, mut seq: V) -> Result<Self::Value, V::Error>
+            where
+                V: SeqAccess<'de>,
+            {
+                let chain_config_or_version: ResolvableChainConfigOrVersion =
+                    seq.next_element()?.unwrap();
```
jbearer, line 103: Do not unwrap here!

types/src/v0/header.rs
```diff
@@ -1,20 +1,166 @@
...
+                match chain_config_or_version.chain_config {
+                    // For v0.1, the first field in the sequence of fields is the first field of the struct, so we call a function to get the rest of
+                    // the fields from the sequence and pack them into the struct.
+                    EitherOrVersion::Left(cfg) => Ok(Header::V1(
+                        v0_1::Header::deserialize_with_chain_config(cfg.into(), seq)?,
+                    )),
+                    EitherOrVersion::Right(commit) => Ok(Header::V1(
+                        v0_1::Header::deserialize_with_chain_config(commit.into(), seq)?,
+                    )),
+                    // For all versions, the first "field" is not actually part of the `Header` struct (you can think of it as the first field of the virtual
```
jbearer, line 113: all _other_ versions?

types/src/v0/header.rs
jbearer: Hmm yeah. The best way might be to grab all the entries and collect into a `HashMap`. Then we can look up specific keys like `version` or `chain_config`

types/src/v0/header.rs
```diff
@@ -1,20 +1,166 @@
...
+                    }
+                    EitherOrVersion::Version(Version { major: 0, minor: 3 }) => {
+                        Ok(Header::V3(map.next_entry::<String, _>()?.unwrap().1))
+                    }
+                    _ => Err(serde::de::Error::custom("invalid version")),
+                }
+            }
+        }
+
+        const FIELDS: &[&str] = &["version", "fields"];
```
imabdulbasit, line 159: So this would be all the fields for the v1 header? Is there  a better way to do this

imabdulbasit: actually its an enum so this is wrong

types/src/v0/header.rs
```diff
@@ -136,50 +170,54 @@ impl<'de> Deserialize<'de> for Header {
...
                     EitherOrVersion::Right(commit) => Ok(Header::V1(
-                        v0_1::Header::deserialize_with_chain_config_map(commit.into(), map)?,
+                        v0_1::Header::deserialize_with_chain_config_map::<V>(
+                            commit.into(),
+                            v1_hashmap,
+                        )?,
                     )),
                     EitherOrVersion::Version(Version { major: 0, minor: 2 }) => {
-                        Ok(Header::V2(map.next_entry::<StructFields, _>()?.unwrap().1))
+                        Ok(Header::V2(serde_json::from_value(fields.unwrap()).unwrap()))
```
jbearer, line 210: Don't unwrap. If `fields` is missing or if it doesn't deserialize into the expected type, it is a reportable error

sequencer/src/state.rs
```diff
@@ -1,413 +1,26 @@
...
-    }
-
-    Ok(())
-}
+use crate::{
+    api::data_source::CatchupDataSource, catchup::SqlStateCatchup,
+    persistence::ChainConfigPersistence, NodeState, SeqTypes,
+};
 
 async fn compute_state_update(
```
tbro, line 25: Are we keeping the `src/*` stuff around for pure utility functions like these. Just trying to understand the relationship between things like `src/state.rs` and `impls/state.rs`.

types/src/v0/impls/header.rs
```diff
@@ -132,6 +77,276 @@ impl Committable for Header {
...
+            }
+        }
+
+        deserializer.deserialize_struct("Header", StructFields::list(), HeaderVisitor)
+    }
+}
+
+impl Header {
+    #[allow(clippy::too_many_arguments)]
+    pub fn create(
```
imabdulbasit, line 262: This is used in some functions where we get a version parameter. I don't know a better way to rewrite this.

types/src/v0/error.rs
```diff
@@ -0,0 +1,104 @@
...
+use serde::{Deserialize, Serialize};
+use snafu::Snafu;
+use thiserror::Error;
+
+use super::{BlockSize, FeeAmount, NsTableValidationError};
+use crate::{BlockMerkleCommitment, FeeMerkleCommitment};
+
+#[derive(Clone, Debug, Snafu, Deserialize, Serialize)]
+pub enum Error {
+    // TODO: Can we nest these errors in a `ValidationError` to group them?
```
tbro, line 13: Not sure, where did this `Error` come from? What is it meant to represent? I don't remember seeing it before.

types/src/v0/error.rs
```diff
@@ -0,0 +1,104 @@
+use core::fmt::Debug;
```
tbro, line 1: I think having a single `error.rs` is likely to grow very large and become very confusing. Can we do `types/src/v0/error/state.rs` instead?

types/src/v0/impls/header.rs
tbro: I only see it being called from tests. If that is so, can we implement it inside the test mod? It might be a little less confusing that way.

types/src/v0/error.rs
tbro: Looks like this was previously found in `lib.rs` and this `TODO` was already present.

tbro: I'm thinking we ultimately don't need these global errors, and we should create an issue to remove this in the future, localize there errors to the modules where they are being used, and replace snafu w/ thiserror.

imabdulbasit: Yeah, we can remove this in a separate PR.

types/src/v0/impls/header.rs
imabdulbasit: It is also called in Header::from_info() and Header::genesis(). 
https://github.com/EspressoSystems/espresso-sequencer/pull/1637/files#diff-f28419d1e3fd809fea4b0c32d1a7ef5223c4fa94d097d31d35368b3906e9779eR786

types/src/v0/error.rs
imabdulbasit: I am now considering moving all these errors into their relevant modules inside impls. We can leave the global Error for now and remove it in a separate PR

types/src/v0/error.rs
tbro: new issue: https://github.com/EspressoSystems/espresso-sequencer/issues/1681

types/src/v0/impls/header.rs
tbro: ok

jbearer (commented):
Looks good, minus a few comments. I think @ggutoski should review `types/src/v0/impls/block/` and `types/src/v0/v0_1/block.rs`, mainly just to sign off on the organization of things; I don't think any logic has changed.

Also, I think we need something like `reference_tests` and `message_compat_tests` for the new versions, so that if we make any breaking changes to *any* existing supported version, a test will fail

types/src/v0/impls/header.rs
```diff
@@ -132,6 +117,276 @@ impl Committable for Header {
...
+            "payload_commitment",
+            "builder_commitment",
+            "ns_table",
+            "block_merkle_tree_root",
+            "fee_merkle_tree_root",
+            "fee_info",
+            "builder_signature",
+        ]
+    }
+}
```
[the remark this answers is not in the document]
imabdulbasit: We can remove the `StructField` but we still need hard coded list of fields which can be passed to `deserialize_struct()`

types/src/v0/impls/header.rs
```diff
@@ -132,6 +117,276 @@ impl Committable for Header {
...
+                    serde_map.insert(key.to_string(), map.next_value()?);
+                }
+
+                if let Some(v) = serde_map.get("version") {
+                    let fields = serde_map
+                        .get(&StructFields::Fields.to_string())
+                        .ok_or_else(|| de::Error::missing_field("fields"))?;
+
+                    let version =
+                        serde_json::from_value::<ResolvableChainConfigOrVersion>(v.clone())
```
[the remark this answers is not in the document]
imabdulbasit: I was getting an error when I replaced it with `Version`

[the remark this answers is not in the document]
imabdulbasit: maybe I have to add serde(untagged)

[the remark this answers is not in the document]
imabdulbasit: This breaks `visit_seq` deserialization as it can not deserialize the first field to `EitherOrVersion::Version` variant

types/src/v0/header.rs
```diff
@@ -0,0 +1,30 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) enum EitherOrVersion<ChainConfig, Commitment> {
+    Left(ChainConfig),
+    Right(Commitment),
+    Version(Version),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct VersionedHeader<Fields> {
+    pub(crate) version: ResolvableChainConfigOrVersion,
```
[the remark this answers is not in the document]
imabdulbasit: This breaks visit_seq deserialization as it can not deserialize the first field to EitherOrVersion::Version variant

types/src/v0/impls/header.rs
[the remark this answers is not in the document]
jbearer: Hmm...technically yes, but actually the list of fields we're passing in isn't even meaningful. These are the union of all fields from all versions of the struct, but we're never going to encounter a struct that has exactly these fields. Luckily, this works, because serde_json ignores the fields, while bincode only cares about the length, which is only an upper bound at that: the bincode deserializer will only consume the next field [when `next_element` is called](https://docs.rs/bincode/latest/src/bincode/de/mod.rs.html#313), which our visitor does the correct number of times. This wouldn't work if, say, the bincode deserializer eagerly consumed one element for each field.

So, in summary, I guess listing all the fields from all versions is...more correct than anything else? But we need some comments explaining exactly what's happening here at a low level for each of the formats we care about (JSON and bincode), because as a reader it's easy to get the wrong impression, and there's a nonzero chance that a future update to bincode or serde_json could break this and we'll need to rely on comments to figure out what happened

ggutoski:
> Looks good, minus a few comments. I think @ggutoski should review `types/src/v0/impls/block/` and `types/src/v0/v0_1/block.rs`, mainly just to sign off on the organization of things; I don't think any logic has changed.

Every field in every struct is now `pub(crate)`. This is a significant break of abstraction. I do not yet have a helpful suggestion on how to recover abstraction. Perhaps this break is a price worth paying. Happy to discuss further.

jbearer:
> > Looks good, minus a few comments. I think @ggutoski should review `types/src/v0/impls/block/` and `types/src/v0/v0_1/block.rs`, mainly just to sign off on the organization of things; I don't think any logic has changed.
> 
> Every field in every struct is now `pub(crate)`. This is a significant break of abstraction. I do not yet have a helpful suggestion on how to recover abstraction. Perhaps this break is a price worth paying. Happy to discuss further.

Can we not just duplicate the old directory structure under `types/src/v0/v0_1`, bring back the `pub(in ...)`, and use public methods to access these types for the trait impls?

types/src/v0/v0_1/header.rs
```diff
@@ -13,18 +15,18 @@ use serde::{
...
-    pub(crate) l1_head: u64,
-    pub(crate) l1_finalized: Option<L1BlockInfo>,
-    pub(crate) payload_commitment: VidCommitment,
-    pub(crate) builder_commitment: BuilderCommitment,
-    pub(crate) ns_table: NsTable,
-    pub(crate) block_merkle_tree_root: BlockMerkleCommitment,
-    pub(crate) fee_merkle_tree_root: FeeMerkleCommitment,
-    pub(crate) fee_info: FeeInfo,
-    pub(crate) builder_signature: Option<BuilderSignature>,
+    pub chain_config: ResolvableChainConfig,
```
jbearer, line 18: Mm why do these need to be public?

types/src/v0/impls/header.rs
```diff
@@ -132,6 +117,276 @@ impl Committable for Header {
...
+            }
+        }
+
+        deserializer.deserialize_struct("Header", StructFields::list(), HeaderVisitor)
+    }
+}
+
+impl Header {
+    #[allow(clippy::too_many_arguments)]
+    pub fn create(
```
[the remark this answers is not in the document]
imabdulbasit: We are using this create() function in reference tests inside sequencer crate.

types/src/v0/v0_1/header.rs
imabdulbasit: We are using it in reference tests. I think we should move reference tests inside the types crate

data/header_v2.json
jbearer, line 1: To keep it clean I would suggest making sub-directories like `data/v2/header.json`, `data/v1/header.json`, etc

types/src/v0/impls/header.rs
```diff
@@ -807,8 +797,10 @@ impl BlockHeader<SeqTypes> for Header {
         let block_merkle_tree_root = block_merkle_tree.commitment();
         let fee_merkle_tree_root = fee_merkle_tree.commitment();
 
+        // The Header is versioned, so we check if there are any upgrades.
+        // If no upgrades are specified, we use the default V1 Header (version 0.1).
+        // If an upgrade is specified, we use the version from the latest upgrade.
         let upgrades = instance_state.upgrades.last_key_value();
```
jbearer: Actually shouldn't the genesis header always be 0.1? Because it would be from before any upgrades had taken effect

imabdulbasit: I think we should use the current_version inside the NodeState.
so if the base version is 0.2, we build V2 Header

jbearer (approved):
😮‍💨

PR #1641, opened 2024-06-25
Refactor the TestNetwork

### LINKED ISSUE #1603: Refactor the constructors of TestNetwork
Currently we use this constructor a lot:
```rust
        pub async fn new(
            opt: Options,
            persistence: [impl PersistenceOptions<Persistence = P>; TestConfig::NUM_NODES],
            l1: Url,
            builder_port: Option<u16>,
        ) -> Self
```
`l1` and `builder_port` here are used to set a different value into the default config. This would be troublesome if we keep adding  parameters in this constructor. For example, we need a `TestNetwork` with a specific `relay-state-url`, so this constructor will be like:
```rust
        pub async fn new(
            opt: Options,
            persistence: [impl PersistenceOptions<Persistence = P>; TestConfig::NUM_NODES],
            l1: Url,
            builder_port: Option<u16>,
            relay_state_url: Option<Url>,
        ) -> Self
```
And every time we change the constructor, we need to do a minor fix on the other places where it is being used. This is also not a good idea for reviewing.

So I prefer using a constructor like this:
```rust
        pub async fn new_with_config(
            opt: Options,
            persistence: [impl PersistenceOptions<Persistence = P>; TestConfig::NUM_NODES],
            config: TestConfig,
        ) -> Self
```
We can invoke the `TestNetwork` after we set the config. It is clear and more elegant. People will easily know how the TestNetwork looks like.

Also we can use the builder pattern to construct a `TestConfig`. That would be like:
```rs
let cfg = TestConfigBuilder::new()
    .defualt()
    .l1_url("http://localhost:8545")
    .builder_port(40000)
    .build();
let network = TestNetwork::new(cfg);
```

ImJeremyHe:
Closes #1603 

### This PR:
- Refactor the `TestNetwork` constructor as the issue above mentioned
- Decrease the node number in espressso-dev-node

### This PR does not:
- tend to do any modification on the `TestNetwork` logic

### Key places to review:
The variable names in the config


sequencer/src/api.rs
```diff
@@ -379,15 +379,95 @@ pub mod test_helpers {
...
+            self.catchup = Some(catchup);
+            self
+        }
+
+        pub fn network_config(mut self, network_config: TestConfig<{ NUM_NODES }>) -> Self {
+            self.network_config = Some(network_config);
+            self
+        }
+
+        pub fn build(self) -> TestNetworkConfig<{ NUM_NODES }, P, C> {
```
jbearer, line 462: What is the point of `TestNetworkConfig`? Why can't we go directly from `TestNetworkConfigBuilder` (or call it `TestNetworkBuilder` to `TestNetwork`?)

sequencer/src/api.rs
```diff
@@ -566,13 +590,14 @@ pub mod test_helpers {
...
-            options,
-            [no_storage::Options; TestConfig::NUM_NODES],
-            l1,
-            None,
-        )
-        .await;
+        let network_config = TestConfigBuilder::default().l1_url(l1).build();
+        let config = TestNetworkConfigBuilder::<5, _, MockStateCatchup>::default()
+            .api_config(options)
+            .persistences([no_storage::Options; 5])
```
jbearer, line 596: Could this be the default so we don't have to specify it explicitly in every test?

sequencer/src/api.rs
```diff
@@ -566,13 +590,14 @@ pub mod test_helpers {
...
+            .network_config(network_config)
+            .catchups(std::array::from_fn(|_| MockStateCatchup::default()))
```
jbearer, line 598: Make this default too?

sequencer/src/api.rs
ImJeremyHe: I think the configuration of a `TestNetwork` is too complecated and it should be more readable and understandable by this. `TestNetwork` provides the accessibility to the running network and `TestNetWorkConfig` defines the network.
As for `Builder` things, actually I do think it is a bit troublesome but builder pattern is prevalent and people can expect what is going on easily. After calling `.build()` the config is fixed and this prevents people modifying the config by `set_xxxx`(I think this is a mess).

sequencer/src/api.rs
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/1641/commits/cc8033aaa057e687ebdd75c67c26df62a1050167

sequencer/src/api.rs
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/1641/commits/cc8033aaa057e687ebdd75c67c26df62a1050167

sequencer/src/api.rs
```diff
@@ -434,24 +430,54 @@ pub mod test_helpers {
         P: PersistenceOptions,
         C: StateCatchup + 'static,
     {
+        /// NOTICE: This function will reset most of the config. Please assign the number of the nodes
+        /// before assigning values.
```
jbearer, line 434: Perhaps this is better as a constructor then, i.e. `TestNetworkConfigBuilder::with_num_nodes::<3>()` rather than say `TestNetworkConfigBuilder::default().num_nodes(3)`. Then we could put all the default values here and implement default as `Self::with_num_nodes::<5>()`

sequencer/src/api.rs
```diff
@@ -434,24 +430,54 @@ pub mod test_helpers {
...
+        }
+
         pub fn states(mut self, state: [ValidatedState; NUM_NODES]) -> Self {
             self.state = state;
             self
         }
 
-        pub fn persistences(mut self, persistence: [P; NUM_NODES]) -> Self {
-            self.persistence = Some(persistence);
-            self
```
jbearer: Why did we replace this assignment with the construction of a whole new object? The new version is more verbose and will have to be modified any time the unrelated fields are changed

ImJeremyHe: Using the new version we can set the `PersistenceOptions` dynamically and it can make the construction more simple. We can `ConfigBuilder::default()` and set the default persistences and then set it with the other type. The old version requires us to assign the type and we can never change it.

jbearer: Ohh I see, clever

sequencer/src/api.rs
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/1641/commits/819528ef6c0d29606805fbe2401966e85695d54d

sequencer/src/api.rs
```diff
@@ -425,25 +425,26 @@ pub mod test_helpers {
...
-        P: PersistenceOptions,
-        C: StateCatchup + 'static,
+    impl<const NUM_NODES: usize>
+        TestNetworkConfigBuilder<{ NUM_NODES }, no_storage::Options, MockStateCatchup>
     {
-        /// NOTICE: This function will reset most of the config. Please assign the number of the nodes
-        /// before assigning values.
-        pub fn num_nodes<const N: usize>(
-            self,
+        pub fn with_num_nodes<const N: usize>(
```
jbearer, line 431: Nit: we don't need this `N` parameter anymore because we can take it from `NUM_NODES`, and then you can do `TestNetworkConfigBuilder::<WhateverNumberOfNodes>::with_num_nodes()`. Actually we could just change `with_num_nodes` to `new` at that point lol. But this is minor

jbearer (approved):

PR #1642, opened 2024-06-25
feat: Namespace table enforce final offset equals payload length

### LINKED ISSUE #1636: Namespace table enforce final offset equals payload length
The final offset in a well-formed namespace table is supposed to equal the payload byte length. Call this property P. Currently we do not enforce P. Indeed, the payload could be much smaller than indicated by the namespace table. We have a rule for how to interpret namespace table offsets that are beyond the payload byte length. But instead we should simply enforce P at consensus. Enforcing P would eliminate the need for this rule, eliminate a number of sneaky edge cases that should never arise in practice anyway, and significantly simplify tests.

Previously we enacted consensus changes to enforce validity of a namespace table in #1607 . But those changes enforce only properties that can be verified by looking only at the namespace table itself. Property P can be verified only if we also know the payload byte length, so it was not included in #1607.

ggutoski:
Closes #1636 

### This PR:

- See commit logs.
- This is a consensus change. Existing blocks already conform to this new condition, so I expect no problems with backward compatibility.

### Key places to review:

- `NsTable::validate`
- `state.rs`
- Enforcement of this new condition in `state.rs` assumes the payload byte length declared in a `VidCommon` is correct. Nowhere do we check against the actual payload byte length because no one (except DA committee nodes) ever sees the entire payload and so cannot know this byte length directly; they must trust whatever's in the `VidCommon`.

### DIFF
--- a/sequencer/src/block/full_payload/ns_table.rs
+++ b/sequencer/src/block/full_payload/ns_table.rs
@@ -411,7 +411,7 @@ impl NsTableBuilder {
 }
 
 /// Index for an entry in a ns table.
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Debug, Display, Eq, Hash, PartialEq)]
 pub struct NsIndex(usize);
 bytes_serde_impl!(NsIndex, to_bytes, [u8; NUM_NSS_BYTE_LEN], from_bytes);
 
--- a/sequencer/src/block/full_payload/ns_table.rs
+++ b/sequencer/src/block/full_payload/ns_table.rs
@@ -221,10 +221,11 @@ impl NsTable {
     /// 1. Byte length must hold a whole number of entries.
     /// 2. All namespace IDs and offsets must increase monotonically. Offsets
     ///    must be nonzero.
-    /// 3. Header consistent with byte length (obsolete after
-    ///    <https://github.com/EspressoSystems/espresso-sequencer/issues/1604>)
-    /// 4. Final offset must equal `payload_byte_len` (obsolete after
-    ///    <https://github.com/EspressoSystems/espresso-sequencer/issues/1604>)
+    /// 3. Header consistent with byte length. (Obsolete after
+    ///    <https://github.com/EspressoSystems/espresso-sequencer/issues/1604>.)
+    /// 4. Final offset must equal `payload_byte_len`. (Obsolete after
+    ///    <https://github.com/EspressoSystems/espresso-sequencer/issues/1604>.)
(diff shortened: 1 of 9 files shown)


sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -306,6 +286,57 @@ impl NsTable {
...
+    /// Helper for [`NsTable::validate`], used in our custom [`serde`]
+    /// implementation.
+    ///
+    /// Checks conditions 1-3 of [`NsTable::validate`]. Those conditions can be
+    /// checked by looking only at the contents of the [`NsTable`].
+    fn validate_deserialization_invariants(&self) -> Result<(), NsTableValidationError> {
+        use NsTableValidationError::*;
+
+        // Byte length for a table with `x` entries must be exactly
+        // `x * NsTableBuilder::entry_byte_len() + NsTableBuilder::header_byte_len()`
```
philippecamacho, line 299: What is described in the comment and the condition of the `if` statement do not seem equivalent.

sequencer/src/block/full_payload/ns_table.rs
```diff
@@ -221,46 +223,24 @@ impl NsTable {
...
-                        return Err(NonIncreasingEntries);
-                    }
-                }
-                if offset <= prev_offset {
-                    return Err(NonIncreasingEntries);
-                }
-                (prev_ns_id, prev_offset) = (Some(ns_id), offset);
+        // condition 4
+        let len = self.len().0;
+        if len > 0 {
```
jbearer, line 239: What if `len == 0`? Do we want to check that `payload_byte_len == 0` as well?

ggutoski: `len` here refers to the number of entries in the namespace table. If it's zero then there is no entry whose offset we must check against `payload_byte_len`. (We need to compute `len-1` just to check anything.)

Perhaps `len` is a confusing name. Maybe `num_nss` or something?

ggutoski: Yes that's a good idea.

sequencer/src/block/full_payload/ns_table.rs
ggutoski: I believe they are equivalent. Explainer comment added in https://github.com/EspressoSystems/espresso-sequencer/pull/1642/commits/50c2abcf2176573b0184cafd834e9b50c03fb870.

philippecamacho: Thanks!

PR #1646, opened 2024-06-26
Make catchup backoff params configurable; adjust defaults

jbearer:
Makes the parameters controlling backoff for catchup requests configurable via env var. Adjusts the defaults to retry more quickly when there are few failures and backoff more dramatically when there are many failures.


sequencer/src/catchup.rs
```diff
@@ -19,33 +21,92 @@ use tide_disco::error::ServerError;
...
+        env = "ESPRESSO_SEQUENCER_CATCHUP_BACKOFF_JITTER",
+        default_value = "1:10"
+    )]
+    jitter: Ratio,
+}
+
+impl Default for BackoffParams {
+    fn default() -> Self {
+        Self::parse_from(std::iter::empty::<String>())
+    }
```
rob-maron, line 64: Does this evaluate to the default `clap` value?

rob-maron (approved):
Looks good to me

PR #1653, opened 2024-06-28
Hotshot/rc 0.5.60

jparr721:
Closes #<ISSUE_NUMBER>


sequencer/src/lib.rs
```diff
@@ -762,6 +767,10 @@ pub mod testing {
             self.builder_port
         }
 
+        pub fn url(&self) -> Url {
+            self.url.clone()
+        }
```
jbearer, line 772: This should probably be called `l1_url` as that's what it really represents. I'm not sure why it had such a generic name as `url` to begin with, as this is a struct representing an entire network and just `url` is really vague

sequencer/src/lib.rs
```diff
@@ -589,23 +589,24 @@ pub mod testing {
...
         port: Option<u16>,
-    ) -> (Box<dyn BuilderTask<SeqTypes>>, Url) {
-        let builder_config = if let Some(port) = port {
-            SimpleBuilderConfig { port }
-        } else {
-            SimpleBuilderConfig::default()
-        };
+    ) -> Box<dyn BuilderTask<SeqTypes>> {
+        let port = port.unwrap_or_else(|| pick_unused_port().expect("No available ports"));
+        url.set_port(Some(port)).expect("Failed to set port");
```
jbearer, line 603: Unclear why we're taking a URL just to change the port, vs just `localhost:{port}`

jbearer (changes requested):

sequencer/src/lib.rs
jparr721: Good point.. honestly not sure why that was there.

sequencer/src/lib.rs
```diff
@@ -596,12 +596,8 @@ pub mod testing {
     const STAKE_TABLE_CAPACITY_FOR_TEST: u64 = 10;
 
     pub async fn run_test_builder<const NUM_NODES: usize>(
-        mut url: Url,
-        port: Option<u16>,
+        url: Url,
```
jbearer, line 599: I think I meant the opposite suggestion. We should pick an unused port (or have one passed in) and the url we use for the builder is just `localhost:{that port}`. Now it looks like we are using the same URL for the builder and L1 in most tests which isn't going to work

jbearer (approved):

PR #1639, opened 2024-06-25
fix: bug on lagrange coeff computation

alxiong:
Closes #1632 

### This PR:

Fix the edge case when `zeta` is one of the evaluation domain elements, and our lagrange coefficient (and also `pi_poly_eval`) is computed incorrectly. 
Tests are also added for these rare edge cases.

PR #1655, opened 2024-06-29
Patch lagrange coeff merge main

philippecamacho:
PR obtained by merging main into #1639.


alxiong (changes requested):
wait, I forgot to mention, we should put everything on `commprefix-patch` branch, isolate our changes, so that CommonPrefix can re-review these changes (without caring about others) in the end. 

so maybe re-target this PR to my branch `patch-lagrange-coeff` (which is targeting `commonprefix-patch`) instead?

philippecamacho:
> so maybe re-target this PR to my branch patch-lagrange-coeff (which is targeting commonprefix-patch) instead?

Done. And CI is green.

alxiong:
Ah! after I merged in, I realize one problem, we shouldn't merge `main` back into our `patch` branch. 

the primary reason we keep a `patch` branch is that it should only contains diff from what CommonPrefix has audited, so that they can re-inspect the little diff.

but now we are introducing so much irrelevant info. 

@philippecamacho do you happen to know which specific commit/PR in the main can fix the CI on my PR?

PR #1659, opened 2024-07-01
multi fees per block (versioned)

tbro:
Re-working multi-fees-per block to account for versioning (#1637).


types/src/v0/impls/state.rs
```diff
@@ -267,22 +267,54 @@ fn charge_fee(
...
-    if !proposed_header.fee_info().account.validate_fee_signature(
-        &signature,
-        fee_amount,
-        proposed_header.metadata(),
-        &proposed_header.payload_commitment(),
-    ) {
-        return Err(BuilderValidationError::InvalidBuilderSignature);
+
+    match proposed_header {
+        Header::V3(proposed_header) => {
```
tbro, line 279: Is there a better way to do this. If we were to match by value how would we inform the compiler which version of header we have?

tbro: Per discussion in Zulip it appears these we are enforcing data structure upgrades on de-serialization, so we don't need the branches for different versions here. We can just handle the latest version.

types/src/v0/impls/header.rs
```diff
@@ -592,6 +604,7 @@ impl ExplorerHeader<SeqTypes> for Header {
     type ProposerId = FeeAccount;
     type NamespaceId = NamespaceId;
 
+    // TODO what are these expected values w/ multiple Fees
```
tbro, line 607: How do we determine `proposer_id` w/ multiple Fees? Can this be resolved in block explorer by propagating the `Vec` upward?

tbro: @Ayiga for your reference.

types/src/v0/impls/state.rs
```diff
@@ -205,12 +205,17 @@ pub fn validate_proposal(
...
-        return Err(ProposalValidationError::InsufficientFee {
-            max_block_size: expected_chain_config.max_block_size,
-            base_fee: expected_chain_config.base_fee,
-            proposed_fee: proposal.fee_info().amount(),
-        });
+    // TODO here we are validating the each fee amount is at least
+    // base_fee * block_size. Seems inappropriate to use current block
+    // to calculate fees for bid on future block.
+    for fee_info in proposal.fee_info().clone() {
+        if fee_info.amount() < expected_chain_config.base_fee * block_size {
```
tbro, line 212: With multiple fees is this still the appropriate way to check if fee is sufficient?

tbro: As discussed in zulip we we need to check against total sum of all amounts.

types/src/v0/header.rs
```diff
@@ -76,8 +76,8 @@ impl Header {
                 ns_table,
                 block_merkle_tree_root,
                 fee_merkle_tree_root,
-                fee_info,
-                builder_signature,
+                fee_info: vec![fee_info],
+                builder_signature: vec![builder_signature],
```
elliedavidson, line 80: Since we now have multiple builders, would it make sense to make the builder signature part of the fee info type?

types/src/v0/impls/header.rs
```diff
@@ -358,10 +358,22 @@ impl Header {
...
 
     /// Fee paid by the block builder
-    pub fn fee_info(&self) -> FeeInfo {
-        *field!(self.fee_info)
+    pub fn fee_info(&self) -> Vec<FeeInfo> {
+        match self {
+            Self::V1(fields) => vec![fields.fee_info],
+            Self::V2(fields) => vec![fields.fee_info],
+            Self::V3(fields) => fields.fee_info.clone(),
+        }
```
elliedavidson, line 366: We need some way to distinguish bid fees from sequencing fees, likely using an enum.  I'm not sure where the best place to put this information is, but it needs to go somewhere :)

types/src/v0/impls/header.rs
elliedavidson: What is the proposer id currently?  Is it the HotShot leader id or the builder id?

types/src/v0/header.rs
tbro: There is discussion around this on zulip: https://espresso.zulipchat.com/#narrow/stream/440749-Marketplace/topic/Multiple.20fees.20per.20block/near/448326448

types/src/v0/impls/header.rs
tbro: Here they are taking the account field of `FeeInfo` (builder account) for use as proposer id. Which I think makes sense in the case of the single fee (since the since that is the account paying the fee). With multiple fees not sure how this would work.

types/src/v0/v0_1/state.rs
```diff
@@ -85,6 +85,8 @@ pub enum ProposalValidationError {
     },
     #[error("Invalid namespace table: {err}")]
     InvalidNsTable { err: NsTableValidationError },
+    #[error("Total fee amount out of range")]
```
tbro, line 88: I think this needs to be moved to `/v0_3/state.rs`.

tbro: Looks like errors don't need minor version, so they will be moved upstream.

tbro:
Superseded by #1696

PR #1670, opened 2024-07-02
prune state history on LC contract

### LINKED ISSUE #1541: escape hatch : purge elements from stateUpdates and hotShotCommitments arrays once it has over 10 days of worth of data
Arising out of this PR https://github.com/EspressoSystems/espresso-sequencer/pull/1493

alysiahuggins:
added functionality to prune the state history using a FIFO approach and added two additional tests

Closes #1541

### This PR:
- ensures that there is only ever 10 days worth of state history stored by deleting the first entries first and storing the first index to use when traversing the state history

### This PR does not:
- truncate the array length to be no greater than 10 days worth of state updates 

### Key places to review:
FIFO approach to pruning and the fact that the array continues to grow though one only ever interacts with the 10 days worth of state updates due to storing the stateHistory first index 

### How to test this PR: 
forge test --match-contract LightClient_StateUpdatesTest --ffi -vvv

### Things tested
- updating max state history allowed
- dealing with state updates greater than the max state history allowed


contracts/src/LightClient.sol
```diff
@@ -373,6 +384,33 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+            // the affected arrays
+            stateHistoryFirstIndex++;
+        } else {
+            // increment the num block states recorded, this increment stops when stateHistoryCount
+            // == maxStateHistoryAllowed
+            stateHistoryCount++;
+        }
+        //add the L1 Block to stateUpdateBlockNumbers &  HotShot commitment for the genesis state
+        stateUpdateBlockNumbers.push(blockNumber);
+        hotShotCommitments.push(HotShotCommitment(state.blockHeight, state.blockCommRoot));
```
ImJeremyHe, line 411: Since we will just push the new elements into the array, why bother deleting the first one?

ImJeremyHe (commented):
What confuses me is that why not just override the old values with the new values? Does that consume more gas?

contracts/src/LightClient.sol
alysiahuggins: so that the storage is emptied and interacting with the array will consume less gas.

The goal is to maintain a history that isn't larger than maxStateHistoryAllowed. So i delete the first element and increment stateHistoryFirstIndex, so that the history is kept within this bound. The amount if storage required is a predictable and we also get a gas refund with the `delete`.

contracts/src/LightClient.sol
```diff
@@ -121,6 +133,16 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         BN254.ScalarField blockCommRoot;
     }
 
+    /// @notice Simplified HotShot commitment struct
+    /// @param l1BlockHeight the block height of l1 when this state update was stored
+    /// @param hotshotBlockHeight The block height of the latest finalized HotShot block
```
sveitser, line 138: The params look outdated.

contracts/src/LightClient.sol
```diff
@@ -207,17 +228,16 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
 
         blocksPerEpoch = numBlockPerEpoch;
 
+        maxStateHistoryAllowed = 10 * 7166; //set the max block states recorded to be 10 days of
```
sveitser, line 231: I wonder how the 7166 was computed.

contracts/src/LightClient.sol
```diff
@@ -84,12 +87,21 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     /// @notice a flag that indicates when a permissioned provrer is needed
     bool public permissionedProverEnabled;
 
-    /// @notice an array to store the L1 Block Heights where the finalizedState was updated
-    uint256[] public stateUpdateBlockNumbers;
+    ///@notice Max number of blockStates to record
+    uint64 public maxStateHistoryAllowed;
+
+    ///@notice number of block states recorded
+    uint64 public stateHistoryCount;
```
sveitser, line 94: I'm not sure that we need this variable. I think we can check if the item sitting at the first non-zero index has an L1 block number that needs to be pruned (over 10 days worth of blocks old) and if yes, remove it.

Even if we don't do it like this I think we could take the difference of the length of stateHistoryCommitments and stateHistoryFirstIndex.

contracts/src/LightClient.sol
```diff
@@ -373,6 +386,35 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
 
+    /// @notice update the stateUpdateBlockNumbers and hotShotCommitments arrays each time a new
+    /// finalized state is added
+    /// @dev only maintains maxStateHistoryAllowed non-zero elements in these arrays
+    /// @dev using a FIFO approach to delete elements at the start of the array and storing the new
+    /// initial index in stateHistoryFirstIndex
+    function updateStateHistory(uint256 blockNumber, LightClientState memory state) internal {
+        if (maxStateHistoryAllowed == 0) revert InvalidMaxStateHistory();
+
+        if (stateHistoryCount == maxStateHistoryAllowed) {
```
sveitser, line 397: The `maxStateHistoryAllowed` name is a bit misleading right now because if the setter `setMaxStateHIstoryAllowed` is used to reduce it then size of the history that is currently larger it won't be reduced because we only ever remove an element if we also add one. Unless I'm misunderstanding something.

sveitser: I really doubt we ever want to set a smaller state history after deployment though so because all that it allows us to do is reclaim some gas. So maybe we can "fix it" by not allowing setting a lower state history?

We could also have functionality to prune multiple entries but the problem with that is that we need to then have potential issues where calls can run out of gas if the iteration is too long. 

One way to resolve this is also to allow for a few (at least 2) pruning operations during an update. This makes the gas usage bounded and still allows the history to shrink.

contracts/src/LightClient.sol
alysiahuggins: that's correct, i've updated it here https://github.com/EspressoSystems/espresso-sequencer/commit/f36a7c41893c792fe780077fd4e23aa57aa988e4

contracts/src/LightClient.sol
alysiahuggins: that's a good improvement - implemented here https://github.com/EspressoSystems/espresso-sequencer/commit/a2e165fc0ed456b36c0a1fe227d86baf61827cbd

contracts/src/LightClient.sol
alysiahuggins: your understanding is correct, it does not considering the reduction of the max state history allowed. Would you allow me to add the ability to prune +1 elements at the same time? and if so, I can make it so that the maxStateHistoryAllowed can be reduced. Otherwise, i can restrict modifications to only increasing the max state history allowed

contracts/src/LightClient.sol
alysiahuggins: That's the number of ethereum blocks per day - which is generally roughly one block per second but i got that figure from here https://ycharts.com/indicators/ethereum_blocks_per_day#:~:text=Ethereum%20Blocks%20Per%20Day%20(I%3AEBC)&text=Ethereum%20Blocks%20Per%20Day%20is,0.21%25%20from%20one%20year%20ago.

sveitser: The block time is 12s which would imply 5 * 60 * 24 = 7200 blocks but it seems there can be empty slots. I think we should use 7200 instead.

But I realize now that this needs to adjusted for the blockchain if the light client contract is deployed on an L2 (something we need to do soon). And if a blockchain has irregular block times we would have to do it differently. Sorry I need to think about this for a bit.

contracts/src/LightClient.sol
sveitser: I think any changes here are somewhat problematic because it potentially affects the result of challenges in the rollups. Increasing the history size definitely seems less problematic though.

contracts/src/LightClient.sol
alysiahuggins: Yea there was a typo in my original response - i meant to write one block per 12 seconds but yea i can use that figure instead.

contracts/src/LightClient.sol
alysiahuggins: ah true - also this light client may be deployed on L2s to support L3s - Would there be a need to shorten the `maxStateHistoryAllowed` in those instances?

alysiahuggins: Currently we hardcode the `maxStateHistoryAllowed` in the `_initializeState` method but considering that the light client would also deploy on L2s to support L3s. Should it be configurable at deployment?

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/pull/1670/commits/9cf843e997cc504a71f4d6cad4bf74688f542389

contracts/src/LightClient.sol
sveitser: L2s are also a lot cheaper so I wonder if it would be acceptable to default to not prune the history on L2s.

sveitser: Another alternative would be to use `block.timestamp` instead of block numbers. That way we could at least set a default that should keep working on L2s as long as their timestamps are correct. I'm currently not 100% clear on how trustworthy `block.timestamp` is, especially on L2s. I would hope that validators would only tolerate small deviations though.

sveitser: Based on the answer here: it sounds like the timestamp is pretty safe to use after the merge. Wdyt @alysiahuggins ?

https://ethereum.stackexchange.com/a/140818

contracts/src/LightClient.sol
```diff
@@ -207,17 +234,15 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
 
         blocksPerEpoch = numBlockPerEpoch;
 
+        maxStateHistoryDuration = 86400; // set to one day epoch by default
```
sveitser, line 237:
```suggestion
        maxStateHistoryDuration = 10 days;
```

Or is there a particular reason for using 1 day and not time units?

sveitser: Also shouldn't we make it possible to pass this to the `initialize` so we don't have to run a second transaction to modify it?

contracts/src/LightClient.sol
```diff
@@ -373,6 +392,53 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+    /// specified index to zero.
+    /// the stateHistoryFirstIndex variable is used as an offset to indicate the starting point for
+    /// reading the array,
+    /// since the length of the array is not reduced even after deletion.
+    function updateStateHistory(
+        uint256 blockNumber,
+        uint256 blockTimestamp,
+        LightClientState memory state
+    ) internal {
+        if (maxStateHistoryDuration < 86400) revert InvalidMaxStateHistory();
```
sveitser, line 415: I think it's a bit unexpected to revert here because we already have a check in `setMaxStateHistoryDuration`.

contracts/src/LightClient.sol
```diff
@@ -121,6 +131,16 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         BN254.ScalarField blockCommRoot;
     }
 
+    /// @notice Simplified HotShot commitment struct
+    /// @param l1BlockHeight the block height of l1 when this state update was stored
+    /// @param l1BlockTimestamp the block timestamp of l1 when this state update was stored
+    /// @param hotShotCommitment The HotShot commitment info of the latest finalized HotShot block
+    struct StateHistoryCommitment {
+        uint256 l1BlockHeight;
```
sveitser, line 139: @alysiahuggins Have you looked at the gas usage of using uint256 vs for example u64 here? The values should both fit safely within a u64 and I think we might save a word or two of storage if we use u64.

alysiahuggins: i mean long live ethereum lol but yes agreed, u65 is more practical , will address

contracts/src/LightClient.sol
alysiahuggins: yes agreed

contracts/src/LightClient.sol
alysiahuggins: yes, will do

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/e7800f0ac35db2b9458dffaaafacd09c6e12a6b5

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/785bf92c65397aca1b323275a5b058bae60810ca

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/ce7e74da95360a0f5abb30248549e6bf329da445

contracts/src/LightClient.sol
```diff
@@ -418,36 +498,45 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        return stateHistoryCommitments.length;
     }
 
-    /// @notice get the number of HotShot block commitments
-    function getHotShotBlockCommitmentsCount() public view returns (uint256) {
-        return hotShotCommitments.length;
+    /// @notice set Max Block States allowed
+    /// @param historySeconds The maximum duration worth of state history updates to store based on
+    /// the block timestamp
+    function setMaxStateHistoryDuration(uint32 historySeconds) public onlyOwner {
```
sveitser, line 537: As I think we previously discussed I think we should not allow lowering the value here since we currently don't have code to actually reduce the size in that case and it might have unintended consequences for rollups.

contracts/src/LightClient.sol
```diff
@@ -397,15 +471,21 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+                prevBlock = stateHistoryCommitments[i].l1BlockHeight;
             }
 
             // We don't consider the lag time for the first two updates
             if (i < 2) {
                 break;
             }
+
+            // We've reached the first recorded block
+            // i >= 2 because we normally clear out the array from a FIFO approach
```
sveitser, line 485: Outdated comment?

contracts/src/LightClient.sol
```diff
@@ -373,6 +402,51 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         }
     }
 
+    /// @notice updates the stateHistoryCommitments array each time a new
+    /// finalized state is added to the LightClient contract.
+    /// Ensures that the array contains only the most recent data and does not
+    /// exceed the maxStateHistoryDuration duration in seconds.
```
sveitser, line 408: I think we don't do this exactly since we at most remove one element.

contracts/src/LightClient.sol
```diff
@@ -418,36 +498,45 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
     }
 
-    /// @notice get the number of HotShot block commitments
-    function getHotShotBlockCommitmentsCount() public view returns (uint256) {
-        return hotShotCommitments.length;
+    /// @notice set Max Block States allowed
+    /// @param historySeconds The maximum duration worth of state history updates to store based on
+    /// the block timestamp
+    function setMaxStateHistoryDuration(uint32 historySeconds) public onlyOwner {
+        if (historySeconds < 86400) revert InvalidMaxStateHistory();
```
sveitser, line 538: Is this meant to throw if the new value is smaller than the current value instead?

contracts/test/LightClient.t.sol
```diff
@@ -17,22 +17,28 @@ import { BN254 } from "bn254/BN254.sol";
 /// @dev Common helpers for LightClient tests
 contract LightClientCommonTest is Test {
     LCMock public lc;
+    LC.LightClientState public genesis;
     uint32 public constant BLOCKS_PER_EPOCH_TEST = 3;
     uint32 public constant DELAY_THRESHOLD = 6;
-    LC.LightClientState public genesis;
+    uint32 public constant MAX_HISTORY_SECONDS = 86400;
+    uint32 oneDaySeconds = 86400;
+    uint32 initialEpoch = oneDaySeconds;
```
sveitser, line 25: Why not write `uint32 foo = 1 weeks;`?

utils/src/deployer.rs
```diff
@@ -245,7 +245,7 @@ pub async fn deploy_light_client_contract<M: Middleware + 'static>(
 pub async fn deploy_mock_light_client_contract<M: Middleware + 'static>(
     l1: Arc<M>,
     contracts: &mut Contracts,
-    constructor_args: Option<(LightClientState, u32)>,
+    constructor_args: Option<(LightClientState, u32, u32)>,
```
sveitser, line 248: I think it makes the code hard to read to have tuples where a single type is used for multiple things.

Suggest creating a struct `LightClientConstructorArgs` with named fields and implement `Default` and `From<LightClientConstructorArgs> for (LightClientState, u32, u32)`.

contracts/src/LightClient.sol
```diff
@@ -84,12 +87,19 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     /// @notice a flag that indicates when a permissioned provrer is needed
     bool public permissionedProverEnabled;
 
-    /// @notice an array to store the L1 Block Heights where the finalizedState was updated
-    uint256[] public stateUpdateBlockNumbers;
+    ///@notice Max number of seconds worth of state commitments to record based on this block
+    /// timestamp
+    uint32 public maxStateHistoryDuration;
```
sveitser, line 92: nit: I think instead of "max" we should call it "keep" (or something like this) because it's the desired amount of history we will keep. Similar for `maxHistorySeconds`, `keepHistorySeconds` IMO expresses more clearly what the variable is supposed to do.

contracts/src/LightClient.sol
alysiahuggins: ah yes, do we want to provide the ability to lower the value or do we think there's a little chance that will be necessary?

contracts/src/LightClient.sol
alysiahuggins: updated comments https://github.com/EspressoSystems/espresso-sequencer/commit/7720ab0478d28e6b5018ec3b6cd86e3e952d511e

contracts/src/LightClient.sol
alysiahuggins: That's another criterion to consider, which I responded to in the comment here (https://github.com/EspressoSystems/espresso-sequencer/pull/1670#discussion_r1716889601). I also believe there should be a minimum state history of one day—unless you think a shorter duration would be more appropriate?

contracts/test/LightClient.t.sol
alysiahuggins: Thank you for pointing that out! I wasn't aware that 1 days existed as a built-in in Solidity. That's a much cleaner approach. Also I have to refactor this test

alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/97d6842ce0695bf3745166629709bb7c2b03ca6f

contracts/src/LightClient.sol
alysiahuggins: That's fair, I renamed it to "stateHistoryRetentionPeriod" https://github.com/EspressoSystems/espresso-sequencer/commit/fac2e76df421bf0d0e584de82297736f7dbfbc89

contracts/src/LightClient.sol
alysiahuggins: we spoke offline and agreed that it's out of scope for now

contracts/src/LightClient.sol
alysiahuggins: done here https://github.com/EspressoSystems/espresso-sequencer/commit/703546faeeee4ea9614ac3057053fd3ec8852767

contracts/src/LightClient.sol
alysiahuggins: sorted here https://github.com/EspressoSystems/espresso-sequencer/commit/703546faeeee4ea9614ac3057053fd3ec8852767

utils/src/deployer.rs
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/1efd5da955490c9670980a9e8380b2936c4cc759

contracts/src/LightClient.sol
alysiahuggins: this has been addressed in a previous commit

contracts/src/LightClient.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/23a7bdc537f956583d6e57b5ce64c03623890899

contracts/src/LightClient.sol
sveitser: Nice

contracts/script/README.md
```diff
@@ -232,8 +232,8 @@ Solution: `export FOUNDRY_PROFILE=default`
 ## LightClient Contract Deployment
 
 ```bash
-forge script contracts/script/LightClient.s.sol:DeployLightClientContractScript $numBlocksPerEpoch $numInitValidators \
---sig 'run(uint32, uint32)' \
+forge script contracts/script/LightClient.s.sol:DeployLightClientContractScript $numBlocksPerEpoch $numInitValidators $maxHistorySeconds \
```
sveitser, line 235: We still use the old name `maxHistorySeconds` here.

contracts/script/LightClient.s.sol
```diff
@@ -7,7 +7,7 @@ import { LightClient as LC } from "../src/LightClient.sol";
 import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 
 contract DeployLightClientContractScript is Script {
-    function run(uint32 numBlocksPerEpoch, uint32 numInitValidators)
+    function run(uint32 numBlocksPerEpoch, uint32 numInitValidators, uint32 maxHistorySeconds)
```
sveitser, line 10:
```suggestion
    function run(uint32 numBlocksPerEpoch, uint32 numInitValidators, uint32 stateHistoryRetentionPeriod)
```
Appears elsewhere too, need to do search/replace.

sveitser (approved):
LGTM missing some renames.

contracts/script/LightClient.s.sol
alysiahuggins: thanks, sorted https://github.com/EspressoSystems/espresso-sequencer/pull/1670/commits/33663d70d64e7f79785d992b90d9dfc97280bc79

contracts/script/README.md
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/pull/1670/commits/33663d70d64e7f79785d992b90d9dfc97280bc79

alxiong (commented):
Q: instead of storing the timestamp for every update, why can't we simply store a single value `oldest_timestamp` which gets updated whenever we evict an element from the FIFO queue?

PR #1673, opened 2024-07-03
Add serialization compatibility test for transaction index

jbearer:


ggutoski (approved):

data/tx_index.json
```diff
@@ -0,0 +1,4 @@
+{
+	"ns_index": [0, 0, 0, 0],
+	"tx_index": [0, 0, 0, 0]
```
ggutoski, line 3: I don't suppose we want a nonzero reference value to catch things like endianness, etc.

jbearer: Yes that's a good idea

jbearer: Nice looks good

PR #1674, opened 2024-07-03
Make backtrace logging configurable: pretty-printed or logged

jbearer:
Adds an option to log panics via the tracing formatter, so they can be parsed and interpeted by Datadog. If `RUST_LOG_FORMAT=json` is _not_ set, panics will be pretty-printed directly to standard out, as they have been until now.

### This PR:
* Makes logging configurable via env var/CLI
* Chooses which panic handler to install based on `RUST_LOG_FORMAT`

### Key places to review:
* `utils/src/logging.rs` has the logging setup code and config parameters
* deleted utility program `count-transactions` as the query service now has this functionality
* all other changes are just replacing
  ```rust
  setup_logging();
  setup_backtrace();
  ```
  with either `logging::Config::init` (for main binaries, to get the config from the command line) or `setup_test`

### Things tested
* demo-native works
* manually forcing a panic gives correct output based on the env
* `git grep setup_logging` and `git grep setup_backtrace` have no results outside the logging utils


Ancient123:
I think it would be cleaner to just use the value in `RUST_LOG_FORMAT` which we have to set to `json` in order to enable json logging output right now anyways.

jbearer:
Done in [2982c72](https://github.com/EspressoSystems/espresso-sequencer/pull/1674/commits/2982c729be0a08326c6e4474c0c2e9142fa54a8f)

imabdulbasit (approved):
LGTM

PR #1675, opened 2024-07-03
test:Gg/1538

ggutoski:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/sequencer/src/block/namespace_payload/ns_payload/test.rs
+++ b/sequencer/src/block/namespace_payload/ns_payload/test.rs
@@ -13,7 +13,7 @@ fn ns_payload_len() {
 
     // ordinary valid ns_payload
     {
-        let ns_payload = ns_payload_with_body(&[10, 20, 30], 30);
+        let ns_payload = NsPayloadOwned::entries_body(&[10, 20, 30], 30);
         let txs = ns_payload.export_all_txs(&ns_id);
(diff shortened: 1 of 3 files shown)


sequencer/src/block/namespace_payload/ns_payload/test.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+#[test]
+fn tx_table_header() {
+    setup_logging();
+    setup_backtrace();
+    let ns_id = NamespaceId::from(69); // dummy
+
+    // header declares 1 fewer txs, tx table bytes appear in tx payloads, wasted
+    // pßayload bytes
```
philippecamacho, line 176:
```suggestion
    // payload bytes
```

philippecamacho:
Would it be worth writing tests mixing abnormal payload size with negative length transactions?

philippecamacho (approved):
I left a question, yet LGTM.

ggutoski:
> Would it be worth writing tests mixing abnormal payload size with negative length transactions?

Absolutely. This PR is just a stepping-stone.

sequencer/src/block/namespace_payload/ns_payload/test.rs
ggutoski: I'll fix this in a future PR so as not to trigger CI

PR #1705, opened 2024-07-12
modify upgrade scripts for non-multisig, non-openzeppelin-defender deployments

### LINKED ISSUE #1704: modify upgrade scripts for non-multisig, non-openzeppelin-defender deployments
The existing script `UpgradeLightClient.s.sol` was created only for testing purposes for the demo LightClientV2. However, since the current LightClient contract has been deployed without OpenZeppelin Defender & without a multisig wallet, we need to maintain upgrade scripts for the current light client deployment.

alysiahuggins:
Closes #1704 

### This PR:
The existing script UpgradeLightClient.s.sol was created only for testing purposes for the demo LightClientV2. However, since the current LightClient contract has been deployed without OpenZeppelin Defender & without a multisig wallet, we need to maintain upgrade scripts for the current light client deployment.

### Key places to review:
- the updated script and tests

### How to test this PR: 
`just sol-test`
To run tests specific to upgrades run
`forge test --match-contract LightClientUpgradeSameContractTest --ffi -vvv`
and
`forge test --match-contract LightClientUpgradeToV2Test --ffi -vvv`

### Things tested
- Deploying the LC with this script `DeployLightClient.s.sol` script
- Upgrading the LC with the `UpgradeSameLightClient.s.sol` script
- An upgrade to the current Light Client that's being updated by provers happened live with @Ancient123


contracts/test/LightClientV2.sol
```diff
@@ -56,6 +56,22 @@ contract LightClientV2 is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+    /// contract is
+    /// in permissioned prover mode. This address is address(0) when the contract is not in the
+    /// permissioned prover mode
+    address public permissionedProver;
+
+    /// @notice a flag that indicates when a permissioned provrer is needed
+    bool public permissionedProverEnabled;
+
+    /// @notice an array to store the L1 Block Heights where the finalizedState was updated
+    uint256[] public stateUpdateBlockNumbers;
```
zacshowa, line 69: Would it be better to have this as `uint128[]`? I'm not 100% sure how solidity lays out memory under the hood but we could potentially half the size of every element in this array if we are okay with the max value of an entry being (2^128) -1. 
However, I also recognize this is a test file and gas costs might not be a factor here.

zacshowa (approved):
Code looks good to me with a first glance. I have one clarifying question but nothing that I think should block merging as the tests are passing locally for me.

contracts/test/LightClientV2.sol
alysiahuggins: actually looking at this again, since we're unsure about the future block sizes and this stores l1 blocks, sticking with `uint256[]` might be best.

PR #1707, opened 2024-07-15
Add anvil and postgres to dev-node image

### LINKED ISSUE #1616: Integrate postgres into dev-node binary
Currently, we are required to run a standalone PostgreSQL instance when using our dev-node binary. To be more developer-friendly, we can provide a dev-node binary integrated with PostgreSQL.

If the relevant environment variables or options are not set, we can check if PostgreSQL has been installed locally:

- If yes, run PostgreSQL via the command line. This is the most common scenario because people tend to run the binary in a Docker container, and we can also build a Docker image for it.
- If no, display an error message indicating that PostgreSQL configuration should be set.

ImJeremyHe:
Closes #1616 

### This PR:
- Download the anvil binary in the `espresso-dev-node.Dockerfile`. It is more convenient based on that we hardly need a pinned `anvil`.
- Use the existing script to run the `postgres`
- Fix the `build-docker-images-native.sh`

### How to test this PR:
Run this command locally
```cmd
./scritps/build-docker-images-native
```
And in this action https://github.com/EspressoSystems/espresso-sequencer/actions/runs/9935660410,
this docker image can be run without setting L1 url and postgres options

### DIFF
--- a/docker/espresso-dev-node.Dockerfile
+++ b/docker/espresso-dev-node.Dockerfile
@@ -14,6 +14,12 @@ RUN curl -LO https://github.com/EspressoSystems/ark-srs/releases/download/v0.2.0
 COPY target/$TARGETARCH/release/espresso-dev-node /bin/espresso-dev-node
 RUN chmod +x /bin/espresso-dev-node
 
+# Download the anvil binary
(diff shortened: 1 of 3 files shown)


sveitser:
@ImJeremyHe looks good to me. Have you tested the script to build the docker image locally? It would be useful to provide this information in the PR description, as the template suggests.

sveitser (approved):

PR #1709, opened 2024-07-15
Add the espresso-dev-node server

### LINKED ISSUE #1679: Run a specific server in dev-node to provide the debug information
A server is needed when debugging with the dev node because developers may wonder about the following things:
- When we run the dev node without assigning ports to `Builder` or `L1`, unused ports will be picked. Knowing the ports or URLs helps developers easily figure things out.
- The contract addresses

We can also use this server to help developers interact with the mock contract, such as `freeze L1 height` and `unfreeze`.

ImJeremyHe:
Closes #1679 

### This PR:
- Expose the prover port
- Run a server for
    - returning debug information to users
    - providing endpoints for users setting the hotshot up/down

This would be helpful in these cases:
- Sometimes users run the `dev-node` without specifying the ports or urls. `/api/dev-info` shows the urls or ports that are being used. That will be helpful when they are going to debug or report the bugs to us
- Easily setting the hotshot is down/up without constructing transactions.

### Key places to review:
`espresso-dev-node.rs` and the toml file.

### DIFF
--- a/sequencer/api/espresso_dev_node.toml
+++ b/sequencer/api/espresso_dev_node.toml
@@ -1,0 +1,29 @@
+[meta]
+NAME = "espresso_dev_node_server"
+DESCRIPTION = "A server for debugging and developing with the espresso dev node"
+FORMAT_VERSION = "0.1.0"
+
+[route.devinfo]
+PATH = ["/dev-info"]
+DOC = """
+Get the debug info
+"""
+
+[route.freeze]
+PATH = ["mock-contract/freeze/:height"]
+":height" = "Integer"
+DOC = """
+Freeze the L1 height in the light client contract.
+
+By doing this, the L1 height in the light contract will be frozen and rollups will detect
+the HotShot failure. This is intended to be used when testing the rollups' functionalities.
+"""
+
+[route.unfreeze]
+PATH = ["mock-contract/unfreeze"]
+DOC = """
+Unfreeze the L1 height in the light client contract.
+
+This is intended to be used when `freeze` has been called previously. By unfreezing the L1 height,
+rollups will detect the reactivity of HotShot.
+"""
--- a/sequencer/api/espresso_dev_node.toml
+++ b/sequencer/api/espresso_dev_node.toml
@@ -7,22 +7,39 @@ FORMAT_VERSION = "0.1.0"
 PATH = ["/dev-info"]
 DOC = """
 Get the debug info
+
+Returns
+```
+{
+    "builder_url": string,
+    "prover_port": integer,
+    "l1_url": string,
+    "light_client_address:" address,
+}
+```
 """
 
-[route.freeze]
-PATH = ["mock-contract/freeze/:height"]
-":height" = "Integer"
+[route.sethotshotdown]
+PATH = ["set-hotshot-down"]
+METHOD = "POST"
 DOC = """
-Freeze the L1 height in the light client contract.
+Set the hotshot down since the given L1 height.
 
+Body:
+```
+{
+    "height": integer,
+}
+```
 By doing this, the L1 height in the light contract will be frozen and rollups will detect
 the HotShot failure. This is intended to be used when testing the rollups' functionalities.
 """
 
-[route.unfreeze]
-PATH = ["mock-contract/unfreeze"]
+[route.sethotshotup]
+PATH = ["set-hotshot-up"]
+METHOD = "POST"
 DOC = """
-Unfreeze the L1 height in the light client contract.
+Set the hotshot up in the light client contract.
 
 This is intended to be used when `freeze` has been called previously. By unfreezing the L1 height,
 rollups will detect the reactivity of HotShot.
--- a/sequencer/src/bin/espresso-dev-node.rs
+++ b/sequencer/src/bin/espresso-dev-node.rs
@@ -177,7 +177,7 @@ async fn main() -> anyhow::Result<()> {
     let retry_interval = Duration::from_secs(2);
     let prover_port = cli_params
         .prover_port
-        .unwrap_or(pick_unused_port().unwrap());
+        .unwrap_or_else(|| pick_unused_port().unwrap());
     let light_client_address = contracts
         .get_contract_address(Contract::LightClientProxy)
         .unwrap();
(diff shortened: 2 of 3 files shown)


sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -148,52 +162,130 @@ async fn main() -> anyhow::Result<()> {
...
+        .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
+
+    api.get("devinfo", move |_, _| {
+        let info = dev_info.clone();
+        async move { Ok(info.clone()) }.boxed()
+    })
+    .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
+
+    let contract = mock_contract.clone();
+    api.get("freeze", move |req, _| {
```
sveitser, line 244: I think these should not be GET but POST. GET isn't supposed to have side effects.

sequencer/api/espresso_dev_node.toml
```diff
@@ -0,0 +1,29 @@
...
+DESCRIPTION = "A server for debugging and developing with the espresso dev node"
+FORMAT_VERSION = "0.1.0"
+
+[route.devinfo]
+PATH = ["/dev-info"]
+DOC = """
+Get the debug info
+"""
+
+[route.freeze]
```
sveitser, line 12: Do we need the extra indirection by introducing new terminology "freeze/unfreeze" or should we just call it set-hotshot-down/up?

ImJeremyHe: Yeah makes sense. Will use `set-hotshot-down/up`

sequencer/api/espresso_dev_node.toml
```diff
@@ -0,0 +1,29 @@
+[meta]
+NAME = "espresso_dev_node_server"
+DESCRIPTION = "A server for debugging and developing with the espresso dev node"
+FORMAT_VERSION = "0.1.0"
+
+[route.devinfo]
+PATH = ["/dev-info"]
+DOC = """
+Get the debug info
```
imabdulbasit, line 9: We should specify what this endpoint returns

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -148,52 +162,130 @@ async fn main() -> anyhow::Result<()> {
...
+        let contract = contract.clone();
+        async move {
+            let height: u64 = req
+                .integer_param("height")
+                .map_err(ServerError::from_request_error)?;
+            contract
+                .set_hot_shot_down_since(U256::from(height))
+                .send()
+                .await
+                .map_err(|err| ServerError::catch_all(StatusCode::FORBIDDEN, err.to_string()))?;
```
imabdulbasit, line 254: Why is the status code `FORBIDDEN` here?

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -395,6 +492,31 @@ mod tests {
...
+                .get::<DevInfo>("api/dev-info")
+                .send()
+                .await
+                .unwrap();
+
+            dev_node_client
+                .get::<()>("api/mock-contract/freeze/3")
+                .send()
+                .await
+                .unwrap();
```
imabdulbasit, line 511: Can we check if the L1 height did freeze or unfreeze?

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -148,52 +162,130 @@ async fn main() -> anyhow::Result<()> {
...
+        .expect("error building wallet")
+        .build()
+        .expect("error opening wallet")
+        .with_chain_id(chain_id);
+
     let update_interval = Duration::from_secs(20);
     let retry_interval = Duration::from_secs(2);
+    let prover_port = cli_params
+        .prover_port
+        .unwrap_or(pick_unused_port().unwrap());
```
imabdulbasit, line 180:
```suggestion
        .unwrap_or_else(|| pick_unused_port().unwrap());
```

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: Maybe `BAD_REQUEST` is better?

imabdulbasit: maybe `INTERNAL_SERVER_ERROR` ?

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: Any reason why we prefer `.unwrap_or_else`?

imabdulbasit: `unwrap_or()` evaluates even if the Option is Some. While, `unwrap_or_else()` only evaluates when the Option is None. unwrap_or() performs eager evaluation, while unwrap_or_else() does lazy evaluation.
So, in this case unwrap_or() would pick a port even if `Option` is `Some`

ImJeremyHe: Thanks. I didn't know that

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

sequencer/api/espresso_dev_node.toml
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

sequencer/api/espresso_dev_node.toml
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: e92c71737d74e7e3db24e017de9bf791f5bfaaac

imabdulbasit (approved):

PR #1711, opened 2024-07-15
Allow nodes to fetch network config from peers

jbearer:
This allows nodes to fetch the network config from the `config` API of another node, instead of/in addition to local storage and the orchestrator. This could be useful when a node in an existing network is migrating to new storage, or when storage has been lost. In particular, Kudasai is currently not running their Cappuccino nodes because (for reasons still unknown) they lost their persisted network config, and now have no way of getting it back since the orchestrator is already shut down.

### This PR:
* Adds a `PublicNetworkConfig` which wraps a `PublicHotShotConfig`
* Has the `config/hotshot` endpoint return the network config instead of just the HotShot config, which is strictly more information
* Adds functions to convert a `PublicNetworkConfig` to a proper `NetworkConfig` based on a given set of private keys
* Adds configuration logic to fetch the network config from a peer on startup

### This PR does not:
* Fix any issues with the network config being lost from persistent storage

### Key places to review:
* New config serialization in `api/data_source.rs`
* Configuration logic in `lib.rs`

### DIFF
--- a/sequencer/src/options.rs
+++ b/sequencer/src/options.rs
@@ -188,6 +188,16 @@ pub struct Options {
     #[derivative(Debug(format_with = "fmt_urls"))]
     pub state_peers: Vec<Url>,
 
+    /// Peer nodes use to fetch missing config
+    ///
+    /// Typically, the network-wide config is fetched from the orchestrator on startup and then
+    /// persisted and loaded from local storage each time the node restarts. However, if the
+    /// persisted config is missing when the node restarts (for example, the node is being migrated
+    /// to new persistent storage), it can instead be fetched directly from a peer.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_CONFIG_PEERS", value_delimiter = ',')]
+    #[derivative(Debug(format_with = "fmt_opt_urls"))]
+    pub config_peers: Option<Vec<Url>>,
+
     /// Exponential backoff for fetching missing state from peers.
     #[clap(flatten)]
     pub catchup_backoff: BackoffParams,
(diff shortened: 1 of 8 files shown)


sequencer/src/options.rs
```diff
@@ -188,6 +188,16 @@ pub struct Options {
...
 
+    /// Peer nodes use to fetch missing config
+    ///
+    /// Typically, the network-wide config is fetched from the orchestrator on startup and then
+    /// persisted and loaded from local storage each time the node restarts. However, if the
+    /// persisted config is missing when the node restarts (for example, the node is being migrated
+    /// to new persistent storage), it can instead be fetched directly from a peer.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_CONFIG_PEERS", value_delimiter = ',')]
+    #[derivative(Debug(format_with = "fmt_opt_urls"))]
+    pub config_peers: Option<Vec<Url>>,
```
rob-maron, line 199: Is there any downside to changing this to `Vec<Url>` and then treating an empty vector as nothing? It may remove some complexity

rob-maron (approved):

sequencer/src/options.rs
jbearer: I personally prefer `Option<Vec>` when empty is qualitatively different from non-empty because the compiler forces you to handle the `None` case explicitly whenever you want to use the vector. Whereas with just `Vec<Url>`, it's pretty easy to forget to check `is_empty` which is bad when there's entirely separate logic you want to do in the empty case

PR #1712, opened 2024-07-15
update upgrades.md

imabdulbasit:
- This PR updates the doc/upgrades.md
- clarifies the difference between migration and version bump
- adds additional parameters for time-based upgrade

### DIFF
--- a/doc/upgrades.md
+++ b/doc/upgrades.md
@@ -1,23 +1,40 @@
 # Upgrades
 
-Hotshot protocol supports upgrades through an Upgrade proposal mechanism. The Upgrade proposal is broadcast separately from the `QuorumProposal`, typically several views in advance of its attachment. The goal is to ensure ample time for nodes to receive and prepare for the upgrade process.
+Hotshot protocol supports upgrades through an Upgrade proposal mechanism. The Upgrade proposal is broadcast separately
+from the `QuorumProposal`, typically several views in advance of its attachment. The goal is to ensure ample time for
+nodes to receive and prepare for the upgrade process.
 
-After enough votes have been collected on the `UpgradeProposal`, an `UpgradeCertificate` is formed. This is attached to the next `QuorumProposal`, and any node that receives an `UpgradeCertificate` in this way re-attaches it to its own `QuorumProposal` until the network has upgraded, or (in rare cases) we failed to reach consensus on the `UpgradeCertificate`.
+After enough votes have been collected on the `UpgradeProposal`, an `UpgradeCertificate` is formed. This is attached to
+the next `QuorumProposal`, and any node that receives an `UpgradeCertificate` in this way re-attaches it to its own
+`QuorumProposal` until the network has upgraded, or (in rare cases) we failed to reach consensus on the
+`UpgradeCertificate`.
 
 An upgrade consists of two parts:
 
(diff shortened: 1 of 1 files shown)


tbro:
Just pushing a change seemed like the easiest way to incorporate my thoughts for improvements. Feel free to revert if you disagree.

imabdulbasit:
> Just pushing a change seemed like the easiest way to incorporate my thoughts for improvements. Feel free to revert if you disagree.

Looks good, thanks!

doc/upgrades.md
```diff
@@ -16,27 +20,39 @@ To enable an upgrade in Hotshot protocol, it is essential to define the base ver
...
+
+View based:
+- **start_voting_view:**  view at which voting for the upgrade proposal starts.
 - **stop_voting_view:**  view at which voting for the upgrade proposal stops. To disable an upgrade, set this parameter to 0 or ensure `stop_voting_view` is less than `start_voting_view`.
 - **start_proposing_view:** the earliest view in which the node can propose an upgrade. This should be set to when an upgrade is intended. If the current view > `start_proposing_view`, the node will send out an `UpgradeProposal`.
 - **stop_proposing_view:** view after which the node stops proposing an upgrade. If the upgrade proposal fails and the current view > stop_proposing_view then the upgrade is never proposed again.
 
-The window between `start_proposing_view` and `stop_proposing_view` should provide sufficient time for nodes to continue proposing the upgrade until successful.
+Time based:
+- **start_voting_time:**  UNIX timestamp at which voting for the upgrade proposal starts.
```
sveitser, line 46: What's the use case for having distinct `start_voting_time` and `start_proposing_time`? Like why can't they be the same?

sveitser (approved):
LGTM, one question, not necessarily related to this PR.

doc/upgrades.md
```diff
@@ -16,27 +20,39 @@ To enable an upgrade in Hotshot protocol, it is essential to define the base ver
...
 }
 ```
 
-These parameters are fetched from the genesis TOML file and set in Hotshot config:
+Hotshot provides two modes for upgrades: view-based and time-based.
+
+View-based parameters allow nodes to vote and propose an upgrade based on the current node's view, while time-based parameters use Unix timestamps for the same purpose.
 
-- **start_voting_view:**  view at which voting for the upgrade proposal starts. In our implementation, this is set to 1 so that voting is enabled as soon as the node is started.
+To simplify configuration, these parameters are fetched from the genesis TOML file and set in the Hotshot config. The TOML file can include either view-based parameters or time-based parameters, but not both.
```
sveitser, line 36: Can the toml file contain view-based parameters for certain upgrades and some time-based parameters for other upgrades? Or do all the upgrades have to use use the same types of parameters?

doc/upgrades.md
imabdulbasit: I don't think this serves a very useful purpose, but it has been provided as an additional setting. This is why we are keeping the voting parameters optional in the TOML file and setting them to 0/u64::MAX by default.

doc/upgrades.md
imabdulbasit: Yes, toml file has array of tables for upgrade so we define version, view/time parameters and the upgrade type separately for each upgrade. This can include different mode view or time based for upgrade

imabdulbasit: I am going to add this in test once we have more upgrades

imabdulbasit: One thing to note is that we don't support multiple upgrades, so view- or time-based settings need to be defined at the start in the hotshot config. Therefore, we just check what upgrade is planned and set those parameters, whether they are view-based or time-based.

sveitser: Ah yes remember that you mentioned we can only do one upgrade at a time. Thanks.

doc/upgrades.md
sveitser: Thanks. I think if it's not useful it would be better not to have it, but at least it's documented now. :D

PR #1718, opened 2024-07-16
making the plonk verifier library external to the light client contract

### LINKED ISSUE #1533: Make PlonkVerifier an external library again
See discussion on zulip: https://espresso.zulipchat.com/#narrow/stream/419254-Contracts/topic/Light.20client.20contract.20size

in https://github.com/EspressoSystems/espresso-sequencer/issues/999#issuecomment-1997942281, we changed the visibility of `verify()` function in `library PlonkVerifier` from `external` to `internal` to placate OZ's defender script. 

Currently just changing the function access from `internal` to `external` results in a compilation error. It does however work with `forge build --via-ir` at the expense of rather lengthy compilation times. So we need to decide how to handle this.

We also need to verify that deploying with defender still works after this change.

See https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/issues/26

alysiahuggins:
Closes #1533 

### This PR:
In https://github.com/EspressoSystems/espresso-sequencer/issues/999#issuecomment-1997942281, we changed the visibility of verify() function in library PlonkVerifier from external to internal to placate OZ's defender script. However that includes the PlonkVerifier bytecode in the LightClient contract which makes the file large. Instead, we have changed the visibility to external again and followed the following steps to be able to deploy the upgradeable light client contract via open zeppelin
- deployed the PlonkVerifier contract via OpenZeppelin & a Safe Multisig wallet
- referenced the address of the deployed PlonkVerifier in the forge script deployment command when deploying the Light Client contract via OpenZeppelin & a Safe Multisig wallet
- removed openzeppelin related secrets into .env.contracts instead so that it can be gitignored since the existing .env file contains configuration variables that are not that secret

### This PR does not:
- even though the PlonkVerifier contract is deployed via OpenZeppelin & a Safe Multisig wallet, it's not upgradeable and thus each time modifications are made to the PlonkVerifier contract, the Light Client Contract has to be updated with the new address of the deployed PlonkVerifier mentioned at deploy time
- automate the inclusion of the  most recent address of the deployed PlonkVerifier in the `forge script` command for deploying or upgrading the LightClient contract
- remove the mnemonic from the .env file and put it into the .env.contracts file since there are other processes that use it so I need to create a separate issue and communicate with other team members so that they are aware of the change. 

### Key places to review:
- PlonkVerifier contract function visibility
- updated README in contracts/script
- new .env.contracts.example file

### How to test this PR:
- to thoroughly test this PR, one would have to follow the instructions in `contracts/src/README` to set up the environment to deploy both the PlonkVerifier contract and the LightClient contract

### Things tested 
- deploying the PlonkVerifier and LightClient contract via `forge script` which involves OpenZeppelin Defender & Multisig Safe Wallet

### Things not tested
- modifying PlonkVerifier methods and upgrading the LightClient contract with the new PlonkVerifier


.env
```diff
@@ -121,10 +121,10 @@ ESPRESSO_SEQUENCER_FETCH_RATE_LIMIT=25
 ESPRESSO_NASTY_CLIENT_PORT=24011
 
 # Openzeppelin Defender Deployment Profile
-DEFENDER_KEY=
-DEFENDER_SECRET=
+DEFENDER_KEY=3vKMu7Dwajrg8GMV3B4qVZpspdocj1Pt
+DEFENDER_SECRET=4TJWVRbw6kWGqR3bTrAQQTBcBQiNduvZvEqUv9o7fAChWZW5pcgqNkkj1RgKjg1b
```
Sneh1999, line 125: Why are we adding the `DEFENDER_SECRET` and `DEFENDER_KEY` to `.env` ?

.env
```diff
@@ -137,3 +137,6 @@ SAFE_ORCHESTRATOR_PRIVATE_KEY=
 # Light Client
 LIGHT_CLIENT_PROXY_CONTRACT_ADDRESS=
 APPROVED_PROVER_ADDRESS=
+
+# Plonk Verification Library Deployment with Defender
+PLONK_VERIFIER_SALT=3
```
Sneh1999, line 142: Just a suggestion - but I think we should not push `.env` to main as its generally not advised and rather we should have a `.env.example` which can be converted to `.env` locally when needed. The reason for this is that someone can by accident add a private key to `.env` file and push it and bots will extract all the assets from that key - this has happened with a lot of people in the industry. Let me know what your thoughts are on this!

contracts/script/Utils.s.sol
```diff
@@ -62,6 +62,36 @@ contract UtilsScript is Script {
...
+        string memory contractName,
+        uint256 contractSalt,
+        address contractAddress,
+        address multisig,
+        string memory approvalProcessId,
+        string memory viaType
+    ) external returns (string memory filePath, string memory data) {
+        string memory outputDir = string.concat(
+            vm.projectRoot(),
+            "/contracts/script/output/defenderDeployments/",
```
Sneh1999, line 75: Should we allow this to be a `.env` variables instead of being hard coded here to make it more configurable?

.env
alxiong: yes, agree. 
we probably version control `.env` somehow by accident in the past.

contracts/script/Utils.s.sol
alysiahuggins: Thanks Sneh. Currently, There's an open issue to deal with defender deployment logs https://github.com/EspressoSystems/espresso-sequencer/issues/1286 but currently when we deploy with openzeppelin defender, the deployment logs aren't written locally (whereas foundry normally does this). So this is a workaround and is not expected to be used more dynamically. I've put that string into a variable since it has been referenced multiple times in the script. https://github.com/EspressoSystems/espresso-sequencer/pull/1718/commits/c80c26ed1147291e29fd86f3ae2818d1eab1304b

.env
alysiahuggins: Thanks for calling this out so we can revise this. Historically, the `.env` file wasn't used for secrets but for config. The openzeppelin related config/secrets have been removed and placed into a gitignored env file called `.env.contracts` with a `.env.contracts.example` file provided as the example. This also resolves the comment above https://github.com/EspressoSystems/espresso-sequencer/pull/1718/commits/0e7ab30d5eb555e36755b9d724c60f76071ddfcd

Sneh1999 (approved):
LGTM!

Issue #1682, opened 2024-07-08
Add documentation for dev-node


sveitser:
- Add a markdown file in `./doc` folder.
- Link to markdown file from gitbook / or publish html documentation in doc folder via mdbook.

sveitser:
Done in #1722

PR #1722, opened 2024-07-17
Add a document for espresso-dev-node

ImJeremyHe:
#1682 

### This PR:
- Add a markdown file for espresso-dev-node
- Ensure the container exits on Ctrl+C

### This PR does not:
Publish this document

### Key places to review:
- To see if any other thing we need to mention in this document
- The changes on the script

### How to test this PR:
Run the container and then Ctrl+C
```
docker run ghcr.io/espressosystems/espresso-sequencer/espresso-dev-node:jh-dev-node-docmusl
```

It will exit successfully.


doc/espresso-dev-node.md
```diff
@@ -0,0 +1,69 @@
...
+  "builder_url": "http://localhost:41003/",
+  "prover_port": 23156,
+  "l1_url": "http://localhost:8545/",
+  "light_client_address": "0xb075b82c7a23e0994df4793422a1f03dbcf9136f"
+}
+```
+
+### POST /api/set-hotshot-down
+
+This endpoint sets the hotshot down at the given L1 height.
```
sveitser, line 53:
```suggestion
This endpoint simulates the effect of a liveness failure of the hotshot consensus protocol in the Light Client smart contract.  
```

Let's try to be more precise.

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,69 @@
...
+
+Parameters
+
+| Name   | Type    | Description                              |
+| ------ | ------- | ---------------------------------------- |
+| height | integer | The L1 height from which hotshot is down |
+
+### POST /api/set-hotshot-up
+
+This endpoint sets the hotshot up in the light client contract.
```
sveitser, line 66: Similarly, let's be more precise and avoid using the terms "hotshut up/down" in the documentation.

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,69 @@
...
+
+Espresso dev node is a node specifically designed for development and testing. It includes various nodes required to run
+a complete Espresso network, such as `builder`, `sequencer`, etc. Developers can use it for development and testing.
+
+## Download
+
+We highly recommend you to use our Docker image. You can install it from the command line:
+
+```cmd
+docker pull ghcr.io/espressosystems/espresso-sequencer/espresso-dev-node:main
```
sveitser, line 11: Suggest putting the `docker run` command instead of the `pull` command.

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## Parameters
+
+| Name                            | Type            | Environment Variable                 | Description                                                                                                                                                             |
```
sveitser, line 16: We should also specify what the default behavior is if arguments are not provided.

scripts/launch-dev-node-with-postgres
```diff
@@ -11,8 +11,22 @@ export POSTGRES_PASSWORD=$ESPRESSO_SEQUENCER_POSTGRES_PASSWORD
 
 export RUST_LOG=${RUST_LOG:-info}
 
+# Function to handle SIGTERM and SIGINT
+cleanup() {
+    echo "Received SIGTERM or SIGINT. Exiting..."
+    kill -TERM "$postgres_pid" 2>/dev/null
+    kill -TERM "$dev_node_pid" 2>/dev/null
+    wait "$postgres_pid"
```
sveitser, line 19: We can use `jobs -p` to get a list of all background processes, then we don't need to track PIDs explicitly.

There are also some oneliners in this SO answer: https://stackoverflow.com/a/2173421

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,71 @@
...
+
+```cmd
+docker run ghcr.io/espressosystems/espresso-sequencer/espresso-dev-node:main
+```
+
+## Parameters
+
+| Name                            | Type            | Environment Variable                 | Default Value                                                 | Description                                                                                                                                  |
+| ------------------------------- | --------------- | ------------------------------------ | ------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| `rpc_url`                       | `Option<Url>`   | `ESPRESSO_SEQUENCER_L1_PROVIDER`     | Automatically launched Avil node if not provided.             | A JSON-RPC endpoint for the L1 to deploy to. If not provided, an Avil node will be launched automatically.                                   |
```
tbro, line 18:
```suggestion
| `rpc_url`                       | `Option<Url>`   | `ESPRESSO_SEQUENCER_L1_PROVIDER`     | Automatically launched Avil node if not provided.             | The JSON-RPC endpoint of the L1. If not provided, an Avil node will be launched automatically.                                   |
```

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,71 @@
...
+| `mnemonic`                      | `String`        | `ESPRESSO_SEQUENCER_ETH_MNEMONIC`    | `test test test test test test test test test test test junk` | Mnemonic for an L1 wallet. This wallet is used to deploy the contracts, so the account indicated by `ACCOUNT_INDEX` must be funded with ETH. |
+| `account_index`                 | `u32`           | `ESPRESSO_DEPLOYER_ACCOUNT_INDEX`    | `0`                                                           | Account index in the L1 wallet generated by `MNEMONIC` to use when deploying the contracts.                                                  |
```
tbro, line 20:
```suggestion
| `account_index`                 | `u32`           | `ESPRESSO_DEPLOYER_ACCOUNT_INDEX`    | `0`                                                           | Account index of the L1 wallet generated from `MNEMONIC`. Used when deploying contracts.                                                  |
```

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,71 @@
...
+By calling this, the L1 height in the light contract will be frozen, and rollups will detect the HotShot failure. This
+is intended for testing rollups' functionalities when HotShot is down.
+
+Parameters
+
+| Name   | Type    | Description                              |
+| ------ | ------- | ---------------------------------------- |
+| height | integer | The L1 height from which hotshot is down |
+
+### POST /api/set-hotshot-up
```
tbro, line 65: I'm curious if we really need both endpoints. could we just have `post /api/hotshot` or similar which which reacted on the basis of current state of hotshot (like a toggle). Also since these are POSTs, it might be reasonable to move the command to the POST body. just a thought

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,71 @@
...
+| `prover_port`                   | `Option<u16>`   | `ESPRESSO_PROVER_PORT`               | An unused port                                                | Port for connecting to the prover. If this is not provided, an available port will be selected.                                              |
+| `dev_node_port`                 | `u16`           | `ESPRESSO_DEV_NODE_PORT`             | `20000`                                                       | Port for the dev node. This is used to provide tools and information to facilitate developers debugging.                                     |
+
+## APIs
+
+Once you have successfully run the dev node, you can access the corresponding ports to call the APIs of the
+[`builder`](https://docs.espressosys.com/sequencer/api-reference/builder-api),
+[`sequencer`](https://docs.espressosys.com/sequencer/api-reference/sequencer-api), and `prover`.
+
+In addition, you can access the `dev_node_port` to retrieve debugging information. Here are the details of the dev node
```
tbro, line 33: Consider linking to the toml specification or generated docs (if we generate documentation from the specification).

doc/espresso-dev-node.md
ImJeremyHe: Yes that makes sense. I don't have any though about which is better. But the parameters of these 2 endpoints are totally different. I think we don't have to combine these into only 1.

doc/espresso-dev-node.md
ImJeremyHe: The toml specification is used to generate an HTML doc for users accessing to a wrong route/endpoint. Of course we can also link the toml file here, but just a bit weird. And `builder api` page doesn't link anything here, so I just followed the conventions.

tbro: builder api documentation should also link to spec :smiley:  (IMO) .

doc/espresso-dev-node.md
tbro: Humm yea but it is POST, so if we do something like RPC we could just have a command input, and the request bodies would all differ based on the command. But anyway it not really import

doc/espresso-dev-node.md
```diff
@@ -0,0 +1,71 @@
...
+
+### POST /api/set-hotshot-down
+
+This endpoint simulates the effect of a liveness failure of the hotshot consensus protocol in the Light Client smart
+contract.
+
+By calling this, the L1 height in the light contract will be frozen, and rollups will detect the HotShot failure. This
+is intended for testing rollups' functionalities when HotShot is down.
+
+Parameters
```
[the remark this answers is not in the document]
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/1722/commits/4e6f01c2f2684a6039db63ca0ea40730f32ba542

tbro (approved):
I'm approving since my comments where mostly suggestions anyway. Hopefully we can keep some of them in mind for future improvements.

PR #1726, opened 2024-07-18
Add Auction Types

tbro:
### This PR:
Adds `BidTx` and friends to `v0_3`. The intention is to provide the necessary elements for first round of Marketplace integration. Since this is all hidden behind a version gate, our priority should be to satisfy integration necessities.

### This PR does not:
Attempt to implement many changes necessary for sequencer `v0_3` (alias Marketplace 0). We only need the basic types for now, and maybe some constructors and such.

### Key places to review:
  * `types/src/v0/v0_3/auction.rs`
  * `types/src/v0/impls/auction.rs` is also included to provide constructors, but most of that logic will not be executed at this time.


types/src/v0/impls/auction.rs
```diff
@@ -0,0 +1,274 @@
...
+            namespaces,
+            url,
+            // TODO gas_price will come from config probably, but we
+            // can use any value for first round of integration
+            ..Self::default()
+        }
+    }
+
+    /// Sign `BidTxBody` and return the signature.
+    pub fn sign(&self, key: &EthKeyPair) -> Result<Signature, SigningError> {
```
shenkeyao, line 68: Do we expect to use this function outside of this crate? If not, can it be a private function, or removed (i.e., `signed` would call `sign_builder_message` directly), to make it clearer that to get `BidTx` we just need to use `signed`?

types/src/v0/impls/auction.rs
```diff
@@ -0,0 +1,274 @@
...
+            namespaces: vec![nsid],
+        }
+    }
+}
+impl Default for BidTx {
+    fn default() -> Self {
+        let body = BidTxBody::default();
+        let key = FeeAccount::test_key_pair();
+        let signature = FeeAccount::sign_builder_message(&key, body.commit().as_ref()).unwrap();
+        Self { signature, body }
```
shenkeyao, line 121: Nit: Can we replace these lines with the `signed` function?

tbro: e0fdf330b8d1133bebf540538bb40737a77c1713

types/src/v0/impls/auction.rs
tbro: Yea, I guess my preference is in early integration is to have too much rather than not enough. df255e6a813d8e403076e00ae573a46108baf2ef

shenkeyao (approved):
May need a second approval in case I'm not caught up with everything yet, but this looks good to me! I mainly focused on reviewing the bid tx part.

tbro:
I'm going to go ahead and merge b/c eve

> May need a second approval in case I'm not caught up with everything yet, but this looks good to me! I mainly focused on reviewing the bid tx part.

I'm going to go ahead and merge b/c everything is gated behind a version anyway.

PR #1759, opened 2024-07-19
[WEEKLY RELEASE] Hotshot 0.5.63

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/types/src/v0/impls/block/full_payload/payload.rs
+++ b/types/src/v0/impls/block/full_payload/payload.rs
@@ -219,14 +219,6 @@ impl QueryablePayload<SeqTypes> for Payload {
 
         TxProof::new(index, self, &common)
     }
-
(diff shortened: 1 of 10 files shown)


types/src/v0/impls/block/full_payload/payload.rs
```diff
@@ -219,14 +219,6 @@ impl QueryablePayload<SeqTypes> for Payload {
 
         TxProof::new(index, self, &common)
     }
-
-    fn transaction(
```
imabdulbasit, line 223: why are we removing this?

bfish713: oh we shouldn't, I tested with this change, just didn't push it, I'll fix it

imabdulbasit (approved):

PR #1762, opened 2024-07-22
`RollCall` changes

rob-maron:
Closes https://github.com/EspressoSystems/HotShot/issues/3455

## This PR:
Adds a message handling framework to handle `External` HotShot events. These are messages that are "external" to HotShot but that HotShot subscribers may want to about. It's like an out-of-band communication layer and prevents us from having to use a separate network to disseminate the information.

Adds `RollCall` functionality. The frontend will use this to aggregate node identities, starting with the public API URL. On startup a node's `RollCallInfo` will be broadcast the network. Past that, we will have an external service occasionally sending these messages (e.g. when it restarts) and providing the data to the frontend.

Adds a Public API URL as an optional argument. If not supplied, we will not send a `RollCall` message.


rob-maron:
We need to wait for HotShot to be updated before this gets merged

imabdulbasit (approved):
LGTM! just a minor comment.

sequencer/src/external_event_handler.rs
```diff
@@ -0,0 +1,368 @@
...
+                    .with_context(|| "Failed to serialize roll call response")?;
+
+                // Send the response
+                self.outbound_message_sender
+                    .try_send(OutboundMessage::Direct(response_bytes, pub_key))
+                    .with_context(|| "External outbound message queue is full")?;
+            }
+
+            _ => {
+                return Err(anyhow::anyhow!("Unknown external message type"));
```
imabdulbasit, line 113: we can use bail!() here

sequencer/src/external_event_handler.rs
```diff
@@ -0,0 +1,368 @@
...
+
+impl ExternalEventHandler {
+    /// Creates a new `ExternalEventHandler` with the given network and roll call info
+    pub fn new<N: ConnectedNetwork<PubKey>>(
+        network: Arc<N>,
+        roll_call_info: RollCallInfo,
+    ) -> Result<Self> {
+        // Create the outbound message queue
+        let (outbound_message_sender, outbound_message_receiver) =
+            async_compatibility_layer::channel::bounded(10);
```
imabdulbasit, line 57: I think the event handling will be quick, but should the channel capacity be increased just to be on the safe side?

PR #1764, opened 2024-07-22
use max base fee in builder

imabdulbasit:
The builder does not have know about upgrades and uses the base fee from the default chain  in the Genesis file. To support chain config upgrades, the builder should use the highest base fees from the list of chain config upgrades and the default chain config.


QuentinI (approved):
LGTM

PR #1770, opened 2024-07-23
Use addmod in function _computeLinPolyConstantTerm.

### LINKED ISSUE #1728: M01: Use addmod for _computeLinPolyConstantTerm
**Affected Code:**  PlonkVerifier.sol (lines 305,310,315,320)

**Summary:** The above-referenced lines use add instead of `addmod` for the parts of the calculations. Such calculations are implicitly performed modulo $2^{256}$, instead of modulo the order of the scalar field, which is
incorrect. In this particular instantiation, due to the size of the BN254 scalar field, the calculations will not overflow (i.e., for all previously reduced $a$ and $b$, $a+b < 2^{256}$), and the addition will take place over the integers. Still, this behavior is specific to BN254 and not documented. Using the same code on a different curve would produce erroneous results.

**Suggestion:** Switch to using addmod throughout, or alternatively add appropriate documentation to indicate a BN254-specific optimization.

**Status:** Pending

philippecamacho:
Closes #1728

### This PR:
Follows suggestion described in #1728.


gas-benchmarks.txt
alxiong, line 1: btw, I removed this file and use `forge snapshot` native tool to generate `.gas-snapshot` instead in https://github.com/EspressoSystems/espresso-sequencer/commit/b3ec621ceeb04f29c8d767f810c69d178f9cebd3

PR #1772, opened 2024-07-24
Send large tx in espresso-dev-node

ImJeremyHe:
Try to send a tx with a size of `19kb` in the `espresso-dev-node`. The contract size limit in `eth` is `24kb`. So this tx should be legal.

But now, this tx will never be accepted, which causes the program to hang there.

Related logs:

> at /Users/jeremyhe/.cargo-nix/git/checkouts/hotshot-events-service d9f56c6d190f3708/efe92b2/src/events_source.rs:68
> 2024-07-24T06:26:26.319533Z  INFO tide_disco::middleware: responding with error: transaction TX~Z9-_-GauVLVYyBYhz7YEnz_8CeDefnE_4qS6eouOoO6k missing or not available
> at /Users/jeremyhe/.cargo-nix/registry/src/index.crates.io-6f17d22bba15001f/tide-disco-0.9.0/src/middleware.rs:89
> 2024-07-24T06:26:26.320121Z  INFO espresso_dev_node::tests: waiting for large tx

Moreover, the `espresso-dev-node` will not accept any tx after receiving the large tx and it looks like some internal stuff has broken.

### This PR:
- Increase the default max block size to 30720 bytes.
- Skip the transactions whose size exceeds the `max_block_size`. Without this, the process of building blocks will keep generating the empty blocks because the large transaction is always the first one.
- Update the testing builder listen address to `0.0.0.0`. This ensures the accessibility of builder enpoints of the `espresso-dev-node` containers

### Key places to review:
The appropriate default value for the `max_block_size`.


sequencer/src/lib.rs
```diff
@@ -420,7 +420,9 @@ pub mod testing {
         (
             <SimpleBuilderImplementation as TestBuilderImplementation<SeqTypes>>::start(
                 NUM_NODES,
-                url.clone(),
+                format!("http://0.0.0.0:{port}")
+                    .parse()
+                    .expect("Failed to parse builder listener"),
```
ImJeremyHe, line 425: This is the listener. `espresso-dev-node` needs this when it is running in a docker container

types/src/v0/impls/block/full_payload/payload.rs
```diff
@@ -68,13 +68,23 @@ impl Payload {
...
+                tracing::warn!(
+                    "skip the transaction to fit in maximum block byte length {max_block_byte_len}, transaction size {tx_size}"
+                );
+                continue;
+            }
+
+            // accounting for block byte length limit
+            block_byte_len += tx_size;
             if block_byte_len > max_block_byte_len {
                 tracing::warn!("transactions truncated to fit in maximum block byte length {max_block_byte_len}");
```
sveitser, line 89: This isn't part of this PR but if we have a large transaction (that is not larger than the block) we skip adding any transactions behind it even if more transactions may fit into the block.

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -438,6 +439,37 @@ mod tests {
                 .await;
         }
 
+        let large_tx = Transaction::new(100_u32.into(), vec![0; 20000]);
```
sveitser, line 442: I think to make it a proper regression test we should also send a transaction that is larger than the limit and check that it doesn't break the dev node.

types/src/v0/impls/block/full_payload/payload.rs
ImJeremyHe: I guess because we cannot change the order of the transactions, so I just keep this logic.

sequencer/src/bin/espresso-dev-node.rs
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/1772/commits/ffecc3bad8b07357f7c4ba6c8057ce1b13007e7b

sveitser (approved):

PR #1783, opened 2024-07-25
marketplace-solver binary

imabdulbasit:


sequencer/src/bin/marketplace-solver.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Port to run the server on.
+    #[clap(short, long, env = "ESPRESSO_MARKETPLACE_SOLVER_API_PORT")]
+    solver_api_port: u16,
+
+    /// Hotshot events service api URL
+    #[clap(short, long, env = "ESPRESSO_SEQUENCER_HOTSHOT_EVENT_API_URL")]
+    events_api_url: String,
+
+    #[clap(flatten)]
+    database_options: DatabaseOptions,
```
imabdulbasit, line 31: I might add the database options here in the future, instead of importing them from the marketplace-solver crate, to keep the env names consistent.

tbro (approved):

PR #1696, opened 2024-07-11
JIT Marketplace

tbro:
Supersedes #1633 and  #1659, bringing in `FullNetworkTx` and multiple fees per block changes.

### This PR:
Adds types and implementations for JIT marketplace.  It includes changes for multiple fees per block. It doesn't currently compile due to pending changes in traits defined in HotShot. 

### Key places to review:
  * `auction.rs`  (impl and v0_3 type)
  * `header.rs` (impl and v0_3 type)
  * `v0_3/chain_config.rs`
  * `impls/state.rs` 

### How to test this PR: 
Currently untestable due to unsatisfied external dependencies.

PR #1784, opened 2024-07-25
Update to hotshot rc-0.5.67 / marketplace

tbro:
Update to hotshot rc-0.5.66

Most changes are being brought in from #1696. Also includes changes
from #1781. Most compile errors are fixed. Depends on
`hotshot-query-service` pointing to the same hotshot tab (see
Cargo.toml patch).


types/src/v0/v0_1/state.rs
```diff
@@ -1,14 +1,12 @@
-use crate::{Header, ResolvableChainConfig};
-
-use super::{FeeAccount, FeeAmount};
-
+use super::{FeeAccount, FeeAmount, Header};
```
imabdulbasit, line 1: oh the issue is here. It is importing v01 Header instead of that versioned header

shenkeyao (approved):

types/src/v0/impls/state.rs
```diff
@@ -283,32 +303,41 @@ impl From<MerkleTreeError> for FeeError {
...
+        state.charge_fee(fee_info, recipient)?;
+        delta.fees_delta.extend([fee_info.account, recipient]);
+    }
     Ok(())
 }
 
-/// Validate builder account by verifying signature
+/// Validate builder accounts by verifying signatures. All fees are
+/// verified against signature by index.
 fn validate_builder_fee(proposed_header: &Header) -> Result<(), BuilderValidationError> {
```
QuentinI, line 318: This should be versioned instead of replaced, pre-marketplace we use old behavior.

tbro: *I think* this is okay.  To support old behavior, Header v1 logic has been updated to wrap fees and signatures in a `Vec`. So I think this should be ok *in principal*. I'll go over it again to look for implementation errors  though.

QuentinI: Oh, I see. I'm seeing `invalid builder fee: Invalid Builder Signature` errors in `demo-native` though so something somewhere doesn't work.

tbro: We were definitely missing a test for this. I've add a test and it is working as expected. Is it possible that signing is being done incorrectly? 4e3b484546acf27888e9699bdfdeb9a7822b67d3

tbro: Sorry I wasn't thinking clearly.  c4a9a8edff1ccfff2ef912e4b0d0b53a13ea778a should fix it though.

types/src/v0/v0_3/chain_config.rs
```diff
@@ -0,0 +1,175 @@
...
+    fn from(
+        &v0_1::ResolvableChainConfig { chain_config }: &v0_1::ResolvableChainConfig,
+    ) -> ResolvableChainConfig {
+        match chain_config {
+            Either::Left(chain_config) => ResolvableChainConfig {
+                chain_config: Either::Left(ChainConfig::from(chain_config)),
+            },
+            // TODO does this work? is there a better way?
+            Either::Right(c) => ResolvableChainConfig {
+                chain_config: Either::Right(Commitment::from_str(&c.to_string()).unwrap()),
```
imabdulbasit, line 115: I also haven't found a better way to handle this. maybe @jbearer would know

types/src/v0/v0_3/chain_config.rs
```diff
@@ -0,0 +1,175 @@
...
+            fee_recipient,
+        }
+    }
+}
+
+impl Default for ChainConfig {
+    fn default() -> Self {
+        Self {
+            chain_id: U256::from(35353).into(), // arbitrarily chosen chain ID
+            max_block_size: 10240.into(),
```
imabdulbasit, line 168: dev node test will fail. I had to change it to 30mb

types/src/v0/impls/auction.rs
```diff
@@ -213,19 +230,61 @@ impl BidTx {
...
-        vec![self.url.clone()]
+        self.winning_bids()
+            .iter()
+            .map(|bid| bid.url())
+            .chain(self.reserve_bids().iter().map(|bid| bid.1.clone()))
+            .collect()
+    }
+}
+
+const SOLVER_URL: &str = "https://solver:1234";
```
imabdulbasit, line 269: this should come from env var?

sequencer/Cargo.toml
```diff
@@ -11,6 +11,9 @@ testing = ["hotshot-testing"]
 libp2p = []
 benchmarking = []
 
+[lints.rust]
```
imabdulbasit, line 14: I think these were already by jeb and merged in main

data/genesis/demo.toml
```diff
@@ -22,3 +22,10 @@ base_fee = '1 wei'
 max_block_size = '1mb'
 fee_recipient = '0x0000000000000000000000000000000000000000'
 fee_contract = '0xa15bb66138824a1c7167f5e85b957d04dd34e468'
+
+[[upgrade]]
+version = "0.4"
+start_proposing_view = 5
+stop_proposing_view = 15
+
+[upgrade.marketplace]
```
imabdulbasit: we can remove it for now as UpgradeType changes are in integrations PR

types/src/v0/impls/auction.rs
tbro: Yes but it doesn't matter yet.

sequencer/Cargo.toml
imabdulbasit: nvm they are moved from bottom to top

types/src/v0/v0_3/chain_config.rs
tbro: I think this is ok to merge for now, but we try to figure out if there is a better way before we turn on marketplace.

types/src/v0/impls/auction.rs
```diff
@@ -137,6 +150,9 @@ pub enum ExecutionError {
     #[error("Could not resolve `ChainConfig`")]
     /// Could not resolve `ChainConfig`.
     UnresolvableChainConfig,
+    #[error("Bid ricipient not set on `ChainConfig`")]
```
imabdulbasit, line 153: typo here: Bid recipient

imabdulbasit (approved):
LGTM! left some minor comments.

data/genesis/demo.toml
imabdulbasit: I think that toml parsing will fail so yeah needs to be removed

PR #1792, opened 2024-07-29
Marketplace integration

imabdulbasit:


imabdulbasit (commented):
Currently, we are using the v1 builder API in the builder binary while the base version is set to v3. We have a marketplace builder for the v3 version, so I have set all the Version parameters to SequencerVersion, which is v1. If we remove the es_crates, then the SequencerVersion needs to be changed to v1. https://github.com/EspressoSystems/espresso-sequencer/issues/1767

tbro:
In case you need something from the old branch, it has be renamed to [maerketplace-integration-0]( https://github.com/EspressoSystems/espresso-sequencer/tree/maerketplace-integration-0)

PR #1800, opened 2024-07-30
DRAFT PR | 1706 improve versioning for the light client contract

### LINKED ISSUE #1706: improve versioning for the light client contract
The light client contract is upgradeable via a proxy and thus has a lot of rules to follow to ensure that upgrading the contract does not mess up storage locations. Currently there's an example of V2 of the contract called `LightClientV2` but it's out of sync with the LightClient contract since upgrades have been made since then.

alysiahuggins:
DRAFT PR
Closes #1706

### This PR:
- Seeks to find the best smart contract design pattern when writing upgradable contracts so that the code is more maintainable and less susceptible to bugs
- The approach considered is using inheritance to make writing new versions more modular 
- the `virtual` keyword was added to functions in the LightClient contract so that they can be overridden in future versions of its implementation

### This PR does not:
- write explicit tests for common bugs in upgradeable contracts such as storage collisions

### Key places to review:
- changes to the LightClient contract
- how extending a struct is handled in the LightClientV2 contract

### How to test this PR: 
`forge test --match-contract LightClientUpgradeToV2Test`
`just sol-test`

### Things tested
- adding a new field works
- adding an extended struct for handling of new struct elements

### To Do
- continue testing for adding more structs in later versions
- add a test for creating a version that adds all the changes of each version
- ensure deployment works with openzeppelin defender to make sure other security checks pass


doc/smart-contract-upgrades.md
```diff
@@ -0,0 +1,194 @@
+# Smart Contract Upgrades
+
+Code on the EVM is immutable but via proxies, smart contracts can be “upgraded”. A proxy contract acts as an
```
Sneh1999, line 3: I think the documentation should be little more specific to our contracts - the `_Example:_` section should have an example of how one should proceed if they need to upgrade `LightClient.sol`. Otherwise I think, this becomes general documentation about smart contract upgradeability which can be found online. What are your thoughts on the same?

alysiahuggins: fair point, i'll add more info specific to LightClient.sol

contracts/test/LightClientUpgradeToVx.t.sol
```diff
@@ -0,0 +1,168 @@
...
+            LCV2.ExtendedLightClientState(0);
+
+        assertEq(abi.encode(lcV2Proxy.getFinalizedState()), abi.encode(expectedLightClientState));
+        assertEq(
+            abi.encode(lcV2Proxy.getExtendedFinalizedState()),
+            abi.encode(expectedExtendedLightClientState)
+        );
+    }
+
+    // that the data remains the same after upgrading the implementation
```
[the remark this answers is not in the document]
alysiahuggins: sorted here https://github.com/EspressoSystems/espresso-sequencer/commit/7c3569c83e067dbb49c1521f67895de69bffed8c

doc/smart-contract-upgrades.md
alysiahuggins: sorted here https://github.com/EspressoSystems/espresso-sequencer/commit/50dec6911459403d8c662ae7ac331fb00730888e

Sneh1999 (approved):
LGTM!

PR #1802, opened 2024-07-31
marketplace solver rollup registration types

imabdulbasit:
- adds rollup registration and rollup update types 
- implements commitable trait for these types

### DIFF
--- a/types/src/v0/impls/solver.rs
+++ b/types/src/v0/impls/solver.rs
@@ -9,13 +9,11 @@ impl Committable for RollupRegistrationBody {
     }
 
     fn commit(&self) -> Commitment<Self> {
-        let active: [u8; 1] = if self.active { [1] } else { [0] };
-
         let mut comm = committable::RawCommitmentBuilder::new(&Self::tag())
             .u64_field("namespace_id", u64::from(self.namespace_id))
             .var_size_field("reserve_url", self.reserve_url.as_str().as_ref())
             .fixed_size_field("reserve_price", &self.reserve_price.to_fixed_bytes())
-            .fixed_size_field("active", &active)
+            .fixed_size_field("active", &[u8::from(self.active)])
             .constant_str("signature_keys");
 
         for key in self.signature_keys.iter() {
@@ -48,9 +46,7 @@ impl Committable for RollupUpdatebody {
         };
 
         if let Some(active) = self.active {
-            let active: [u8; 1] = if active { [1] } else { [0] };
-
-            comm = comm.fixed_size_field("active", &active);
+            comm = comm.fixed_size_field("active", &[u8::from(active)]);
         }
 
         if let Some(keys) = &self.signature_keys {
--- a/types/src/v0/impls/solver.rs
+++ b/types/src/v0/impls/solver.rs
@@ -5,7 +5,7 @@ use super::v0_3::{RollupRegistrationBody, RollupUpdatebody};
 
 impl Committable for RollupRegistrationBody {
     fn tag() -> String {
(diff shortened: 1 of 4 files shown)


types/src/v0/impls/solver.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use super::v0_3::{RollupRegistrationBody, RollupUpdatebody};
+
+impl Committable for RollupRegistrationBody {
+    fn tag() -> String {
+        "ROLLUP_REG".to_string()
+    }
+
+    fn commit(&self) -> Commitment<Self> {
+        let active: [u8; 1] = if self.active { [1] } else { [0] };
```
tbro, line 12:
```suggestion
        let active: [u8; 1] = [u8::from(self.active)];
```

types/src/v0/impls/solver.rs
```diff
@@ -0,0 +1,67 @@
+use committable::{Commitment, Committable};
+use hotshot::types::SignatureKey;
+
+use super::v0_3::{RollupRegistrationBody, RollupUpdatebody};
+
+impl Committable for RollupRegistrationBody {
+    fn tag() -> String {
+        "ROLLUP_REG".to_string()
```
tbro, line 8: Unless we get some benefit from the shorter string, I think we should spell it out:

```suggestion
        "ROLLUP_REGISTRATION".to_string()
```

types/src/v0/impls/solver.rs
```diff
@@ -0,0 +1,67 @@
+use committable::{Commitment, Committable};
+use hotshot::types::SignatureKey;
+
+use super::v0_3::{RollupRegistrationBody, RollupUpdatebody};
+
+impl Committable for RollupRegistrationBody {
```
tbro, line 6: I'm wondering if it would be better to do the implementation on `RollupRegistration`, since that is the thing that will be getting passed around. `commit()` could just ignore the signature.

types/src/v0/impls/solver.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        comm = comm
+            .var_size_field("signature_key", &self.signature_key.to_bytes())
+            .var_size_field("text", self.text.as_bytes());
+
+        comm.finalize()
+    }
+}
+
+impl Committable for RollupUpdatebody {
```
tbro, line 31: Here again I think it may be better to have the implementation on the outer type, to avoid exposing the internals.

types/src/v0/impls/solver.rs
imabdulbasit: done in 5e5c80536f400ed1adb96c7a01c23792c3dbb6b4

tbro (commented):
I think this is reasonable for moving us forward. But I think we will need constructors and getters in the not so distant future. Is that the plan?

types/src/v0/impls/solver.rs
imabdulbasit: So the signature is created by signing the commit of registration body. If the commit() is implemented over the the outer struct then I would need to insert a random signature -> sign the commit() and then overwrite the signature or am I missing something?

imabdulbasit:
> I think this is reasonable for moving us forward. But I think we will need constructors and getters in the not so distant future. Is that the plan?

Yeah but these types may change maybe so I am exposing all the fields for now

types/src/v0/impls/solver.rs
tbro: Yea, silly me.

tbro (approved):

PR #1804, opened 2024-07-31
Remove es_version crate and use SeqTypes::Base version

imabdulbasit:
We have consistencies where we use the es_version crate's SequencerVersion in most places, and also SeqTypes::Base version. This did not cause any issues as both versions were the same. However, updating the base version would cause some tests to fail.

This PR
- removes es_crate completely and use SeqTypes::Base version

future work:
- parameterize SeqTypes<Base, Upgrade> and then instantiate the sequencer at the top level with the appropriate Base and Upgrade version based on runtime configuration (e.g. genesis file) so that we are still able to run staging tests with 0.1 and 0.2 with the same binary
- add message compat tests for all versions as changing base version will break them

### DIFF
--- a/builder/src/bin/permissionless-builder.rs
+++ b/builder/src/bin/permissionless-builder.rs
@@ -3,15 +3,14 @@ use std::{num::NonZeroUsize, path::PathBuf, str::FromStr, time::Duration};
 use builder::non_permissioned::{build_instance_state, BuilderConfig};
 use clap::Parser;
 use cld::ClDuration;
-use espresso_types::{eth_signature_key::EthKeyPair, SeqTypes};
+use espresso_types::eth_signature_key::EthKeyPair;
 use hotshot::traits::ValidatedState;
-use hotshot_builder_core::testing::basic_test::NodeType;
 use hotshot_types::{data::ViewNumber, traits::node_implementation::ConsensusTime};
 use sequencer::{Genesis, L1Params};
 use sequencer_utils::logging;
 use snafu::Snafu;
 use url::Url;
-use vbs::version::StaticVersionType;
+use vbs::version::{StaticVersion, StaticVersionType};
 
 #[derive(Parser, Clone, Debug)]
 struct NonPermissionedBuilderOptions {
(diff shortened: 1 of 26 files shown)


builder/src/bin/permissionless-builder.rs
```diff
@@ -123,7 +122,7 @@ async fn main() -> anyhow::Result<()> {
         genesis.chain_config,
         l1_params,
         opt.state_peers,
-        <SeqTypes as NodeType>::Base::instance(),
+        StaticVersion::<0, 1>::instance(),
```
jbearer, line 125: Why change from using `SeqTypes` to hard-coding?

builder/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub fn run_builder_api_service(url: Url, source: ProxyGlobalState<SeqTypes>) {
     let private_mempool_api = hotshot_builder_api::v0_1::builder::submit_api::<
         ProxyGlobalState<SeqTypes>,
         SeqTypes,
-        <SeqTypes as NodeType>::Base,
+        StaticVersion<0, 1>,
```
jbearer, line 89: Why hard code?

builder/src/lib.rs
```diff
@@ -98,7 +98,7 @@ pub fn run_builder_api_service(url: Url, source: ProxyGlobalState<SeqTypes>) {
     app.register_module("txn_submit", private_mempool_api)
         .expect("Failed to register the private mempool API");
 
-    async_spawn(app.serve(url, <SeqTypes as NodeType>::Base::instance()));
+    async_spawn(app.serve(url, StaticVersion::<0, 1>::instance()));
```
jbearer, line 101: Why hard code?

builder/src/bin/permissionless-builder.rs
imabdulbasit: I think  builder uses v1 api so I added V1 version. I can change it to Base and see if tests passes

imabdulbasit: Actually, now I remember I was getting version mismatch errors and had to change to v1

jbearer: Oh builder API version is unrelated to sequencer protocol...fair enough

jbearer (approved):

PR #1806, opened 2024-07-31
Marketplace builder

QuentinI:
Closes #0000

### This PR:
- Adds marketplace builder binary
- Adds it to process-compose
- Fixes an error in header validation for marketplace version

### This PR does not:
- Add tests for marketplace builder


marketplace-builder/src/hooks.rs
```diff
@@ -0,0 +1,104 @@
...
+    tracing::info!("Builder client connected to the solver api");
+
+    Some(client)
+}
+
+/// Builder hooks for espresso sequencer
+/// Provides bidding and transaction filtering on top of base builder functionality
+pub(crate) struct EspressoHooks {
+    /// ID of namespace to filter and bid for
+    pub(crate) namespace_id: NamespaceId,
```
shenkeyao, line 45: In the previous implementation, this is an `Option<Vec<...>>`. Did we decide to always filter one namespace for this version?

shenkeyao (approved):
I was mainly focusing on reviewing `marketplace-builder` changes and I haven't tested this out, but the changes look good.

marketplace-builder/src/hooks.rs
QuentinI: I don't think it was a `Vec` before, but I do plan to make it a `Vec` as it's quite an easy change and gives us more flexibility
For `Option` part, it was first there because originally it was supposed to be merged into hotshot-builder-core and the idea was that when running legacy builder we'd want to set it to None and filter nothing. I imagine for generic builder we'd want to make it an Option again

shenkeyao (approved):

docker-compose.yaml
```diff
@@ -484,6 +484,64 @@ services:
       permissionless-builder:
         condition: service_healthy
 
+  marketplace-reserve-builder:
+    image: ghcr.io/espressosystems/espresso-sequencer/marketplace-builder:main
```
imabdulbasit, line 488: can we also add the docker file and CI to include the new binary?

marketplace-builder/src/hooks.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub namespaces: Vec<NamespaceId>,
+    /// Amount to bid.
+    pub amount: FeeAmount,
+}
+
+pub async fn connect_to_solver(
+    solver_api_url: Url,
+) -> Option<Client<hotshot_events_service::events::Error, <SeqTypes as NodeType>::Base>> {
+    let client = Client::<hotshot_events_service::events::Error, <SeqTypes as NodeType>::Base>::new(
+        solver_api_url.join("v0/marketplace-solver/").unwrap(),
```
imabdulbasit, line 40: why do we need to specify v0 here?

marketplace-builder/src/bin/marketplace-builder.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// If not, it's a fallback buidler that only builds for unregistered rollups.
+    #[clap(short, long, env = "ESPRESSO_MARKETPLACE_BUILDER_IS_RESERVE")]
+    is_reserve: bool,
+
+    /// URL of hotshot events API running on Espresso Sequencer DA committee node
+    /// The builder will subscribe to this server to receive hotshot events
+    #[clap(
+        long,
+        env = "ESPRESSO_SEQUENCER_HOTSHOT_EVENT_STREAMING_API_URL",
+        default_value = "http://localhost:8081"
```
imabdulbasit, line 32: why is the default port 8081? Shouldn't we keep it the same as in .env (22001)

QuentinI: Makes sense, it's just copypasta from legacy builder. I'll change both.

docker-compose.yaml
QuentinI: Sure! I added those, although I'm not that well-versed in docker, so mostly by analogue with existing files, hope it all makes sense.

marketplace-builder/src/hooks.rs
QuentinI: We don't, fixed

marketplace-builder/src/hooks.rs
```diff
@@ -82,7 +83,7 @@ impl BuilderHooks<SeqTypes> for EspressoReserveHooks {
             let bid_tx = match BidTxBody::new(
                 self.bid_key_pair.fee_account(),
                 self.bid_amount,
-                view_number + 3, // We submit a bid three views in advance.
+                view_number + 10, // We submit a bid ten views in advance.
```
shenkeyao, line 86: Did we end up deciding to submit 10 views in advance? Last time I checked it's 3 views: https://espresso.zulipchat.com/#narrow/stream/440749-Marketplace/topic/Async.20Standup/near/452603085, but maybe I missed a more recent discussion.

shenkeyao (commented):
Having a question about the view number for bid submission.

marketplace-builder/src/hooks.rs
QuentinI: No, that's my bad, I was testing something and forgot to remove this. 3 views is what we settled on, I reverted this

marketplace-builder/src/bin/marketplace-builder.rs
```diff
@@ -0,0 +1,205 @@
...
+        opt.hotshot_event_streaming_url,
+        builder_server_url,
+        api_response_timeout_duration,
+        buffer_view_num_count,
+        txn_timeout_duration,
+        base_fee,
+        bid_config,
+        opt.solver_url,
+    );
+    let _builder_config = init.await;
```
imabdulbasit, line 199: Why is the Result ignored here?

marketplace-builder/src/hooks.rs
```diff
@@ -0,0 +1,158 @@
...
+            };
+
+            if let Err(e) = solver_client
+                .post::<()>("submit_bid")
+                .body_json(&bid_tx)
+                .unwrap()
+                .send()
+                .await
+            {
+                error!("Failed to submit the bid: {:?}.", e);
```
imabdulbasit, line 111: return here?

docker-compose.yaml
```diff
@@ -484,6 +484,64 @@ services:
...
+      - ESPRESSO_BUILDER_TX_CHANNEL_CAPACITY
+      - ESPRESSO_BUILDER_EVENT_CHANNEL_CAPACITY
+      - ESPRESSO_BUILDER_BOOTSTRAPPED_VIEW
+      - ESPRESSO_BUILDER_WEBSERVER_RESPONSE_TIMEOUT_DURATION
+      - ESPRESSO_BUILDER_BUFFER_VIEW_NUM_COUNT
+      - ESPRESSO_BUILDER_INIT_NODE_COUNT
+      - RUST_LOG
+      - RUST_LOG_FORMAT
+      - ASYNC_STD_THREAD_COUNT
+    depends_on:
```
imabdulbasit, line 512: We are using sequencer0 event streaming URL so this should also depend on sequencer0

docker-compose.yaml
```diff
@@ -484,6 +484,64 @@ services:
...
```
imabdulbasit, line 541: same here

marketplace-builder/src/builder.rs
```diff
@@ -0,0 +1,390 @@
...
+            base_fee
+                .as_u64()
+                .context("the base fee exceeds the maximum amount that a builder can pay (defined by u64::MAX)")?,
+            Arc::new(instance_state),
+            Duration::from_secs(60),
+            Arc::new(validated_state),
+        );
+
+        // spawn the builder event loop
+        async_spawn(async move {
```
imabdulbasit, line 172: Not sure if ignoring all the handles from these async tasks is okay.

marketplace-builder/src/hooks.rs
```diff
@@ -0,0 +1,158 @@
...
+) -> Option<Client<SolverError, <SeqTypes as NodeType>::Base>> {
+    let client = Client::<SolverError, <SeqTypes as NodeType>::Base>::new(
+        solver_api_url.join("marketplace-solver/").unwrap(),
+    );
+
+    if !(client.connect(None).await) {
+        return None;
+    }
+
+    tracing::info!("Builder client connected to the solver api");
```
imabdulbasit, line 48: should also include the api url here

imabdulbasit (changes requested):

docker-compose.yaml
QuentinI: It already indirectly depends on sequencer through fund-builder, but explicitly stating it is probably safer

marketplace-builder/src/builder.rs
QuentinI: Due to how the builder-core works, this handle will actually eventually fork and exit, so we're fine with it exiting.

imabdulbasit (approved):

marketplace-builder/src/builder.rs
sveitser: How about adding this as a comment?

marketplace-builder/src/bin/marketplace-builder.rs
```diff
@@ -0,0 +1,205 @@
...
+use sequencer::{Genesis, L1Params};
+use snafu::Snafu;
+use url::Url;
+use vbs::version::StaticVersionType;
+
+#[derive(Parser, Clone, Debug)]
+struct NonPermissionedBuilderOptions {
+    /// Whether this is a reserve builder.
+    ///
+    /// If not, it's a fallback buidler that only builds for unregistered rollups.
```
sveitser, line 23:
```suggestion
    /// If not, it's a fallback builder that only builds for unregistered rollups.
```

PR #1810, opened 2024-08-01
Add generic builder

shenkeyao:
Closes https://github.com/EspressoSystems/marketplace-builder-core/issues/36.
Note: This PR is to be merged to `ag/builder-bin` and will be merged to main as part of https://github.com/EspressoSystems/espresso-sequencer/pull/1806.

### This PR:
* Adds hooks for generic builder.
  * Generic builder is named as "fallback builder".
  * Existing builder is renamed as "reserve builder".
* Adds a flag to indicate the builder type upon initiation.
* Updates `yaml` files.

### This PR does not:
* Change existing builder logic.

### Key places to review:
* `builder.rs`.
* `hooks.rs`.


QuentinI (approved):
LGTM!

QuentinI:
@shenkeyao @elliedavidson wdyt of making reserve and fallback non-exclusive options? I.e. allowing the builder binary to be run with any combination of those? I imagine having a `--fallback` flag that enables fallback behaviour, i.e. when filtering transactions also requesting list of registered namespaces and including all non-registered transactions. And if one wanted to run purely fallback builder, they'd just set namespace list to empty.

elliedavidson:
> @shenkeyao @elliedavidson wdyt of making reserve and fallback non-exclusive options? I.e. allowing the builder binary to be run with any combination of those? I imagine having a `--fallback` flag that enables fallback behaviour, i.e. when filtering transactions also requesting list of registered namespaces and including all non-registered transactions. And if one wanted to run purely fallback builder, they'd just set namespace list to empty.

Yes, that makes sense to me.

shenkeyao:
> @shenkeyao @elliedavidson wdyt of making reserve and fallback non-exclusive options? I.e. allowing the builder binary to be run with any combination of those? I imagine having a `--fallback` flag that enables fallback behaviour, i.e. when filtering transactions also requesting list of registered namespaces and including all non-registered transactions. And if one wanted to run purely fallback builder, they'd just set namespace list to empty.

Sounds good! I’ll make the change.

marketplace-builder/src/builder.rs
```diff
@@ -202,31 +203,64 @@ impl BuilderConfig {
...
-                },
-                events_url,
-            )
-            .await;
-            tracing::error!(?res, "builder service exited");
-            if res.is_err() {
-                panic!("Builder should restart.");
-            }
-        });
+        if fallback {
```
QuentinI, line 206: I don't think marketplace-builder-core will behave correctly if we spawn two `BuilderState`'s for the same chain.
Can we unify hooks as well and put a flag for fallback there?

QuentinI (changes requested):

QuentinI (approved):

PR #1814, opened 2024-08-02
Use nix-direnv

QuentinI:
Closes #0000

### This PR:
Adds nix-direnv to .envrc. This makes subsequent invocations of direnv on unchanged flake.nix much faster


tbro (approved):
Thanks that startup delay was very annoying.

PR #1817, opened 2024-08-04
Fix the size of the public input vector to 8.

philippecamacho:
Closes https://github.com/EspressoSystems/espresso-sequencer/issues/1733

This PR:
Fixes the size of the public input of the circuit to 8.


philippecamacho:
@alxiong There are still a few TODOs yet could you have a look at this one: https://github.com/EspressoSystems/espresso-sequencer/blob/047c748fd518040116697167d06a02ac933c9702/contracts/src/libraries/PolynomialEval.sol#L158

It seems there is some problem handling memory in assembly like you mentioned in https://github.com/EspressoSystems/espresso-sequencer/pull/1801#issuecomment-2264504954

philippecamacho:
@alxiong Also there are no gas savings (actually benchmarks suggest more gas is spent now...). So we might as well keep the size of inputs variable do more validations as suggested in the ticket. WDYT?

alxiong:
> @alxiong Also there are no gas savings (actually benchmarks suggest more gas is spent now...). So we might as well keep the size of inputs variable do more validations as suggested in the ticket. WDYT?

yes, let's pause this. Maybe come back later to investigate why. 
maybe when we have a more typed Transcript, we can use assembly to `memcpy` from fixed bytes public inputs to transcript more efficiently. 

I don't fully understand why a fixed array increases cost atm.

We can safely turn this into a draft PR for now.

alxiong:
I'm working on this today, I know why tests failed. 
couple of reasons:
- need to update serde code in `diff-test`
- need to update `evaluatePiPoly()` assembly code which access PI as a dynamic array layout instead of fixed array
- etc.

PR #1819, opened 2024-08-05
refactor: add G2 point from SRS to verifying key and transcript

### LINKED ISSUE #1731: M04: Transcript does not include common preprocessed input, or any SRS elements
**Affected Code:**
• Transcript.sol (line 81)
• IPlonkVerifier.sol (line 52)
**Summary:** The transcript currently does not include the entirety of the common preprocessed input (CPI) as specified in the Plonk paper. Rather, it includes the verification key in its place. This is partly acceptable as the elements of the verification key serve as commitments to the elements of the CPI. However, the struct `VerifyingKey` structure lacks any representative of the SRS, introducing an unexpected and undesired degree of freedom. In the paper, the $[x]_2$ is part of the verifier key, whereas in the code, it is represented as a “magic” value.
**Suggestion:** The $[x]_2$ value should be included in the struct `VerifyingKey` structure.
**Status:** Pending

alxiong:
Closes #1731 

### This PR:

- add two `bytes32` in `struct VerifyingKey` to include the compressed serialized G2 point from SRS
- append it to transcript (according update to jellyfish made in https://github.com/EspressoSystems/jellyfish/pull/647)

### This PR does not:

Since the serialized bytes is a fixed value (as I have hardcoded some `assert_eq!()` check in the `adapter/jellyfish.rs`) we could also potentially hardcod these values to the contract itself. 

To avoid the slightly increased cost of appending 2 more bytes32 to transcript, we can hardcode at deployment time as suggested in the [original issue here](https://github.com/EspressoSystems/espresso-sequencer/issues/1731#issuecomment-2243079479).

But I will leave this as future work. This PR only focuses on getting the logic refactored.


Cargo.toml
```diff
@@ -73,7 +73,7 @@ cdn-marshal = { git = "https://github.com/EspressoSystems/Push-CDN", features =
   "global-permits",
 ], tag = "0.4.2", package = "cdn-marshal" }
 
-jf-plonk = { git = "https://github.com/EspressoSystems/jellyfish", tag = "jf-plonk-v0.5.0", features = [
+jf-plonk = { git = "https://github.com/EspressoSystems/jellyfish", branch = "cp-m04", features = [
```
alxiong, line 76: I'll update this once the cooresponding PR on jellyfish is merged. 

at the end of the re-review, we will merge `commonprefix-patch` branch on jellyfish into `main` and release a new tag. and update here.

philippecamacho (approved):
LGTM

PR #1825, opened 2024-08-06
add marketplace-builder to builder-docker-images scripts

imabdulbasit:


process-compose.yaml
```diff
@@ -490,7 +490,6 @@ processes:
       failure_threshold: 100
 
   marketplace-reserve-builder:
-    disabled: true
```
QuentinI, line 493: This (and removing `[marketplace]` in `docker-compose.yml`) will cause conflicts, as both services bind to `ESPRESSO_BUILDER_SERVER_PORT`.

imabdulbasit: do we want to run both builders at the same time or only one of those?

imabdulbasit: fallback has a separate port in .env file?

QuentinI: We want to run either both marketplace builders when we're upgraded to marketplace or one legacy builder otherwise

imabdulbasit: ok I will revert it for now

PR #1827, opened 2024-08-06
Attempt to remove function evaluateLagrangeOne.

### LINKED ISSUE #1752: I16: Redundant function evaluateLagrangeOne
**Affected Code:** PolynomialEval.sol (line 115)

**Summary:** The function evaluateLagrangeOne computes the first Lagrange polynomial evaluation. However, this evaluation is also calculated as an intermediate result in the function `evaluatePiPoly`.

**Suggestion:** We recommend removing the `evaluateLagrangeOne` function and modifying `evaluatePiPoly` to return both the first Lagrange polynomial evaluation and the public input polynomial evaluation. This change would optimize gas usage and reduce code complexity.

philippecamacho:
Closes #1752

### This PR:

Still draft PR. This is an attempt to remove the function evaluateLagrangeOne. While all the Plonk verifier tests are passing, some PolynomialEval tests are still failing. Not clear to me why.

### DIFF
--- a/contracts/src/libraries/PolynomialEval.sol
+++ b/contracts/src/libraries/PolynomialEval.sol
@@ -195,19 +195,20 @@ library PolynomialEval {
         uint256[] memory pi,
         uint256 zeta,
         uint256 vanishingPolyEval
-    ) internal view returns (uint256 res) {
-        uint256 length = pi.length;
+    ) internal view returns (uint256 res, uint256 evalLagrangeOne) {
+        // TODO document better that pi.length=8
+
         uint256 p = BN254.R_MOD;
 
         if (vanishingPolyEval == 0) {
             uint256 group = 1;
-            for (uint256 i = 0; i < length; i++) {
+            for (uint256 i = 0; i < 8; i++) {
                 if (zeta == group) {
-                    return pi[i];
+                    return (pi[i],0); // TODO
                 }
                 group = mulmod(group, self.groupGen, p);
             }
-            return 0;
+            return (0,0); // TODO
         }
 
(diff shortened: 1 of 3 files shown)


contracts/src/libraries/PolynomialEval.sol
```diff
@@ -195,19 +195,20 @@ library PolynomialEval {
...
-            for (uint256 i = 0; i < length; i++) {
+            for (uint256 i = 0; i < 8; i++) {
                 if (zeta == group) {
-                    return pi[i];
+                    return (pi[i], 1);
                 }
                 group = mulmod(group, self.groupGen, p);
             }
-            return 0;
+            return (0, 0);
```
alxiong, line 211: this is wrong logic btw, when `vanishingPolyEval == 0`, the zeta could be one of the domain element: if it's the first domain element, then `lagrange_1_eval = 1`

alxiong (commented):
so the cost of `evaluateLagrangeOne()` is roughly 2.5k. not a lot, should we sacrifice cleaner modularity for this?

contracts/src/libraries/PolynomialEval.sol
philippecamacho: Thanks. Some tests still fail though.

philippecamacho:
> so the cost of `evaluateLagrangeOne()` is roughly 2.5k. not a lot, should we sacrifice cleaner modularity for this?

I agree, especially after Common Prefix's optimization, the code would look less clear if we removed `evaluateLagrangeOne()`.

So let us keep this function as is.

PR #1801, opened 2024-07-30
 refactor!: use big endian and simplify transcript logic

alxiong:
Part of  #1739 
Closes #1797

### This PR:

- Point to the latest release of `jf-plonk` which modify the `PlonkTranscript` logic
- Remove endian reversal for field elements
- For G1Point, instead of using compressed form via `g1Serialize`, simply concatenate two base fields to the transcript byte array
- Update `_computeChallenges()` to reflect the same change

### Gas benchmark

```
@@ -1,2 +1,2 @@
-LightClient_newFinalizedState_Test:testCorrectUpdateBench() (gas: 594565)
-PlonkVerifier_verify_Test:test_verify_succeeds() (gas: 507806)
\ No newline at end of file
+LightClient_newFinalizedState_Test:testCorrectUpdateBench() (gas: 661721)
+PlonkVerifier_verify_Test:test_verify_succeeds() (gas: 496676)
```
I'm not 100% sure why the cost of PlonkVerifier goes down  but the LightClient cost goes up. 
(see if you have any idea @philippecamacho ?)

PR #1829, opened 2024-08-07
improve challenge generation 

### LINKED ISSUE #1735: L03: Challenge generation logic deviates from the protocol specification
**Affected Code:** Transcript.sol (line 45)

**Summary:** The transcript is a log of public inputs, verifying key, and all the messages exchanged between the prover and the verifier. According to the PLONK paper and the Fiat Shamir heuristic, challenges are generated by hashing the current state of the transcript. However, the current implementation prepends the hash of the current transcript to the transcript itself and then hashes that to generate the new challenge. This additional step deviates from the protocol specification. Additionally, the function `getAndAppendChallenge` copies both the hash and the transcript to a new memory location each time it is called. This practice is inefficient and unnecessary.

**Suggestion:** We suggest directly hashing the transcript to align with the standard protocol specifications. Additionally, when generating multiple challenges for the same round (such as $\beta$ and $\gamma$), concatenate an index to the transcript and hash the result. For example:
$\beta = H(transcript, 0)$
$\gamma = H(transcript, 1)$

alxiong:
Closes #1735, 
Correspond to https://github.com/EspressoSystems/jellyfish/pull/648

### This PR:

**Not exactly implementing the suggestion from L03**, 
- dropped unnecessary "append challenge back to transcript" logic after generating as suggested
- use `state` (but properly this time where transcript is emptied each time a challenge is squeezed), see longer explanation in https://github.com/EspressoSystems/espresso-sequencer/issues/1735#issuecomment-2270316275

This PR partially reverts some changes made in #1801 by introducing back the `state`, but this time, the transcript is no longer an ever-growing append-only bytes.

Additionally, given the design, I optimized the `genChallenge()` logic and bring down gas cost by another 36k 🥂


contracts/src/libraries/Transcript.sol
```diff
@@ -40,29 +41,39 @@ library Transcript {
...
-            ret := mod(keccak256(dataPtr, len), p)
-
-            // same as self.transcript = abi.encodePacked(self.transcript, ret);
-            mstore(transcript, newLen)
-            mstore(add(dataPtr, len), ret)
-
-            // update free memory pointer since we extend the dynamic array
-            // to prevent potential overwrite
-            mstore(0x40, add(mload(0x40), 32))
+            let lenPtr := mload(add(self, 0x20))
```
philippecamacho, line 65: Maybe rename `lentPtr` to `transcriptPtr`?

contracts/src/libraries/Transcript.sol
```diff
@@ -40,29 +41,39 @@ library Transcript {
...
+    // 3. challenge = bytes_to_field(state)
+    //
+    // Note: every challenge generation is implicitly domain-separated, thus it's safe
+    // to call it multiple times in a row (e.g. multiple challenges in a single round)
+    // to get multiple different challenges.
+    function getChallenge(TranscriptData memory self) internal pure returns (uint256 ret) {
+        // memory layout:
+        // offset 0x00: state
+        // offset 0x20: pointer to transcript (i.e. 0x40)
+        // offset 0x40: length of transcript
```
philippecamacho, line 55: Not sure this is the case even though I like a lot the documentation. When I change the line

`let lenPtr := mload(add(self, 0x20))`

for

`let lenPtr := add(self, 0x40)`

The tests don't pass anymore. Am I missing something?

philippecamacho (approved):
A question regarding the documentation. Otherwise LGTM. Great assembly trick that yielded some nice gas savings!

contracts/src/libraries/Transcript.sol
alxiong: done in [c28459c](https://github.com/EspressoSystems/espresso-sequencer/pull/1829/commits/c28459cbc1655b33e5ead1dc14d58ba686ec4d31)

PR #1831, opened 2024-08-08
feat: Efficient computeChallenges in pure assembly

### LINKED ISSUE #1738: I01: Inefficient memory allocation in transcript generation logic
**Affected Code:**
• PlonkVerifier.sol (line 173)
• Transcript.sol (line 81)

**Summary:** The current implementation of the function `_computeChallenges` in PlonkVerifier.sol and `appendVkAndPubInput` in Transcript.sol inefficiently manages memory by repeatedly calling `abi.encodePacked` to concatenate elements to the transcript. Each call to `abi.encodePacked` results in a new memory allocation and copying of the existing transcript, leading to excessive gas consumption.

**Suggestion:** Create a struct containing all the fields that are appended to the transcript. Hash the parts of the struct using low-level memory access to generate the relevant challenges. To allow for tight packing of the transcript one can also use a bytes array instead of struct and low-level memory writes to populate the array.

alxiong:
Closes #1738 

### This PR:

- Rewrite `_computeChallenges()` in pure assembly:
  - we no longer uses `Transcript.sol` even, in `PlonkVerifier.sol` (but I didn't delete it just for reference or in case we want to use it or cross-compare with it)
  - all tests pass, just a drop-in replacement of the original, less efficient `_computeChallenges()`, **gas saving: 66k** (from 491667, 2 commits ago, to 425555 now) 🚀 💰 

### This PR does not:

I could have further hardcode VK and avoid using `getVk()` to load into memory and save some memory copying, but I benchmark it: only 2.8k gas. 
Considering that I will have to a) write more assembly; b) update the rust code that generate the current vk contract into rust code that populate my assembly solidity. 
The ROI is low, thus I didn't implement that.


contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -171,84 +168,241 @@ library PlonkVerifier {
...
-        transcript.transcript = abi.encodePacked(
-            transcript.transcript, proof.split3.x, proof.split3.y, proof.split4.x, proof.split4.y
-        );
+        assembly {
+            // use free memory space for scratch pad, 0x40: free memory ptr
+            let statePtr := mload(0x40)
+            let dataPtr := add(statePtr, 0x20)
+
+            // Start of transcript (unit: bytes)
+            // sizeInBits (4)  | domainSize (8) | numInputs (8) | pad (12)
```
philippecamacho, line 183: Should it be `sizeInBytes` instead of `sizeInBits`?

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -171,84 +168,241 @@ library PlonkVerifier {
...
+            }
 
-        res.zeta = transcript.getChallenge();
+            // challenge: gamma
+            {
+                // preimage len: state(0x20) + transcript(0x0)
+                mstore(dataPtr, keccak256(statePtr, 0x20))
+                // update new state (by updating state pointer)
+                statePtr := dataPtr
+                dataPtr := add(statePtr, 0x20)
```
philippecamacho, line 317: Add comment `// empty transcript` to be consistent with the code above?

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -171,84 +168,241 @@ library PlonkVerifier {
...
-            proof.prodPermZetaOmegaEval
-        );
+            // challenge: alpha, alpha2, alpha3
+            {
+                // preimage len: state(0x20) + transcript(0x40)
+                let alpha := keccak256(statePtr, 0x60)
+                // update new state (by updating state pointer)
+                statePtr := add(dataPtr, 0x40)
+                mstore(statePtr, alpha)
+                dataPtr := add(statePtr, 0x20)
```
philippecamacho, line 333: Add comment `// empty transcript` to be consistent with the code above?

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -171,84 +168,241 @@ library PlonkVerifier {
...
+            mstore(add(dataPtr, 0x100), mload(split4Ptr)) // split4.x
+            mstore(add(dataPtr, 0x120), mload(add(split4Ptr, 0x20))) // split4.y
+
+            // challenge: zeta
+            {
+                // preimage len: state(0x20) + transcript(0x140)
+                mstore(add(dataPtr, 0x140), keccak256(statePtr, 0x160))
+                // update new state (by updating state pointer)
+                statePtr := add(dataPtr, 0x140)
+                dataPtr := add(statePtr, 0x20)
```
philippecamacho, line 365: Add comment `// empty transcript` to be consistent with the code above?

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -171,84 +168,241 @@ library PlonkVerifier {
...
+            mstore(add(dataPtr, 0x100), mload(add(proof, 0x2a0))) // sigmaEval3
+            mstore(add(dataPtr, 0x120), mload(add(proof, 0x2c0))) // prodPermZetaOmegaEval
+
+            // challenge: v
+            {
+                // preimage len: state(0x20) + transcript(0x140)
+                mstore(add(dataPtr, 0x140), keccak256(statePtr, 0x160))
+                // update new state (by updating state pointer)
+                statePtr := add(dataPtr, 0x140)
+                dataPtr := add(statePtr, 0x20)
```
philippecamacho, line 387: Add comment `// empty transcript` to be consistent with the code above?

philippecamacho (approved):
Awesome work!

contracts/src/libraries/PlonkVerifier.sol
alxiong: sizeInBits which is why the value is 254. 
it's a data type of uint32 thus 4 bytes.

philippecamacho: I see, thanks.

contracts/src/libraries/PlonkVerifier.sol
alxiong: add in [5f74f10](https://github.com/EspressoSystems/espresso-sequencer/pull/1831/commits/5f74f10918a3fc28a5b53c034e342215abe4cb07) 
similar for all following comments.

PR #1833, opened 2024-08-08
Fix compose file configuration for node-validator and block-explorer

### LINKED ISSUE #1832: The Block Explorer Configuration for process-compose.yaml and docker-compose.yaml is not set for the Node Validator
With the completion of the Node Validator API and the updates provided for the Node Validator UI it is clear that the configuration for the Block Explorer has not been updated to point at the Demo Node Validator.  Additionally, there are misconfigurations of the node-validator itself that has it pointing at the external ports of the `docker-compose` configuration, instead of the internal ports.

The end result is that the `Nodes Page` of the `Node Validator UI` contained within the `Block Explorer` is not reflecting the local data produced by the demo.  In order to fix this, the configurations need to be corrected.

Ayiga:
Fix docker-compose and process-compose for node-validator

The `node-validator` section of the `docker-compose.yaml` file has the environment variable
`ESPRESSO_NODE_VALIDATOR_INITIAL_NODE_PUBLIC_BASE_URLS` pointing at the incorrect
ports for all of the sequencers that are not `sequencer0`.  As a result, the other sequencers are
not correctly scraped and are not reflected in the data collected.  Additionally, the port that the
`node-validator` is meant to expose is not being forwarded to the container via the environment
variable `ESPRESSO_NODE_VALIDATOR_PORT`.  Finally, the local demo service doesn't produce
any visible logs while running.

Forwards the relevant needed environment variables for logging, and modifies the the ports in
the node public base URLs to correct the issue.

Add node validator configuration to process-compose and docker-compose

The block-explorer now has support for the node validator API, but has not been updated to point
to the relevant base URL for the service.  As a result, the Nodes Page will still generate fake data.

Updates the process-compose.yaml and docker-compose.yaml configuration to supply the
environment variable `NODE_VALIDATOR_URI` so that it points to the `node-validator` as a
websocket connection.

Closes #1832

### This PR:

Corrects the configuration of the `node-validator` by ensuring that logging is enabled for the service.  Additionally, it corrects the list of initial public api urls supplied, by utilizing the internal port in the `docker-compose` file instead of the external exposed port.

Adds the node-validator url to the block-explorer in order to for the `Nodes Page` to reflect the data
of the local demo instead of just producing a bunch of fake data.

### Key places to review:

- process-compose.yaml
- docker-compose.yaml

### How to test this PR:

```sh
just demo-native
```

or

```sh
docker-compose up -d
```

With docker-compose, make sure you have the latest images by running the following:

```sh
docker-compose pull
```

Then, once things are up and running, visit the [block-explorer](http://localhost:3000/), and navigate to the [Nodes Page](http://localhost:3000/nodes) to see updating data for the local block-chain and nodes.  Please note that we've supplied fake node identity information for the nodes.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -662,9 +662,12 @@ services:
     ports:
       - "$ESPRESSO_NODE_VALIDATOR_PORT:$ESPRESSO_NODE_VALIDATOR_PORT"
     environment:
+      - RUST_LOG
+      - RUST_LOG_FORMAT
+      - ESPRESSO_NODE_VALIDATOR_PORT
       - ESPRESSO_NODE_VALIDATOR_STAKE_TABLE_SOURCE_BASE_URL=http://sequencer0:$ESPRESSO_SEQUENCER_API_PORT/v0/
       - ESPRESSO_NODE_VALIDATOR_LEAF_STREAM_SOURCE_BASE_URL=http://sequencer0:$ESPRESSO_SEQUENCER_API_PORT/v0/
-      - ESPRESSO_NODE_VALIDATOR_INITIAL_NODE_PUBLIC_BASE_URLS=http://sequencer0:$ESPRESSO_SEQUENCER_API_PORT,http://sequencer1:$ESPRESSO_SEQUENCER1_API_PORT,http://sequencer2:$ESPRESSO_SEQUENCER2_API_PORT,http://sequencer3:$ESPRESSO_SEQUENCER3_API_PORT,http://sequencer4:$ESPRESSO_SEQUENCER4_API_PORT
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -733,7 +736,8 @@ services:
     ports:
       - "$ESPRESSO_BLOCK_EXPLORER_PORT:3000"
     environment:
-      - QUERY_SERVICE_URI:http://localhost:$ESPRESSO_SEQUENCER1_API_PORT/v0/
+      - QUERY_SERVICE_URI=http://localhost:$ESPRESSO_SEQUENCER1_API_PORT/v0/
```
imabdulbasit, line 739: We are specifying the version here, which requires updating it every time we change the base version in the sequencer. Can we omit this and specify it on the client side? surf disco client takes a version and forwards it

imabdulbasit (approved):

docker-compose.yaml
Ayiga: This is a configuration for the Block Explorer which doesn't use `rust` and cannot use `surf-disco` at all.  Anything not prefixed with the version results in a redirect status code.  So it seemss more simple to have the version suffixed.

imabdulbasit: Gotcha

PR #1834, opened 2024-08-08
deploy light client contracts to alt chains and send proofs

imabdulbasit:
closes https://github.com/EspressoSystems/espresso-sequencer/issues/1820
This PR:
- adds a functionality to support alternative chains for light client contracts in additon to base L1 which we already use. This can be useful for L3s
- also adds endpoints to prover service and dev node service to get those alt chain light client addresses


hotshot-state-prover/src/service.rs
```diff
@@ -54,7 +56,19 @@ use crate::snark::{generate_state_update_proof, Proof, ProvingKey};
 type F = ark_ed_on_bn254::Fq;
 
 /// A wallet with local signer and connected to network via http
-pub type L1Wallet = SignerMiddleware<Provider<Http>, LocalWallet>;
+pub type SignerWallet = SignerMiddleware<Provider<Http>, LocalWallet>;
+
+#[derive(Debug, Clone)]
+pub struct AltChainConfig {
```
sveitser, line 62: Currently I'm not sure if handling multiple chains in the same prover service is advantageous. I think we could run multiple prover services for multiple chains. Or does that create any issues?

With multiple prover services they can be configured independently and use different intervals for different chains which is somewhat important because the gas cost can vary a lot between chains.

imabdulbasit: makes sense! I have pushed the changes

sveitser:
@imabdulbasit i think it's missing updates to the documentation markdown file

sequencer/api/espresso_dev_node.toml
```diff
@@ -28,6 +35,7 @@ Set the hotshot down since the given L1 height.
 Body:
 ```
 {
+    "chain_id: Option<u64>
```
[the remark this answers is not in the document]
imabdulbasit: the idea was to use the base L1 light client contract in case the chain_id is not provided.

Sneh1999 (approved):

PR #1835, opened 2024-08-08
refactor!: change publicInputs from dynamic to fixed array

### LINKED ISSUE #1733: L01: Inefficient handling of public inputs

alxiong:
Closes #1733 
Supersede #1811 

(I'll change the target/base branch to `commonprefix-patch` when `cp-i01` is merged)

### This PR:

- Change `uint256[] memory publicInput` to `uint256[8] memory publicInput` everywhere. hardcoded the length `8` in lots of places. 
- Removed `PolynomialEval.domainElements()` since it's no longer needed, I simply inline assembly the logic of getting the 8 domain elements 
  - similarly also inlined the `_computeSuffixProduct()` directly 
 
**Gas saving is only 1.4k**.

### This PR does not:

There are further optimization opportunities (to be done in future PRs)

- `evaluatePiPoly()` is mixture of assembly and non-assembly code, maybe we can turn it into a huge blob of assembly
- `commScalars` and `commBases` can also be turned into fixed array.


.gas-snapshot
```diff
@@ -1,2 +1,2 @@
-LightClient_newFinalizedState_Test:testCorrectUpdateBench() (gas: 591187)
-PlonkVerifier_verify_Test:test_verify_succeeds() (gas: 425555)
\ No newline at end of file
+LightClient_newFinalizedState_Test:testCorrectUpdateBench() (gas: 589370)
```
philippecamacho, line 1: Interesting, the gas cost for  updating the state of the light client contract looks reasonable now.

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -269,14 +269,14 @@ library PlonkVerifier {
 
             // public inputs
             // TODO: update these when change PI from dynamic array to fixed length
```
philippecamacho, line 271: Shall we remove this TODO?

philippecamacho (approved):
LGTM

.gas-snapshot
alxiong: not really. the diff (saving in plonk verifier) was offseted and not 1:1 to diff in light client. still need to investigate later.

contracts/src/libraries/PlonkVerifier.sol
alxiong: done in [cf37205](https://github.com/EspressoSystems/espresso-sequencer/pull/1835/commits/cf3720528e22e7a2b74249a1281615818ecb9928)

PR #1836, opened 2024-08-08
Remove redundant transcript files.

### LINKED ISSUE #1743: I07: Inconsistent definition and usage of COSET_Ki constants
**Affected Code:**
• Transcript.sol (line 104-L119)
• PlonkVerifier.sol (line 28-L35)

**Summary:** The `COSET_Ki` constants are used across multiple contracts. In the PlonkVerifier contract, these are defined as constants, whereas in the Transcript contract, they appear as magic numbers. This can lead to difficulties in maintaining the codebase.

**Suggestion:** We recommend refactoring the code to reference a single source of truth for the `COSET_Ki` constants. This can be achieved by defining these constants in a common library or a shared contract.

philippecamacho:
Closes #1743

### This PR:

After #1829 and #1835, the Transcript.sol library is not needed anymore and thus the corresponding files have been deleted. This also solve the code duplication issue mentioned in #1743 as no the COSET_Ki constants are defined in a single place.


alxiong (commented):
I thought about this too. but I prefer keeping the Transcript and its test file, because we might want to refer to it, or cross-compare with this later?
I don't mind moving them a special subfolder like `contracts/src/legacy/` or "deprecated" or "archived".
(since we didn't use it, the original issue can also be considered as resolved)

alxiong:
can you also change these line: https://github.com/EspressoSystems/espresso-sequencer/blob/commonprefix-patch/contracts/src/libraries/PlonkVerifier.sol#L197-L209

to use `COSET_K1` ... instead of using hardcoded value?

philippecamacho:
> can you also change these line: https://github.com/EspressoSystems/espresso-sequencer/blob/commonprefix-patch/contracts/src/libraries/PlonkVerifier.sol#L197-L209
> 
> to use `COSET_K1` ... instead of using hardcoded value?

Done in 2f40358d7bc1cd0c8b7073a7db43220d407aa45f.

philippecamacho:
> I thought about this too. but I prefer keeping the Transcript and its test file, because we might want to refer to it, or cross-compare with this later? I don't mind moving them a special subfolder like `contracts/src/legacy/` or "deprecated" or "archived". (since we didn't use it, the original issue can also be considered as resolved)

Yes, done in a6af61d781071fa31ba970e1c19aceaf89f5483b.

alxiong (approved):

PR #1837, opened 2024-08-08
moved etherscan api key to env contracts file

### LINKED ISSUE #1799: Improve the handling of config variables related to contracts
After issue #1533 we created another environment file called .env.contracts but we should continue to improve the handling of config variables related to smart contracts

alysiahuggins:
Closes #1799 

### This PR:
Moves the ETHERSCAN_API_KEY into the .env.contracts file since the .env file is currently used for config variables and not secrets

### Key places to review:
the justfile


alxiong (approved):
don't have objection. 

can you point me to the discussion of this separation (what goes into which file and why)? I'd like to catch up. 

e.g. i wonder why aren't we avoid version controlling `.env` file
and why not put all secrets into a separate file. (i.e. separating public and secret configs)

Ancient123:
>  separating public and secret configs

I actually really like this, because config values in our production environment are set differently between the two. I helps with the jump to cloud deployment when I can know what stuff is a secret from just looking at the configs.

Ancient123 (approved):

alysiahuggins:
> e.g. i wonder why aren't we avoid version controlling `.env` file and why not put all secrets into a separate file. (i.e. separating public and secret configs)

from what I see,  all the secrets from the `.env` were contract deployment related, the remaining MNEMONICs are never changed and they only use the test account. 

> don't have objection.
> 
> can you point me to the discussion of this separation (what goes into which file and why)? I'd like to catch up.
> 
Sure I'll ping you

alysiahuggins:
> > separating public and secret configs
> 
> I actually really like this, because config values in our production environment are set differently between the two. I helps with the jump to cloud deployment when I can know what stuff is a secret from just looking at the configs.

ok great @Ancient123 , and we can also have a conversation about handling these deployments going forward.

Issue #1816, opened 2024-08-02
Increase of gas consumption in the light client contract


philippecamacho:
After this commit a93b964612c96158040a62ec855db4bfa3e104dc, the light client contract uses around 660k gas despite the plonk verifier being more gas efficient.

Understand the cause of this sudden increase.

See #1801 for more context.

alxiong:
I looked into this today, now I know why:

<img width="935" alt="Screenshot 2024-08-12 at 4 42 27 PM" src="https://github.com/user-attachments/assets/2bfb7588-71ed-48ad-8b9f-f56c6eaffc84">


```solidity
// ======= before
        // check plonk proof
        verifyProof(newState, proof);

        // upon successful verification, update the latest finalized state
        states[finalizedState] = newState;
        emit NewState(newState.viewNum, newState.blockHeight, newState.blockCommRoot);

// ====== after or now:
        // check plonk proof
        verifyProof(newState, proof);

        // upon successful verification, update the latest finalized state
        states[finalizedState] = newState;

        //add the L1 Block to stateUpdateBlockNumbers for the new finalized state
        stateUpdateBlockNumbers.push(block.number);

        //add the blockheight and blockCommRoot to hotShotCommitments for the new finalized state
        hotShotCommitments.push(HotShotCommitment(newState.blockHeight, newState.blockCommRoot));

        emit NewState(newState.viewNum, newState.blockHeight, newState.blockCommRoot);
```

in short, this is expected. 
our saving in PlonkVerifier is reducing LightClient cost. (but admittedly the LC contract need serious optimization 🤦 )

philippecamacho:
> our saving in PlonkVerifier is reducing LightClient cost. (but admittedly the LC contract need serious optimization 🤦 )

Awesome @alxiong ! Yes, I think @alysiahuggins has been working on some light client optimizations.

PR #1844, opened 2024-08-11
feat!: fully assemblify preparePcsInfo() and verifyOpeningProof

### LINKED ISSUE #1737: L05: Suboptimal verifier code due to missing MSM implementation
**Affected Code:**
• PlonkVerifier.sol
• BN254.sol (line 173)

**Summary:** In the verifier code, particularly within the `_preparePolyCommitments` function, arrays of bases and scalars are allocated and passed between functions with the expectation of a valid Multi-Scalar Multiplication (MSM) implementation. However, the `multiScalarMul` function implementation only performs the naive iterative scalar multiplication and addition. This results in the following issues:

• The allocation (PlonkVerifier (lines L154,L482) ) and copying of arrays (PlonkVerifier.sol (line 488 - L493) ) result in unnecessary memory overhead.
• Introduce severe overhead in code complexity and gas cost for `_batchVerifyOpeningProofs` function.

**Suggestion:** We recommend refactoring the code to calculate the dot products and cumulative sums immediately instead of passing around the scalars and bases arrays and then applying the naive MSM algorithm.

alxiong:
Closes #1737 

**PlonkVerifier gas cost to 385k 😊 (another 38k saving from last commit, 122k gas saving from the audit version)**

~It's still unclear to me why LightClient bench is worse. To be investigated in #1816~ (resolved)

Backstory: 

the original issue suggested that we shouldn't carry the `uint256[] commScalars` and `G1Point[] commBases` around and passing through so many intermediate functions (memory inefficient). since the final MSM is a naive mul-and-add MSM.
Namely, we might as well just compute the scalarMul on each new `commScalar` and `commBase` and only carry this acculuated MSM result forward. 

^^ to make this change is already major headache and refactoring, so I was compelled to completely break down the original code and centralize all the logic into a single assembly block. 

the carried forward MSM intermediate result is stored directly at `bPtr` where `b` is the G1 point in the final pairing check `e(a, [x]_2) =?= e(b, [1]_2)`, and each new scalar and basis is stored as `let scalar`, `let base` in the assembly code.

another nicety of this approach is the computation of `eval` is just alongside the last 10 scalars and bases. Avoiding repeated memory loading of the original code when `eval` and `commScalars` are computed separately.

### This PR:

This is perhaps the biggest refactoring on `commonprefix-patch` branch yet. 

- removed `struct PcsInfo` ! 
  - removed `_preparePcsInfo()` and `_verifyOpeningProof()`
  - instead, centralize and flatten the logic in a single, pure assembly block
- Reduce reliance on `BN254` API, instead inline most of the assembly logic in the `verify()` function directly
  - `negate()`, `scalarMul()` and `multiScalarMul()` are all just local assembly functions
  - functions like `BN254.P2()` are moved in as constants
  - inside assembly, we avoid excessive `.unwrap()` and `.wrap()` for type checker
  
I suggest we keep this version as separate contract since it's also easier for CP to re-review and cross-compare with the old logic. 
We can archive or remove the original PlonkVerifier when we are ready to merge into `main`

### Key places to review:

Due to the major refactoring and large (hard to review) diff, I decide to start `PlonkVerifier2.sol`.
There's no meaningful change before the assembly block inside `verify()` (i.e. unchanged until end of `linPolyConstant` computation)
The assembly block is the only place reviewer need to look.

(note that I will make further PRs to further assemblify `verify()`, but in this PR, I want to restrict the scope of change)


alxiong:
locally I have tried to further assemblify the `validateXX()` and `computeLinPolyTerm()` with the eventual goal of a single assembly blob in the `verify()` like most other verifier contract out there. 

but I notice little (<100) improvement with each, since they are already mostly in assembly. So I'd say we keep the current code for now.

contracts/src/libraries/PlonkVerifier2.sol
```diff
@@ -0,0 +1,927 @@
...
+            // credit: error function from Succinct's SP1 code
+            function error_verify() {
+                let ptError := mload(0x40)
+                mstore(ptError, ERROR_STRING_ID) // selector for function Error(string)
+                mstore(add(ptError, 0x4), 0x20)
+                mstore(add(ptError, 0x24), 0xc)
+                mstore(add(ptError, 0x44), "error verify")
+                revert(ptError, 0x64)
+            }
+            function error_pairing() {
```
philippecamacho, line 465: This function is used only once, why not inline it?

alxiong: we could, but i think this is clearer and easier to audit, and won't cost much extra, just 2~3 more opcode, (push this function selector to stack, jump, then pop off (~10 gas)

philippecamacho (approved):
Fantastic work!

Pushed a couple of very minor gas optimizations.

alxiong:
Thanks for the extra optimization, makes sense!

PR #1850, opened 2024-08-12
add marketplace upgrade type 

imabdulbasit:


sequencer/src/genesis.rs
```diff
@@ -56,10 +56,8 @@ impl Genesis {
         let upgrades: Vec<&Upgrade> = self.upgrades.values().collect();
 
         for upgrade in upgrades {
-            match upgrade.upgrade_type {
-                UpgradeType::ChainConfig { chain_config } => {
-                    base_fee = std::cmp::max(chain_config.base_fee, base_fee);
-                }
+            if let UpgradeType::ChainConfig { chain_config } = upgrade.upgrade_type {
```
tbro, line 59: So this `Vec<&Upgrade>` holds the actions we need to perform per type per upgrade? Is that right? So we are saying that for every upgrade we need to set base_fee to the maximum base fees of all upgrades. I think implicit in this is that all base_fee can only ever increment. Just checking my understanding. thanks!

types/src/v0/impls/header.rs
```diff
@@ -747,13 +747,11 @@ impl BlockHeader<SeqTypes> for Header {
         let mut validated_state = parent_state.clone();
 
         let chain_config = if version > instance_state.current_version {
-            match instance_state
-                .upgrades
-                .get(&version)
-                .map(|upgrade| match upgrade.upgrade_type {
+            match instance_state.upgrades.get(&version) {
+                Some(upgrade) => match upgrade.upgrade_type {
                     UpgradeType::ChainConfig { chain_config } => chain_config,
```
tbro, line 752: Not looking for a change here,  just trying to improve my understanding. These semantics are confusing to me. Why do we have an `UpgradeType:ChainConfig`. Couldn't we just be upgrading between versions, and have the version define what is included in the upgrade? For example, I would imaging that the upgrade type would be `Marketplace` and `ChainConfig` upgrade would happen "under the hood" since it is a dependency of marketplace.

tbro (approved):
Looks good. I've made a couple of comments but they are not to request changes, but are just to help me understand.

types/src/v0/impls/header.rs
imabdulbasit: hmm I think renaming this upgrade type to FeeUpgrade would be better. So, fee upgrade would require a new chain config that is provided in the toml file.

sequencer/src/genesis.rs
imabdulbasit: Vec<Upgrade> is basically array of tables from the genesis toml file. Yes, we set base fee to max base fee if genesis toml file has fee upgrades (chain config upgrades) specified. This is required by builder as builder takes base_fee parameter which can not be updated as it doesn't know about validated state

types/src/v0/impls/header.rs
tbro: Ah! yes I think so too.

PR #1852, opened 2024-08-12
reject transactions with size greater than  max block size

imabdulbasit:
Closes https://github.com/EspressoSystems/espresso-sequencer/issues/1777

This PR

Rejects any transaction whose payload length is more than the max block size. The max block size parameter is found in the chain config, which can be updated through an upgrade. Therefore, we use the chain config from the validated state first, as the validated state is the mutable state of the node and the chain config, if present, will be the latest chain config. 

If the chain config is not present in the validated state, which may occur if the node has not seen an upgrade, we use the chain config from the node state. The node will automatically catch up and have an updated chain config in the validated state when building the header for the next block.

### DIFF
--- a/node-metrics/src/service/client_state/mod.rs
+++ b/node-metrics/src/service/client_state/mod.rs
@@ -1308,7 +1308,7 @@ pub mod tests {
         let (internal_client_message_sender, internal_client_message_receiver) = mpsc::channel(1);
         let (server_message_sender_1, mut server_message_receiver_1) = mpsc::channel(1);
         let (server_message_sender_2, mut server_message_receiver_2) = mpsc::channel(1);
-        let _process_internal_client_message_handle = InternalClientMessageProcessingTask::new(
+        let process_internal_client_message_handle = InternalClientMessageProcessingTask::new(
             internal_client_message_receiver,
             data_state,
             client_thread_state,
@@ -1359,6 +1359,8 @@ pub mod tests {
                 voters_1, voters_2
             ]))),
         );
+
+        drop(process_internal_client_message_handle)
     }
 
     #[async_std::test]
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -168,12 +168,36 @@ impl<N: ConnectedNetwork<PubKey>, Ver: StaticVersionType + 'static, P: Sequencer
     SubmitDataSource<N, P> for ApiState<N, P, Ver>
 {
     async fn submit(&self, tx: Transaction) -> anyhow::Result<()> {
-        self.consensus()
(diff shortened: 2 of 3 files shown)


node-metrics/src/service/client_state/mod.rs
```diff
@@ -1359,6 +1359,8 @@ pub mod tests {
                 voters_1, voters_2
             ]))),
         );
+
+        drop(process_internal_client_message_handle)
```
sveitser, line 1363: Shouldn't it be dropped when it goes out of scope anyway?

imabdulbasit: Yeah it should. The test was failing with an error like stream has ended. I changed it to explicitly drop it, but this might be unrelated

imabdulbasit: I did debug this test but didn't find anything

sveitser (approved):

node-metrics/src/service/client_state/mod.rs
sveitser: I'll  try running the test

imabdulbasit: the exact error was `InternalClientMessageProcessingTask stream closed, unable to process new requests from clients.`
cc @Ayiga

sveitser: @imabdulbasit doesn't fail for me without the drop. I don't mind adding it but I think it's a good idea to leave a comment as to why it was added because it looks a bit out of place.

imabdulbasit: @sveitser yeah I couldn't replicate it either locally. I'll revert this change and see if CI passes. I think its a flaky test

Ayiga: Oh! Actually, this might be erroring occasionally due to a clean up of the underlying channels. If the thread scheduling happens in an inconvenient order, then the channels will drop first, and that will cause process_internal_client_message_handle to fail reading from the channel which will result in a panic! as the task is a critical task for the node-validator's continued operation.

I think the better thing to do, rather than drop(process_internal_client_message_handle) is to just cancel the task and await it. This will ensure that it doesn't get scheduled anymore and will avoid the panic entirely.

It seems like a hacky work-around, but I'm not sure how else to maintain the duality of, "this is a critical task and should panic when running", but it is not possible to catch in testing without a specialized entry point to help tie into the panic.

sveitser: https://github.com/EspressoSystems/espresso-sequencer/issues/1862

PR #1854, opened 2024-08-13
Tb/next test or not

### LINKED ISSUE #1848: Reduce CI runtime
Currently the "test" run takes almost 40 minutes. We should reduce this. Below are some ideas.

First collect information on how long all the required CI jobs take.

I think it would be good to split the up the test step into one step for building and one for running the tests:

https://github.com/EspressoSystems/espresso-sequencer/blob/f1270db9f82be0516790207044513893c47b4aef/.github/workflows/test.yml#L55-L57

This gives more visibility on which step takes how long.

Then it might make sense to get an idea on which tests are the slowest and see what we can do about it: for example, rewrite them, not run them on PRs by default. It would also be good if the CI made it easy to see which tests are slow. `nextest` would also take care of that.

If the time it takes to run tests is longer than the time it takes to build them we can consider splitting up the test job across github runners. `nextest` also makes this straightforward.

Another issue is that some tests are not well isolated. For this we can either fix those tests or use some combination of retries and timeouts to make the CI reliable enough. I think it should (falsely) fail in less than 1% of cases.

tbro:
Closes #1848 

Trying to bring down runtime of tests. Just gathering information currently.


.github/workflows/slowtest.yaml
```diff
@@ -0,0 +1,63 @@
...
+    tags:
+      # YYYYMMDD
+      - "20[0-9][0-9][0-1][0-9][0-3][0-9]*"
+  schedule:
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
sveitser, line 17: We probably shouldn't cancel the job on `main` so we would know which commit broke it: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches

I also think we can run it on PRs just not make it required for merging. I think if you add a new workflow it will be not required by default.

.github/workflows/test.yml
```diff
@@ -47,12 +44,21 @@ jobs:
...
         timeout-minutes: 90
+
+      - name: Test
+        env:
+          CARGO_TERM_COLOR: always
+        # Build test binary with `testing` feature, which requires `hotshot_example` config
+        run: |
+          export RUSTFLAGS="$RUSTFLAGS --cfg hotshot_example"
+          export PATH="$PWD/target/release:$PATH"
+          cargo nextest run --locked --release --workspace --all-features --retries 2 --verbose -E '!test(slow_)'
```
sveitser, line 63: I think it would be nice to add just recipies to mirror the new test commands locally since it's also nice to run the tests quickly locally. And sometimes we want to re-run the same tests as in a CI job locally.

sveitser (approved):
LGTM, some suggestions

.github/workflows/slowtest.yaml
sveitser: image

indeed doesn't seem to be required

.github/workflows/slowtest.yaml
```diff
@@ -0,0 +1,64 @@
+name: SlowEst
```
sveitser, line 1:
```suggestion
name: Slow Tests
```

.github/workflows/slowtest.yaml
```diff
@@ -0,0 +1,64 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  RUSTFLAGS: '--cfg async_executor_impl="async-std" --cfg async_channel_impl="async-std"'
+  RUST_LOG: info,libp2p=off,node=error
+
+jobs:
+  slowest:
```
sveitser, line 25: Or is "slowest" chosen on purpose?

tbro: Yes. I was it was a play on `nextest` plus these are in fact the slowest tests.

.github/workflows/test.yml
tbro: a7d14fdaa48aecfe46083c956658bf89dbcb4a85

.github/workflows/slowtest.yaml
tbro: 39d2e888b70e375be9727249f2dc46ea70617ae9 and 5a609dfc9d20643873302fe82f09a443ff710ac8

PR #1859, opened 2024-08-13
simplify interaction with multisig wallets in deployment scripts

### LINKED ISSUE #1838: simplify interaction with multisig wallets in deployment scripts
Simplify the deployment & upgrade scripts to use the multisig wallet only in these situations:
- executing the `upgradeToAndCall` method on the proxy to upgrade to a new implementation contract
-calling methods that are restricted to onlyAdmin e.g. `setPermissionedProver`, `disablePermissionedProverMode` etc
This assumes the multisig was set as the owner of the proxy as soon as it has been deployed.

Add the following checks
- the admin of the proxy is usable and indeed a multisig wallet

alysiahuggins:
Closes #1838 

### This PR:
- provides scripts to deploy the FeeContract, LightClient Contract and PlonkVerifier library without OpenZeppelin Defender and without a multisig wallet creating the transaction (thus via an EOA)
- provides scripts to upgrade the FeeContract and LightClient using an EOA to deploy the new implementation contracts and the Safe SDK to create a transaction on the Safe interface so that a Safe multisig wallet can be used to upgrade the proxy contracts as its the owner of these proxies
- consolidates scripts, improve comments across scripts and uses a stricter naming convention for scripts & their containing functions 

### Key places to review:
- updated README
- updated scripts
- the use of vm.ffi to interact with the Safe wallet (this was used since that interaction flow was already set up for a previous admin modification)

### How to test this PR: 
- Follow the instructions in the README to deploy the the FeeContract and LightClient proxies
- Follow the instructions in the README to upgrade the FeeContract and LightClient

### Things tested
- deploying the contracts mentioned above using the forge script command
- upgrading the respective proxies using the forge script command 

### Things not tested
- interacting with functions on the PlonkVerifier library once deployed


contracts/script/FeeContract.s.sol
```diff
@@ -0,0 +1,207 @@
...
+        vm.stopBroadcast();
+
+        return (payable(proxyAddress), implementationAddress);
+    }
+}
+
+/// @notice Upgrades the fee contract first by deploying the new implementation
+/// and then executing the upgrade via the Safe Multisig wallet using the SAFE SDK.
+contract UpgradeFeeContractScript is Script {
+    string internal originalContractName = "FeeContract.sol";
```
Ayiga, line 47: If we want to use this upgrade more than once, should we also make it so that the `originalContractName` can be derived from an environment variable?

alysiahuggins: thanks @Ayiga , i did that in this update  https://github.com/EspressoSystems/espresso-sequencer/commit/4ac6ca7cfd7492ad82a83177b50cb2a25a00f9fc however, the actual contract still needs to be imported into the script as we see here https://github.com/EspressoSystems/espresso-sequencer/blob/4ac6ca7cfd7492ad82a83177b50cb2a25a00f9fc/contracts/script/FeeContract.s.sol#L67 so there will still need to be changes to the script when the time comes

contracts/broadcast/FeeContract.s.sol/11155111/run-1723219248.json
```diff
@@ -0,0 +1,59 @@
+{
```
zacshowa, line 1: For my own understanding, Are these run files used by something in the repo like CI or tests?

If we do need to be committing them to the repo, what are we using them for? If not should we add the `/contracts/broadcast` directory to the .gitignore and remove ones that currently reside in the repo?

zacshowa (commented):
one comment / nit for my own understanding :smile:

contracts/broadcast/FeeContract.s.sol/11155111/run-1723219248.json
sveitser: The idea is to only commit the files for public deployments to the repo. It's to make life easy for anyone who wants to interact with the deployed contracts.

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
```diff
@@ -0,0 +1,157 @@
...
+  dotenv.config();
+
+  try {
+    const upgradeData: UpgradeData = processCommandLineArguments();
+
+    /**TODO
+     * change from SEPOLIA_RPC_URL to production URL when deploying to production
+     */
+    // Initialize web3 provider using the RPC URL from environment variables
+    const web3Provider = new ethers.JsonRpcProvider(getEnvVar("SEPOLIA_RPC_URL"));
```
Sneh1999, line 28: nit: could just name it `RPC_URL` then we can just change the RPC_URL in .env and no code changes needed

Sneh1999 (approved):

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
alysiahuggins: i like, will do

alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/391ce1ce4870422cba5ccea8303c15ec19eb554d

.env
```diff
@@ -88,6 +88,8 @@ ESPRESSO_SEQUENCER_STAKE_TABLE_CAPACITY=10
 # Foundry
 # The mnemonic used by foundry to deploy contracts.
 MNEMONIC="test test test test test test test test test test test junk"
+# The mnemonic offset, starts at zero
+MNEMONIC_OFFSET=0
 # The RPC URL for deploying to the sepolia network.
 SEPOLIA_RPC_URL=
```
Sneh1999, line 94: nit: should change this as well to `RPC_URL`

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
Sneh1999: Nice, added a few extra comments related to this

.env
alysiahuggins: Currently, the `SEPOLIA_RPC_URL` is specific to Sepolia gas profiling tests. I think we can consider network-labeled RPC URLs as we expand tests to other networks. If managing these becomes unwieldy, we can then consolidate to `RPC_URL`. What do you think?

Sneh1999: okay that makes sense! yea lets keep it `SEPOLIA_RPC_URL` here for now

Sneh1999 (approved):
LGTM!

Issue #1756, opened 2024-07-18
I05: Inefficient runtime calculation of domain elements and eval


philippecamacho:
**Affected Code:** PolynomialEval.sol (line 235-L258)

**Summary:** Currently, domain elements and eval domain are computed at runtime, which leads to unnecessary gas consumption. This approach is inefficient if the public input size and the domain size remain fixed.

**Suggestion:** For a given verifying key, we recommend computing the domain elements and eval domain in the constructor and storing it as immutable constants. Similar to the `getVk` function, we can add `getDomainElements` and `getEvalDomain` functions which can load these constants to the memory.

philippecamacho:
As discussed, add a rust script that generates the hard coded values for domain elements. Remove domain sizes that are not needed (all but 32 and 2^20).

philippecamacho:
Closed in #1860.

PR #1860, opened 2024-08-14
Attempt to reduce gas cost by precomputing domain elements

philippecamacho:
Related to #1756 

### This PR:
As discussed in https://github.com/EspressoSystems/espresso-sequencer/issues/1756#issuecomment-2284096087, the domain elements are precomputed. However the gas cost [slightly increases](https://github.com/EspressoSystems/espresso-sequencer/pull/1860/commits/9068eea5b687e8925b8dae5c6c4034a2073cbe23#diff-11538116afca6cc6a316ec7e5532f95ee5b404085a676b3d186ac5776ef99712)...


contracts/src/libraries/PolynomialEval.sol
```diff
@@ -14,7 +14,8 @@ library PolynomialEval {
     struct EvalDomain {
         uint256 logSize; // log_2(self.size)
         uint256 sizeInv; // Inverse of the size in the field
-        uint256 groupGen; // A generator of the subgroup
+        uint256 groupGen; // A generator g of the subgroup
+        uint256[8] domainElements; // 1, g, g^2, ..., g^7
```
alxiong, line 18: - we can remove `groupGen`
- we can reduce to only 7 and starts with `g` since 1 is the same, we don't have to store it everywhere.

contracts/src/libraries/PolynomialEval.sol
```diff
@@ -150,26 +165,7 @@ library PolynomialEval {
...
-            tmp := mulmod(tmp, groupGen, p)
-            mstore(add(localDomainElements, 0x80), tmp) // g^4
-            tmp := mulmod(tmp, groupGen, p)
-            mstore(add(localDomainElements, 0xa0), tmp) // g^5
-            tmp := mulmod(tmp, groupGen, p)
-            mstore(add(localDomainElements, 0xc0), tmp) // g^6
-            tmp := mulmod(tmp, groupGen, p)
-            mstore(add(localDomainElements, 0xe0), tmp) // g^7
-        }
+        uint256[8] memory localDomainElements = self.domainElements;
```
alxiong, line 168: this is extra memory copying, we can directly access them in the logic below via `self.domainElements` instead of via `localDomainElements`

contracts/src/libraries/PolynomialEval.sol
alxiong: > we can reduce to only 7 and starts with g since 1 is the same, we don't have to store it everywhere.

this is bad advice, because it complicates the code, affect readability. an extra uint256 in memory is no big deal.

alxiong: done in [5708d75](https://github.com/EspressoSystems/espresso-sequencer/pull/1860/commits/5708d754124d98b3f1472ae5919633cb8d7419a0)

contracts/src/libraries/PolynomialEval.sol
alxiong: done in [5708d75](https://github.com/EspressoSystems/espresso-sequencer/pull/1860/commits/5708d754124d98b3f1472ae5919633cb8d7419a0)

alxiong (approved):
I addressed my own comments. Ready to merge.

contracts/src/libraries/PolynomialEval.sol
philippecamacho: Thanks!

PR #1861, opened 2024-08-14
build speed

tbro:
Experiment using mold linker in CI


tbro:
closing. no apparent effect.

PR #1867, opened 2024-08-14
Update loggings to make sure builders are run correctly

shenkeyao:
Closes https://github.com/EspressoSystems/marketplace-builder-core/issues/42.

### This PR:
* Updates loggings to make sure we are running fallback and/or reserve builders correctly.

### This PR does not:
* Change any logic.

### How to test this PR:
* Run the two builders on demo-native and observe the different loggings.


imabdulbasit (approved):

QuentinI (approved):
LGTM! I think we can actually target main with this, as it doesn't touch process-compose

PR #1871, opened 2024-08-15
register rollups for native demo

imabdulbasit:


process-compose.yaml
```diff
@@ -662,3 +662,12 @@ processes:
     depends_on:
       sequencer1:
         condition: process_healthy
+
+  dev-rollup:
+    command:
+      dev-rollup register --ns 1; dev-rollup register --ns 2; dev-rollup register --ns 3
```
imabdulbasit, line 668: There might be a better name for the binary than dev-rollup. I'm also considering something like dev-rollup register --ns 1 --ns 2 --ns 3, but I'm not sure if there's a clean way to implement it and if that would be better, as each register also has other optional parameters. I might keep it as it is for now.

sequencer/src/bin/dev-rollup.rs
```diff
@@ -0,0 +1,216 @@
...
+        long,
+        env = "ESPRESSO_MARKETPLACE_RESERVE_BUILDER_URL",
+        default_value_t = Url::from_str("http://localhost").unwrap()
+    )]
+    pub reserve_url: Url,
+
+    #[clap(long, default_value_t = 200)]
+    pub reserve_price: u64,
+
+    #[clap(long, default_value_t = true)]
```
elliedavidson, line 52: Can we change this default to false for now, since we aren't testing the auction yet.  I want to make sure we are testing rollups using reserve sequencers and not participating in the auction.

elliedavidson (changes requested):
Super small change requested!   Otherwise LGTM!

elliedavidson (approved):

PR #1877, opened 2024-08-19
Tj/fee contract improvements

0xkato:
### This PR:
Minor gas optimizations

### DIFF
--- a/contracts/src/FeeContract.sol
+++ b/contracts/src/FeeContract.sol
@@ -21,9 +21,9 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     //
     /// @notice max amount allowed to be deposited to prevent fat finger errors
     // @TODO confirm this amount with product
-    uint256 public constant maxDepositAmount = 1 ether;
+    uint256 public maxDepositAmount = 1 ether;
 
-    uint256 public constant minDepositAmount = 0.001 ether;
+    uint256 public minDepositAmount = 0.001 ether;
 
     // === Errors ===
     //
--- a/contracts/src/FeeContract.sol
+++ b/contracts/src/FeeContract.sol
@@ -16,11 +16,11 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
 
     /// @notice upgrade event when the proxy updates the implementation it's pointing to
     event Upgrade(address implementation);
+
     // === Constants ===
     //
     /// @notice max amount allowed to be deposited to prevent fat finger errors
     // @TODO confirm this amount with product
-
     uint256 public constant maxDepositAmount = 1 ether;
 
     uint256 public constant minDepositAmount = 0.001 ether;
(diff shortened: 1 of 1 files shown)


contracts/src/FeeContract.sol
```diff
@@ -13,18 +13,17 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
-    event Log(string func, uint256 gas);
 
     /// @notice upgrade event when the proxy updates the implementation it's pointing to
     event Upgrade(address implementation);
+
     // === Constants ===
     //
     /// @notice max amount allowed to be deposited to prevent fat finger errors
     // @TODO confirm this amount with product
+    uint256 public constant maxDepositAmount = 1 ether;
```
philippecamacho, line 24: Can we change this value easily when upgrading? cc @alysiahuggins

philippecamacho (approved):
LGTM but please check with @alysiahuggins regarding the upgradeability of the contract.

contracts/src/FeeContract.sol
0xkato: Corrected. Those were not meant to be constant.

philippecamacho (changes requested):
Some tests fails locally on my machine:

```
> just sol-test

[FAIL. Reason: DepositTooLarge()] test_depositForManyDifferentUsers() (gas: 44817)
[FAIL. Reason: DepositTooLarge()] test_depositManyTimesForTheSameUser() (gas: 40448)
[PASS] test_depositMaxAmount() (gas: 27816)
[FAIL. Reason: panic: arithmetic underflow or overflow (0x11)] test_depositMinAmount() (gas: 16785)
[FAIL. Reason: Error != expected error: 0xc56d46d3 != 0x702b3d90] test_invalidUserAddress() (gas: 24845)
```

contracts/src/FeeContract.sol
```diff
@@ -13,18 +13,17 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
-    event Log(string func, uint256 gas);
 
     /// @notice upgrade event when the proxy updates the implementation it's pointing to
     event Upgrade(address implementation);
+
     // === Constants ===
     //
     /// @notice max amount allowed to be deposited to prevent fat finger errors
     // @TODO confirm this amount with product
+    uint256 public maxDepositAmount = 1 ether;
```
alysiahuggins, line 24: as it's an upgradable contract - we can't assign values to state variables on the implementation contract since we actually use the storage on the proxy contract. Only set these values in the `initialize` function

contracts/src/FeeContract.sol
```diff
@@ -49,8 +48,6 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
     function initialize(address multisig) public initializer {
         __Ownable_init(multisig); //sets owner to msg.sender
         __UUPSUpgradeable_init();
-        maxDepositAmount = 1 ether;
```
alysiahuggins, line 52: see comment above but these values have to set here as opposed to assigning it upon declaration of the state variable.

alysiahuggins:
> LGTM but please check with @alysiahuggins regarding the upgradeability of the contract.

thanks for tagging me @philippecamacho , i left comments

PR #1872, opened 2024-08-15
[WEEKLY RELEASE] HotShot - rc-0.5.71

rob-maron:

PR #1878, opened 2024-08-20
dynamic versioning

imabdulbasit:
This PR:

- Adds both base and upgrade versions to the genesis.toml file and deserializes them from string format into the Version struct.
- Initializes the base and upgrade versions in binaries  from the genesis.toml file.
- separates the API version from the protocol versions. The `SequencerApiVersion` is currently hardcoded as V0.1 for all APIs.
- Renames the `main` function to `run` in binaries that require a protocol version, run() accepts a generic parameter V that implements the Versions trait. This allows matching and initializing versions from the genesis.toml file using the appropriate type. This can be replaced with a macro
- adds message compat test for each version

Files to review:
all files mostly but primary focus on genesis.toml config, SequencerVersions type and main functions in all binaries.

test:
Run the demo native locally to ensure everything functions as expected.


builder/src/lib.rs
```diff
@@ -65,7 +65,7 @@ use sequencer::{
     context::{Consensus, SequencerContext},
     network,
     state_signature::{static_stake_table_commitment, StakeTableCommitmentType, StateSigner},
-    L1Params, NetworkParams, Node,
+    L1Params, NetworkParams, Node, SequencerApiVersion,
```
tbro, line 68: Good name :)

tbro (commented):
Looks good overall. There is a little bit of duplication w/ #1872, so I'd like to look through one more time after that gets merged. Should be a smaller change set at that time.

types/src/v0/mod.rs
```diff
@@ -135,25 +137,39 @@ impl NodeType for SeqTypes {
...
+    pub fn new() -> Self {
+        Self {
+            _pd: Default::default(),
+        }
+    }
+}
 
-#[derive(Debug, Copy, Clone)]
-pub struct SequencerVersions;
+pub type BaseV01UpgradeV02 = SequencerVersions<StaticVersion<0, 1>, StaticVersion<0, 2>>;
```
tbro, line 153: Since we are defining aliases for these, it might be a little easier to keep in sync if we use the alias everywhere.

imabdulbasit: this is used for test network mostly

imabdulbasit: I think I should rename it to MockSequencerVersions

tbro (approved):

PR #1880, opened 2024-08-20
Tb/test marketplace upgrade

tbro:
Test marketplace upgrade

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -1531,7 +1531,7 @@ mod test {
 
         upgrades.insert(UpgradeVersion::VERSION, Upgrade { mode, upgrade_type });
 
-        let stop_voting_view = u64::MAX;
+        let stop_voting_view = 30u64;
 
         const NUM_NODES: usize = 5;
         let config = TestNetworkConfigBuilder::<NUM_NODES, _, _>::with_num_nodes()
@@ -1585,21 +1585,23 @@ mod test {
                 .send()
                 .await
                 .unwrap();
-
+            dbg!(&stop_voting_view);
+            dbg!(&height);
             for peer in &network.peers {
                 let state = peer.consensus().read().await.decided_state().await;
 
                 match state.chain_config.resolve() {
                     Some(cf) => {
-                        if cf != chain_config_upgrade && height as u64 > stop_voting_view {
-                            panic!("failed to upgrade chain config");
+                        dbg!(&cf);
+                        dbg!(&chain_config_upgrade);
+                        if height as u64 > stop_voting_view {
+                            assert_eq!(cf, chain_config_upgrade);
+                            break 'outer;
                         }
                     }
                     None => continue 'outer,
                 }
             }
-
-            break;
         }
 
         network.server.shut_down().await;
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -1531,8 +1531,6 @@ mod test {
 
         upgrades.insert(UpgradeVersion::VERSION, Upgrade { mode, upgrade_type });
 
-        let stop_voting_view = 30u64;
-
         const NUM_NODES: usize = 5;
         let config = TestNetworkConfigBuilder::<NUM_NODES, _, _>::with_num_nodes()
             .api_config(
@@ -1607,7 +1605,7 @@ mod test {
                     break;
                 }
             }
-            sleep(Duration::from_secs(1));
+            sleep(Duration::from_secs(1)).await;
         }
 
         network.server.shut_down().await;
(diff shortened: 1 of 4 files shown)


sequencer/src/api.rs
```diff
@@ -1575,21 +1611,23 @@ mod test {
...
                 .unwrap();
-
+            dbg!(&stop_voting_view);
+            dbg!(&height);
             for peer in &network.peers {
                 let state = peer.consensus().read().await.decided_state().await;
 
                 match state.chain_config.resolve() {
                     Some(cf) => {
-                        if cf != chain_config_upgrade && height as u64 > stop_voting_view {
```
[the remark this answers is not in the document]
imabdulbasit: isn't the break for the outer loop? . The idea is that when the fee upgrade is complete, we check that the chain config in the validated state is updated. It should always be present in this test, but in reality, if it is not present, it will catch up automatically. If the config is not updated*and the block height is greater than the stop proposing height, we panic.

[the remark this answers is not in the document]
imabdulbasit: I think the problem is with stop_voting_view. It needs to be updated to take it from the upgrade mode parameter that we are passing. So for the view, it should be height > stop_proposing_view 
and for time it should be current_time > stop_proposing_time. I forgot to update it when we introduced time based parameters.

types/src/v0/v0_1/instance_state.rs
```diff
@@ -8,7 +8,7 @@ use crate::{v0_3::ChainConfig, Timestamp};
 #[serde(rename_all = "snake_case")]
 pub enum UpgradeType {
     Fee { chain_config: ChainConfig },
-    Marketplace {},
+    Marketplace { chain_config: ChainConfig },
```
imabdulbasit, line 11: why are we passing chain config here

imabdulbasit: oh nvm I think its `bid_recipient`

sequencer/src/api.rs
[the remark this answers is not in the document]
imabdulbasit: Also, we check chain config upgrade for each node in the network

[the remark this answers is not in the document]
tbro: Yes `break` is for outer loop. As soon as we iterate through the first set of peers the loop exits. I think we need to `break` on a successful match of the two chain_configs like in 97b377d38f074ce49c239d840f1c025b897316e6. I'm not sure just yet how to test all the peers.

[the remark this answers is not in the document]
tbro: I think 5050a85787a01c3ea02da5733f8ddad7e86158bf is what we discused.

sequencer/src/api.rs
```diff
@@ -1575,24 +1611,32 @@ mod test {
...
-
             for peer in &network.peers {
                 let state = peer.consensus().read().await.decided_state().await;
 
                 match state.chain_config.resolve() {
                     Some(cf) => {
-                        if cf != chain_config_upgrade && height as u64 > stop_voting_view {
-                            panic!("failed to upgrade chain config");
+                        // Fail test if we've waited long enough
+                        if height as u64 > stop_voting_view {
```
imabdulbasit, line 1620: this should be inside the if statement  `if cf != chain_config_upgrade`

tbro: I think If we exceed the number of configured views, we should fail the test. If chain_config gets upgraded *after* that number, still a failure. No?

types/src/v0/v0_1/instance_state.rs
tbro: humm but I guess we should add header in there as well

sequencer/src/api.rs
imabdulbasit: I am not sure if this is an accurate way to determine that chain config upgraded after that block height. The api call may take some time ?

tbro: On the other hand I'm thinking we should avoid writing such mind bending tests. I think part of the reason we didn't catch the bug in the test is that it was so hard to understand. So maybe we can think of going in a different direction.

sequencer/src/genesis.rs
```diff
@@ -601,13 +601,20 @@ mod test {
...
             hash = "0x80f5dd11f2bdda2814cb1ad94ef30a47de02cf28ad68c89e104c00c4e51bb7a5"
 
-
             [[upgrade]]
             version = "0.3"
             start_proposing_view = 1
             stop_proposing_view = 10
 
             [upgrade.marketplace]
+            [upgrade.marketplace.chain_config]
```
tbro, line 610: marketplace genesis upgrade

tbro:
superseded by #1927

PR #1889, opened 2024-08-22
fee upgrade fixes

tbro:
Fee upgrade tests in the current form do not properly test the upgrade. See [this thread](https://github.com/EspressoSystems/espresso-sequencer/pull/1880/files#r1722867590). 

### This PR:
This PR fixes the bug and the tests and make some attempt at making the test more easy to follow (in order to make it easier to spot bugs).

### Key places to review:
  * espresso-sequencer/sequencer/src/api.rs 

### How to test this PR: 
Convince yourself that 
  * the test will only succeed if chain_config is upgraded. 
  * it will fail if chain_config is not upgraded

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -1549,14 +1549,14 @@ mod test {
 
             // ChainConfigs will eventually be resolved
             if let Some(configs) = configs {
-                if height > new_version_first_view {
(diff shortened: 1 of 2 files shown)


imabdulbasit (approved):
LGTM! thanks

PR #1891, opened 2024-08-22
make reserve_url optional for rollup registration

imabdulbasit:


marketplace-solver/src/state.rs
```diff
@@ -271,7 +270,14 @@ impl UpdateSolverState for GlobalState {
             Vec::new(),
             rollups
                 .into_iter()
-                .map(|r| (r.body.namespace_id, r.body.reserve_url))
+                .map(|r| {
+                    (
+                        r.body.namespace_id,
+                        r.body
```
imabdulbasit, line 276: I think `SolverAuctionResults.reserve_bids` should be `Vec<(NamespaceId, Option<Url>)>,`

imabdulbasit: cc @elliedavidson

PR #1898, opened 2024-08-23
update toml files.

imabdulbasit:


data/genesis/cappuccino.toml
```diff
@@ -1,3 +1,6 @@
+base_version = "0.1"
```
imabdulbasit, line 1: I might add a versions table and include the base and upgrade field in it in future.

Ancient123 (approved):

PR #1900, opened 2024-08-23
add versions section to toml file

imabdulbasit:
This PR changes formatting of  the genesis.toml file by introducing a `versions` section, moving the `base_version` and `upgrade_version` fields into `base` and `upgrade` fields under this new section.
Testing

To test, simply run the native demo, as it uses the `demo.toml` file .


tbro:
I think it's fine, but I don't really see much of a difference. Can you comment on the reason for the change?

imabdulbasit:
> I think it's fine, but I don't really see much of a difference. Can you comment on the reason for the change?



> I think it's fine, but I don't really see much of a difference. Can you comment on the reason for the change?

Everything in toml file has a section so I decided to create a separate section for versions

tbro (approved):

imabdulbasit:
Going to hold it for now as this changes the genesis configuration. cc @Ancient123

imabdulbasit:
think this is not important, and don't want to change toml file schema now.

PR #1903, opened 2024-08-24
Marketplace builder in tests

QuentinI:
Closes #<ISSUE_NUMBER>

### This PR:
- Adds marketplace builder based on marketplace-builder-core to be run in `TestNetwork`
- Adds config option to specify the builder's URL
- Builder is started if we either start with or upgrade to `MarketplaceVersion`
- Changes builder API hooks not to block transaction submission
- As a drive-by, fixes solver client version in auction implementation

### This PR does not:
- Add any explicit tests targeting marketplace behaviour

### DIFF
--- a/marketplace-builder/Cargo.toml
+++ b/marketplace-builder/Cargo.toml
@@ -11,9 +11,9 @@ libp2p = ["sequencer/libp2p"]
 
 [dependencies]
 anyhow = { workspace = true }
-async-broadcast = "0.7.0"
+async-broadcast = { workspace = true }
 async-compatibility-layer = { workspace = true }
-async-lock = "3.4.0"
+async-lock = "2.2"
 async-std = { workspace = true }
 async-trait = { workspace = true }
 clap = { workspace = true }
--- a/marketplace-builder/src/builder.rs
+++ b/marketplace-builder/src/builder.rs
@@ -115,11 +115,11 @@ impl BuilderConfig {
         // spawn the builder service
         tracing::info!("Running builder against hotshot events API at {events_api_url}",);
 
-        let stream =
-            marketplace_builder_core::utils::EventServiceStream::<SeqTypes, V0_1>::connect(
-                events_api_url,
-            )
-            .await?;
+        let stream = marketplace_builder_core::utils::EventServiceStream::<
+            SeqTypes,
+            SequencerApiVersion,
+        >::connect(events_api_url)
+        .await?;
 
         async_spawn(async move {
             let res = run_builder_service::<SeqTypes>(hooks, senders, stream).await;
--- a/marketplace-builder/src/builder.rs
+++ b/marketplace-builder/src/builder.rs
@@ -11,8 +11,9 @@ use async_compatibility_layer::{
 use async_lock::RwLock;
 use async_std::sync::Arc;
 use espresso_types::{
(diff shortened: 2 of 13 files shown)


marketplace-builder/Cargo.toml
```diff
@@ -11,9 +11,9 @@ libp2p = ["sequencer/libp2p"]
 
 [dependencies]
 anyhow = { workspace = true }
-async-broadcast = "0.7.0"
+async-broadcast = { workspace = true }
 async-compatibility-layer = { workspace = true }
-async-lock = "3.4.0"
+async-lock = "2.2"
```
imabdulbasit, line 16: Why don't we update async-lock to 3.4 in all deps?

sequencer/src/api.rs
```diff
@@ -524,11 +528,31 @@ pub mod test_helpers {
...
+                let (task, url) =
+                    run_test_builder::<{ NUM_NODES }>(cfg.network_config.builder_port()).await;
+                builder_tasks.push(task);
+                legacy_builder_url = url;
+            };
+
+            if <V as Versions>::Upgrade::VERSION >= MarketplaceVersion::VERSION {
+                println!("Running marketplace builder!");
+                let (task, url) = run_marketplace_builder::<{ NUM_NODES }>(
+                    cfg.network_config.builder_port(),
```
imabdulbasit, line 545: we can run both builders simultaneously so different ports?

sequencer/src/api.rs
```diff
@@ -524,11 +528,31 @@ pub mod test_helpers {
             bind_version: V,
         ) -> Self {
             let mut cfg = cfg;
-            let (builder_task, builder_url) =
-                run_test_builder::<{ NUM_NODES }>(cfg.network_config.builder_port()).await;
+            let mut builder_tasks = Vec::new();
+            let mut legacy_builder_url = "http://example.com".parse().unwrap();
+            let mut marketplace_builder_url = "http://example.com".parse().unwrap();
+
+            if <V as Versions>::Base::VERSION <= MarketplaceVersion::VERSION {
```
tbro, line 535: should this be just `<`, since we want to get marketplace builder when version `==` marketplace_version.

sequencer/src/api.rs
QuentinI: Yes, for example in case we upgrade from 0.2 -> 0.3

marketplace-builder/Cargo.toml
QuentinI: We can, but we should start with HotShot, otherwise we run into incompatibilities

sequencer/src/api.rs
QuentinI: Oh, I misinterpreted your comment. Fixed!

tbro (approved):

marketplace-builder/src/builder.rs
```diff
@@ -170,75 +212,51 @@ impl BuilderConfig {
...
-        );
-
-        // start the hotshot api service
-        run_builder_api_service(hotshot_builder_apis_url.clone(), proxy_global_state);
-
-        // spawn the builder service
-        let events_url = hotshot_events_api_url.clone();
-        tracing::info!("Running permissionless builder against hotshot events API at {events_url}",);
+        // Start builder event loop
+        builder_state.event_loop();
```
shenkeyao, line 216: Why do we no longer need the `async_spawn(async move { ... })` wrap?

PR #1908, opened 2024-08-26
Update hotshot-builder-core to fix deadlock

Ayiga:
### This PR:

Updates the dependency of `hotshot-builder-core` to fix a deadlock that would prevent the builder from generating new blocks.  This allows `demo-native` and `docker` demos to work and make progress with new blocks without stopping almost immediately.

### This PR does not:

Address any other currently known issues within the builder.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4141,8 +4141,8 @@ dependencies = [
 
 [[package]]
 name = "hotshot-builder-core"
-version = "0.1.39"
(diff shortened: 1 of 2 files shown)


imabdulbasit (commented):
can you also push the lock file?

Ayiga:
> can you also push the lock file?

Oops.  Missed that one.  Thanks for the heads up!

imabdulbasit (approved):

PR #1910, opened 2024-08-26
Fix namespace filtering for the reserve builder

### LINKED ISSUE #1909: Fix namespace filtering for the reserve builder
We are using the namespaces from the bid config for the reserve builder, rather than the registered namespace list from the solver server, which may cause incorrect transaction filtering.

Code: https://github.com/EspressoSystems/espresso-sequencer/blob/ag/integration/marketplace-builder/src/builder.rs#L199.

shenkeyao:
Closes #1909.

### This PR:
* Fixes transaction filtering for the reserve builder to only build and bid for namespaces that have registered with a reserve builder URL, rather than rely on the bid configuration.
* Replaces `BidConfig` with just the bid amount, not reading namespaces from the configuration.
* Combines registration fetching code for both reserve and fallback builders.
* Removes unused `namespaces` option and `ESPRESSO_MARKETPLACE_BUILDER_NAMESPACE`.


elliedavidson (changes requested):
I think we may want to implement this differently.  The reserve builders should take in a namespace from a config file.  We may have multiple reserve builders running at the same time that are building for different namespaces each.  The reserve builder should just know that it is a reserve builder for a particular namespace, and shouldn't need to contact the solver.   In this case, what if we just changed the namespace that the example rollups register for in demo-native so that they match the namespace in the builder bid config?

shenkeyao:
> I think we may want to implement this differently. The reserve builders should take in a namespace from a config file. We may have multiple reserve builders running at the same time that are building for different namespaces each. The reserve builder should just know that it is a reserve builder for a particular namespace, and shouldn't need to contact the solver. In this case, what if we just changed the namespace that the example rollups register for in demo-native so that they match the namespace in the builder bid config?

As [discussed on Zulip](https://espresso.zulipchat.com/#narrow/stream/440749-Marketplace/topic/Fallback.20Builder.20Building.20for.20Correct.20Namespaces/near/465244599), the namespace change is done in the integration branch, which will be merged to main soon. I'm going to close this PR.

PR #1914, opened 2024-08-27
add marketplace.chain_config to demo.toml

tbro:
We need market place configuration for `demo_native` tests. This also updates the serialization unit tests.

### DIFF
--- a/data/genesis/demo.toml
+++ b/data/genesis/demo.toml
@@ -35,9 +35,9 @@ stop_proposing_view = 15
 
 [upgrade.marketplace]
 [upgrade.marketplace.chain_config]
-chain_id = 12345
-max_block_size = 30000
-base_fee = 1
+chain_id = 999999999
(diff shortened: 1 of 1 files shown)


data/genesis/demo.toml
```diff
@@ -33,4 +33,11 @@ version = "0.3"
 start_proposing_view = 5
 stop_proposing_view = 15
 
-[upgrade.marketplace]
\ No newline at end of file
+[upgrade.marketplace]
+[upgrade.marketplace.chain_config]
+chain_id = 12345
```
imabdulbasit, line 38: can we please keep the chain id and the formatting consistent with the above chain configs

tbro: a81aa0f44645f47cfb2735851b2674bb0cefcdba

imabdulbasit (approved):

PR #1919, opened 2024-08-27
remove unused deps

tbro:
Removes unused dependencies from workspace members.


sequencer/Cargo.toml
```diff
@@ -112,3 +108,6 @@ normal = ["hotshot-testing"]
 unexpected_cfgs = { level = "warn", check-cfg = [
     'cfg(async_executor_impl, values("tokio", "async-std"))',
 ] }
+
+[package.metadata.cargo-machete]
```
imabdulbasit, line 112: are we using cargo machete in CI?

imabdulbasit: should we remove the `[package.metadata.cargo-udeps.ignore]` from the toml file then? and also replace  `udeps` with `cargo-machete` in CI

imabdulbasit (approved):

PR #1926, opened 2024-08-28
Only store voted view on vote and propose

### LINKED ISSUE #1932: Quick fix for view sync after restart
As a temporary, faster fix, we can simply not update our persisted view on actions that could push our node ahead of the quorum, such as DA actions. But this leads to a very unlikely vulnerability where we vote for DA in one view, shut down, restart, and vote again in the same view, which is Byzantine behavior. Since this is highly unlikely in practice, this is probably a fine temporary fix.

bfish713:
Closes #1932 

### This PR:

This is a temporary fix which prevents nodes from restarting in different views such that no view sync certificate can ever be formed.  It introduces the possibility for nodes to vote twice in a view for a DA proposal, but this case is very unlikely.  Will follow up with a fix, see: #1931

### DIFF
--- a/sequencer/src/persistence/sql.rs
+++ b/sequencer/src/persistence/sql.rs
@@ -557,6 +557,7 @@ impl SequencerPersistence for Persistence {
         view: ViewNumber,
         action: HotShotAction,
     ) -> anyhow::Result<()> {
+        // Todo Remove this after https://github.com/EspressoSystems/espresso-sequencer/issues/1931
         if !matches!(action, HotShotAction::Propose | HotShotAction::Vote) {
             return Ok(());
         }
--- a/sequencer/src/persistence/sql.rs
+++ b/sequencer/src/persistence/sql.rs
@@ -555,8 +555,11 @@ impl SequencerPersistence for Persistence {
     async fn record_action(
         &mut self,
         view: ViewNumber,
-        _action: HotShotAction,
+        action: HotShotAction,
     ) -> anyhow::Result<()> {
+        if !matches!(action, HotShotAction::Propose | HotShotAction::Vote) {
+            return Ok(());
+        }
         let stmt = "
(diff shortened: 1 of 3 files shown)


types/src/v0/traits.rs
```diff
@@ -437,10 +437,7 @@ pub trait SequencerPersistence: Sized + Send + Sync + 'static {
         // between `highest_voted_view` and `leaf.view_number`. This prevents double votes from
         // starting in a view in which we had already voted before the restart, and prevents
         // unnecessary catchup from starting in a view earlier than the anchor leaf.
-        let mut view = max(highest_voted_view, leaf.view_number());
-        if view != ViewNumber::genesis() {
-            view += 1;
-        }
```
jbearer, line 443: Without this, we start in the view we last voted, not after it. So it seems like we are no longer preventing double votes either for consensus or DA. Just want to be sure we're ok with that as a temporary fix, and that we have something better in the pipeline.

sequencer/src/persistence/sql.rs
```diff
@@ -555,8 +555,11 @@ impl SequencerPersistence for Persistence {
     async fn record_action(
         &mut self,
         view: ViewNumber,
-        _action: HotShotAction,
+        action: HotShotAction,
     ) -> anyhow::Result<()> {
+        if !matches!(action, HotShotAction::Propose | HotShotAction::Vote) {
```
jbearer, line 560: Can we add a comment that this should be removed once we have #1931 , to prevent possible double voting across restarts?

types/src/v0/traits.rs
bfish713: Yeah it does introduce this, but it's ok for now, I don't think there is any exploit of the potential for double vote

jbearer (approved):

PR #1927, opened 2024-08-28
test marketplace upgrade

tbro:
Test marketplace upgrade

supersedes #1880


sequencer/src/lib.rs
```diff
@@ -641,18 +641,15 @@ pub mod testing {
         }
 
         pub fn upgrades(mut self, upgrades: BTreeMap<Version, Upgrade>) -> Self {
+            for key in upgrades.keys() {
```
imabdulbasit, line 644: so we are overwriting config parameters if there are multiple upgrades. Why don't we pass the Versions generic parameter and get the upgrade version from there

tbro: I was mainly trying to avoid updating every use of `TestConfigBuilder`. But I can just add it to the `upgrades` method like so: 8e602bafe590a14ce2542d23eb684a84ac618d47

tbro: Side note, I don't believe we ever have a case where there is more than one upgrade in the `upgrades` map.

imabdulbasit: yes, I was thinking to remove this BTreeMap and cleanup some stuff but maybe in a separate PR

imabdulbasit (approved):
LGTM! I think we should extend both the fee and marketplace upgrade tests to check that we are producing blocks containing transactions after the upgrade. 

https://github.com/EspressoSystems/espresso-sequencer/issues/1996

PR #1934, opened 2024-08-28
remove epoch related code from light client contract

### LINKED ISSUE #1928: simplify light client contract: remove epoch related code
subtask of https://github.com/EspressoSystems/espresso-sequencer/issues/1912

- remove `blocksPerEpoch` constructor/initializer argument
- remove `nextEpoch` computation and checks
- remove `_advanceEpoch()`
- update related solidity tests, rust tests and deployment code

alysiahuggins:
Closes #1928 

### This PR:
Simplifies the Light Client contract by removing any functionality related to epochs since epochs have not been implemented yet. Code Changes:
- Removes the following from the LightClient contract and related code
  - the `blocksPerEpoch` variable
  - removes `currentEpoch` variable
- Removes the following from the StakeTable contract
  - `currentEpoch()`
  - removes the tests as the stake table isn't used and the current version heavily depends on epochs which have not been implemented yet
- Update the diff-test code
 

### This PR does not:
- remove epoch the StakeTable contracts as they are not used, instead the relevant tests are ignored when running `just sol-test`

### Key places to review:
- Updated Tests
- Implementation Code
- Deployment Scripts
- Modified Rust Scripts


contracts/src/StakeTable.sol
```diff
@@ -90,10 +90,11 @@ contract StakeTable is AbstractStakeTable {
         return keccak256(abi.encode(blsVK.x0, blsVK.x1, blsVK.y0, blsVK.y1));
     }
 
+    /// TODO handle this logic more appropriately when epochs are re-introduced
     /// @dev Fetches the current epoch from the light client contract.
     /// @return current epoch (computed from the current block)
-    function currentEpoch() public view returns (uint64) {
-        return lightClient.currentEpoch();
+    function currentEpoch() public pure returns (uint64) {
```
philippecamacho, line 96: Why do we need to keep the function `currentEpoch()`?

contracts/test/LightClient.t.sol
```diff
@@ -21,7 +21,7 @@ contract LightClientCommonTest is Test {
     uint32 public constant BLOCKS_PER_EPOCH_TEST = 3;
     uint32 public constant DELAY_THRESHOLD = 6;
     uint32 public constant MAX_HISTORY_SECONDS = 1 days;
-    uint32 initialEpoch = 1 days;
+    uint32 public initialEpoch = 1 days;
```
philippecamacho, line 24: Why do we need to keep this variable?

contracts/src/StakeTable.sol
alysiahuggins: at the time i made the change, i wanted to keep the related tests with as little breaking changes as possible - but now that i removed the stateTable tests, i can remove that function

contracts/test/LightClient.t.sol
alysiahuggins: ah this epoch isn't used for testing the block per epoch functionality but for testing the retention period for storing the state history on the contract

philippecamacho: I see, should we change the name to avoid confusion?

contracts/src/StakeTable.sol
alxiong: i also agree to keep this, a dummy logic is perfect here. Plus we are not using or auditing this contract anyway.

contracts/test/LightClient.t.sol
```diff
@@ -369,45 +353,41 @@ contract LightClient_newFinalizedState_Test is LightClientCommonTest {
...
+            //                // first block of a new epoch (from epoch 2) should update the
+            // following
+            //                bytes32 genesisComm = lc.computeStakeTableComm(genesis);
+            //                LC.LightClientState memory lastBlockInFirstEpoch = states[i - 1];
+            //                bytes32 firstEpochComm =
+            // lc.computeStakeTableComm(lastBlockInFirstEpoch);
+            //                assertEq(lc.votingStakeTableCommitment(), genesisComm);
+            //                assertEq(lc.frozenStakeTableCommitment(), firstEpochComm);
+            //                assertEq(lc.votingThreshold(), genesis.threshold);
+            //                assertEq(lc.frozenThreshold(), lastBlockInFirstEpoch.threshold);
```
alxiong, line 366: let's just safely remove these, can always add back or find these through git history. let's keep the testing logic clean.

contracts/test/LightClient.t.sol
```diff
@@ -369,45 +353,41 @@ contract LightClient_newFinalizedState_Test is LightClientCommonTest {
...
     /// @dev Test happy path for updating after skipping a few blocks (but not an epoch)
     /// forge-config: default.fuzz.runs = 4
     /// forge-config: quick.fuzz.runs = 1
     /// forge-config: ci.fuzz.runs = 10
-    function test_UpdateAfterSkippedBlocks(uint32 numBlockSkipped, uint32 numBlockPerEpoch)
-        external
-    {
-        numBlockPerEpoch = uint32(bound(numBlockPerEpoch, 2, 10));
-        numBlockSkipped = uint32(bound(numBlockSkipped, 1, numBlockPerEpoch - 1));
+    function test_UpdateAfterSkippedBlocks(uint32 numBlockSkipped) external {
```
alxiong, line 381: update the `@dev` code comment, remove the "(but not an epoch)" part.

hotshot-state-prover/src/mock_ledger.rs
```diff
@@ -48,6 +48,9 @@ type SchnorrSignKey = jf_signature::schnorr::SignKey<ark_ed_on_bn254::Fr>;
 /// Stake table capacity used for testing
 pub const STAKE_TABLE_CAPACITY: usize = 10;
 
+/// Number of blocks per epoch for testing
+pub const BLOCKS_PER_EPOCH: u32 = 3;
+
```
alxiong, line 53: can we set to `u32::MAX` for now?

contracts/test/StakeTable.t.sol
alxiong, line 1: instead of deleting tests for stake table, can't we just ignore them in `foundry.toml`?

alxiong (commented):
Thanks for the work! ❤️

hotshot-state-prover/src/mock_ledger.rs
alysiahuggins: since the blocks_per_epoch is used to determine how many states to generate, which is then used in the tests, I would say we need to choose a small number. I would opt for 5 though

contracts/test/LightClient.t.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/5acfd9472fc649eb6c8d6c76507789a318e55a99

hotshot-state-prover/src/mock_ledger.rs
alxiong: okay, sounds good.

contracts/test/LightClient.t.sol
alysiahuggins: done here https://github.com/EspressoSystems/espresso-sequencer/commit/5acfd9472fc649eb6c8d6c76507789a318e55a99

contracts/test/StakeTable.t.sol
alysiahuggins: addressed here https://github.com/EspressoSystems/espresso-sequencer/commit/3f6ecc2a206ebc21619a346c38dbb00486bc1184

alysiahuggins:
> LGTM, one comment is: instead of forge test --no-match-contract, can we edit foundry.toml to ignore? Because locally I won't have to remember to add that flag to forge commands all the time.
> 
> Thanks!

I agree, however, `ignored_warnings_from` in the `foundry.toml` didn't work. https://github.com/foundry-rs/foundry/issues/4414

philippecamacho (approved):
LGTM

PR #1936, opened 2024-08-28
Fix namespace filtering for the fallback builder

### LINKED ISSUE #1921: Fix namespace filtering for the fallback builder
1. For the fallback builder, when namespaces_to_skip is None, we just [process an empty transaction list](https://github.com/EspressoSystems/espresso-sequencer/blob/main/marketplace-builder/src/hooks.rs#L144).

  * I fixed this locally by reverting the logic, i.e., process all transactions if there's no namespace to skip.
2. We use connect_to_server to [connect to the server](https://github.com/EspressoSystems/espresso-sequencer/blob/main/marketplace-builder/src/hooks.rs#L164).

  * This doesn't seem to work anymore, but could be easily fixed by using surf_disco::Client::<...>::new instead.
3. We construct a fallback builder with [no skipped namespace](https://github.com/EspressoSystems/espresso-sequencer/blob/main/marketplace-builder/src/builder.rs#L240), and we only query the solver to [update the namespace list](https://github.com/EspressoSystems/espresso-sequencer/blob/main/marketplace-builder/src/hooks.rs#L157) every 20 views.

  * This itself isn't too bad, we just won't skip any namespace in the first 20 views. However, because we have issue 1 as well, we end up skipping all transactions in the first 20 views.
There are several ways to fix this, e.g., fetching the solver when constructing the fallback builder, or fetching it every time before processing transactions. I chose the former locally, but can switch to a different approach if needed.

(Zulip discussion starts from https://espresso.zulipchat.com/#narrow/stream/419378-Block-Builder/topic/Reserve.20and.20fallback.20builders/near/465593591.)

shenkeyao:
Closes #1921.

### This PR:
* Fetches the namespaces from the solver when constructing the fallback builder, not relying on the next fetch on view 20.
* Adds a helper function to connect to the solver and get the registration info.

### This PR does not:
* Fix the API inconsistency between real and mock solvers, as mentioned in the TODO comment.
  * Issue: https://github.com/EspressoSystems/espresso-sequencer/issues/1935.

### DIFF
--- a/marketplace-builder/src/builder.rs
+++ b/marketplace-builder/src/builder.rs
@@ -52,7 +52,9 @@ use surf_disco::Client;
 use tide_disco::{app, method::ReadState, App, Url};
 use vbs::version::{StaticVersion, StaticVersionType};
 
-use crate::hooks::{self, BidConfig, EspressoFallbackHooks, EspressoReserveHooks};
+use crate::hooks::{
+    self, fetch_namespaces_to_skip, BidConfig, EspressoFallbackHooks, EspressoReserveHooks,
+};
 
 #[derive(Clone, Debug)]
 pub struct BuilderConfig {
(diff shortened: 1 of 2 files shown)


QuentinI (approved):
LGTM!

PR #1940, opened 2024-08-29
perf: avoid carrying around commScalars and commBases

### LINKED ISSUE #1737: L05: Suboptimal verifier code due to missing MSM implementation

alxiong:
Closes #1737

### This PR:

- Instead of carrying around two arrays of 30 elements: `commScalars` and `commBases`, we continuously calculate the intermediate MSM results, which are used in the final pairing check
- merge `_preparePcsInfo()` and `_verifyPcsOpening()` into a single function `_verify()` which basically prepares the elements `[A]1, [B]1` to do pairing check
  - removed `struct PcsInfo` as we don't need to carry those
  - inline `_prepareEvaluations()` to `_preparePolyCommitments()` since the computation are both using the last 10 entries of the original 30-item arrays, so we avoid recomputing the `vPow`, powers of `v` in Step 10 & 11 of Plonk verifier

This change is on the original PV, not new PV2 (introduced in #1844), thus not a complete rewrite, easier to audit the diff. Surprisingly, the gas is even lower than our PV2.


contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -134,37 +113,57 @@ library PlonkVerifier {
...
-        res = PcsInfo(
-            chal.u, zeta, zetaOmega, eval, commScalars, commBases, proof.zeta, proof.zetaOmega
-        );
+        BN254.G2Point memory betaH = BN254.G2Point({
+            x0: BN254.BaseField.wrap(BETA_H_X1),
+            x1: BN254.BaseField.wrap(BETA_H_X0),
+            y0: BN254.BaseField.wrap(BETA_H_Y1),
+            y1: BN254.BaseField.wrap(BETA_H_Y0)
+        });
+        // Check e(A, [x]2) ?= e(B, [1]2)
```
philippecamacho, line 164:
```suggestion
        // Check e(A, [x]2) =?= e(B, [1]2)
```

philippecamacho (approved):
A typographic suggestion. LGTM.

contracts/src/libraries/PlonkVerifier.sol
alxiong: done in [025cd8a](https://github.com/EspressoSystems/espresso-sequencer/pull/1940/commits/025cd8aae06b9fd9771bfce047517560626aa31f)

PR #1941, opened 2024-08-29
reset solver storage

imabdulbasit:
This PR
- adds solver reset subommand to reset_storage
- combines pubkey, keygen and reset_storage binaries into a single bianry:  `utils`


sequencer/src/bin/reset-storage.rs
```diff
@@ -21,6 +21,13 @@ struct Options {
 
 #[derive(Clone, Debug, Subcommand)]
 enum Command {
```
tbro, line 23: Do we really need another binary for this? Can we just add the `reset` option to the binary that is using the db, as w/ solver?

imabdulbasit: I think its easy for devops engineers to reset using this cli. I am not sure how we can add this reset functionality in sequencer and solver binary. cc @jbearer

imabdulbasit: If we are not resetting the storage while running the sequencer then maybe we can remove this binary?

tbro: I think sequencer could just have a --reset-db switch, and the only thing it would do is reset the db and exit.

imabdulbasit: yeah we can do that. but lets keep it for now to unblock the deployment and we can remove it later if everyone agrees

tbro: I'm concerned about the amount of binaries we currently have. I don't think we need this one. And I think deployment will get unblocked in the same amount of time whether this is a switch on sequencer or if it is a stand alone binary.

jbearer: I think it's a matter of personal preference. I like small and self-contained binaries that have a single responsibility. Of course a reset switch on the sequencer that resets and exits would also work, but if all it does is reset and exit, without running the actual sequencer logic, one might wonder why it is in the same binary as the sequencer.

Only really crucial thing here, though, is that if we make this a switch in the sequencer, it must exit after resetting. We used to have this as just a startup option on the sequencer, where it would reset the storage and then run the sequencer. This is bad because if the container then restarts for any reason, it will have the same config and thus will reset storage again unintentionally

tbro: Agree that there is no hard and fast rule. I wouldn't be as concerned if there weren't already so many binaries.  This topic has been brought up in other conversations, so I think it is fair to begin to request that new binaries have sufficient justification.  I think it is fairly common-place to provide a way for a state-full application to clean up after itself. To me it seems strange to require someone to download a second binary to do this sort of house keeping. Also agree that it should exit after wipe.

imabdulbasit: Maybe we can combine keygen, pub-key, and reset-storage into a single binary `utils`

elliedavidson: A binary is better for DevOps than a switch in the sequencer.  It's much easier to track and deploy, so we should keep the binaries, but it is fine to combine them into a single binary as Abdul said above.

elliedavidson (approved):

marketplace-solver/src/database.rs
```diff
@@ -72,6 +73,17 @@ impl PostgresClient {
...
+        if reset {
+            sqlx::query("DROP SCHEMA public CASCADE;")
+                .execute(&connection)
+                .await
+                .unwrap();
+            sqlx::query("CREATE SCHEMA public;")
+                .execute(&connection)
+                .await
+                .unwrap();
+        }
```
sveitser, line 85: Previous comments by Jeb and Tim were that we should exit if we do the reset. Currently it looks like we don't do that, but I could be wrong.

imabdulbasit: That was for the sequencer binary. In the utils binary, we just call connect(), which resets and returns a client, but we don't do anything after that.

sveitser (approved):

PR #1943, opened 2024-08-29
perf: reuse free mem for each chal

### LINKED ISSUE #1917: Address observations on Plonk Verifier fixes

alxiong:
Closes #1917 

This is the last comments in the above issue, and a follow-up of #1738.

### This PR:

Instead of continuously expanding the free memory space, with every challenge generation, we reuse the free memory. 

according to Jakov from CommonPrefix, this could leads to gas efficiency: 

> Yes, there is. The key reason lies in how memory allocation and expansion work. Initially, memory is allocated in chunks, and as more memory is allocated, the gas cost increases.
> 
> Given that the stored data is no longer needed after hashing, you can simply reuse the already allocated memory. This avoids unnecessary memory expansion and results in lower gas costs.

in our particular case, we didn't observe any improvement, but it's still a good advice to implement and adhere to in the future.

### DIFF
--- a/contracts/src/libraries/PlonkVerifier.sol
+++ b/contracts/src/libraries/PlonkVerifier.sol
@@ -286,11 +286,8 @@ library PlonkVerifier {
             {
                 mstore(statePtr, 0x0) // init state
                 // preimage len: state(0x20) + transcript(0x7c0)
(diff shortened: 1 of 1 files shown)


philippecamacho (approved):
LGTM

PR #1947, opened 2024-08-30
add endpoints for version information

### LINKED ISSUE #1888: Native demo test is not asserting that upgrades succeed
We had an issue where the native-demo test was successful but the upgrade failed. The native demo should have stricter assertions to prevent it from passing if an upgrade fails.

tbro:
Fixes #1888 

This PR adds endpoints relevant to version/upgrades. The intention is to use these in deme-native smoke-test to check that upgrade succeeded. It is also thought that the ability the confirm the version/config of a running node is generally useful. I'm not very happy with the names chosen in this PR, but we need to work around previous choices.


sveitser:
Nice. Wondering if we should call this something like "procotol version" in order to not confuse it with the software versions?

sequencer/api/node_state.toml
```diff
@@ -0,0 +1,15 @@
+[route.version]
+PATH = ["/current_version"]
+METHOD = "GET"
+DOC = "Get the current version for the node."
+
+[route.chain_config]
+PATH = ["/chain_config"]
```
imabdulbasit, line 7: shouldn't we return validated state chain config because node state's chain config is the base version's chain config

sequencer/src/api/endpoints.rs
```diff
@@ -215,6 +215,28 @@ where
...
+    _: ApiVer,
+) -> Result<Api<S, Error, ApiVer>>
+where
+    S: 'static + Send + Sync + ReadState,
+    S::State: Send + Sync + NodeStateDataSource,
+{
+    let toml = toml::from_str::<toml::Value>(include_str!("../../api/node_state.toml"))?;
+    let mut api = Api::<S, Error, ApiVer>::new(toml)?;
+    api.get("current_version", move |_, state| {
+        async move { Ok(state.get_node_state().await.current_version) }.boxed()
```
imabdulbasit, line 228: Sorry, I should named it better but this current version in the node state is actually the base version

imabdulbasit: I think it should be renamed to base version to avoid confusion. If we really need the current version of the node then it should be stored in validated state and retrieved from it

tbro: Oh yea I misunderstood! thanks for the clarification!

tbro: If I understand, we first need to add the version field to validated_state, since it is not currently present. Then we need to pass validated_state into `ApiState` somehow. Not quite as easy as I had hoped. Maybe we should just go w/ chain_config now, since we can tell by the values in chain_config what version we are on.

tbro: Nevermind, we have the same problem with chain_config :face_exhaling:

PR #1961, opened 2024-09-03
assert header version

### LINKED ISSUE #1888: Native demo test is not asserting that upgrades succeed

tbro:
Closes #1888

This PR adds a quick and dirty way to test that upgrade succeeded. I think we should do something better than this, but I think something more sophisticated will require some thought. Given all the output from smoke test, it isn't easy to tell whether the upgrade assertion is actually executing, but I think it should be.


tbro:
#1888 will be addressed here https://github.com/EspressoSystems/espresso-sequencer/issues/2053

tbro:
https://github.com/EspressoSystems/espresso-sequencer/issues/1888 will be addressed here https://github.com/EspressoSystems/espresso-sequencer/issues/2053

PR #1964, opened 2024-09-03
fix mp native demo

imabdulbasit:


process-compose-mp.yml
```diff
@@ -4,67 +4,111 @@ environment:
   - ESPRESSO_SEQUENCER_GENESIS_FILE=data/genesis/demo-marketplace.toml
   - ESPRESSO_BUILDER_GENESIS_FILE=data/genesis/demo-marketplace.toml
 processes:
+
+  marketplace-reserve-builder:
```
imabdulbasit, line 8: I had to move these here because the builders were disabled. seems like disabled:false setting is not working for overridden file

PR #1970, opened 2024-09-04
Test reserve and fallback builder transaction submission via private mempool

### LINKED ISSUE #1966: Test reserve builder transaction submission via private mempool
The builders should correctly filter the transactions they receive.

### LINKED ISSUE #1967: Test fallback builder transaction submission via private mempool
The builders should correctly filter the transactions they receive.

shenkeyao:
Closes #1966, closes #1967.

### This PR:
* Refactors the existing builder test into two tests for the reserve and fallback builders.
* Adds helper functions for testing setup, including URL picking, solver connection, and transaction submission.
* Modifies the use of the solver base URL to make it consistent in builder and solver code.

### Key places to review:
* `marketplace-builder/src/builder.rs`.

### How to test this PR: 
* Run the two tests.


marketplace-builder/src/builder.rs
```diff
@@ -221,7 +223,7 @@ impl BuilderConfig {
             let bid_config = bid_config.expect("Missing bid config for the reserve builder.");
             let hooks = Arc::new(hooks::EspressoReserveHooks {
                 namespaces: bid_config.namespaces.into_iter().collect(),
-                solver_api_url,
+                solver_base_url,
```
elliedavidson, line 226: Good change!  This name is clearer.

marketplace-builder/src/builder.rs
```diff
@@ -332,54 +360,79 @@ mod test {
...
-                    .hotshot_events(HotshotEvents {
-                        events_service_port: event_port,
-                    }),
-            )
-            .network_config(network_config)
-            .build();
-        let _network = TestNetwork::new(config, MockSequencerVersions::new()).await;
+        let private_key =
+            <BLSPubKey as SignatureKey>::PrivateKey::generate(&mut rand::thread_rng());
+        let signature_key = BLSPubKey::from_private(&private_key);
```
elliedavidson, line 388: It might just be how GitHub is displaying the diff, but these seems like the same logic that was at line 381? What is the purpose of the logic here?

marketplace-builder/src/builder.rs
```diff
@@ -332,54 +360,79 @@ mod test {
...
-        let _builder_config = init.await;
+        // Sign the registration body
+        let reg_signature = <SeqTypes as NodeType>::SignatureKey::sign(
+            &private_key,
+            reg_ns_1_body.commit().as_ref(),
+        )
+        .expect("failed to sign");
+
+        let reg_ns_1 = RollupRegistration {
+            body: reg_ns_1_body.clone(),
```
elliedavidson, line 411: Small nit: Does this clone need to be here?  It seems like reg_ns_1 can take ownership of reg_ns_body?

marketplace-builder/src/builder.rs
```diff
@@ -332,54 +360,79 @@ mod test {
...
+        (mock_solver, solver_base_url)
+    }
 
+    /// Connect to the builder.
+    async fn connect_to_builder(
+        urls: QueryEventAndBuilderUrls,
+    ) -> Client<ServerError, MarketplaceVersion> {
         // Wait for at least one empty block to be sequenced (after consensus starts VID).
-        let sequencer_client: Client<ServerError, SequencerApiVersion> = Client::new(query_api_url);
+        let sequencer_client: Client<ServerError, SequencerApiVersion> = Client::new(urls.0);
```
elliedavidson, line 435: Super small nit: can we add a comment explaining what urls.0 and urls.2 represent?  I know it's explained above in the type definition, but it would be helpful here as well.

elliedavidson: Or perhaps we can just define each of the 3 urls as their own type with a descriptive name?

elliedavidson (commented):
Just some small nits is all.

marketplace-builder/src/builder.rs
shenkeyao: `signature_keys` is the list of keys that can update the registration, while the `signature_key` without "s" is the key to sign this registration. Alternatively, I could just create a list of keys and use one of them as the `signature_key`, but I was using the same implementation as in the mock solver test.

Now that I think about this, this helper function could be helpful for both builder and solver tests, so I'm going to simplify the logic here and move the function to the solver crate.

marketplace-builder/src/builder.rs
shenkeyao: I'll give each URL a name, good suggestion!

marketplace-builder/src/builder.rs
shenkeyao: > Alternatively, I could just create a list of keys and use one of them as the `signature_key`, but I was using the same implementation as in the mock solver test.

I realized that this alternative approach wasn't really better than the current implementation, since we would need to keep track of the index in the list and make a copy of the private key associated with `signature_key`. The current implementation, though looking a bit redundant, actually does the minimal work needed. I'll just keep it as is unless you think it could be optimized. 😃

marketplace-builder/src/builder.rs
```diff
@@ -332,54 +368,83 @@ mod test {
...
+    ///
+    /// Returns the solver and its base URL.
+    async fn init_mock_solver_and_register_rollup() -> (MockSolver, Url) {
+        let mock_solver = MockSolver::init().await;
+        let solver_base_url = mock_solver.solver_url.clone();
+        let client = connect_to_solver(solver_base_url.clone());
+
+        // Create a list of signature keys for rollup registration data
+        let mut signature_keys = Vec::new();
+        for _ in 0..10 {
```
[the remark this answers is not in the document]
shenkeyao: They are different. Here 10 is just an arbitrary number we used to represent the number of keys for the registration data.

marketplace-builder/src/builder.rs
elliedavidson: Makes sense!

elliedavidson (approved):

PR #1983, opened 2024-09-05
Use the marketplace builder in dev node

ImJeremyHe:
Currently in this [dae071b](https://github.com/EspressoSystems/espresso-sequencer/pull/1983/commits/dae071bf41dccd58248c5019e5315b876baf6c81) we are referring to a branch not a tag from the `marketplace-builder-core`. Should replace this with the tag after [this PR](https://github.com/EspressoSystems/marketplace-builder-core/pull/123) has been merged.

### This PR:
- Specify the sequencer version to use the marketplace builder in dev node
- Fix the builder listener so that the builder API is accessible when running in a docker container


sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -188,7 +188,11 @@ async fn main() -> anyhow::Result<()> {
         .network_config(network_config)
         .build();
 
-    let network = TestNetwork::new(config, MockSequencerVersions::new()).await;
+    let network = TestNetwork::new(
+        config,
+        SequencerVersions::<SequencerApiVersion, MarketplaceVersion>::new(),
```
imabdulbasit, line 193: Do we want to run the dev node with the marketplace version or have an argument to run v1/fee/marketplace?

ImJeremyHe: I am not sure. dev-node is for the developers who want to integrate their rollups with espresso. If you think they have to use this function, then it will be necessary for us to add this function.

Cargo.toml
```diff
@@ -58,7 +58,7 @@ hotshot = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.74",
 # Hotshot imports
 hotshot-builder-api = { git = "https://github.com/EspressoSystems/HotShot.git", tag = "0.5.74" }
 hotshot-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.1.47" }
-marketplace-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.1.47" }
+marketplace-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", branch = "ab/update-module-path" }
```
sveitser, line 61: This should be updated before merge so we don't use a branch.

imabdulbasit: done in 2f636f46f22e5c1a80d633326ec8456f11fb4281

sequencer/src/bin/espresso-dev-node.rs
tbro: I think we should support any version. Marketplace will not be live for some time.

imabdulbasit: This was discussed in standup and I think we only need marketplace version here cc @sveitser

tbro: Ok. I think it limits the usefulness of devnode, but I also don't remember the conversation. I go ahead a approve and if we need it we can add versions later.

tbro (approved):

sequencer/src/bin/espresso-dev-node.rs
imabdulbasit: I do agree tho that we should have support for pre marketplace version so I'll take a look at it next week. I think this blocks some other work so should be merged

imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/issues/2033

PR #1984, opened 2024-09-05
refactor!: PlonkVerifier public input reduced from 8 to 4

alxiong:
### This PR:

As described by the title

### How to test this PR: 

Since there are many other pending changes that fail CI, to test the correctness of this change:
```
cargo build --release --bin diff-test
forge test --mc PlonkVerifier
```

all plonk related tests should pass


alxiong:
@philippecamacho can you work on updating this PR:

As mentioned in the meeting, instead of 4, we should reduce to 7 only. the change would be similar to mine. 

make sure that `forge test --mc PlonkVerifier` will all pass.

contracts/src/LightClient.sol
```diff
@@ -291,7 +291,7 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         IPlonkVerifier.VerifyingKey memory vk = VkLib.getVk();
 
         // Prepare the public input
-        uint256[8] memory publicInput;
+        uint256[7] memory publicInput;
```
alxiong, line 294: you didn't push more items to the public input, right now there is just 4 items.

reminder for us to edit this later.

contracts/test/mocks/LightClientMock.sol
```diff
@@ -35,7 +35,7 @@ contract LightClientMock is LC {
         IPlonkVerifier.VerifyingKey memory vk = VkLib.getVk();
 
         // Prepare the public input
-        uint256[8] memory publicInput;
+        uint256[7] memory publicInput;
```
alxiong, line 38: similarly here, the publicInputs only filled with 4 elements

alxiong (commented):
thanks for hte work! need to resync with base branch

alxiong:
I'll just merge into my branch on https://github.com/EspressoSystems/espresso-sequencer/pull/1986 and then fix issues and regenerate the VK contracts there

contracts/src/LightClient.sol
alxiong: fixed in [59243df](https://github.com/EspressoSystems/espresso-sequencer/pull/1986/commits/59243df5edec5d83aff09fe3d9ae35b95dad3dcf)

PR #1985, opened 2024-09-05
Lr/restart tests

lukaszrzasik:
Closes #<ISSUE_NUMBER>


.github/workflows/slowtest.yaml
```diff
@@ -22,7 +22,7 @@ env:
   RUST_LOG: info,libp2p=off,node=error
 
 jobs:
-  slow-tests:
+  slowest:
```
tbro, line 25: I think some unwanted change may have been pulled in here.

jbearer: Seems likely, I can revert this

tbro:
How is this test intended to be run? It is not linked into the module tree so things like `cargo test` will not find it.

tbro (approved):
There's a lot in there. Looks good as far as I can tell.

jbearer:
CI here is no more broken than on main, and I checked that the restart tests themselves are passing, so I'm gonna merge this

PR #1939, opened 2024-08-29
switch from states mapping to two state variables for the genesis and…

alysiahuggins:
… finalized state

Closes #1938 

### This PR:
- removes states mapping
- added global variable for genesisState and finalizedState
- WIP

PR #1986, opened 2024-09-05
update light client state prover

alxiong:
### This PR:

updates the `hotshot-state-prover` with the new split between the static `StakeTableState` and the dynamic `LightClientState`

Currently there are still errors in my code.


hotshot-state-prover/src/circuit.rs
```diff
@@ -476,9 +448,8 @@ mod tests {
             .is_err());
 
         // bad path: total weight doesn't meet the threshold
-        // bit vector with total weight 23
         let bad_bit_vec = [
-            true, true, true, true, true, false, false, true, false, false,
+            false, false, true, false, true, false, false, true, false, false,
```
alysiahuggins, line 452: can you explain this change a bit?

contracts/src/libraries/PlonkVerifier.sol
```diff
@@ -262,31 +261,30 @@ library PlonkVerifier {
...
             let wire1Ptr := mload(add(proof, 0x20))
-            mstore(add(dataPtr, 0x6c0), mload(wire1Ptr)) // wire1.x
-            mstore(add(dataPtr, 0x6e0), mload(add(wire1Ptr, 0x20))) // wire1.y
+            mstore(add(dataPtr, 0x6a0), mload(wire1Ptr)) // wire1.x
+            mstore(add(dataPtr, 0x6c0), mload(add(wire1Ptr, 0x20))) // wire1.y
             let wire2Ptr := mload(add(proof, 0x40))
-            mstore(add(dataPtr, 0x700), mload(wire2Ptr)) // wire2.x
-            mstore(add(dataPtr, 0x720), mload(add(wire2Ptr, 0x20))) // wire2.y
+            mstore(add(dataPtr, 0x6e0), mload(wire2Ptr)) // wire2.x
+            mstore(add(dataPtr, 0x700), mload(add(wire2Ptr, 0x20))) // wire2.y
```
alysiahuggins, line 274: can you also explain this value change?

hotshot-state-prover/src/circuit.rs
```diff
@@ -335,7 +301,10 @@ where
     circuit.enforce_true(sig_ver_result.0)?;
 
     circuit.finalize_for_arithmetization()?;
-    Ok((circuit, public_inputs.into()))
+    Ok((
+        circuit,
+        GenericPublicInput::new(lightclient_state.clone(), stake_table_state.clone()),
```
alysiahuggins, line 306: yes I was thinking of doing something like this 👏

hotshot-state-prover/src/mock_ledger.rs
```diff
@@ -201,25 +188,26 @@ impl MockLedger {
...
+    //     }
 
-        self.sync_stake_table(num_reg, num_exit);
-    }
+    //     self.sync_stake_table(num_reg, num_exit);
+    // }
 
     /// Return the light client state and proof of consensus on this finalized state
-    pub fn gen_state_proof(&mut self) -> (GenericPublicInput<F>, Proof, GenericStakeTableState<F>) {
+    pub fn gen_state_proof(&mut self) -> (GenericPublicInput<F>, Proof) {
```
alysiahuggins, line 210: love this

hotshot-state-prover/src/service.rs
```diff
@@ -590,11 +570,11 @@ mod test {
         Vec<(StateSignKey, StateVerKey)>,
         StakeTable<BLSPubKey, StateVerKey, CircuitField>,
     ) {
-        let pp = MockSystemParam::init(BLOCKS_PER_EPOCH);
+        let pp = MockSystemParam::init();
```
alysiahuggins, line 573: 👍

hotshot-state-prover/src/circuit.rs
alxiong: ah, the making sure the accumulated weight of the signers is lower than threshold. previously we set a higher threshold, now it's computed as 1/3 +1, so I let fewer nodes to vote.

contracts/src/libraries/PlonkVerifier.sol
alxiong: yes, so we removed a public input field (from 8 to 7), 
so all these values, which were laying out consecutively in the memory space, are moved up in their memory location by `0x20` (one field element is 32 bytes)

alxiong:
This marks the closure of the #1939.

Please "merge with commit" instead of squash merge here, lots of useful history.

alxiong:
@mrain can you review any changes under `hotshot-state-prover`? 
thanks!

alxiong:
ah, and I don't quite understand the gas increase for the PlonkVerifier, but we can investigate in another PR. 
I'll look at it over the weekend

cc @philippecamacho

mrain (approved):
state prover part LGTM

philippecamacho (approved):
LGTM

PR #1987, opened 2024-09-05
add test config file

tbro:
Move test options to profiles. This helps clean up the command line a little, and should add more flexibility for future changes. Also addes `--no-fail-fast` option to run all the tests even after one fails.

### DIFF
--- a/.github/workflows/test-demo-native-marketplace.yml
+++ b/.github/workflows/test-demo-native-marketplace.yml
@@ -10,7 +10,6 @@ on:
       - "20[0-9][0-9][0-1][0-9][0-3][0-9]*"
   schedule:
     - cron: "0 0 * * 1"
-  pull_request:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-demo-native-marketplace.yml
```diff
@@ -10,7 +10,6 @@ on:
       - "20[0-9][0-9][0-1][0-9][0-3][0-9]*"
   schedule:
     - cron: "0 0 * * 1"
-  pull_request:
```
imabdulbasit, line 13: why this change?

tbro: The thought was that we can just run marketplace test manually for now.

tbro: reverted 4e9f7428f7b2a5d7190bcd64b73ce301f7c77496

imabdulbasit (approved):

PR #1988, opened 2024-09-06
[WEEKLY RELEASE] HotShot 0.5.74

jparr721:
Closes #<ISSUE_NUMBER>


hotshot-state-prover/src/circuit.rs
```diff
@@ -68,10 +66,6 @@ impl LightClientStateVar {
                 circuit.create_public_variable(view_number_f)?,
                 circuit.create_public_variable(block_height_f)?,
                 circuit.create_public_variable(state.block_comm_root)?,
-                circuit.create_public_variable(state.fee_ledger_comm)?,
-                circuit.create_public_variable(state.stake_table_comm.0)?,
-                circuit.create_public_variable(state.stake_table_comm.1)?,
-                circuit.create_public_variable(state.stake_table_comm.2)?,
```
alxiong, line 74: these changes are wrong. 

we have correctly update these part in https://github.com/EspressoSystems/espresso-sequencer/pull/1939

should we consider pointing to latest tag of hotshot in that PR instead?
otherwise, we need to cherry-pick many changes from that PR.

sequencer/src/state_signature.rs
```diff
@@ -45,18 +45,14 @@ pub struct StateSigner<ApiVer: StaticVersionType> {
     /// The most recent light client state signatures
     signatures: RwLock<StateSignatureMemStorage>,
 
-    /// Commitment for current fixed stake table
-    stake_table_comm: StakeTableCommitmentType,
-
```
alxiong, line 50: indeed, we are not updating this anymore, but I also don't see any harm of keeping it (as a static member field of this struct

PR #1989, opened 2024-09-07
[WEEKLY RELEASE] HotShot 0.5.74

alxiong:
### This PR:

Another attempt to update hotshot version to `rc-0.5.74`, aims to supercede #1988

The plan is to merge into `1938-xxx` branch and merge into `main`, this commit would still appear in the `main` history thus easy to locate.


alxiong:
should we also try to replace `branch = "hotshot/rc-0.5.74"` with a new tag on all hotshot-family repo?

@jparr721

types/src/v0/traits.rs
```diff
@@ -467,6 +467,7 @@ pub trait SequencerPersistence: Sized + Send + Sync + 'static {
             state,
             validated_state,
             view,
+            highest_voted_view,
```
alxiong, line 470: Can we change v0 struct like this? @sveitser 

We need this to placate the tests, but not sure if we should edit the v0

imabdulbasit: I think its fine to change it since we don't define it at sequencer side and serialize it.

Cargo.toml
```diff
@@ -54,32 +54,32 @@ dotenvy = "0.15"
...
 # Hotshot imports
-hotshot-builder-api = { git = "https://github.com/EspressoSystems/HotShot.git", tag = "0.5.73" }
-hotshot-builder-core = { git = "https://github.com/EspressoSystems/hotshot-builder-core", tag = "0.1.46" }
-marketplace-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.0.9" }
-hotshot-events-service = { git = "https://github.com/EspressoSystems/hotshot-events-service.git", tag = "0.1.45" }
-hotshot-orchestrator = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.73" }
-hotshot-query-service = { git = "https://github.com/EspressoSystems/hotshot-query-service", tag = "0.1.55" }
-hotshot-stake-table = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.73" }
+hotshot-builder-api = { git = "https://github.com/EspressoSystems/HotShot.git", tag = "rc-0.5.74" }
+hotshot-builder-core = { git = "https://github.com/EspressoSystems/hotshot-builder-core", branch = "hotshot/rc-0.5.74" }
```
imabdulbasit, line 60: point to the new tag?

Ayiga (commented):
Are these contract-bindings artifacts meant to be committed?

Cargo.toml
alxiong: yes, we should. so tags are already out? I'll update then.

alxiong:
> Are these contract-bindings artifacts meant to be committed?

yes, for now. https://github.com/EspressoSystems/espresso-sequencer/issues/1068#issuecomment-1938660743

Cargo.toml
alxiong: done in [44cb934](https://github.com/EspressoSystems/espresso-sequencer/pull/1989/commits/44cb934c390d83592afcbf2f323397bf29536b4a)

PR #1991, opened 2024-09-09
Update to new transaction-oriented query service

jbearer:
This PR moves us to the new transaction-oriented version of the query service. By moving most operations away from the main data source into isolated transactions, we massively improve concurrency and eliminate bottlenecks due to exclusive locks on the API state.

This also prepares us to switch to the sqlx backend and then replace the file system persistence with sqlite.

### This PR:
* Updates the query service dependency
* Updates uses of the query service storage types to use transactions for all reads and writes
* Removes `RwLock` from the API state, since all operations are now immutable, with concurrency handled by the underlying storage medium
* Depends on https://github.com/EspressoSystems/hotshot-query-service/pull/674


sequencer/src/state.rs
```diff
@@ -130,43 +131,50 @@ async fn store_state_update(
...
     let parent_chain_config = parent_state.chain_config;
 
     let (state, delta) = compute_state_update(parent_state, instance, parent_leaf, proposed_leaf)
         .await
         .context("computing state update")?;
 
-    let mut storage = storage.write().await;
-    if let Err(err) = store_state_update(&mut *storage, proposed_leaf.height(), &state, delta).await
-    {
-        storage.revert().await;
```
imabdulbasit, line 149: Shouldn't we keep it in case the transaction fails to commit?

imabdulbasit: Oh I think it automatically reverts when the tx is dropped without being committed

imabdulbasit (approved):
LGTM

PR #1998, opened 2024-09-11
These limits seem far to conservite.

tbro:
I wonder if any thing explodes if I bump them up a little.


sveitser:
I vaguely recall the async std thread count limit is set to 1 to highlight any issues from accidentally blocking an entire thread for too long. @jbearer might remember

jparr721:
Closing due to being extremely stale/high number of conflicts. No longer using Async std, so this PR is no longer relevant.

PR #2000, opened 2024-09-12
check whether light client and fee contract is a proxy in relevant pl…

### LINKED ISSUE #1527: Sequencer nodes should do a sanity checks to check if the L1 contract addresses are wrong
Currently setting for example a wrong fee contract will result in never getting any fee deposits but no visible errors.

alysiahuggins:
…aces - to verify

Closes #1527 

### This PR:
- adds a check to see if the fee contract or the implementation contract is a proxy

### This PR does not:
- do anything if the contract is not a proxy
- elegantly handle failures if it's not a proxy

### Key places to review:
- all the changes to the rust code
- should we validate if it's a zero address?

 ### How to test this PR:  
- `cargo test genesis --all-features`


sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -259,6 +260,19 @@ async fn main() -> anyhow::Result<()> {
             .get_contract_address(Contract::LightClientProxy)
             .unwrap();
 
+        // confirm that the light_client_address is a proxy
+        // using the implementation slot, 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc, which is the keccak-256 hash of "eip1967.proxy.implementation" subtracted by 1
+        let hex_bytes =
```
sveitser, line 265: Could we extract this check into a function to re-use the code? We could add  method to the L1 client with this check.

Currently it's not quite clear where would be a good place to perform the check so that if one starts up a local demo or tries to apply an upgrade with a bad fee contract address the user would see an error.

Maybe we could check here

https://github.com/EspressoSystems/espresso-sequencer/blob/31cd49fde6530c71b98197fbb86cc033c4101738/types/src/v0/impls/state.rs#L515

Or maybe there is a better place? 

cc @imabdulbasit

imabdulbasit: I think that we can add this check to Genesis struct maybe and call it when the node starts?

imabdulbasit: so we panic on the startup

imabdulbasit: yes

imabdulbasit: in future we can rename this to `validate()` and validate other things including the contract

imabdulbasit: Just to clarify, I don't see light client contract address anywhere for sequencer main binary. We only have fee contract address in genesis toml file. So, I think we should validate this in separate function and just call it in sequencer main binary

utils/src/deployer.rs
```diff
@@ -399,6 +418,25 @@ fn should_deploy(group: ContractGroup, only: &Option<Vec<ContractGroup>>) -> boo
     }
 }
 
+pub async fn is_proxy_contract(
```
imabdulbasit, line 421: this looks good to me. We can use it to validate each contract address in toml file  (base + upgrades) by adding validate() to Genesis.

sequencer/src/bin/espresso-dev-node.rs
sveitser: Yeah the LC contract is used by the prove, so we should check the LC contract when starting up the prover.

alysiahuggins: where should the LC check be? here? https://github.com/EspressoSystems/espresso-sequencer/blob/1527-sequencer-nodes-should-do-a-sanity-checks-to-check-if-the-l1-contract-addresses-are-wrong/hotshot-state-prover/src/bin/state-prover.rs#L98

alysiahuggins: the check for the fee contract in sequencer main is here  https://github.com/EspressoSystems/espresso-sequencer/blob/1527-sequencer-nodes-should-do-a-sanity-checks-to-check-if-the-l1-contract-addresses-are-wrong/sequencer/src/main.rs#L28 and the implementation of that contract is here https://github.com/EspressoSystems/espresso-sequencer/blob/1527-sequencer-nodes-should-do-a-sanity-checks-to-check-if-the-l1-contract-addresses-are-wrong/sequencer/src/genesis.rs#L80https://github.com/EspressoSystems/espresso-sequencer/blob/1527-sequencer-nodes-should-do-a-sanity-checks-to-check-if-the-l1-contract-addresses-are-wrong/sequencer/src/genesis.rs#L80

utils/src/deployer.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/pull/2000#discussion_r1763274927

sequencer/src/genesis.rs
```diff
@@ -74,6 +76,25 @@ impl Genesis {
     }
 }
 
+impl Genesis {
+    pub async fn validate_contract(&self, l1_rpc_url: String) -> anyhow::Result<()> {
+        let provider = Provider::<Http>::try_from(l1_rpc_url)?;
+        let accounts_vec: Vec<_> = self.accounts.keys().collect();
```
imabdulbasit, line 82: These accounts are prefunded and used for testing purposes such as for builder to have enough funds to build blocks. the fee contract address is in `chain_config`.
So, we should validate` self.chain_config.fee_contract` as  well as fee_contract in chain config of all the upgrades

alysiahuggins: ah thank you https://github.com/EspressoSystems/espresso-sequencer/pull/2000/commits/50d4b9c46e5d29949af3d6a40851919d8c0c1362

sequencer/src/bin/espresso-dev-node.rs
alysiahuggins: made the check for the light client contract [here](https://github.com/EspressoSystems/espresso-sequencer/pull/2000/commits/3c7614d7c0047d40a007cdab0f1ddde9e7497cb6#diff-53f5d56b703d684326ffd2a4011ae3fcf58e68deb773b916efd4028f52b4c1c5R114) and the implementation of the verification is [here](https://github.com/EspressoSystems/espresso-sequencer/pull/2000/commits/3c7614d7c0047d40a007cdab0f1ddde9e7497cb6#diff-8bcfa1727c03a3fa83a1b636f2ba693ff0dc52770828f180dabf4b541236c0e3R85).

sequencer/src/genesis.rs
```diff
@@ -74,6 +76,25 @@ impl Genesis {
...
+
+        if let Some(fee_contract_address) = self.chain_config.fee_contract {
+            if !is_proxy_contract(provider, fee_contract_address)
+                .await
+                .expect("Failed to determine if fee contract is a proxy")
+            {
+                panic!("Fee contract's address is not a proxy");
+            }
+        } else {
+            panic!("Fee contract's address is missing from the chain config!");
```
imabdulbasit, line 91: shouldn't panic if the contract address is not provided. It is optional for V1.

sequencer/src/genesis.rs
```diff
@@ -74,6 +76,25 @@ impl Genesis {
...
+        }
+
```
imabdulbasit, line 93: Also would be great if we can add the check for upgrades. So, we iterate over each upgrade -> get the chain config -> validate the contract address

imabdulbasit (approved):

sequencer/src/genesis.rs
alysiahuggins: should i also ignore the check if the fee_contract it set to the zero address?

sequencer/src/genesis.rs
alysiahuggins: ah gotcha, done here with supporting tests https://github.com/EspressoSystems/espresso-sequencer/pull/2000

PR #2001, opened 2024-09-12
pass V0 for unknown upgrade version

imabdulbasit:


imabdulbasit:
so this PR actually removes the macro specified here https://github.com/EspressoSystems/espresso-sequencer/pull/1955 but the CI still fails. cc @sveitser

jbearer (approved):

PR #2003, opened 2024-09-12
Add smoke-test integration test

tbro:
Adds `tests` crate w/ a single integration test. This test re-implements `demo-smoke-test` bash script in rust, with some minor additions. In order to avoid code duplication, some logic from `espresso-bridge` is factored into a `client` crate, and imported from there for use in the test. `espresso-bridge` is also updated to use this `client`. 

This primary goal of this PR is parity with `smoke-test` bash script. But, there are a couple of improvements:

  * the number of blocks to wait before declaring the test successful is now configurable. This may be useful in scenarios like weekly hostshot integrations.
  * separate waits for for service readiness has been parallelized
  * adds a justfile entry
  * host that tests are run against is configurable (run against remote host)
  
The intention is to use this as the first step toward an integration suite. [The followup will be upgrade test](https://github.com/EspressoSystems/espresso-sequencer/issues/2053).


.config/nextest.toml
```diff
@@ -1,10 +1,19 @@
 [profile.default]
 slow-timeout = "2m"
-default-filter = 'not test(slow_)'
+default-filter = 'not test(slow_) and not package(tests)'
```
sveitser, line 3: Does this mean anyone running `cargo test` will see a failure if they didn't start demo-native?

tests/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
sveitser, line 1: Is there any advantage to use a crate versus a vanilla rust integration test?

sveitser: I guess I implicitly expect tests to be self contained and not require an external service to started for just one test. We could also make the test a binary then it would be more straightforward to customize it's functionality using clap.

.config/nextest.toml
tbro: Yes. You need to run `cargo nextest run` now.

client/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+use anyhow::Context;
```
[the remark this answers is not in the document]
tbro: Maybe. I also not sure `utils` is a good fit. Humm I guess it could go there for now, and if we add more to it later we could move it to its own crate.

tests/Cargo.toml
tbro: Yea I'm not married to the separate crate. But the test does cover more than just sequencer crate, so id does seem workspace level to me. Keep in mind that the follow up to this is upgrade tests, so there will be more in there than just one test. The idea is that there is a suit of integration tests that require native-demo to be running, this is just the first one. I don't love the idea of making this a binary. I don't think we need clap, or much customization. If we need another test, we write another test. Making it a binary will mean we either have to shoehorn new tests in to a single binary or create a new binary for every test.

client/src/lib.rs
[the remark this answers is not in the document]
tbro: On the other hand, I can see some general utility to a single espresso-client package that someone could install from crates.io and use to interact with sequencer API.

.config/nextest.toml
tbro: Or maybe we can address this in cargo configuration. Maybe an alias: https://doc.rust-lang.org/cargo/reference/config.html

.config/nextest.toml
```diff
@@ -1,10 +1,19 @@
...
 retries = 2
 
 [profile.slow]
+slow-timeout = "2m"
```
imabdulbasit, line 7: Slow tests take more time than 2 minutes, so I think this timeout should be increased

tests/demo_native.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        // Number of blocks to wait before deeming the test successful
+        let expected_block_height = dotenvy::var("SMOKE_TEST_EXPECTED_BLOCK_HEIGHT")?
+            .parse::<u64>()
+            .unwrap();
+
+        let provider =
+            Provider::<Http>::try_from(l1_provider_url)?.interval(L1_PROVIDER_RETRY_INTERVAL);
+
+        let espresso = EspressoClient::new(Url::from_str(&sequencer_api_url).unwrap());
```
imabdulbasit, line 138: maybe we should rename the client to `SequencerClient`

tests/demo_native.rs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            panic!("Balance not conserved");
+        }
+
+        if start.elapsed().as_secs() > SEQUENCER_BLOCKS_TIMEOUT {
+            panic!("Timeout waiting for block height, transaction count, and light client updates to increase.");
+        }
+
+        // test that we progress EXPECTED_BLOCK_HEIGHT blocks from where we started
+        if new.block_height.unwrap()
```
imabdulbasit, line 242: I think we should check multiple times that block height is increasing like maybe for 1-2minutes.

imabdulbasit (changes requested):

client/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    socket::{Connection, Unsupported},
+    Url,
+};
+use vbs::version::StaticVersion;
+
+type SequencerApiVersion = StaticVersion<0, 1>;
+// pub type EspressoClient = surf_disco::Client<ClientError, SequencerApiVersion>;
+
+#[derive(Clone, Debug)]
+pub struct EspressoClient(surf_disco::Client<ClientError, SequencerApiVersion>);
```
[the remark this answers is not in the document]
imabdulbasit: maybe we should rename the client to SequencerClient

.env
```diff
@@ -132,3 +132,6 @@ ESPRESSO_SEQUENCER_FETCH_RATE_LIMIT=25
 
 # Query service stress test
 ESPRESSO_NASTY_CLIENT_PORT=24011
+
+# Set this to the number of blocks you would like to confirm the sequencer can reach
+SMOKE_TEST_EXPECTED_BLOCK_HEIGHT=1
```
imabdulbasit, line 137: why is this 1? Can we set it to a higher number like 300/500?

client/src/lib.rs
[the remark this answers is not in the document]
tbro: > I think we should replace clients in sequencer crate where we are passing `SequencerApiVersion` with this new client ( `EspressoClient` )

I agree, but I'd rather do it in a follow up and  concentrate on getting the integration test setup merged in this PR. And

.config/nextest.toml
tbro: I'll double-check, but I believe this is just the time before a test is marked slow, not the time before a test is killed.

.env
tbro: So on anvil we get a block / second, right? so 300 blocks should be reasonable.

tbro:
Regarding 495e59edc3d685d7a4035acbdd2ef84098063a9f, it seems our formatting tool not supporting expected cargo format. Maybe its not a good tool then?

.env
tbro: dee59d4de40d1ac9d0e281b4b957acf6a22f76bc

client/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    pub async fn get_transaction_count(&self) -> anyhow::Result<u64> {
+        self.0
+            .get::<u64>("node/transactions/count")
+            .send()
+            .await
+            .context("getting Espresso transaction count")
+    }
+
+    pub async fn subscribe<FromServer: DeserializeOwned>(
```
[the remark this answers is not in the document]
tbro: 63ebacad4a2ed57f66c7226487b23801a3e11ea7

client/src/lib.rs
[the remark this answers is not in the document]
tbro: > maybe we should rename the client to SequencerClient

f2254bc2b3b613a95bf179691f14c8a58d358c40

.env
tbro: After some thinking I think `300` is to much for CI. I've set at `10` for now. Its configurable so we can move that locally if we want.

client/src/lib.rs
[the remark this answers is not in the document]
imabdulbasit: ok lets create an issue for the follow up on this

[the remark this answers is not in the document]
tbro: https://github.com/EspressoSystems/espresso-sequencer/issues/2052

client/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::time::Duration;
+use surf_disco::{
+    error::ClientError,
+    http::convert::DeserializeOwned,
+    socket::{Connection, Unsupported},
+    Url,
+};
+use vbs::version::StaticVersion;
+
+type SequencerApiVersion = StaticVersion<0, 1>;
```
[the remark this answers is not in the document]
tbro: https://github.com/EspressoSystems/espresso-sequencer/issues/2052

tests/demo_native.rs
```diff
@@ -0,0 +1,249 @@
...
+use std::{
+    fmt,
+    str::FromStr,
+    time::{Duration, Instant},
+};
+use surf_disco::Url;
+
+const L1_PROVIDER_RETRY_INTERVAL: Duration = Duration::from_secs(1);
+// TODO add to .env or derive
+const BUILDER_ADDRESS: &str = "0x23618e81E3f5cdF7f54C3d65f7FBc0aBf5B21E8f";
```
imabdulbasit, line 17: why is this hard coded?

tbro: It is hard coded in the smoke-test bash script

imabdulbasit (approved):

tests/demo_native.rs
tbro: https://github.com/EspressoSystems/espresso-sequencer/issues/2055

PR #2011, opened 2024-09-13
Test builder transaction submission via public mempool

### LINKED ISSUE #1968: Test reserve builder transaction submission via public mempool

### LINKED ISSUE #1969: Test fallback builder transaction submission via public mempool

shenkeyao:
Closes #1968. Closes #1969.

### This PR:
* Add tests for transaction submissions via public mempool with reserve and fallback builders.
* Combine and refactor tests for public and private mempools.

### This PR does not:
* Test complicated scenarios, which will be added later.

### Key places to review:
* Whether the retries make sense.
* Whether `sleep` is needed put in the right places.
* Whether the submission and bundle fetching are properly done.

### How to test this PR: 
* Run the four `test_marketplace_{reserve/fallback}_builder_with_{public/private}_mempool` tests.


marketplace-builder/src/builder.rs
```diff
@@ -502,8 +506,47 @@ mod test {
...
+    /// Verify that the bundle associated with the received quorum proposal contains the expected
+    /// transaction.
+    ///
+    /// # Returns
+    /// * `true` if the verification succeeds.
+    /// * `false` if the expected quorum proposal isn't received yet.
+    ///
+    /// # Panics
+    /// * If the verification fails.
+    async fn verify_bundle(
```
QuentinI, line 518: Nit: I think the control flow would be a bit clearer if this function returned Option<Vec<Transactions>> and assert would be relegated to the caller

marketplace-builder/src/builder.rs
```diff
@@ -563,40 +606,55 @@ mod test {
...
-
-        // Verify the bundle.
-        let start = Instant::now();
-        loop {
-            if start.elapsed() > Duration::from_secs(10) {
-                panic!("Didn't get a quorum proposal in 10 seconds");
+        if public_mempool {
+            let server = &network.server;
+            for transaction in transactions {
+                let _ = server.submit_transaction(transaction).await;
```
QuentinI, line 612: Shouldn't we call unwrap on the result? If we fail to submit transactions the test will fail later anyways, and this way the reason will be clearer

QuentinI (approved):
LGTM with minor comments

marketplace-builder/src/builder.rs
```diff
@@ -563,40 +606,55 @@ mod test {
...
-                        )
-                        .as_str(),
-                    )
-                    .send()
-                    .await
-                    .unwrap();
-                assert_eq!(bundle.transactions, vec![registered_transaction]);
-                break;
+            let mut events = server.event_stream().await;
+            task::sleep(std::time::Duration::from_millis(1000)).await;
```
QuentinI, line 615: I think we should remove the sleep, we'll wait in the loop regardless

marketplace-builder/src/builder.rs
```diff
@@ -501,8 +505,38 @@ mod test {
...
+                    format!(
+                        "bundle_info/bundle/{parent_view_number}/{parent_commitment}/{}",
+                        parent_view_number + 1
+                    )
+                    .as_str(),
+                )
+                .send()
+                .await
+            {
+                return Some(bundle);
```
shenkeyao, line 531: I'm not sure whether it's correct to return `Some` without checking if `bundle.transactions` is empty. The public mempool tests now fail pretty consistently after I changed it to this way.

marketplace-builder/src/builder.rs
shenkeyao: Good suggestion. I had to change it to `get_bundle` rather than `get_transactions` as you suggested, in order to make it compatible with the fee assertions recently added to the tests.

However, as mentioned in [this comment](https://github.com/EspressoSystems/espresso-sequencer/pull/2011/files#r1776047858), I also changed the logic a bit to handle empty transactions the same as non-empty transactions as long as the bundle fetching succeeds. This caused the public mempool tests to fail the transaction assertion.

marketplace-builder/src/builder.rs
Ayiga: I think one potential cause of this comes from assumping that the next event we get from the Network is a proposal.  I think we should be waiting for a `Transaction` event before we wait for this `Quorum Proposal` in order to see the transactions.

We might not actually want to wait for the `Quorum Proposal` either in these cases.  As receiving the `Transaction` event will cause any requests for subsequent `Quorum Proposal`s to return the transaction list until they are included.

marketplace-builder/src/builder.rs
```diff
@@ -565,75 +599,74 @@ mod test {
...
 
-                break;
-            }
-        }
+        assert_eq!(bundle.sequencing_fee, sequencing_fee);
     }
 
-    #[async_std::test]
-    async fn test_marketplace_fallback_builder() {
+    async fn test_marketplace_fallback_builder(public_mempool: bool) {
```
Ayiga, line 669: I think it would be better to rewrite this test to replace the `bool` in this test with an `enum` that can add clarity to each of these cases.  Since both forks end with an `Event` stream in either case, it would also be a natural place for the conditional to exist instead of within the body of this function.

shenkeyao: > Since both forks end with an `Event` stream in either case, it would also be a natural place for the conditional to exist instead of within the body of this function.

Could you clarify this? Besides replacing the `bool` with an `enum`, are you suggesting some other change?

marketplace-builder/src/builder.rs
shenkeyao: Looks like we can't use a `Transaction` event directly here, because we need the parent commitment to fetch the bundle. But I agree we shouldn't wait for the `QuorumProposal`. I'll think about what might be a good alternative.

shenkeyao: Oh, maybe I misunderstood your point. Do you mean we just do transaction comparison so there's no need to fetch the bundle?

Ayiga: I should add that I tried adding this yesterday, but it didn't seem to result in a pass... instead I received errors about not being able to receive a bundle.

shenkeyao:
The latest [commit](https://github.com/EspressoSystems/espresso-sequencer/pull/2011/commits/1255ef9580c51de25a0182b85481028f32677652) is just to sync with main.

marketplace-builder/src/builder.rs
shenkeyao: As discussed I moved the quorum proposal handling earlier, so it corresponds to the parent view now, and the tests pass!

marketplace-builder/src/builder.rs
shenkeyao: All make sense! I replaced `bool` with `enum` and reorganized functions. There are some parts that can't be combined or extracted easily, e.g., due to the public and private mempools using different event streams, but I added some helper functions to simplify things.

PR #2013, opened 2024-09-16
Return blockHeight for historical state query

alxiong:
> Mathis: The batch poster attaches what we call a justification that allows validating that the namespace is finalized by espresso. Previously it would use the height of the first hotshot root that includes the finalized block on L1 (which was returned by the function getHotShotCommitment before the changes). Since we're using an append only merkle tree I'm thinking it should work with any root after that though.
> 
> Mathis: So in this line https://github.com/EspressoSystems/nitro-espresso-integration/blob/8df6661b4449754c5a171b7f722c555da24f27ac/arbnode/batch_poster.go#L523
>
> I think we could for example use the current height in the LC contract instead of snapshot.height

### This PR:

- Add back the hotshot block height as the returned value during `getHotshotCommitment()`


alysiahuggins:
LGTM

alysiahuggins (approved):

PR #2026, opened 2024-09-19
validate chain config received from the state peer catchup response

imabdulbasit:
This PR adds a check for the received chain config from the peer.  It ensures that the commitment of the received chain config matches the commitment of the requested chain config


sveitser:
Is it reasonably straightforward to add a regression test for this?

imabdulbasit:
> Is it reasonably straightforward to add a regression test for this?

Not straightforward because I need a dishonest peer but I can take a look at it later.

sveitser:
I think it's also fine to factor out a small pure function that you can call conveniently with the necessary inputs instead. The most important thing is that there's a test for it this behavior.

sequencer/src/catchup.rs
```diff
@@ -193,7 +194,16 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
                 .await
             {
                 Ok(cf) => {
-                    return Ok(cf);
+                    if cf.commit() == commitment {
```
tbro, line 197: Where is this this method called from?

imabdulbasit: In get_chain_config(). Which is called in validate_and_apply_header() and Header::new()

imabdulbasit: also in Payload::from_transactions(). Basically, wherever we need chain config, we validate and do catchup if required

imabdulbasit: if all the peers are dishonest then we should panic in that case. This will be part of this issue (https://github.com/EspressoSystems/espresso-sequencer/issues/2020) after we discuss it more. If only one peer is dishonest and others don't have chain config at that time then we retry. Hotshot is going to cancel the tasks after some timeout.

tbro: Ok. I think along w/ those changes we could consider how we structure this. It feels weird to me doing data validation in the fetcher... I feel these errors should get bubbled up and to a higher level validation routine that has some checks we can test individually in unit tests and all together in an integration test. As a rule of thumb, I'm thinking validation errors like this could get bubbled up as far as possible w/out breaking public API. Anyway these are just some thoughts. I'll approve and we can discuss further elsewhere.

tbro (approved):
Approving. I left some comments for future consideration.

sequencer/src/catchup.rs
imabdulbasit: Yeah,  the idea was to try another peer if we receive a bad chain config, instead of returning an error from here or returning a bad chain config and  doing validation in get_chain_config() function which would return an error

PR #2028, opened 2024-09-19
add multisig wallet and prover mode to deployment steps in the deployer utils

### LINKED ISSUE #2023: update the deploy.rs script
Add the following updates
- set the admin to the multisig address after deployments of the light client contract and the fee contract
- add the permissionedProver for the light client address using the method on the contract `setPermissionedProver`

"You can't deploy the light client contract until we know the stake table, usually we deploy the network first and then run the deploy sc"

alysiahuggins:
Closes #2023 

### This PR: 
- sets the admin to the multisig address after deployments of the light client contract and the fee contract
- adds the permissionedProver for the light client address using the method on the contract setPermissionedProver


utils/src/deployer.rs
```diff
@@ -389,7 +410,7 @@ pub async fn deploy(
             .await?;
         let fee_contract = FeeContract::new(fee_contract_address, l1.clone());
         let data = fee_contract
-            .initialize(owner)
+            .initialize(deployer)
```
alysiahuggins, line 413: the multisig can be set as the owner here since we do not want to call other functions on the fee contract with the deployer wallet, but lines 430 to 442 also works to accomplish that

utils/src/deployer.rs
```diff
@@ -440,7 +474,7 @@ pub async fn is_proxy_contract(
 #[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
 pub enum ContractGroup {
     #[clap(name = "hotshot")]
-    HotShot,
+    HotShot, // TODO: confirm whether we keep HotShot in the contract group
```
alysiahuggins, line 477: do we still want to deploy HotShot?

utils/src/deployer.rs
jbearer: Yeah good point

utils/src/deployer.rs
jbearer: We can get rid of it but I think that can be a separate change, just to keep this as small as possible

.env
```diff
@@ -58,9 +58,10 @@ ESPRESSO_BUILDER_ETH_ACCOUNT_INDEX=8
 ESPRESSO_DEPLOYER_ACCOUNT_INDEX=9
 
 # Contracts
-ESPRESSO_SEQUENCER_HOTSHOT_ADDRESS=0xb19b36b1456e65e3a6d514d3f715f204bd59f431
-ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0x0c8e79f3534b00d9a3d4a856b665bf4ebc22f2ba
+ESPRESSO_SEQUENCER_HOTSHOT_ADDRESS=0x8ce361602b935680e8dec218b820ff5056beb7af
+ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0xed1db453c3156ff3155a97ad217b3087d5dc5f6e
```
tbro, line 62: Is it possible native-demo failure is related to this change? https://github.com/EspressoSystems/espresso-sequencer/blob/5be401ab99cbf0845337a677026c3cea46ae03a0/scripts/smoke-test-demo#L69

jbearer (approved):

PR #2040, opened 2024-09-20
Update `tide-disco`

rob-maron:
Fixes an issue where we couldn't use non-root users in the docker containers


jbearer (approved):
I fear tide-disco might need to be updated in HotShot/query service too for this to work. Maybe I'm wrong, CI will tell us

rob-maron:
> I fear tide-disco might need to be updated in HotShot/query service too for this to work. Maybe I'm wrong, CI will tell us

Yeah, CI passed and I didn't see the old version in the lock. But I'm going to tag this commit and try changing the root user

jbearer:
Oh, because this is a non-breaking change....thank god we finally have proper semver on these dependencies

PR #2049, opened 2024-09-23
test marketplace builder bundle fees and signature

imabdulbasit:
This PR validates the bundle fees and signature fields in both reserve and fallback builder tests


shenkeyao (approved):
LGTM

PR #2051, opened 2024-09-23
Fix node-validator's state becoming stale due to disconnected leaf stream

### LINKED ISSUE #2002: Modify `node-metrics` so that it handles when the `ProcessProduceLeafStreamTask` exits its `process_consume_leaf_stream` method
Currently when the `node-validator` service doesn't die when the source `Leaf` `Stream` fails to connect, or dies.  For now, this should be modified to better handle this loop exit.  Ideally this would attempt to reconnect to the stream.  However, it can be easier to just panic initially until this can be handled better.

Ayiga:
Closes #2002

### This PR:

Addresses an issue that occurs when the source Leaf Stream HotShot Query service becomes unavailable, resulting in the node-validator service ending up in a Stale state with no future updates being delivered.

This is implemented with connection re-attempts to acquire a new Leaf Stream should the previous one complete.  The re-attempts themselves will loop to try and ensure a new Leaf Stream is returned with an exponential backkoff up to a maximum limit of 100 times, at which point it will panic as it seems that the service is not going to be coming back immediately.

### DIFF
--- a/node-metrics/src/api/node_validator/v0/mod.rs
+++ b/node-metrics/src/api/node_validator/v0/mod.rs
@@ -5,7 +5,7 @@ use crate::service::client_message::{ClientMessage, InternalClientMessage};
 use crate::service::data_state::{LocationDetails, NodeIdentity};
 use crate::service::server_message::ServerMessage;
 use async_std::task::JoinHandle;
-use espresso_types::SeqTypes;
+use espresso_types::{BackoffParams, SeqTypes};
 use futures::channel::mpsc::SendError;
 use futures::future::Either;
 use futures::{
@@ -20,7 +20,6 @@ use hotshot_types::traits::{signature_key::StakeTableEntryType, stake_table::Sta
 use hotshot_types::PeerConfig;
 use prometheus_parse::{Sample, Scrape};
 use serde::{Deserialize, Serialize};
-use std::cmp::max;
 use std::fmt;
 use std::future::Future;
 use std::io::BufRead;
@@ -597,6 +596,9 @@ impl ProcessProduceLeafStreamTask {
     where
         R: LeafStreamRetriever<Item = Leaf<SeqTypes>>,
     {
+        let backoff_params = BackoffParams::default();
+        let mut delay = Duration::ZERO;
+
         for attempt in 1..=100 {
             let leaves_stream_result = leaf_stream_receiver.retrieve_stream(None).await;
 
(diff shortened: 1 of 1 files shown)


node-metrics/src/api/node_validator/v0/mod.rs
```diff
@@ -540,28 +553,93 @@ impl ProcessProduceLeafStreamTask {
...
 
+    async fn connect_and_process_leaves<R, K>(leaf_stream_retriever: R, leaf_sender: K)
+    where
+        R: LeafStreamRetriever<Item = Leaf<SeqTypes>>,
+        K: Sink<Leaf<SeqTypes>, Error = SendError> + Clone + Send + Sync + Unpin + 'static,
+    {
+        // We want to try and ensure that we are connected to the HotShot Query
+        // Service, and are consuming leaves.
+        // - If we are able to connect, then we can start relaying Leaves into
+        //   the leaf sender.  If we are unable
```
imabdulbasit, line 564:
```suggestion
        //   the leaf sender.
```

node-metrics/src/api/node_validator/v0/mod.rs
```diff
@@ -540,28 +553,93 @@ impl ProcessProduceLeafStreamTask {
...
+                        "attempt {attempt} to connect to leaf stream failed with error {error}"
+                    );
+
+                    // Our retry penalty will be a minimum of 100ms, and a maximum
+                    // of 5 seconds.
+                    // For every failed iteration, we will double our delay, up
+                    // to the maximum of 5 seconds.
+
+                    let retry_penalty = max(
+                        Duration::from_millis(100) * 2u32.pow(attempt),
```
imabdulbasit, line 618: if attempt is large like 99 then wouldn't this result in an overflow?

imabdulbasit: Also we have `BackoffParams` so maybe we can re use it here

Ayiga: Hmm... the `BackoffParams` seem to indicate that they are used specifically for catchup.  At least that's what the environment variables indicate.  We could still potentially use the default here though.

Also, yeah, you're right... it would overflow as soon as it hits 30 attempts or so.

Ayiga: `BackOffParams`'s api seems inconvenient at the moment as well.  It assumes an underlying that is being mutated, and requires `anyhow` to operate.

Ayiga: Though it does look like we might be able to achieve what `BackOffParams::retry` is doing with `backoff` itself.

imabdulbasit: ok lets leave BackOffParams

imabdulbasit (approved):

PR #2064, opened 2024-09-25
Patch HotShot 0.5.75

ss-es:
Patches hotshot 0.5.75 to include all commits that were patched into the version we are deploying with.

This should fix any leaf commitment issues introduced into `demo-native-marketplace` in the previous hotshot update, 
but there may be lingering issues (which are out of scope for this PR)


tbro:
What is the reasoning of patching hotshot in this way?

tbro (approved):
Running integration test w/ marketplace setup passes. I think the failure of MP demo native is flakyness, so I'm approviing.

data/v1/messages.json
```diff
@@ -38,11 +38,11 @@
               "justify_qc": {
                 "_pd": null,
                 "data": {
-                  "leaf_commit": "COMMIT~eaBGKF8-lw-t211wxLq7tcXEEqacbvDVcFFR9aCddO3G"
```
tbro, line 41: updated messages

PR #2067, opened 2024-09-25
Also test marketplace through integration test

tbro:
Runs integration test a second time w/ marketplace setup. The build from first run should be reused, so it should run fairly quickly. This may make marketplace demo native obsolete, but we can keep them both around for a bit to see if we learn anything.


imabdulbasit (approved):
LGTM! the CI is failing because of different events stream client versions and requires a fix in hotshot-builder-core

PR #2079, opened 2024-09-26
assert transactions are incrementing in integration test

tbro:
assert transactions are incrementing in integration test. I also put in a change for disabling incremental builds in CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@ concurrency:
 env:
   RUSTFLAGS: '--cfg async_executor_impl="async-std" --cfg async_channel_impl="async-std"'
   RUST_LOG: info,libp2p=off,node=error
+  CARGO_INCREMENTAL: 0
 
 jobs:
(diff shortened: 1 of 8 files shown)


imabdulbasit (approved):

.github/workflows/build.yml
```diff
@@ -20,6 +20,7 @@ concurrency:
 env:
   RUSTFLAGS: '--cfg async_executor_impl="async-std" --cfg async_channel_impl="async-std"'
   RUST_LOG: info,libp2p=off,node=error
+  CARGO_INCREMENTAL: 0
```
imabdulbasit, line 23: isn't it okay to use incremental compilation ?

tbro: I don't think it is doing us much good in our current configuration. But I reverted that for now, we can look into it more later. A little context: https://rustmagazine.org/issue-2/optimize-rust-comptime/#step-3:-disable-incremental-compilation

PR #2085, opened 2024-09-27
remove marketplace-demo-native workflow

tbro:
This PR disables and removes marketplace-demo-native workflow.

Marketplace demo-native fails always currently. I suspect this is due to the slower speed which which marketplace processes blocks. That's speculation, but I don't want to spend time debugging it b/c we already have [a more reliable test](https://github.com/EspressoSystems/espresso-sequencer/pull/2067) in the integration run.


imabdulbasit (approved):
LGTM!

Issue #2055, opened 2024-09-23
Get builder address from API in tests


tbro:
See comment: https://github.com/EspressoSystems/espresso-sequencer/pull/2003#discussion_r1771972157

PR #2086, opened 2024-09-30
Add upgrade test

### LINKED ISSUE #2053: Add upgrade integration test
We need to assert upgrades in our integration test suite. This may require some refactoring of the test suite itself, such as moving common functionality to a shared module. This is follow up to https://github.com/EspressoSystems/espresso-sequencer/pull/2003#

tbro:
Closes #2053 and #2055

### This PR:
Add integration test for upgrades. Collect multiple sequencer streams into one stream and and assert version in the headers. First few blocks are asserted to be base version. Then we check that all the sequencers are upgraded within a timeout period. Currently we only check 2 sequencers, since that is how many have the API enabled currently. The same test can be run against any consecutive base/upgrade version pairs.

  * factor common logic to module
  * add sequencer version env var for setting version to be tested
  * get builder address from API (instead of hard coding)
  * small change to `SequencerClient` to avoid requiring generics


tests/common/mod.rs
```diff
@@ -94,24 +71,54 @@ fn url_from_port(port: String) -> Result<String> {
...
                 "ESPRESSO_SUBMIT_TRANSACTIONS_PRIVATE_FALLBACK_PORT",
-            )?)?;
-        }
+            )?)?
+        } else {
+            url_from_port(dotenvy::var("ESPRESSO_SUBMIT_TRANSACTIONS_PRIVATE_PORT")?)?
+        };
+
+        let builder_url = if sequencer_version >= 03 {
+            let url = url_from_port(dotenvy::var("ESPRESSO_RESERVE_BUILDER_SERVER_PORT")?)?;
```
imabdulbasit, line 94: should we use fallback server port here as the load generator has fallback port?

tbro: Sounds correct.

tests/upgrades.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Test is limited to those sequencers with correct modules
+    // enabled. It would be less fragile if we could discover them.
+    let subscriptions = vec![
+        clients[0].subscribe_headers(0).await?,
+        clients[1].subscribe_headers(0).await?,
+    ];
+    let subscriptions_size = subscriptions.len();
+    let mut streams = stream::iter(subscriptions)
+        .flatten_unordered(usize::try_from(SEQUENCER_BLOCKS_TIMEOUT).ok());
```
imabdulbasit, line 37: why are we passing `SEQUENCER_BLOCKS_TIMEOUT` here

tests/upgrades.rs
```diff
@@ -0,0 +1,64 @@
...
+        if header.version() == versions.1 {
+            upgraded_nodes += 1;
+        }
+
+        if upgraded_nodes == subscriptions_size {
+            println!("Upgrade succeeded @ height {}!", header.height());
+            break;
+        }
+
+        if header.height() > SEQUENCER_BLOCKS_TIMEOUT {
```
imabdulbasit, line 59: can we also check that we are producing blocks after the upgrade is successful

tests/common/mod.rs
```diff
@@ -5,57 +5,34 @@ use espresso_types::FeeAmount;
...
+pub struct TestConfig {
+    pub load_generator_url: String,
+    pub provider: Provider<Http>,
+    pub espresso: SequencerClient,
+    pub builder_address: Address,
+    pub recipient_address: Address,
+    pub light_client_address: Address,
+    pub prover_url: String,
+    pub expected_block_height: u64,
+    pub timeout: f64,
```
imabdulbasit, line 27: why is it f64? can we make it Duration?

tests/upgrades.rs
tbro: It provides a limit, but I had misread the docs as to what it was limiting. 

https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.flatten_unordered

tests/common/mod.rs
tbro: Just to provide that little extra time on the timeout for larger numbers.  `f64` has `.sqrt()` fn. Not sure how we would do the same w/ `Duration`.

tests/common/mod.rs
tbro: eb3cd72eea26ab5f2218fdb53f68f6b24ff882a7

tests/upgrades.rs
tbro: eb3cd72eea26ab5f2218fdb53f68f6b24ff882a7

tests/common/mod.rs
tbro: e42ba141277096d99e36cc6e51e8cf48183c9b20

tbro: So that actually breaks it: https://github.com/EspressoSystems/espresso-sequencer/pull/2086/commits/e42ba141277096d99e36cc6e51e8cf48183c9b20

tests/upgrades.rs
tbro: 504de74a7bc2858122ae36607a0b42b2ba206314

imabdulbasit (approved):

PR #2089, opened 2024-10-01
Validation bugfix

### LINKED ISSUE #968: Validate existing header fields once HotShot supports validated state
* `height` increases by 1
* `timestamp` is within some delta of the current time and not decreasing
* L1 head is within some delta of the current L1 head and not decreasing
* L1 finalized is accurate, within some delta of the current L1 head, not in the future, and not decreasing

### LINKED ISSUE #2088: [HEADER-1] - Malicious leader can stop replicas from reading new depositors
The issue stems from the fact that we assume that l1_finalized, l1.head and timestamp are trusted values.

New depositors are kept track of by checking if there are any new deposits from the `parent_header` by taking a snapshot of the L1 client and checking if there are any new depositors in that time period. This snapshot occurs in both `new_marketplace` and `new_legacy`.

The problem arises due to the fact that a malicious leader can set the `l1_finalized` to some arbitrarily large number. This will not cause a problem until that proposal has been voted on, after which the malicious proposal is now the parent reference as seen here:

```Rust
if l1.finalized < parent_header.l1_finalized() {
    tracing::warn!(
        "L1 finalized {:?} behind parent {:?}, L1 client may be lagging",
        l1.finalized,
        parent_header.l1_finalized()
    );
    l1.finalized = parent_header.l1_finalized();
}
```
The malicious proposal now is referred to as the parent, and it will not include any new deposits due to the early return if the new snapshots’ `l1_finalized < parent l1_finalized`. The mechanism here is small. Since we are always taking the larger `l1_finalized` value, we override the proposals’ `l1_finalized` if the new snapshots’ `l1_finalized < parent l1_finalized` so we will *always* ignore the new depositors in this case.

This issue originates in the `from_info` method in `header.rs`, there are multiple blocks of `if` statements here, and the same concept applies to `l1.head` and `timestamp` of a proposal.

tbro:
Closes #2088 
Closes #968 

### This PR:
Adds validation to replicas.

### DIFF
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -24,6 +24,7 @@ use num_traits::CheckedSub;
 use serde::{Deserialize, Serialize};
 use std::{ops::Add, time::Duration};
 use thiserror::Error;
+use time::OffsetDateTime;
 use vbs::version::Version;
 
 use super::{
@@ -95,9 +96,9 @@ pub enum ProposalValidationError {
     InvalidNsTable { err: NsTableValidationError },
     #[error("Some fee amount or their sum total out of range")]
     SomeFeeAmountOutOfRange,
-    #[error("Invalid timestamp: parent:={parent_timestamp}, proposal={proposal_timestamp}")]
+    #[error("Invalid timestamp: local:={local_timestamp}, proposal={proposal_timestamp}")]
     InvalidTimestamp {
-        parent_timestamp: u64,
+        local_timestamp: u64,
         proposal_timestamp: u64,
     },
     #[error("l1_finalized has `None` value")]
@@ -316,15 +317,16 @@ pub fn validate_proposal(
         });
     }
 
-    // Check if timestamp is increasing (with some tolerance).
-    if proposal.timestamp() + 10 <= parent_header.timestamp() + 8 {
+    // Check if timestamp is increasing.
+    let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
+    if proposal.timestamp() < system_time - 2 {
         return Err(ProposalValidationError::InvalidTimestamp {
-            parent_timestamp: parent_header.timestamp(),
             proposal_timestamp: proposal.timestamp(),
+            local_timestamp: proposal.timestamp(),
         });
     }
 
-    if proposal.l1_head() <= parent_header.l1_head() {
+    if proposal.l1_head() < parent_header.l1_head() {
         return Err(ProposalValidationError::NonIncrementingL1Head {
             parent: parent_header.l1_head(),
             proposal: proposal.l1_head(),
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -696,27 +696,25 @@ impl HotShotState<SeqTypes> for ValidatedState {
             .expect("Chain Config not found in validated state");
 
         // Validate l1_finalized.
-        let Some(proposed_finalized) = proposed_header.l1_finalized() else {
-            tracing::error!("L1 finalized has None value.");
-            return Err(BlockError::InvalidBlockHeader);
-        };
-
-        let parent_finalized = parent_leaf.block_header().l1_finalized().unwrap().number();
-        if proposed_finalized.number() < parent_finalized {
+        let proposed_finalized = proposed_header.l1_finalized();
+        let parent_finalized = parent_leaf.block_header().l1_finalized();
+        if proposed_finalized < parent_finalized {
             tracing::error!(
-                "L1 finalized not incrementing. parent: {}, proposal: {}",
+                "L1 finalized not incrementing. parent: {:?}, proposal: {:?}",
                 parent_finalized,
-                proposed_finalized.number()
+                proposed_finalized,
             );
         }
-        let finalized = instance
-            .l1_client
-            .wait_for_finalized_block(proposed_finalized.number())
-            .await;
+        if let Some(proposed_finalized) = proposed_finalized {
+            let finalized = instance
+                .l1_client
+                .wait_for_finalized_block(proposed_finalized.number())
+                .await;
 
-        if finalized != proposed_finalized {
-            tracing::error!("Invalid proposal: l1_finalized mismatch");
-            return Err(BlockError::InvalidBlockHeader);
+            if finalized != proposed_finalized {
+                tracing::error!("Invalid proposal: l1_finalized mismatch");
+                return Err(BlockError::InvalidBlockHeader);
+            }
         }
 
         // Validate `l1_head`.
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -94,6 +94,11 @@ pub enum ProposalValidationError {
     InvalidNsTable { err: NsTableValidationError },
     #[error("Some fee amount or their sum total out of range")]
     SomeFeeAmountOutOfRange,
+    #[error("Invalid timestamp: parent:={parent_timestamp}, proposal={proposal_timestamp}")]
+    InvalidTimestamp {
+        parent_timestamp: u64,
+        proposal_timestamp: u64,
+    },
 }
 
 impl StateDelta for Delta {}
@@ -303,6 +308,14 @@ pub fn validate_proposal(
         return Err(ProposalValidationError::InvalidHeight {
             parent_height: parent_header.height(),
             proposal_height: proposal.height(),
+        });
+    }
+
+    // Check if timestamp is increasing (with some tolerance).
+    if proposal.timestamp() <= parent_header.timestamp() - 2 {
+        return Err(ProposalValidationError::InvalidTimestamp {
+            parent_timestamp: parent_header.timestamp(),
+            proposal_timestamp: proposal.timestamp(),
         });
     }
 
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -95,10 +95,15 @@ pub enum ProposalValidationError {
     InvalidNsTable { err: NsTableValidationError },
     #[error("Some fee amount or their sum total out of range")]
     SomeFeeAmountOutOfRange,
-    #[error("Invalid timestamp: local:={local_timestamp}, proposal={proposal_timestamp}")]
-    InvalidTimestamp {
-        local_timestamp: u64,
+    #[error("Invalid timestamp: proposal={proposal_timestamp}, parent={parent_timestamp}")]
+    InvalidTimestampNonIncrementing {
         proposal_timestamp: u64,
+        parent_timestamp: u64,
+    },
+    #[error("Invalid timestamp: local:={local_timestamp}, proposal={proposal_timestamp}")]
+    InvalidTimestampDrift {
+        proposal_timestamp: u64,
+        local_timestamp: u64,
     },
     #[error("l1_finalized has `None` value")]
     L1FinalizedNotFound,
@@ -316,12 +321,18 @@ pub fn validate_proposal(
         });
     }
 
-    // Check if timestamp is increasing.
+    // Validate timestamp increasing.
+    if proposal.timestamp() < parent_header.timestamp() {
+        return Err(ProposalValidationError::InvalidTimestampNonIncrementing {
+            proposal_timestamp: proposal.timestamp(),
+            parent_timestamp: parent_header.timestamp(),
+        });
+    }
+
+    // Validate timestamp hasn't drifted too much from system time.
     let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
-    if proposal.timestamp() < parent_header.timestamp()
-        || proposal.timestamp().abs_diff(system_time) > 2
-    {
-        return Err(ProposalValidationError::InvalidTimestamp {
+    if proposal.timestamp().abs_diff(system_time) > 2 {
+        return Err(ProposalValidationError::InvalidTimestampDrift {
             proposal_timestamp: proposal.timestamp(),
             local_timestamp: system_time,
         });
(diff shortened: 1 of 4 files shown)


types/src/v0/impls/state.rs
```diff
@@ -306,6 +316,21 @@ pub fn validate_proposal(
...
 
+    // Check if timestamp is increasing (with some tolerance).
+    if proposal.timestamp() <= parent_header.timestamp() - 2 {
+        return Err(ProposalValidationError::InvalidTimestamp {
+            parent_timestamp: parent_header.timestamp(),
+            proposal_timestamp: proposal.timestamp(),
+        });
+    }
+
+    if proposal.l1_head() <= parent_header.l1_head() {
```
jbearer, line 327: We should also wait for the `l1_head` to actually exist (ie wait for L1 block number >= `proposal.l1_head()`) as we do for the finalized block. Otherwise a malicious proposer can set `l1_head` arbitrarily far in the future, and it ceases to be a useful field

types/src/v0/impls/state.rs
```diff
@@ -663,6 +688,32 @@ impl HotShotState<SeqTypes> for ValidatedState {
             .resolve()
             .expect("Chain Config not found in validated state");
 
+        let Some(l1_finalized) = proposed_header.l1_finalized() else {
+            tracing::error!("L1 finalized has None value.");
+            return Err(BlockError::InvalidBlockHeader);
+        };
+
+        let parent_finalized = parent_leaf.block_header().l1_finalized().unwrap().number();
+        if l1_finalized.number() <= parent_finalized {
```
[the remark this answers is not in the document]
tbro: I'm trying to keep to the minimum changes. `validate_proposal` is currently *not* `async`. And I thought I would go ahead and do everything that needs to be done w/ `l1_finalized` in the same place. I would rather have enum variants for all these errors, like in `validate_proposal`, but I'm more concerned w/ getting it done than making it pretty.

types/src/v0/impls/state.rs
```diff
@@ -306,6 +316,21 @@ pub fn validate_proposal(
         });
     }
 
+    // Check if timestamp is increasing (with some tolerance).
+    if proposal.timestamp() <= parent_header.timestamp() - 2 {
```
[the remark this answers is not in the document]
tbro: 22febe1222c1c0155b49d7981f83c4475f0812ea

types/src/v0/impls/state.rs
```diff
@@ -663,6 +688,32 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+            );
+        }
+        // Wait for l1_finalized with a timeout.
+        if let Err(e) = future::timeout(Duration::from_secs(2), async {
+            instance
+                .l1_client
+                .wait_for_finalized_block(l1_finalized.number())
+                .await;
+        })
+        .await
```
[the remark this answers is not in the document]
tbro: 56097092bc76cc92e96e639c70168e42ddc81eca

types/src/v0/impls/state.rs
```diff
@@ -316,15 +316,16 @@ pub fn validate_proposal(
...
 
-    // Check if timestamp is increasing (with some tolerance).
-    if proposal.timestamp() <= parent_header.timestamp() - 2 {
+    // Check if timestamp is increasing.
+    let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
+    if proposal.timestamp() < system_time - 2 {
         return Err(ProposalValidationError::InvalidTimestamp {
-            parent_timestamp: parent_header.timestamp(),
             proposal_timestamp: proposal.timestamp(),
+            local_timestamp: proposal.timestamp(),
```
jbearer, line 324: Should this be `system_time`?

types/src/v0/impls/state.rs
```diff
@@ -316,15 +316,16 @@ pub fn validate_proposal(
         });
     }
 
-    // Check if timestamp is increasing (with some tolerance).
-    if proposal.timestamp() <= parent_header.timestamp() - 2 {
+    // Check if timestamp is increasing.
+    let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
+    if proposal.timestamp() < system_time - 2 {
```
jbearer, line 321: We _also_ want to enforce the timestamp is increasing relative to the parent. Ie the condition for validity should be something like `proposal.timestamp() >= parent_header.timestamp() && proposal.timestamp().abs_diff(system_time) <= 2`

types/src/v0/impls/state.rs
```diff
@@ -688,30 +689,30 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+            .l1_client
+            .wait_for_finalized_block(proposed_finalized.number())
+            .await;
+
+        if finalized.hash() != proposed_finalized.hash() {
+            tracing::error!("Invalid proposal: l1_finalized hash mismatch");
+        }
+        if finalized.timestamp() != proposed_finalized.timestamp() {
+            tracing::error!("Invalid proposal: l1_finalized timestamp mismatch");
         }
```
jbearer: Nit: I think we can just do `if finalized != proposed_finalized`. Less error prone in case we ever add fields to the `L1BlockInfo` struct

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
jbearer: Sounds good. Can you create issues for the tech debt cleanup so we can tackle it when we're less rushed?

types/src/v0/impls/state.rs
tbro: dc197d664cf7cc2368500b08ed584a52b4c9398f

types/src/v0/impls/state.rs
tbro: dc197d664cf7cc2368500b08ed584a52b4c9398f

types/src/v0/impls/state.rs
tbro: dc197d664cf7cc2368500b08ed584a52b4c9398f

types/src/v0/impls/state.rs
tbro: dc197d664cf7cc2368500b08ed584a52b4c9398f

types/src/v0/impls/state.rs
```diff
@@ -663,6 +688,32 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+        let parent_finalized = parent_leaf.block_header().l1_finalized().unwrap().number();
+        if l1_finalized.number() <= parent_finalized {
+            tracing::error!(
+                "L1 finalized not incrementing. parent: {}, proposal: {}",
+                parent_finalized,
+                l1_finalized.number()
+            );
+        }
+        // Wait for l1_finalized with a timeout.
+        if let Err(e) = future::timeout(Duration::from_secs(2), async {
```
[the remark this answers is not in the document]
tbro: 56097092bc76cc92e96e639c70168e42ddc81eca

types/src/v0/impls/state.rs
```diff
@@ -702,18 +698,29 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+
+        // Validate `l1_head`.
+        if proposed_header.l1_head() < parent_leaf.block_header().l1_head() {
+            tracing::error!("Invalid proposal: l1_head not increasing");
+            return Err(BlockError::InvalidBlockHeader);
+        }
+
+        let l1_head = instance
+            .l1_client
+            .wait_for_finalized_block(proposed_header.l1_head())
```
jbearer, line 719: We don't want to wait for the l1 head to be finalized, we just want to wait for it to exist (ie for the L1 block height to be >= l1_head). Might need a new l1_client method for this.

types/src/v0/impls/state.rs
```diff
@@ -318,17 +318,12 @@ pub fn validate_proposal(
 
     // Check if timestamp is increasing.
     let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
-    if proposal.timestamp() < system_time - 2 {
+    if proposal.timestamp() >= parent_header.timestamp()
+        && proposal.timestamp().abs_diff(system_time) <= 2
```
jbearer, line 322: Sorry, this is backwards. This condition is _validity_, but as written we're checking for _in_validity, so we would need to wrap this in `!( )` or flip the inequalities or something

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
tbro: https://github.com/EspressoSystems/espresso-sequencer/issues/2090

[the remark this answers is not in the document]
tbro: Ideally we add assertions for this new validation as well. Our `validate_proposal_error_cases` has become a little difficult to deal with though.

types/src/v0/impls/state.rs
tbro: Sorry. I'm typing faster than I'm thinking

tbro: 4f474014341baaac1e22af1fe7b3a7e58f4e2c90

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
jbearer: Yeah, maybe we can break that down into smaller units somehow

types/src/v0/impls/state.rs
tbro: 064eda5e3dadd3bdf9a1f71e2bf9b313b2f16914

jbearer (approved):

types/src/v0/impls/state.rs
```diff
@@ -306,6 +321,23 @@ pub fn validate_proposal(
...
+    if proposal.timestamp() < parent_header.timestamp() {
+        return Err(ProposalValidationError::InvalidTimestampNonIncrementing {
+            proposal_timestamp: proposal.timestamp(),
+            parent_timestamp: parent_header.timestamp(),
+        });
+    }
+
+    // Validate timestamp hasn't drifted too much from system time.
+    let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
+    if proposal.timestamp().abs_diff(system_time) > 2 {
```
sveitser, line 334: Maybe a comment that we think 2 seconds is enough for nodes that are reasonably configured and that we may want to make this configurable later?

types/src/v0/impls/state.rs
```diff
@@ -663,6 +695,41 @@ impl HotShotState<SeqTypes> for ValidatedState {
             .resolve()
             .expect("Chain Config not found in validated state");
 
+        // Validate l1_finalized.
+        let Some(proposed_finalized) = proposed_header.l1_finalized() else {
+            tracing::error!("L1 finalized has None value.");
```
sveitser, line 700: Does this mean the proposer proposed an invalid block? I think in that case this should be `warn` at most? We should only log `error` if this is something the application cannot handle and/or it's worth waking people up for.

types/src/v0/impls/state.rs
```diff
@@ -663,6 +695,41 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+            .await;
+
+        if finalized != proposed_finalized {
+            tracing::error!("Invalid proposal: l1_finalized mismatch");
+            return Err(BlockError::InvalidBlockHeader);
+        }
+
+        // Validate `l1_head`.
+        if proposed_header.l1_head() < parent_leaf.block_header().l1_head() {
+            tracing::error!("Invalid proposal: l1_head not increasing");
```
sveitser, line 724: Since we allow the same `l1_head`:

```suggestion
            tracing::error!("Invalid proposal: l1_head decreasing");
```

types/src/v0/impls/state.rs
```diff
@@ -663,6 +695,41 @@ impl HotShotState<SeqTypes> for ValidatedState {
             .resolve()
             .expect("Chain Config not found in validated state");
 
+        // Validate l1_finalized.
+        let Some(proposed_finalized) = proposed_header.l1_finalized() else {
+            tracing::error!("L1 finalized has None value.");
+            return Err(BlockError::InvalidBlockHeader);
+        };
+
+        let parent_finalized = parent_leaf.block_header().l1_finalized().unwrap().number();
```
sveitser, line 704: Since we're not adding tests now please open an issue for that (if we don't have it already) and link it in the code.

sveitser (approved):
LGTM, some nits.

types/src/v0/impls/state.rs
tbro: appears to have been addressed here: https://github.com/EspressoSystems/espresso-sequencer/pull/2089/commits/8f0d2291f21808ef96e3671a99b069b9ffcfaa31

types/src/v0/impls/state.rs
tbro: https://github.com/EspressoSystems/espresso-sequencer/pull/2089/commits/0f4ebc0f1310cdc4e524908f6567ae369e06fcf4

types/src/v0/impls/state.rs
tbro: 0becffbc6e88029fa5afec89a8bd0f94ae05390e

types/src/v0/impls/state.rs
```diff
@@ -663,6 +698,40 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
+        if proposed_header.l1_head() < parent_leaf.block_header().l1_head() {
+            tracing::error!("Invalid proposal: l1_head decreasing");
+            return Err(BlockError::InvalidBlockHeader);
+        }
+
+        let _ = instance
+            .l1_client
+            .wait_for_block(proposed_header.l1_head())
+            .await;
+
```
rob-maron, line 734: Anchor linking

PR #2091, opened 2024-10-01
Validate fee account proof matches the requested account during catchup

imabdulbasit:
This PR adds a check to validate that the peer response for the account matches the requested account proof.

### DIFF
--- a/sequencer/src/catchup.rs
+++ b/sequencer/src/catchup.rs
@@ -136,13 +136,14 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
                 .await
             {
                 Ok(res) => {
-                    if res.proof.account == account.into() {
-                        match res.proof.verify(&fee_merkle_tree_root) {
-                            Ok(_) => return Ok(res),
-                            Err(err) => tracing::warn!("Error verifying account proof: {}", err),
-                        }
-                    } else {
+                    if res.proof.account != account.into() {
                         tracing::warn!("Invalid proof received from peer {:?}", client.url);
+                        continue;
+                    }
+
+                    match res.proof.verify(&fee_merkle_tree_root) {
+                        Ok(_) => return Ok(res),
+                        Err(err) => tracing::warn!("Error verifying account proof: {}", err),
                     }
                 }
                 Err(err) => {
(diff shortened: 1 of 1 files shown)


sveitser:
I think we should add a test for this. If it's hard and we don't have time now please open an issue for it and link to it from code comments.

imabdulbasit:
> I think we should add a test for this. If it's hard and we don't have time now please open an issue for it and link to it from code comments.

https://github.com/EspressoSystems/espresso-sequencer/issues/2102

sequencer/src/catchup.rs
```diff
@@ -135,10 +135,16 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
                 .send()
                 .await
             {
-                Ok(res) => match res.proof.verify(&fee_merkle_tree_root) {
-                    Ok(_) => return Ok(res),
-                    Err(err) => tracing::warn!("Error verifying account proof: {}", err),
-                },
+                Ok(res) => {
+                    if res.proof.account == account.into() {
```
tbro, line 139: I think this would be clearer as early return:

```rust
if res.proof.account != account.into() {
  tracing::warn!("Invalid proof received from peer {:?}", client.url);
}
```

sequencer/src/catchup.rs
```diff
@@ -235,17 +241,36 @@ impl<T> StateCatchup for SqlStateCatchup<T>
...
+
+        if res.proof.account != account.into() {
+            panic!(
+                "Critical error: Mismatched fee account proof: expected {:?}, got {:?}.
+                This may indicate a compromised database",
+                account, res.proof.account
+            );
+        }
+
+        match res.proof.verify(&fee_merkle_tree_root) {
```
tbro, line 267: I think `verify()` method should not return `anyhow` and  `FeeAccountProof` should have a dedicated error. This way we can clean up caller logic and also write better tests.

sequencer/src/catchup.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2091/commits/38d714e6170ce43614b753fbe8255af0ac6d5b7a

sequencer/src/catchup.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/issues/2146

tbro (approved):
LGTM

Issue #2076, opened 2024-09-26
I05: Inability to reduce stateHistoryRetentionPeriod


alysiahuggins:
Affected Code: LightClient.sol (line 441)

Summary: The function setstateHistoryRetentionPeriod wrongly when
reducing the retention period, which can lock in a large value if set
by mistake.

Suggestion: Instead of reverting when the retention period decreases,
updateStateHistory could be called with the last finalized state to delete
the newly outdated states.

Comments: This was intentional for this version of the escape hatch but will address later.

alysiahuggins:
This will not be addressed right now however we have created an issue to set a max state history retention period allowed https://github.com/EspressoSystems/espresso-sequencer/issues/2114

PR #2092, opened 2024-10-02
I01 validate state history upon initialization

### LINKED ISSUE #2072: I01: stateHistoryRetentionPeriod not validated
Affected Code: LightClient.sol (line 210)

Summary: The property stateHistoryRetentionPeriod is not validated
during initialization, but it requires a minimum of 1 hour when updat-
ing it (on lines 439-445). This inconsistency could result in the period
being initialized to less than 1 hour, leading to an unexpected state.

Suggestion: We suggest to validate whether the property
stateHistoryRetentionPeriod is greater than 1 hour at initialization time.
This would result in removing the revert condition whether
stateHistoryRetentionPeriod is less than 1 hour in
setstateHistoryRetentionPeriod if the value can only be increased.

alysiahuggins:
Closes #2072 

### This PR:
- validates the state history retention period upon contract initialization

### This PR does not:
- set a max state history retention period, it will be addressed in #2076 

### Key places to review:
- `initializeState` method

### How to test this PR:
- `forge test --match-test test_revertWhen_InvalidStateHistoryRetentionPeriodOnSetUp`

### Things tested
- stateHistoryRetentionPeriod less than 1 hour reverts upon contract initialization


contracts/src/LightClient.sol
```diff
@@ -182,27 +182,31 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
         LightClientState memory _genesis,
         StakeTableState memory _genesisStakeTableState,
         uint32 _stateHistoryRetentionPeriod
     ) internal {
-        // stake table commitments and threshold cannot be zero, otherwise it's impossible to
-        // generate valid proof to move finalized state forward.
-        // Whereas blockCommRoot can be zero, if we use special value zero to denote empty tree.
-        // feeLedgerComm can be zero, if we optionally support fee ledger yet.
+        // The viewNum and blockHeight in the genesis state must be zero to indicate that this is
+        // the
```
sveitser, line 194: This paragraph is strangely formatted.

contracts/src/LightClient.sol
```diff
@@ -182,27 +182,31 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        // generate valid proof to move finalized state forward. The stateHistoryRetentionPeriod
+        // must be
+        // at least 1 hour to ensure proper state retention.
         if (
             _genesis.viewNum != 0 || _genesis.blockHeight != 0
                 || BN254.ScalarField.unwrap(_genesisStakeTableState.blsKeyComm) == 0
                 || BN254.ScalarField.unwrap(_genesisStakeTableState.schnorrKeyComm) == 0
                 || BN254.ScalarField.unwrap(_genesisStakeTableState.amountComm) == 0
-                || _genesisStakeTableState.threshold == 0
+                || _genesisStakeTableState.threshold == 0 || _stateHistoryRetentionPeriod < 1 hours
```
sveitser, line 205: Can we define a constant for the minimum period of `1 hours` and then use that here and in `setstateHistoryRetentionPeriod`?

alysiahuggins: sure, we can define a constant. Historically, I've been trying to minimize the use of constants in upgradeable contracts to avoid storage layout issues (even though it's stored in the contract's bytecode, if the same variable is no longer a constant in a future version of the contract, it would mess up the storage layout). However, since we use the inheritance pattern for handling upgrades, we shouldn't run into that issue. Will proceed

contracts/src/LightClient.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/58f43df708454d978a40e7614f154abd819ecdb4

contracts/src/LightClient.sol
alysiahuggins: we spoke offline and agreed not to set a constant since constants can't be upgraded and this constant is one that does have the possibility of changing. Instead, extra comments were added to ensure that developers and users are aware of the min state history retention period https://github.com/EspressoSystems/espresso-sequencer/commit/58f43df708454d978a40e7614f154abd819ecdb4

alysiahuggins:
These changes were also added to this PR which was merged
https://github.com/EspressoSystems/espresso-sequencer/pull/2117/files#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292

PR #2095, opened 2024-10-02
validate chain config from db

imabdulbasit:
This PR validates the chain config fetched from the database during catchup

### DIFF
--- a/sequencer/src/catchup.rs
+++ b/sequencer/src/catchup.rs
@@ -276,7 +276,17 @@ where
         &self,
         commitment: Commitment<ChainConfig>,
     ) -> anyhow::Result<ChainConfig> {
-        self.db.get_chain_config(commitment).await
+        let cf = self.db.get_chain_config(commitment).await?;
+
+        if cf.commit() != commitment {
+            bail!(
+                "Invalid chain config: expected {:?}, got {:?}",
+                commitment,
+                cf.commit()
+            )
+        }
+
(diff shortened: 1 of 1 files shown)


sequencer/src/catchup.rs
```diff
@@ -276,7 +276,17 @@ where
         &self,
         commitment: Commitment<ChainConfig>,
     ) -> anyhow::Result<ChainConfig> {
-        self.db.get_chain_config(commitment).await
+        let cf = self.db.get_chain_config(commitment).await?;
+
+        if cf.commit() != commitment {
+            bail!(
```
imabdulbasit, line 282: maybe we panic in this case but this should never be true.

sveitser: I also think here it makes sense to panic because something went wrong in ways we do not anticipate and we currently cannot handle.

If we panic we should also tell node operators what to do next or what to investigate.

Also as usual this should be covered by tests.

imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/issues/2103

imabdulbasit: replaced with panic: https://github.com/EspressoSystems/espresso-sequencer/pull/2095/commits/4690af82e8251166f5b6cbc29976ca5269d7051a

sveitser (approved):

PR #2101, opened 2024-10-02
Add backport action

sveitser:
This makes it easier to backport PRs to other branches


tbro (approved):
Honestly I think cherry-picking is pretty easy. I guess this will do a better job of tracking these though. No harm in trying it out I guess.

sveitser:
Yes exactly I think it's good for visibility and saves the trouble of opening PRs. We used it in the nitro fork before and it worked well as far as I remember.

PR #2107, opened 2024-10-03
L02 Implementation of lagOverEscapeHatchThreshold deviates from comment

### LINKED ISSUE #2071: L02: Implementation of lagOverEscapeHatchThreshold deviates from comment
Affected Code: LightClient.sol (line 355-L399)

Summary: The current implementation allows for index 1 to be checked
and will mark prevUpdateFound as true if the condition is satisfied.

Suggestion: Consider either moving the if condition (on lines 381-383)
at the start of the while loop (on line 375), or modifying the code sim-
ilarly to the provided implementation which also improves on read-
ability

alysiahuggins:
Closes #2071 

### This PR:
- moves if statement so that it captures edge case accurately
- modified comments


contracts/src/LightClient.sol
```diff
@@ -362,40 +361,44 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
-            // We don't consider the lag time for the first two updates
+            // Skip the first two updates as per logic mentioned above.
             if (i < 2) {
                 break;
             }
 
-            // We've reached the first recorded block
-            if (i == stateHistoryFirstIndex) {
+            // Stop if we've already assessed the first recorded state in the history.
+            if (i < stateHistoryFirstIndex) {
```
Sneh1999, line 382: should this be  `if (i <= stateHistoryFirstIndex)`

alysiahuggins: not "=" because I still want to evaluate when the index  is equal to the stateHistoryFirstIndex

contracts/src/LightClient.sol
```diff
@@ -362,40 +361,44 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         uint256 updatesCount = stateHistoryCommitments.length;
 
         // Handling Edge Cases
-        // Edgecase 1: The block is in the future or
-        // before HotShot was live, allow for at least two updates before considering HotShot live
+        // Edgecase 1: blockNumber is greater than current block.number or
+        // less than 3 updates exist which means the history is insufficient to determine a lag.
         if (blockNumber > block.number || updatesCount < 3) {
```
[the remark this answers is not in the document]
alysiahuggins: it can be but the idea here is that we're giving after the first two updates we can consider the contract stable and thus we expect the updates to continue - i don't think it's necessary to be configurable

contracts/src/LightClient.sol
```diff
@@ -362,40 +361,44 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
         uint256 i = updatesCount - 1;
         while (!prevUpdateFound) {
-            if (stateHistoryCommitments[i].l1BlockHeight <= blockNumber) {
-                prevUpdateFound = true;
-                prevBlock = stateHistoryCommitments[i].l1BlockHeight;
-            }
-
-            // We don't consider the lag time for the first two updates
+            // Skip the first two updates as per logic mentioned above.
             if (i < 2) {
```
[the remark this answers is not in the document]
alysiahuggins: same response as the comment about

Sneh1999 (approved):
LGTM!

contracts/src/LightClient.sol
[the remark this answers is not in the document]
alxiong: I was little confused when reading the number 3, we should add more explanation of where does this number come from? does it come from HotShot consensus protocol or just a heuristic?

contracts/src/LightClient.sol
```diff
@@ -362,40 +361,44 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+            if (i < stateHistoryFirstIndex) {
                 break;
             }
+
+            // Find the first update where the block height is less than or equal to the given
+            // blockNumber.
+            if (stateHistoryCommitments[i].l1BlockHeight <= blockNumber) {
+                prevUpdateFound = true;
+                prevBlock = stateHistoryCommitments[i].l1BlockHeight;
+            }
```
alxiong, line 391: logic error, we should immediately `break;` once found, right?

alysiahuggins: yes that's right! there should be a break there

alysiahuggins:
@alxiong for your comment here - https://github.com/EspressoSystems/espresso-sequencer/pull/2107#discussion_r1808185748, the number is chosen as a buffer when hotShot first starts. The idea here is that during the buffer time (3 blocks), there isn't enough history to start checking for lags, since hotshot was only just starting up. But if you think it's uneccesssary, let me know

contracts/src/LightClient.sol
alysiahuggins: created a PR for this here https://github.com/EspressoSystems/espresso-sequencer/pull/2189

alxiong:
> The idea here is that during the buffer time (3 blocks), there isn't enough history to start checking for lags, since hotshot was only just starting up. But if you think it's uneccesssary, let me know

tbh, i find this number 3 a little contrived. I don't see a principled reason why this buffer time is necessary. 
It's not super clear from the code doc why 3 is necessary or sufficient (why not 5 for example?)
we can either a) remove this; b) add more comments

I incline towards option a)
what's your opinion @philippecamacho ?

alysiahuggins:
> > The idea here is that during the buffer time (3 blocks), there isn't enough history to start checking for lags, since hotshot was only just starting up. But if you think it's uneccesssary, let me know
> 
> tbh, i find this number 3 a little contrived. I don't see a principled reason why this buffer time is necessary. It's not super clear from the code doc why 3 is necessary or sufficient (why not 5 for example?) we can either a) remove this; b) add more comments
> 
> I incline towards option a) what's your opinion @philippecamacho ?

that's fair, here are code changes for option a. https://github.com/EspressoSystems/espresso-sequencer/commit/8889a1758a5dc66bcbf84e8d4b95d42eee30c368
@alxiong

PR #2108, opened 2024-10-03
CI: contracts, new cache

sveitser:
Currently `diff-test` fails to build in the contracts CI job on the  `release-gambit` branch.

Suspect failure to build `diff-test` is due to something being wrong with the cache because removing the `./target` directory made the error disappear locally.


sveitser:
Fixed on `release-gambit` now.

PR #2109, opened 2024-10-03
[WEEKLY RELEASE] HotShot - rc-0.5.77

lukeiannucci:
Weekly Release

### This PR:

Weekly merge include:
1.) File changes for the Vote Commitment for UTs to pass (see https://github.com/EspressoSystems/espresso-sequencer/pull/2111)
2.) Libp2p refactoring
3.) Adding table to store upgrade certificate that hotshot is able to load after a restart. Add a test to ensure we can store/load the upgrade certificate.


types/src/v0/traits.rs
```diff
@@ -615,6 +616,13 @@ impl<P: SequencerPersistence> Storage<SeqTypes> for Arc<P> {
     ) -> anyhow::Result<()> {
         (**self).append_quorum_proposal(proposal).await
     }
+
+    async fn update_decided_upgrade_certificate(
+        &self,
+        _decided_upgrade_certificate: Option<UpgradeCertificate<SeqTypes>>,
+    ) -> anyhow::Result<()> {
+        Ok(())
```
ss-es, line 624: I think we want to add `store_upgrade_certificate` to the `SequencerPersistence` trait then call that here

types/src/v0/traits.rs
```diff
@@ -461,6 +461,7 @@ pub trait SequencerPersistence: Sized + Send + Sync + 'static {
                 highest_voted_view,
                 saved_proposals,
                 high_qc,
+                None,
```
ss-es, line 464: I think we want a `load_upgrade_certificate` method in `SequencerPersistence` to be able to call in this function, so that we can pass the stored certificate here

ss-es (commented):
After adding the methods to the `SequencerPersistence` trait, I think you'll just need to update `sequencer/src/persistence/` to implement them. Someone from the sequencer could probably help with this

builder/src/non_permissioned.rs
```diff
@@ -195,7 +195,7 @@ impl BuilderConfig {
         let global_state_clone = global_state.clone();
         tracing::info!("Running permissionless builder against hotshot events API at {events_url}",);
         async_spawn(async move {
-            let res = run_non_permissioned_standalone_builder_service::<_, V>(
+            let res = run_non_permissioned_standalone_builder_service::<_, StaticVersion<0, 1>>(
```
imabdulbasit, line 198: please merge main into this branch. We are now passing `SequencerApiVersion` here

data/v1/messages.json
```diff
@@ -42,7 +42,7 @@
                 },
                 "signatures": null,
                 "view_number": 0,
-                "vote_commitment": "COMMIT~tzMwZg2x5QJTx7Di9eEvNC0P8CZ_iNX7LJ01DHA9XEdO"
+                "vote_commitment": "COMMIT~f11eqw2KFTy-i-EaXUfEC1v0CPo8gPcd2My10Y5zSNok"
```
sveitser, line 45: Why does this change, and it is intentional?

sequencer/api/migrations/V38__upgrade_certificate.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE upgrade_certificate (
+     -- The ID is always set to 0. Setting it explicitly allows us to enforce with every insert or
+    -- update that there is only a single entry in this table: the latest known state.
+    id INT PRIMARY KEY,
```
sveitser, line 4: @imabdulbasit I think we could enforce at the DB level that there is at most 1 upgrade certificate. Should we do that?

sequencer/src/persistence/fs.rs
```diff
@@ -698,6 +702,41 @@ impl SequencerPersistence for Persistence {
...
+        if !path.is_file() {
+            return Ok(None);
+        }
+        let bytes = fs::read(&path).context("read")?;
+        Ok(Some(
+            bincode::deserialize(&bytes).context("deserialize upgrade certificate")?,
+        ))
+    }
+
+    async fn store_upgrade_certificate(
```
sveitser, line 720: I think we should add some tests for this:

1. to verify that storing and loading works
2. to verify that we store at most 1 upgrade certificate 
3. to verify that cleanup works

@imabdulbasit can you advise @lukeiannucci on how to add these tests.

sequencer/api/migrations/V38__upgrade_certificate.sql
imabdulbasit: I think the only way to do it is to add a trigger which is fine but I don't know how frequently we are storing the upgrade certificate as the trigger will do a query on the table and then insert

data/v1/messages.json
ss-es: This is intentional! We missed a HotShot PR when cherry-picking into `release-gambit` which had a vote commitment fix, but we did not catch this until after we deployed `release-gambit` for cappuccino. Since we can't change what we've deployed, we reverted that change on HotShot `main`.

Note that the vote commitment here matches what's required to get the sequencer test to pass: https://github.com/EspressoSystems/espresso-sequencer/pull/2111

sequencer/api/migrations/V38__upgrade_certificate.sql
ss-es: It should be accessed very rarely. We (should) write exactly once per upgrade, and read only on node restart.

sveitser: @imabdulbasit What about something like what's proposed here: https://stackoverflow.com/a/25393923 ?

imabdulbasit: yeah, this should work fine. cc @lukeiannucci

lukeiannucci: made changes, can you verify if this looks good @imabdulbasit / @sveitser . CI seems good also added a test

data/v1/messages.json
sveitser: @ss-es thanks! This would be good information to add to the PR or commit message.

sequencer/api/migrations/V38__upgrade_certificate.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE upgrade_certificate (
+    id bool PRIMARY KEY DEFAULT true,
+    data BYTEA
+);
+REVOKE DELETE, TRUNCATE ON upgrade_certificate FROM public;
```
imabdulbasit, line 5: So this wouldn't allow garbage_collect() to work as we are revoking the delete permissions

lukeiannucci: yes we don't need to delete this certificate, I deleted the code for that after talking with @ss-es

imabdulbasit (approved):

PR #2122, opened 2024-10-04
M01 State History Update Comparison to upcoming block timestamp instead of last one in the array

### LINKED ISSUE #2119: M01: State history update error
Affected Code: LightClient.sol (line 322-L340)

Summary: updateStateHistory function incorrectly computes updates by
comparing the first and last elements instead of the first and newly
added element, potentially retaining outdated data. Additionally, only
the first outdated element is removed despite there possibly being
more outdated elements that are not checked.

alysiahuggins:
Closes #2119 

### This PR:
- changes the comparison in the `updateStateHistory` function to the upcoming block timestamp instead of the most recent one in the array


Sneh1999 (approved):
LGTM!

PR #2132, opened 2024-10-08
FIx prover panic

Sneh1999:
Closes #<ISSUE_NUMBER>

Fixes prover panic on main:
image

### This PR:

It uses the light client proxy address instead of light client implementation address


imabdulbasit:
I think we don't need this PR as proxy address is set to implementation contract address for light client in the env.
The prover panic is fixed by this PR
https://github.com/EspressoSystems/espresso-sequencer/pull/2141

PR #2133, opened 2024-10-08
improve dev build times for everyone

tbro:
Adds dev profile to `Cargo.toml`. It improves build times considerably. I don't think it hurts anything. So probably ok if CI passes. Feel free to run some cargo commands to see how you like it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,19 +144,6 @@ paste = "1.0"
 rand = "0.8.5"
 time = "0.3"
 trait-set = "0.3.0"
-
-[patch."https://github.com/EspressoSystems/HotShot.git"]
-hotshot = { git = "https://www.github.com/EspressoSystems/HotShot.git", tag = "0.5.75-patch1" }
-hotshot-builder-api = { git = "https://www.github.com/EspressoSystems/HotShot.git", tag = "0.5.75-patch1" }
-hotshot-orchestrator = { git = "https://www.github.com/EspressoSystems/HotShot.git", tag = "0.5.75-patch1" }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -156,3 +156,12 @@ hotshot-testing ={ git = "https://www.github.com/EspressoSystems/HotShot.git", t
 hotshot-types = { git = "https://www.github.com/EspressoSystems/HotShot.git", tag = "0.5.75-patch1" }
 libp2p-networking = { git = "https://github.com/EspressoSystems//HotShot.git", tag = "0.5.75-patch1" }
 hotshot-example-types = { git = "https://www.github.com/EspressoSystems/HotShot.git", tag = "0.5.75-patch1" }
+
+[profile.dev]
+opt-level = 1
```
ss-es, line 161: Is there a reason not to set this (and the other one) to just `opt-level = 0`?

Cargo.toml
```diff
@@ -156,3 +156,12 @@ hotshot-testing ={ git = "https://www.github.com/EspressoSystems/HotShot.git", t
...
+# Skip compiling the debug information in the first place.
+debug = 0
+# Skip linking debug information.
+strip = "debuginfo"
```
ss-es, line 165: is there a reason not to just do `strip = true`? (I don't actually know how significant this is)

tbro: I know that we don't want `debuginfo`. I'm not sure if we want `symbols` or not.

Cargo.toml
tbro: I don't remember why I set it to 1.

f329bdf1e61d7729d06d48934265b7957b46edaa

sveitser (approved):

PR #2134, opened 2024-10-09
increase timestamp tolerance to 4 seconds

sveitser:


tbro:
Should this PR also move timestamp validation earlier?

tbro:
> Should this PR also move timestamp validation earlier?

never mind I see that is here: https://github.com/EspressoSystems/espresso-sequencer/pull/2137

tbro (approved):
I don't see how this could cause any harm.

sveitser:
I think we will just "backport" the push to the release branch that sets it to 12 seconds instead.

PR #2148, opened 2024-10-10
fix: lower priority fee 

### LINKED ISSUE #2147: Set a lower priority fee for light client update
Currently we are setting a 3 Gwei priority fee on top of the base fee (e.g. [this tx](https://etherscan.io/tx/0x2ca703529390c357fff0249af60c481dc7083201739be7b6577ad9291d12b2df)), while in the same block there are more frugal tx that still got included with 0 priority fee (like [this](https://etherscan.io/tx/0x34ffe08eaa237d5ee12e3b03fb062f3e1984d76129cef859a3add2512c415657))

Since we don't care about strict or fast update on light client contract, we should be save money by setting a low or zero priority fee.

Other resources:
- etherscan's [gas oracle](https://docs.etherscan.io/api-endpoints/gas-tracker#get-gas-oracle) (e.g. [API response](https://api.etherscan.io/api?module=gastracker&action=gasoracle))
- EIP 1559 variables and definitions [here](https://hackmd.io/@tvanepps/1559-wallets#1559-Variables-and-their-functions)

we will need to test this out in testnet first before deploying anything to production.

alxiong:
Closes #2147 

### This PR:

- ~~Switch to the right version of PlonkVerifier (in the `deployer.rs` we incorrectly deployed PlonkVerifier2.sol which is secure but less efficient)~~
- Use lower priority fee based on live gas oracle when sending out light client update transaction

### Comments
- there is no dedicated price oracle for testnet (e.g. sepolia.etherscan.io), to avoid complicated logic, we simply keep using the same mainnet stats-informed gas pricing.


utils/src/deployer.rs
```diff
@@ -190,7 +190,7 @@ pub async fn deploy_light_client_contract<M: Middleware + 'static>(
     let plonk_verifier = contracts
         .deploy_tx(
             Contract::PlonkVerifier,
-            PlonkVerifier2::deploy(l1.clone(), ())?,
+            PlonkVerifier::deploy(l1.clone(), ())?,
```
sveitser, line 193: @alxiong @alysiahuggins I think this now needs to be done via a smart contract upgrade (resp. downgrade). Changing it here in the code is not enough to use it on decaf and mainnet.

alxiong: update: Since we are redeploying it, we would just run this.
(i did a project wise search, I don't see any accidental use of PV2 anymore)

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +316,31 @@ pub async fn submit_state_and_proof(
...
     let proof: ParsedPlonkProof = proof.into();
     let new_state: ParsedLightClientState = public_input.into();
-    let tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    // frugal gas price: set to SafeGasPrice based on live price oracle
+    // safe to be included with low priority position in a block
+    let gas_info = reqwest::get("https://api.etherscan.io/api?module=gastracker&action=gasoracle")
+        .await?
+        .json::<serde_json::Value>()
+        .await?;
```
sveitser, line 325: I think we should handle the case where etherscan we get an error when calling the etherscan API.

sveitser (changes requested):

hotshot-state-prover/src/service.rs
alxiong: make sense, changed in [f0cf7aa](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/f0cf7aa08b69693e3fdd11d5bc17f4b8dbf7be26)

now, only set the price explicitly when etherscan successfully return; else simply ignore and create tx with default gas (whatever `ethers` decide to do)

alxiong: allow for falling back the default when Etherscan changes their API: [c835231](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/c835231830979279594e789707cf0a04d3239c59)

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +340,39 @@ pub async fn submit_state_and_proof(
...
+            if let TypedTransaction::Eip1559(inner) = &mut tx.tx {
+                // both safe unwraps below assuming correct etherscan stream
+                let safe_max_fee: U256 = parse_units(res.result.safe_max_fee.clone(), "gwei")
+                    .unwrap()
+                    .into();
+                let base_fee: U256 = parse_units(res.result.base_fee.clone(), "gwei")
+                    .unwrap()
+                    .into();
+                inner.max_fee_per_gas = Some(safe_max_fee);
+                inner.max_priority_fee_per_gas = Some(safe_max_fee - base_fee);
```
sveitser, line 365: @alxiong It looks like this could underflow it would probably lead to the transaction being rejected because we don't have the funds but let's not rely on that.

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +340,39 @@ pub async fn submit_state_and_proof(
...
+        // parse etherscan response,
+        // if etherscan API changes, we log the warning and fallback to default (by falling through)
+        if let Ok(res) = res.json::<EtherscanPriceOracleResponse>().await {
+            if let TypedTransaction::Eip1559(inner) = &mut tx.tx {
+                // both safe unwraps below assuming correct etherscan stream
+                let safe_max_fee: U256 = parse_units(res.result.safe_max_fee.clone(), "gwei")
+                    .unwrap()
+                    .into();
+                let base_fee: U256 = parse_units(res.result.base_fee.clone(), "gwei")
+                    .unwrap()
```
sveitser, line 362: Suggest removing unwraps, maybe add a helper function to parse the API response into a type that is convenient for us to work with.

alxiong: should we just return error (propagate err up) or still use unwrap but hide inside some helper function?

hotshot-state-prover/src/service.rs
alxiong: see if you like [6106914](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/6106914b3c282978b1e5a4036c548a51c0d117fa)

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +368,37 @@ pub async fn submit_state_and_proof(
...
-    let tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    let mut tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    // TODO: (alex) centralize this to a dedicated ethtxmanager with gas control
+    // frugal gas price: set to SafeGasPrice based on live price oracle
+    // safe to be included with low priority position in a block.
+    // ignore this if etherscan fail to return
+    if let Ok(res) =
+        reqwest::get("https://api.etherscan.io/api?module=gastracker&action=gasoracle").await
```
jbearer, line 379: This needs to be a configurable URL

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +368,37 @@ pub async fn submit_state_and_proof(
...
+                    if let TypedTransaction::Eip1559(inner) = &mut tx.tx {
+                        inner.max_fee_per_gas = Some(gas_info.safe_max_fee);
+                        inner.max_priority_fee_per_gas =
+                            Some(gas_info.safe_max_fee - gas_info.base_fee);
+                        tracing::info!("Setting maxFeePerGas to: {}", gas_info.safe_max_fee);
+                    }
+                }
+            }
+        } else {
+            tracing::warn!("!! Etherscan Price Oracle API changed !!")
```
jbearer, line 399: Probably want to include the returned error in this message

hotshot-state-prover/src/service.rs
jbearer: Ideally it should be an optional URL. The logic really isn't that complicated if the L1 doesn't have a gas price oracle: we just skip this block if the configured URL is none (using the default gas price estimation from ethers.rs). This would be a much better behavior for Sepolia than following completely unrelated gas trends from mainnet

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +368,37 @@ pub async fn submit_state_and_proof(
...
     let new_state: ParsedLightClientState = public_input.into();
-    let tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    let mut tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    // TODO: (alex) centralize this to a dedicated ethtxmanager with gas control
+    // frugal gas price: set to SafeGasPrice based on live price oracle
+    // safe to be included with low priority position in a block.
+    // ignore this if etherscan fail to return
+    if let Ok(res) =
```
jbearer, line 378: If this call returns `Err` it looks like we silently ignore it? We should log the error if this happens

hotshot-state-prover/src/service.rs
```diff
@@ -315,7 +368,37 @@ pub async fn submit_state_and_proof(
...
+        // parse etherscan response,
+        // if etherscan API changes, we log the warning and fallback to default (by falling through)
+        if let Ok(res) = res.json::<EtherscanPriceOracleResponse>().await {
+            tracing::info!("Etherscan gas oracle: {:?}", res.result);
+            match TryInto::<GasInfo>::try_into(res.result) {
+                Err(e) => {
+                    tracing::warn!("{}", e)
+                }
+                Ok(gas_info) => {
+                    if let TypedTransaction::Eip1559(inner) = &mut tx.tx {
```
jbearer, line 390: Are we sure this is the right transaction variant returned by the `new_finalized_state` binding? I feel like if the transaction variant doesn't match we should at least log something

alxiong: Yes it's our tx type. All tx after  London upgrade (since [Aug 2021](https://ethereum.org/en/history/#london)) is this type.

hotshot-state-prover/src/service.rs
jbearer: Sounds good. Just to clarify, we don't care _that_ much about testnet gas, but it is still a scarce resource, and moreover the mainnet and testnet gas prices are completely unrelated, so we might have transactions failing/not getting mined if testnet gas price is spiking and we are setting too low of a price, or if mainnet gas price is spiking we might burn through our sep eth for no reason

alxiong:
I just found out that ethers-rs already has [some code](https://github.com/gakonst/ethers-rs/blob/6e2ff0ef8af8c0ee3c21b7e1960f8c025bcd5588/ethers-etherscan/src/gas.rs) that fetch from etherscan price oralce. I'll try to use those instead of reinventing the wheel.

alxiong:
this is annoying: ethers-middleware define [this field](https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/gas_oracle/blocknative.rs#L46) from BlockNative's feed to be `u64`, however, it's actually a float64 (e.g. [sample](https://api.blocknative.com/gasprices/blockprices), see the `price` field is a float).
I'm not sure if Blocknative changed its feed or ethers-rs has never tested their code. Either way, we need to choose a different one.

hotshot-state-prover/src/service.rs
alxiong: done in [db9ab26](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/db9ab26f48ec7b4f2f37faa0cc815d7f20c5ae6d)

hotshot-state-prover/src/service.rs
alxiong: done in [db9ab26](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/db9ab26f48ec7b4f2f37faa0cc815d7f20c5ae6d)

hotshot-state-prover/src/service.rs
alxiong: refactored in [384fc7f](https://github.com/EspressoSystems/espresso-sequencer/pull/2148/commits/384fc7f9c80eb0221d0057e4efee95d83b5fb831)

1. only invoke gas oracle if the provider is detected to be on mainnet
2. use blocknative code (directly copied from ethers-rs, with minimal changes to make it in sync with the latest feed data format, see https://github.com/EspressoSystems/espresso-sequencer/pull/2148#issuecomment-2415359785). The URL is still hardcoded, but we can easily switch to a different price oracle in the future if we want.

hotshot-state-prover/src/service.rs
alxiong: this is outdated, see the latest commits that use blocknative and `trait PriceOracle` from ethers-rs

jbearer (approved):

hotshot-state-prover/src/service.rs
```diff
@@ -306,16 +308,34 @@ pub async fn read_contract_state(
...
+                    inner.max_priority_fee_per_gas = Some(priority_fee);
+                    tracing::info!(
+                        "Setting maxFeePerGas: {}; maxPriorityFeePerGas to: {}",
+                        max_fee,
+                        priority_fee
+                    );
+                }
+            }
+            Err(e) => {
+                tracing::warn!("!! BlockNative Price Oracle failed: {}", e);
```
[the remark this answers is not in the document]
alxiong: by default, there's a tx filler. which will have some heuristic of choosing gas price, but will be less frugal (i.e. instead of choosing `SafeLow`, it will choose `Standard`)

hotshot-state-prover/src/service.rs
```diff
@@ -306,16 +308,34 @@ pub async fn read_contract_state(
...
-
     // prepare the input the contract call and the tx itself
     let proof: ParsedPlonkProof = proof.into();
     let new_state: ParsedLightClientState = public_input.into();
-    let tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    let mut tx = contract.new_finalized_state(new_state.into(), proof.into());
+
+    // only use gas oracle for mainnet
+    if contract.client_ref().get_chainid().await?.as_u64() == 1 {
```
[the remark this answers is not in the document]
alxiong: right now, the prover is only sending light client updates to L1, not yet to Arbitrum, we can add a case for Arbitrum when we also deploy a light client clone on ArbOne.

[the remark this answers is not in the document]
alysiahuggins: got it - where will be the best place to leave a reminder? in the comments?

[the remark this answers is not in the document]
alxiong: i have added a reminder in https://github.com/EspressoSystems/espresso-sequencer/issues/2142#issue-2577095967, IMO, it should be part of the alloy-rs migration effort, which have mid-priority level.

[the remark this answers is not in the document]
alysiahuggins: ok great!

alysiahuggins (approved):

alysiahuggins:
a nice to have would be a test for this added functionality

utils/src/blocknative.rs
```diff
@@ -0,0 +1,150 @@
...
+use ethers::{
+    middleware::gas_oracle::{from_gwei_f64, GasCategory, GasOracle, GasOracleError, Result},
+    types::U256,
+};
+use reqwest::{header::AUTHORIZATION, Client};
+use serde::Deserialize;
+use std::collections::HashMap;
+use url::Url;
+
+const URL: &str = "https://api.blocknative.com/gasprices/blockprices";
```
alysiahuggins, line 15: checking out rate limits on [blocknative](https://docs.blocknative.com/gas-prediction/gas-platform#rate-limits), they say the following
```
Each endpoint is limited to provide updated results once per second for paid tier API keys, or every 5 seconds for free tier API keys. Polling faster than these rates may deliver stale data.
``` 
Are we on the free tier and i reckon we're not likely to hit the rate limit?

alxiong: the rate limit is not great, but our prover is only sending once per hour, we are not hitting this limit if we are just querying it once and send tx with it. But if later we decided to send multiple queries in sequence to wait for favorable gas, then that logic might require paid tier. 

honestly, right now, i just want to merge something simple that at least provide some lower gas price setting. We can also constantly monitor the datadog log, if we see too many rate limited errors, then we can consider upgrading from free tier.

alysiahuggins: got it! the rate limit for the free tier sounds sufficient for now

sveitser (approved):

alxiong:
@jbearer should I merge this or should I wait for https://github.com/EspressoSystems/espresso-sequencer/pull/2189 to fix the slow test?

PR #2154, opened 2024-10-10
2127 test rollback strategies for smart contracts

### LINKED ISSUE #2127: test rollback strategies for smart contracts
- e.g. upgrade of LightClientV1 to LightClientV2 and then a rollback to LightClientV1

alysiahuggins:
Closes #2127 
The light client contract and fee contract can be upgraded but we need to have a roll back strategy. The act of upgrading a contract means that the contract proxy now points to a new implementation address. So an upgradable contract usually has two addresses, a proxy (the address users interact with) and an implementation (the contract with the business logic). 
Let's say the contract is upgraded and we noticed degraded performance, we may want to rollback to the previous implementation. 

The aim of this PR is to use foundry tests to prove that rolling back a contract is possible. 

### This PR:
- adds downgrade test script for rolling back
- performs rolling back in tests
- performs failing roll back in test so show that you can only rollback to a valid implementation address
- changes some require statements to revert statements in the test scirpts (this is unrelated to the above)

### This PR does not:
- add documentation about rolling back
- add auto checks or auto rollbacks

### Key places to review:
- downgrade script
- downgrade tests

### How to test this PR: 
- `forge test --match-test test_expectRevert_RollbackV2toInvalidAddress`
- `forge test --match-test testRollbackV2toV1`


Sneh1999 (approved):
LGTM!
As discussed, it can be nice if we have a E2E test which works on a forked network and is able to downgrade and upgrade the contracts

alysiahuggins:
Created issue for E2E tests https://github.com/EspressoSystems/espresso-sequencer/issues/2174

PR #2160, opened 2024-10-10
Fix fee catchup after restart

jbearer:
This PR reproduces and will fix an issue discovered in staging, in which, after a staggered restart of all nodes, the network has forgotten the fee state for all accounts except those which actively produced blocks during the restart. Since the state is needed for undecided blocks, it cannot directly be obtained from persisted merklized state, which is only available for decided blocks.

# This PR:
* Adds a restart-style test reproducing the issue
* Adds a new catchup endpoint to fetch multiple accounts at once, for performance
* Modifies SQL-based fee account catchup to reconstruct fee state by searching back to the last decided state and replaying headers
* Caches results of successful reconstruction in HotShot's validated state map
* Decouples storage from API layer by creating `CatchupStorage` trait, distinct from `CatchupDataSource`. This allows us to get more information when querying storage, such as the leaf corresponding to fetched account state, which in turn ensures we are always able to cache fetched state in the HotShot view map
* Adds a background task to fetch missing quorum proposals that are parents of proposals we have. This ensures that every node will eventually have every proposal in a chain between valid proposals and the anchor, which is required for replay to work

# Key places to review
* [Regression test](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-f1cd73a8c7e576c13792f84c6438621c0452e49ff29b62214faec7c135008581R170)
* [State reconstruction from decided state](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-f76aae9b3a41d6643e9fa0e19e5eee26c71ccd6262e4063a55da0dda1d3a1226R251)
* [Proposal fetching task](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-c0044626ec1b765993f43c033167cd10f0a8317ad2e320033f8929db64bba565R412)
* [Caching of successful lookups in catchup endpoint](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-9c5c25f61c89b1c60988253125d7d03fb13becb07c9242f5966669ecfa00d373R236)


bfish713 (commented):
I reviewed the code which uses the new `request_proposal` fn provided by Hotshot and it looks correct to me

jbearer:
@bfish713 can you also look at the [state map update](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-9c5c25f61c89b1c60988253125d7d03fb13becb07c9242f5966669ecfa00d373R236-R291) during catchup? (This basically acts as a cache for successful catchup requests)

sequencer/api/catchup.toml
```diff
@@ -27,6 +27,20 @@ the proof is a Merkle _non-membership_ proof.
 ```
 """
 
+[route.accounts]
+PATH = ["/:height/:view/accounts"]
+":height" = "Integer"
+":view" = "Integer"
+METHOD = "POST"
+DOC = """
+Bulk version of `/:height:view/account`. The request body should be a JSON array consisting of
```
sveitser, line 36:
```suggestion
Bulk version of `/:height/:view/account`. The request body should be a JSON array consisting of
```

bfish713:
> @bfish713 can you also look at the [state map update](https://github.com/EspressoSystems/espresso-sequencer/pull/2160/files#diff-9c5c25f61c89b1c60988253125d7d03fb13becb07c9242f5966669ecfa00d373R236-R291) during catchup? (This basically acts as a cache for successful catchup requests)

Took a look and that all looks fine to me

bfish713 (approved):

PR #2165, opened 2024-10-11
Adding API point to get fee balance per account

### LINKED ISSUE #2105: API endpoint to get current balance of a fee account
Should return the balance as a plain integer, not a tagged base64 encoded field element

fkrell:
Closes #2105 

### This PR:
Adds get_fee_balance endpoint to the sequencer API to retrieve the current balance of an account.
This allows to detect whenever the Builder is low on funds and issue the corresponding alert

### Key places to review:
sequencer/api/merklized_state.toml: defines new endpoint.
sequencer/src/api/endpoints.rs: extends merklized state API with the new endpoint and define its behaviour.
In particular, given address as an input it retrieves current height from state, and uses  height and address to retrieve merkle path from merklized state. Balance is extracted from leaf of this path.

Manual test is to run the native demo locally, retrieve the builder address from http://localhost:31003/v0/block_info/builderaddress, then use the address to call the new endpoint
http://localhost:24001/v0/fee-state/fee_balance/<ADDRESS>

### DIFF
--- a/sequencer/api/merklized_state.toml
+++ b/sequencer/api/merklized_state.toml
@@ -1,0 +1,4 @@
+[route.getfeebalance]
+PATH = ["/fee_balance/:address"]
+":address" = "Literal"
+DOC = "Get current balance in address account."


sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
...
+    api.get("getfeebalance", move |req, state| {
+        async move {
+            let address = req.string_param("address")?;
+            let height = state.get_last_state_height().await?;
+            let snapshot = Snapshot::Index(height as u64);
+            let key = address
+                .parse()
+                .map_err(|_| merklized_state::Error::Custom {
+                    message: "failed to parse Key param".to_string(),
+                    status: StatusCode::INTERNAL_SERVER_ERROR,
```
sveitser, line 75: This should be a 400 / BadRequest error because it's the client's fault.

sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
...
 
+pub(super) fn get_balance<State, Ver>() -> Result<Api<State, merklized_state::Error, Ver>>
+where
+    State: 'static + Send + Sync + ReadState,
+    Ver: 'static + StaticVersionType,
+    <State as ReadState>::State: Send
+        + Sync
+        + MerklizedStateDataSource<SeqTypes, FeeMerkleTree, 256>
+        + MerklizedStateHeightPersistence,
+    //for<'a> <<UniversalMerkleTree<FeeAmount,Sha3Digest,FeeAccount,256,Sha3Node> as hotshot_query_service::merklized_state::MerklizedState<Types, ARITY>>::Commit as TryFrom<&'a TaggedBase64>>::Error: Display,
```
sveitser, line 57: Remove the comment?

sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
...
+
+    api.get("getfeebalance", move |req, state| {
+        async move {
+            let address = req.string_param("address")?;
+            let height = state.get_last_state_height().await?;
+            let snapshot = Snapshot::Index(height as u64);
+            let key = address
+                .parse()
+                .map_err(|_| merklized_state::Error::Custom {
+                    message: "failed to parse Key param".to_string(),
```
sveitser, line 74: Failed to parse address?

sequencer/api/merklized_state.toml
```diff
@@ -0,0 +1,4 @@
+[route.getfeebalance]
+PATH = ["/fee_balance/:address"]
+":address" = "Literal"
+DOC = "Get current balance in address account."
```
sveitser, line 4: "Balance in fee state" or something like that?

sequencer/api/merklized_state.toml
```diff
@@ -0,0 +1,4 @@
+[route.getfeebalance]
+PATH = ["/fee_balance/:address"]
```
jbearer, line 2: As a convention we usually use hyphens in URL segments eg `fee-balance`

sequencer/src/api/endpoints.rs
jbearer, line 57: Delete before merging

sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
     pub transactions: Vec<Transaction>,
 }
 
+pub(super) fn get_balance<State, Ver>() -> Result<Api<State, merklized_state::Error, Ver>>
+where
+    State: 'static + Send + Sync + ReadState,
+    Ver: 'static + StaticVersionType,
+    <State as ReadState>::State: Send
+        + Sync
+        + MerklizedStateDataSource<SeqTypes, FeeMerkleTree, 256>
```
sveitser, line 55: Define a constant for the 256 arity?

sequencer/api/merklized_state.toml
sveitser: And mention that we are expecting an ethereum address in hex format.

sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
...
+    let extension = toml::from_str(include_str!("../../api/merklized_state.toml"))?;
+    options.extensions.push(extension);
+
+    let mut api =
+        merklized_state::define_api::<State, SeqTypes, FeeMerkleTree, Ver, 256>(&options)?;
+
+    api.get("getfeebalance", move |req, state| {
+        async move {
+            let address = req.string_param("address")?;
+            let height = state.get_last_state_height().await?;
```
sveitser, line 69: I think `height` should be a parameter one can specify. But it can default to latest.

sequencer/src/api/options.rs
```diff
@@ -356,7 +356,7 @@ impl Options {
             // Initialize merklized state module for fee merkle tree
             app.register_module(
                 "fee-state",
-                endpoints::merklized_state::<N, P, _, FeeMerkleTree, _, 256>()?,
+                endpoints::get_balance::<_, SequencerApiVersion>()?,
```
sveitser, line 359: It's probably fine but we are replacing and API so some endpoints will be removed. I guess nobody needs them but then I'm wondering why we had it in the first place 🤔

sequencer/src/api/options.rs
```diff
@@ -356,7 +356,7 @@ impl Options {
             // Initialize merklized state module for fee merkle tree
```
sveitser, line 356: Outdated comment

sveitser (changes requested):

sequencer/src/api/endpoints.rs
```diff
@@ -46,6 +46,42 @@ pub struct NamespaceProofQueryData {
     pub transactions: Vec<Transaction>,
 }
 
+pub(super) fn get_balance<State, Ver>() -> Result<Api<State, merklized_state::Error, Ver>>
```
sveitser, line 49: Add at least a TODO comment that mentions that this code is currently not covered by any tests.

sveitser:
Did manually test it and that looks good 👍

sequencer/src/api/options.rs
jbearer: The new API should have all the endpoints of the original one, because the new endpoint is added as an extension to the hotshot-query-service `merklized_state` API

sequencer/src/api/endpoints.rs
fkrell: I cannot compile use FeeMerkleTree::ARITY

error[E0747]: type provided when a constant was expected

sequencer/src/api/endpoints.rs
fkrell: there is a test, as part of slow_test_merklized_state_api test

sequencer/src/api/endpoints.rs
fkrell: I think Mathis is right, we need to define the constant.
even in the type alias FeeMerkleTee we use 256 as the type parameter for ARITY

sequencer/api/merklized_state.toml
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

sequencer/src/api/endpoints.rs
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

sequencer/src/api/endpoints.rs
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

sequencer/src/api/endpoints.rs
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

sequencer/src/api/endpoints.rs
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

sequencer/api/merklized_state.toml
fkrell: done in d19a139fb2599f6147bd6d5413adc5f0891338e3

tbro (approved):
Looks like all the comments  were addressed.

sequencer/src/api/endpoints.rs
fkrell: @sveitser I'll put this in another ticket so we can merge the current version. I changed the api path to "/fee-balance/latest/:address" so that we don't have to change it later when we add the height as parameter as well.

sveitser (approved):

PR #2170, opened 2024-10-11
Replica Validation Refactor

### LINKED ISSUE #2090: Review and refactor state validation logic
Some quick fixes to `validate_and_apply_header` in #2089 have left it in a disorganized state. We now have some `async` calls which need to be dealt with, but the primary validation function is synchronous. We should think about the best way to organize this. Ideally each validation step would be represented by a variant of `ProposalValidationError` and in `validate_and_apply_header` we would log those as we currently do with `validate_proposal`.

tbro:
Closes #2090 

### This PR:
Refactors contents of `validate_and_apply_header` by introducing two new internal (non-pub) types. This should facilitate:

  * comprehension
  * documentation
  * testing

Basically this PR:

  * adds `Proposal` and `ValidatedTransition` types
  * moves validation steps to isolated methods.
  * adds doc comments to methods
  * updates error enum
  * adds many tests

## Review

Most important changes are isolated `./types/src/v0/impls/state.rs`. Many unit tests have been added there. Tests should pass and you should be convinced that there is sufficient coverage. Small changes were made to `./types/src/v0/impls/header.rs` where we previously had testing for this validation in the form of high level *success cases* and *eror cases* tests. The success case test has been renamed and the error case test removed in favor of the more comprehensive error case tests in the new unit tests. 

## Review Documentation

You can run `just doc` and navigate here in your browser:
`./target/doc/espresso_types/v0/impls/state/struct.ValidatedTransition.html`


zacshowa (commented):
I know this is a WIP, but I left some general discussion points from my first read. I think that this will be a great use of the type system to make validation more readable.

types/src/v0/impls/state.rs
```diff
@@ -252,122 +298,244 @@ impl ValidatedState {
...
+        self.validate()?;
+        self.wait_for_l1_head().await;
+        self.wait_for_finalized_block().await?;
+        Ok(self)
     }
+    /// Top level validation routine. Performs all validation units in
+    /// the given order.
+    // TODO add order to doc
+    fn validate(&self) -> Result<(), ProposalValidationError> {
+        // TODO review order
```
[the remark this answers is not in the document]
tbro: Generally speaking, I think you want the things you can invalidate the most quickly to come first. This way you avoid doing a bunch of work on steps that pass, only to invalidate at the end. There are other considerations though. Timestamps are time sensitive, so validation could potentially be affected by steps that take time. Fee validation requires network calls and waiting. So some waiting is required for some validation steps. There is similar complexity to catchup. So there is some complexity we have to navigate.

types/src/v0/impls/state.rs
```diff
@@ -252,122 +298,244 @@ impl ValidatedState {
...
-    let ValidatedState {
-        block_merkle_tree,
-        fee_merkle_tree,
-        ..
-    } = state;
+        // Validate timestamp hasn't drifted too much from system time.
+        let system_time: u64 = OffsetDateTime::now_utc().unix_timestamp() as u64;
+        // TODO 2 seconds of tolerance should be enough for reasonably
+        // configured nodes, but we should make this configurable.
+        if self.proposal.header.timestamp().abs_diff(system_time) > 2 {
```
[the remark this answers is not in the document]
tbro: Requires some though. Putting it in chain config means we require an upgrade to modify it. Maybe we can just pass it in on the command line.

types/src/v0/impls/state.rs
```diff
@@ -680,69 +845,39 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
-
-            if finalized != proposed_finalized {
-                tracing::error!("Invalid proposal: l1_finalized mismatch");
+        // validate the proposal.
+        // TODO review clones and ownership. We
+        // should only have 1 clone of validated state. It might be passed back here.
+        let validated_state = match ValidatedTransition::new(
+            validated_state,
+            parent_leaf.clone(),
+            instance.l1_client.clone(),
```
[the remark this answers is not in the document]
tbro: I don't believe `l1_client` makes any network connections on its own... you are required to call one of the provided methods for it to connect to anything. I also don't know if we ultimately want this clone. We could possibly pass `&NodeState` at appropriate places. Also keep in mind the types here are pretty hap hazard so far. We may still change some things, so probably not worth going down a rabbit hole over this. But before it gets merged we definitely need to pin down all the shapes.

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
zacshowa: Ah, I wasn't aware the chain config was immutable between upgrades. What's the reasoning for that?

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
zacshowa: I agree that it's fine for now and as things settle a bit more we can work on optimizing that front. It seems like ValidatedTransition doesn't need many mutable borrows, so we likely don't need to take ownership over most of the data it holds. Then the question becomes memory trade off vs the complexity of reading lifetimes.

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
zacshowa: Hmm. This makes sense. Thanks for the insight!

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
tbro: It's meant to be immutable configuration for the block_chain. Part of blockchain the safety concerns I think.

types/src/v0/impls/state.rs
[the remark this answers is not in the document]
tbro: I think so. 1ccf3c092a7e42cdfc04fb671b5a281a383758a1

types/src/v0/impls/header.rs
```diff
@@ -1612,23 +1490,23 @@ mod test_headers {
...
-            &proposal.clone(),
-            &vid_common,
-        )
-        .unwrap();
 
-        assert_eq!(
-            proposal_state.block_merkle_tree.commitment(),
-            proposal.block_merkle_tree_root()
-        );
+        // ValidatedTransition::new(
```
sveitser, line 1494: Is this intentionally left commented out?

types/src/v0/impls/state.rs
```diff
@@ -253,130 +303,329 @@ impl ValidatedState {
...
+    }
+    /// Ensure that L1 Head on proposal is not decreasing.
+    fn validate_l1_head(&self, parent_l1_head: u64) -> Result<(), ProposalValidationError> {
+        if self.header.l1_head() < parent_l1_head {
+            return Err(ProposalValidationError::DecrementingL1Head);
         }
+        Ok(())
+    }
+    /// Validates [`ChainConfig`] of proposal against that stored in
+    /// state by comparing commitments.
```
sveitser, line 325:
```suggestion
    /// Validates [`ChainConfig`] of proposal is equal to the one stored in
    /// state by comparing commitments.
```

types/src/v0/impls/state.rs
```diff
@@ -253,130 +303,329 @@ impl ValidatedState {
...
+                proposal_root: self.header.block_merkle_tree_root(),
+            });
+        }
+
+        Ok(())
+    }
+}
+/// Type to hold cloned validated state and provide validation methods.
+#[derive(Debug)]
+pub(crate) struct ValidatedTransition<'a> {
```
sveitser, line 390: A bit of a nitpick: the transition is only validated if `validate()` does not return an error but given current naming one might expect it to be validated upon creation unless one actually inspects the return type. I'm totally fine with the current name just thinking maybe we can make it even clearer at some point.

types/src/v0/impls/state.rs
```diff
@@ -253,130 +303,329 @@ impl ValidatedState {
...
+        self.proposal
+            .validate_timestamp_non_dec(self.parent.timestamp())?;
 
-    let ValidatedState {
-        block_merkle_tree,
-        fee_merkle_tree,
-        ..
-    } = state;
+        // Validate timestamp hasn't drifted too much from system time.
+        // Do this check first so we don't add unnecessary drift.
```
sveitser, line 560: Nit: comment is now potentially misleading because in this function the drift check is done last.

types/src/v0/impls/state.rs
```diff
@@ -253,130 +303,329 @@ impl ValidatedState {
         Ok(())
     }
 }
+/// Block Proposal to be verified and applied.
+#[derive(Debug)]
+pub(crate) struct Proposal<'a> {
+    header: &'a Header,
+    block_size: u32,
```
sveitser, line 310: There is also a Proposal type in hotshot that we use in this repo, so I think we can use a different name here. `ProposalValidationInput`? A bit of a mouthful :roll_eyes:

types/src/v0/impls/state.rs
```diff
@@ -253,130 +303,329 @@ impl ValidatedState {
...
+    pub(crate) fn validate(self) -> Result<Self, ProposalValidationError> {
+        self.validate_timestamp()?;
+        self.validate_builder_fee()?;
+        self.validate_height()?;
+        self.validate_chain_config()?;
+        self.validate_block_size()?;
+        self.validate_fee()?;
+        self.validate_fee_merkle_tree()?;
+        self.validate_block_merkle_tree()?;
+        self.validate_proposed_finalized()?;
```
sveitser, line 483:
```suggestion
        self.validate_l1_finalized()?;
```

types/src/v0/impls/state.rs
```diff
@@ -516,12 +771,9 @@ impl ValidatedState {
...
 
         let mut delta = Delta::default();
+        validated_state.apply_proposal(&mut delta, parent_leaf, l1_deposits);
 
-        let mut validated_state =
-            apply_proposal(&validated_state, &mut delta, parent_leaf, l1_deposits);
-
-        charge_fee(
-            &mut validated_state,
+        validated_state.charge_fees(
```
sveitser, line 776: Should we do the fee charging as part of `apply_proposal`?

types/src/v0/impls/state.rs
```diff
@@ -895,9 +1057,110 @@ mod test {
...
+                Header::V1(_) => panic!("You called `Header.next()` on unimplemented version (v1)"),
+                Header::V2(parent) => Header::V2(v0_2::Header {
+                    height: parent.height + 1,
+                    timestamp: OffsetDateTime::now_utc().unix_timestamp() as u64,
+                    ..parent.clone()
+                }),
+                Header::V3(_) => panic!("You called `Header.next()` on unimplemented version (v3)"),
+            }
+        }
+        /// Replaces builder signature w/ invalid one.
```
sveitser, line 1097:
```suggestion
        /// Replaces builder signature w/ a valid one from a random key pair
```
?

types/src/v0/impls/state.rs
```diff
@@ -895,9 +1057,110 @@ mod test {
...
+                }),
+                Header::V3(_) => panic!(
+                    "You called `Header.invalid_builder_signature()` on unimplemented version (v3)"
+                ),
+            }
+        }
+    }
+
+    impl<'a> ValidatedTransition<'a> {
+        async fn mock(instance: NodeState, parent: &'a Header, proposal: Proposal<'a>) -> Self {
```
sveitser, line 1152: Remove async?

types/src/v0/impls/state.rs
```diff
@@ -1259,4 +1855,112 @@ mod test {
 
         validate_builder_fee(&header).unwrap();
     }
+
+    // #[async_std::test]
```
sveitser, line 1859: Commented block

sveitser (approved):
LGTM. I think this is a great improvement in terms of readability. Just some suggestions.

I also think @0xkato should review this PR as well.

types/src/v0/impls/header.rs
tbro: Humm I thought this was previously commented in relation to the above TODO... Maybe I got confused will double check.

types/src/v0/impls/state.rs
tbro: Its a tough one. It's a type that only exists to facilitate a state transition by encapsulating validation logic. The state is actually held in `ValidatedState`. Open to ideas, but I don't have any at the time being.

types/src/v0/impls/state.rs
tbro: :thinking:

types/src/v0/impls/state.rs
tbro: 092eba7ec44db27fd2d1fda28d8b7f53df5f800c

types/src/v0/impls/state.rs
tbro: bf5c57228d3936dda0de969f2cc80e158abb313a

types/src/v0/impls/state.rs
tbro: The intent is convey that signature validation will fail because we are adding a signature that doesn't match the fee_info. It is therefore invalid (relative to what it will be validated against).

types/src/v0/impls/state.rs
tbro: cea4a80a5beb971559ee661cc4af193233eb1e9f

types/src/v0/impls/state.rs
tbro: 104b12150f288edcd801e9e2fc3a216808b894e1

types/src/v0/impls/state.rs
tbro: Thinking again I see your point... The signature is structurally valid but will fail verification.

PR #2172, opened 2024-10-14
Add option to wait for L1 genesis by timestamp instead of block number

jbearer:
This allows us to specify the `l1_finalized` in our genesis file by timestamp, instead of block number. This makes it much easier to configure the network to start up from a future L1 block, as we typically do in testnet and mainnet deployments. The alternative is to calculate the expected L1 block number based on an average of 12 seconds per L1 block, which is tedious, error prone, and subject to drift due to L1 reorgs.

### This PR:
* Adds an option to specify the `l1_finalized` section in the genesis file by timestamp only
* Adds an L1 client function `wait_for_finalized_block_with_timestamp` to get the earliest finalized block with timestamp greater than or equal to a specified timestamp

### Key places to review:
* `wait_for_finalized_block_with_timestamp`, convince yourself it is deterministic

### DIFF
--- a/types/src/v0/impls/l1.rs
+++ b/types/src/v0/impls/l1.rs
@@ -12,6 +12,7 @@ use futures::{
     join,
     stream::{self, StreamExt},
 };
+use time::OffsetDateTime;
 use url::Url;
 
 use super::L1BlockInfo;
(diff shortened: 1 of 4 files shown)


types/src/v0/impls/l1.rs
```diff
@@ -172,6 +173,40 @@ impl L1Client {
...
+
+        // The finalized block jumps by more than 1 at a time, so we might not have found the
+        // earliest block with the desired timestamp. Work backwards until we find it.
+        loop {
+            let prev = self.wait_for_block(block.number - 1).await;
+            if prev.timestamp < timestamp {
+                return block;
+            }
+            block = prev;
+        }
```
rob-maron, line 207: ✅

rob-maron (approved):
My only concern was little clock deltas potentially causing us to select two different blocks (due to the sleep). But it seems adequately mitigated with the final reversing step

types/src/v0/impls/l1.rs
```diff
@@ -172,6 +173,40 @@ impl L1Client {
...
     }
 
+    /// Get information about the first finalized block with timestamp greater than or equal
+    /// `timestamp`.
+    pub async fn wait_for_finalized_block_with_timestamp(&self, timestamp: U256) -> L1BlockInfo {
+        let interval = self.provider.get_interval();
+
+        // Sleep until approximately the right time for the desired block to appear.
+        let now = U256::from(OffsetDateTime::now_utc().unix_timestamp() as u64);
+        if timestamp > now {
```
sveitser, line 183: Suggest to add a bit of info logging to this function otherwise it might sleep and or retry for a long time without logging anything.

sveitser (approved):

types/src/v0/impls/l1.rs
```diff
@@ -172,6 +173,47 @@ impl L1Client {
...
+                "waiting for finalized block with sufficient timestamp"
+            );
+            sleep(interval).await;
+        };
+
+        // The finalized block jumps by more than 1 at a time, so we might not have found the
+        // earliest block with the desired timestamp. Work backwards until we find it.
+        loop {
+            let prev = self.wait_for_block(block.number - 1).await;
+            if prev.timestamp < timestamp {
```
alysiahuggins, line 210: if the prev.timestamp is > timestamp, shouldn't we break? as that means that the block we retrieved in the previous loop is the earliest block with the desired timestamp?

jbearer: We're iterating backwards. So if `prev.timestamp > timestamp`, it is possible that `prev.prev.timestamp > timestamp` as well, so we must keep going. This is why when we finally return, we return `block`, *not* `prev`: we keep going backwards until we see a block with timestamp which is *too small*, and then we return the last block we saw before that one

jbearer: I think `parent` is probably a less confusing name than `prev` here, it's the "parent" in the chain, not the block. from the "previous" iteration of the loop. I'll make that change

alysiahuggins: ahh makes sense - got it!

PR #2173, opened 2024-10-15
fix: use correct PlonkVerifier in deployer.rs

alxiong:
separate the part of the changes originally included in #2148

### This PR:

- Switch to the right version of PlonkVerifier (in the deployer.rs we incorrectly deployed PlonkVerifier2.sol which is secure but less efficient)


alxiong:
let's get this uncontested logic backported to release-gambit first, the gas oracle logic can take longer. @jbearer @sveitser

jbearer (approved):

PR #2183, opened 2024-10-17
Enable frontier catchup from storage

jbearer:


sequencer/src/api/sql.rs
```diff
@@ -333,9 +333,16 @@ async fn reconstruct_state<'a>(
...
     let mut parent = from_leaf;
     let mut state = ValidatedState::from_header(parent.block_header());
+
     // Pre-load the state with the accounts we care about to ensure they will be present in the
     // final state.
-    state.fee_merkle_tree = load_accounts(tx, from_height, accounts)
+    let mut accounts = accounts.iter().copied().collect::<HashSet<_>>();
+    // Add in all the accounts we will need to replay any of the headers, to ensure that we don't
+    // need to do catchup recursively.
+    accounts.insert(instance.chain_config.fee_recipient);
```
jbearer, line 342: Slightly cleaner if we move this into `fee_account_dependencies`

sequencer/src/catchup.rs
```diff
@@ -387,3 +387,44 @@ where
...
+    async fn try_fetch_accounts(
+        &self,
+        _instance: &NodeState,
+        _height: u64,
+        _view: ViewNumber,
+        _fee_merkle_tree_root: FeeMerkleCommitment,
+        _account: &[FeeAccount],
+    ) -> anyhow::Result<FeeMerkleTree> {
+        bail!("state catchup is disabled");
+    }
```
rob-maron, line 408: We're using `NullStateCatchup` because we technically don't need to catchup on top of catchup correct?

sequencer/src/api/sql.rs
```diff
@@ -333,9 +333,16 @@ async fn reconstruct_state<'a>(
...
+    accounts.extend(fee_account_dependencies(instance, parent, &leaves).await);
```
bfish713, line 343: Does this change what parent the parent leaf is, and is later used in `compute_state_update`

sequencer/src/catchup.rs
jbearer: Yes, and if there's some logic error leading to recursive catchup I would rather it just error out than try something weird

sequencer/src/api/sql.rs
```diff
@@ -344,26 +351,68 @@ async fn reconstruct_state<'a>(
...
-    // For catchup that is required during the following state replay, use the local database via
-    // this transaction.
-    let peers = SqlStateCatchup::new(Arc::new(tx), Default::default());
+    // We need the blocks frontier as well, to apply the STF.
+    let frontier = load_frontier(tx, from_height)
+        .await
+        .context("unable to reconstruct state because frontier is not available at origin")?;
+    match frontier
+        .proof
+        .first()
```
rob-maron, line 360: The first element of the proof should be the leaf we're looking for right?

sequencer/src/api/sql.rs
```diff
@@ -344,26 +351,68 @@ async fn reconstruct_state<'a>(
...
     }
 
     tracing::info!(from_height, ?to_view, "successfully reconstructed state");
     Ok((state, to_leaf))
 }
 
+/// Get the set of all fee accounts on which the STF depends, applied to the given list of headers.
+async fn fee_account_dependencies(
+    instance: &NodeState,
+    mut parent: &Leaf,
```
bfish713, line 395: I don't think this only lets you reassign parent to a different reference in this function but wanted to double check

jbearer: That's the idea, I don't want to modify `parent` in the caller because we don't want to lose track of where catchup is starting from, but we need `parent` to advance in the callee. Effectively this function is doing like a dry run of the whole replay process, not actually executing but only looking for dependencies

sequencer/src/api/sql.rs
jbearer: Nope, can't change `parent` if we're passing it in by value. Rust is helpful here

PR #2184, opened 2024-10-17
[WEEKLY RELEASE] HotShot - rc-0.5.78

ss-es:


shenkeyao (commented):
As mentioned on Zulip I think we should use the new HotShot tag rather than the branch name in `Cargo.toml`. Everything else looks good!

shenkeyao (approved):

builder/src/bin/permissionless-builder.rs
```diff
@@ -161,7 +151,7 @@ async fn run<V: Versions>(
         opt.hotshot_event_streaming_url,
         builder_server_url,
         api_response_timeout_duration,
-        buffer_view_num_count,
+        ChainConfig::default().max_block_size.into(),
```
jbearer, line 165: Uhh @shenkeyao why are we defaulting this? This should come from the actual chain config, e.g. `instance_state.chain_config.max_block_size`

jbearer (changes requested):

builder/src/bin/permissionless-builder.rs
shenkeyao: Good catch! Using `instance_state.chain_config` also allows us to simplify the function by removing this argument, since we already have `instance_state` as an arg.

jbearer (approved):

PR #2185, opened 2024-10-17
Cleanup builder.

tbro:
Builder had defined `#![alllow(unused_imports)]` at the top. Remove and clean them up.


mrain:
Thanks for the cleaning up!

Should we also remove `jf_merkle_tree` deps from Cargo.toml and move `jf_signature` to dev-dependencies?
Not sure about other dependencies.

shenkeyao (approved):

tbro:
> Thanks for the cleaning up!
> 
> Should we also remove `jf_merkle_tree` deps from Cargo.toml and move `jf_signature` to dev-dependencies? Not sure about other dependencies.

I'll finish this up after weekly release is merged.

PR #2187, opened 2024-10-18
CDN upgrade

rob-maron:
Adds support for both in parallel:
- `Quic` and `TCP+TLS`. We are going to be switching to `TCP+TLS` singularly 
- Namespaced and non-namespaced signatures. We will be switching to only namespaced

Adds `todo_by`s which will throw an error if not switched to support only the new version by next month


sequencer/src/network/cdn.rs
```diff
@@ -43,22 +44,39 @@ impl<T: SignatureKey> SignatureScheme for WrappedSignatureKey<T> {
     type PublicKey = Self;
 
     /// Sign a message of arbitrary data and return the serialized signature
-    fn sign(private_key: &Self::PrivateKey, message: &[u8]) -> anyhow::Result<Vec<u8>> {
-        let signature = T::sign(private_key, message)?;
-        // TODO: replace with rigorously defined serialization scheme...
-        // why did we not make `PureAssembledSignatureType` be `CanonicalSerialize + CanonicalDeserialize`?
+    fn sign(
+        private_key: &Self::PrivateKey,
+        namespace: &str,
```
sveitser, line 49: Since we have "namepaces" elsewhere in this codebase might be good to have a type or some documentation what this namespace is used for.

PR #2189, opened 2024-10-21
lagoverescapehatchthreshold updates

alysiahuggins:
related to #2071 and this post-merge comment https://github.com/EspressoSystems/espresso-sequencer/pull/2107#discussion_r1808194099

We should break once `prevUpdate` (now called `eligibleBlock`) is found. 

### This PR:
- appropriately breaks once an eligible (blockFound <= desired block number) update found in the function that determines if there was a lag in hotShot updates

### Key places to review:
- lag over threshold function
- places in the rust code where the lag over threshold function is called, e.g. in the slow tests https://github.com/EspressoSystems/espresso-sequencer/blob/0cc5db6cd26c04ad6bd77625cef32586e2428aea/sequencer/src/bin/espresso-dev-node.rs#L1008

### How to test this PR:  
- to test the solidity code run `forge test --match-contract LightClient_StateUpdatesTest` in the root of this directory
- to test the slow dev test run `cargo test --release --all-features --workspace -- --nocapture slow_dev_node_test`


contracts/src/LightClient.sol
```diff
@@ -361,42 +362,46 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        if (
+            blockNumber > block.number || updatesCount == 0
+                || blockNumber < stateHistoryCommitments[stateHistoryFirstIndex].l1BlockHeight
+        ) {
             revert InsufficientSnapshotHistory();
         }
 
-        uint256 prevBlock;
-        bool prevUpdateFound;
+        uint256 eligibleBlock; // the eligible block is <= blockNumber
```
sveitser, line 377: I think something like "lastStateUpdateBeforeBlockNumber" might be clearer?

alysiahuggins: well the last state update can also be the same block number as the one being assessed. So I wanted to remove terminology like `prev` or `before` so that it's not misleading when reading the code

contracts/test/LightClient.t.sol
```diff
@@ -822,93 +820,50 @@ contract LightClient_StateUpdatesTest is LightClientCommonTest {
...
 
-        LC.StateHistoryCommitment[] memory stateHistoryCommitments =
-            new LC.StateHistoryCommitment[](numUpdates);
+        uint64[] memory blockNumberUpdates = new uint64[](numUpdates);
+        blockNumberUpdates[0] = 1;
+        blockNumberUpdates[1] = blockNumberUpdates[0] + DELAY_THRESHOLD / 2; // 4
+        blockNumberUpdates[2] = blockNumberUpdates[1] + DELAY_THRESHOLD / 2; // 7
+        blockNumberUpdates[3] = blockNumberUpdates[2] + DELAY_THRESHOLD + 5; // 18
+        blockNumberUpdates[4] = blockNumberUpdates[3] + DELAY_THRESHOLD / 2; // 21
+        blockNumberUpdates[5] += 1; //22
```
sveitser, line 854: Isn't this 0 + 1 = 1?

contracts/src/LightClient.sol
sveitser: How about `eligibleStateUpdateBlockNumber` then?

contracts/test/LightClient.t.sol
```diff
@@ -822,93 +820,50 @@ contract LightClient_StateUpdatesTest is LightClientCommonTest {
...
-        updates[0] = 1;
-        updates[1] = updates[0] + DELAY_THRESHOLD / 2; // 4
-        updates[2] = updates[1] + DELAY_THRESHOLD / 2; // 7
+        uint8 numUpdates = 6;
 
-        LC.StateHistoryCommitment[] memory stateHistoryCommitments =
-            new LC.StateHistoryCommitment[](numUpdates);
+        uint64[] memory blockNumberUpdates = new uint64[](numUpdates);
+        blockNumberUpdates[0] = 1;
+        blockNumberUpdates[1] = blockNumberUpdates[0] + DELAY_THRESHOLD / 2; // 4
```
sveitser, line 850: It's not clear why we use `DELAY_THRESHOLD` here with it's value at 6 and then later specify it to be `1`.

sveitser:
I think the changes in logic to the contract function are fine but I find some of the tests rather difficult to read.

contracts/src/LightClient.sol
alysiahuggins: udpated here https://github.com/EspressoSystems/espresso-sequencer/commit/3d5ae7aefea737d7611d26f78a02c08144029481#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292R377

contracts/test/LightClient.t.sol
alysiahuggins: thanks for spotting that, updated here https://github.com/EspressoSystems/espresso-sequencer/commit/3d5ae7aefea737d7611d26f78a02c08144029481#diff-08d922b027356f5ca532486fcea6a356b593fadcc63440be9268b59f2503d5d8R856

contracts/test/LightClient.t.sol
alysiahuggins: i just use that variable as a buffer for calculating the block numbers in the state update but i can see how it's confusing here, made some changes to address that but i can create another issue to improve on this https://github.com/EspressoSystems/espresso-sequencer/commit/3d5ae7aefea737d7611d26f78a02c08144029481#diff-08d922b027356f5ca532486fcea6a356b593fadcc63440be9268b59f2503d5d8R846

alysiahuggins:
> I think the changes in logic to the contract function are fine but I find some of the tests rather difficult to read.

Thanks for the feedback, I can create another issue to improve on this

PR #2192, opened 2024-10-21
Libp2p and overflow patches

rob-maron:


Cargo.toml
```diff
@@ -153,3 +154,15 @@ debug = false
 strip = "debuginfo"
 [profile.dev.package."*"]
 opt-level = 3
+
+[patch."https://github.com/EspressoSystems/HotShot.git"]
```
jbearer, line 158: Can we remove the `patch` before merging? Just actually do the HotShot update in `dependencies`?

rob-maron: I can try yeah

rob-maron: I'm getting compile errors 😢

jbearer: Oh right because without the `patch` we'd have to go update the HotShot version in all the dependencies

jbearer: I suppose we can merge with the patch to unblock staging testing for the next release, but we should quickly go back and make this an official hotshot update

jbearer (approved):

Cargo.toml
rob-maron: Yeah it is already on HS `main`, but I have to update the weekly release

PR #2194, opened 2024-10-22
Build the commitment without bincode

ImJeremyHe:
This PR builds the commitment of v0.3 header without bincode, which helps other language be align with the Rust code.

### This PR:
- Implements the `Committable` trait for some types required in v0.3 Header
- Modifies the `commit` in v0.3 Header
- Updates the reference test

### This PR does not:
- Change any other logic

### Relevant PR:
https://github.com/EspressoSystems/espresso-sequencer-go/pull/52


types/src/v0/impls/auction.rs
```diff
@@ -29,11 +29,24 @@ impl FullNetworkTx {
...
 
-impl Committable for BidTxBody {
+impl Committable for BidTx {
     fn tag() -> String {
         "BID_TX".to_string()
     }
 
+    fn commit(&self) -> Commitment<Self> {
+        let comm = committable::RawCommitmentBuilder::new(&Self::tag())
+            .field("body", self.body.commit())
```
tbro, line 39: I think it was probably a previous mistake to have an explicit commitment for `BidTxBody`. The `BidTx` can just access `body` methods (or fields) directly in its own commitment. Having a `body` is a structural necessity but behavior should act as if all the fields are on `BidTx`.

types/src/v0/impls/auction.rs
```diff
@@ -29,11 +29,24 @@ impl FullNetworkTx {
...
+            .fixed_size_field("signature", &self.signature.into());
+        comm.finalize()
+    }
+}
+
+impl Committable for BidTxBody {
+    fn tag() -> String {
+        "BID_TX_BODY".to_string()
+    }
+
```
tbro, line 49: It doesn't appear in the diff, but on line `57` there is another call to `bincode`.

tbro:
FYI this code will not be enabled this year, so this probably shouldn't be a big priority.

tbro (commented):
FYI this code will not be enabled this year, so this probably shouldn't be a big priority.

types/src/v0/impls/auction.rs
ImJeremyHe: https://github.com/EspressoSystems/espresso-sequencer/pull/2194/commits/23f640c710e737319a6cff0580af793319ec38b7

ImJeremyHe:
@tbro I wonder which version will be used on our mainnet? v0.2?

tbro (approved):
LGTM

PR #2195, opened 2024-10-22
[Backport release-gambit] lower priority fee for light client update 

alxiong:
## Description

Backport of #2148 to `release-gambit`.


alxiong:
the contract CI error is weird, I can run `just sol-test` just fine locally. `error[E0463]: can't find crate for `num_enum` sounds strange.

sveitser:
I would try erasing the github cache by bumping the cache key.

sveitser:
@alxiong seems to have built diff-test successfully.

alxiong:
On current `release-gambit`, "Test" job succeed, but here we seems to time out on 
`TRY 1 SLOW [>240.000s] sequencer api::sql::generic_tests::test_namespace_query`

I'm not familiar with that code. also should we fix the Slow Test and TestDemoNative failure as well?

jbearer:
@alxiong we're not on release-gambit anymore, we're on release-rogue which forked off main recently. You can double check this change is in there

PR #2209, opened 2024-10-24
Test CI overhaul

tbro:
### This PR:
Speeds up tests in CI. Total duration is now < 20 min. Currently on `main` tests take about twice as long. You can see the successful workflow [here](https://github.com/EspressoSystems/espresso-sequencer/actions/runs/11606244980).

  * instead of `--release` we set a lower optimization level
    * packages that require higher optimization are specified individually
  * application and test binaries are built in parallel
  * unit tests and integration/e2e tests are run in parallel
  
Switching from `--release` to `test` profile has the benefit of shorter build times as well as catching more errors. There is  a fix included  for builder using duplicated `-m` short options. This sort of error will cause some tests to fail now with these changes. And availability options were added to more process-compose services to catch more failures early. An additional improvement would be to run slow tests in parallel w/ unit tests. I'm not currently aware of a way to include slow tests in this workflow w/out making it required for merge.

### How to test this PR:
If tests pass faster than what we see in earlier commits,  its a winner.


sveitser:
@tbro I noticed when working on https://github.com/EspressoSystems/espresso-sequencer/pull/2232, that I need to set the opt-level to 1 for some tests to pass. Not sure that would be helpful here but in case you haven't tried.

.github/workflows/test.yml
```diff
@@ -18,12 +18,93 @@ concurrency:
...
 jobs:
+  build-test-artifacts:
+    name: Build test artifacts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: rui314/setup-mold@v1
+
+      - uses: actions/checkout@v4
+        with:
+          # By default actions/checkout checks out a merge commit. Check out the PR head instead.
```
sveitser, line 34: Why do we want to check out the PR head instead?

.github/workflows/test.yml
```diff
@@ -18,12 +18,93 @@ concurrency:
...
+          # https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
+          ref: ${{ github.event.pull_request.head.sha }}
+          submodules: recursive
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+
+      - name: Enable Rust Caching
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
```
sveitser, line 44: Is this a good idea? Doesn't it mean that if we have a failure we only get a partial cache and if we afterwards get a successful build for the same cache key we don't save it and keep a partial cache?

.github/workflows/test.yml
sveitser: I might be missing something but I thought we _want_ to merge the PR into main and use that to run the CI so that we are confident things work _after_ a future merge into main.

.github/workflows/slowtest.yaml
```diff
@@ -41,25 +43,23 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Configure Environment
+        run: PATH="$PWD/target/debug:$PATH"
```
sveitser, line 47: I'm surprised this works. I thought one has to write it to $GITHUB_ENV to persist between step

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#passing-values-between-steps-and-jobs-in-a-workflow

scripts/demo-native
```diff
@@ -7,7 +7,7 @@ if [ -z "${IN_NIX_SHELL-}" ]; then
     REPO_ROOT="$(dirname "$(dirname "$(readlink -fm "$0")")")"
     # Default to CARGO_TARGET_DIR if set, otherwise use the default target directory.
     TARGET_DIR="${CARGO_TARGET_DIR:-${REPO_ROOT}/target}"
-    export "PATH=${TARGET_DIR}/release:$PATH"
+    export "PATH=${TARGET_DIR}/debug:$PATH"
```
sveitser, line 10: Hmm I think this will confuse people who have been building with release. I think we should log something prominent to the terminal about this change.

.github/workflows/test.yml
tbro: I don't understand why it would prevent the cache from being saved on a successful run. But I'll remove it isn't adding much any way.

.github/workflows/slowtest.yaml
tbro: We made this change in `test.yaml` a while back. I think it works b/c we are running all the steps on the same host... so we don't need any additional persistence. Also we are not creating a new variable, we are simply appending to `PATH` and `PATH` is always exported on the host. For it *not* to work I think github would have to do some variable resetting between steps.

.github/workflows/test.yml
tbro: Yea I guess I'm confused about what is happening there.

scripts/demo-native
tbro: e7ce1264aee825b9c13c63e1021ef520790389f1

.github/workflows/test.yml
tbro: fbd508ef5b21976bce3edcd32100a35a58f65640

.github/workflows/test.yml
tbro: cb8f7e458bfb7482911b1ecbc65478d2fd0d517e

sveitser (approved):

PR #2215, opened 2024-10-24
Do not load entire leaf just to get anchor view

jbearer:
We have seen heavy load on water DBs due to loading the anchor leaf in the proposal fetching task. This is because the leaves are very large because we store the payload in the leaf, which we should not do. That is a separate bug with a separate fix. This fix addresses the fact that we shouldn't need to load the leaf at all in this task, since we only care about the view number.

### This PR:
* Adds a dedicated `load_anchor_view` persistence function
* Implements this function efficiently for SQL storage
 
### This PR does not:
* Implement `load_anchor_view` efficiently for FS storage, since to get the view number we'd have to load the whole file and deserialize the leaf anyways, because view is not stored separately. Thus, it is easier to just call `load_anchor_leaf` and pull out the view number

### DIFF
--- a/sequencer/src/context.rs
+++ b/sequencer/src/context.rs
@@ -533,9 +533,8 @@ async fn fetch_proposal_chain<N, P, V>(
 
 async fn load_anchor_view(persistence: &impl SequencerPersistence) -> ViewNumber {
     loop {
-        match persistence.load_anchor_leaf().await {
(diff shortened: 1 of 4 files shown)


tbro (approved):
LGTM

imabdulbasit (approved):

PR #2221, opened 2024-10-25
Ensure consistency of decide events

jbearer:
The logic for moving data from consensus storae to archive storage was sometimes generating an invalid decide event, because it made a chain of non-consecutive leaves originating from multiple separate events. This fixes the issue and prevents it from happening again by generating a separate decide event for each leaf in consensus storage.

Also adds a regression test.


jbearer:
Tested on a water node that was having this issue. Also deployed and tested on decaf

tbro (approved):
Approving since it is unit test is included and has been tested elsewhere.

PR #2231, opened 2024-10-29
CI: skip ARM build on PRs

sveitser:
The arm build uses buildjet and costs money. There is relatively little utility in running it on every PR. If necessary it can be triggered explicitly via workflow dispatch.

It will build the docker images in PRs but only for x86. This way we would still catch most issues in the dockerfiles before merge to main, even though we aren't building them for arm.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -145,8 +145,8 @@ jobs:
   build-dockers:
     runs-on: ubuntu-latest
     needs: [build]
-    # run if `build` succeeded and build-arm succeeded or was skipped
-    if: success() && (needs.build.result == 'success' && (needs.build-arm.result == 'success' || needs.build-arm.result == 'skipped'))
+    # run if `build` succeeded
+    if: ${{ always() && !cancelled() && needs.build.result == 'success' }}
     outputs:
       sequencer-tag: ${{ steps.sequencer.outputs.tags }}
       cdn-broker-tag: ${{ steps.cdn-broker.outputs.tags }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -142,7 +144,9 @@ jobs:
 
   build-dockers:
     runs-on: ubuntu-latest
-    needs: [build, build-arm]
+    needs: [build]
```
jbearer, line 147: If I'm reading right this will allow the job to start before `build-arm` completes when running on `main`? But that won't work

sveitser, line 147: I guess this will break if build_arm is slower than build and the artifact might not be available.

.github/workflows/build.yml
```diff
@@ -142,7 +144,9 @@ jobs:
...
+    # run if `build` succeeded
+    if: ${{ always() && !cancelled() && needs.build.result == 'success' }}
```
jbearer, line 149: What does this mean? Isn't `always() && X` basically the same as `X`? Also, isn't this expression mirroring the default semantics?

.github/workflows/build.yml
sveitser: Yes, commented that here too: https://github.com/EspressoSystems/espresso-sequencer/pull/2231#discussion_r1820960902

sveitser:
The graph looks promising on the latest PR: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/11576703665 ...

.github/workflows/build.yml
sveitser: It's from here: https://stackoverflow.com/a/70959885

It probably has some effect, not sure what exactly though.

sveitser:
This seems to work as desired on PRs now.

Ancient123 (approved):

PR #2232, opened 2024-10-29
WIP: basic mutation testing with cargo mutants

### LINKED ISSUE #2227: Mutation testing to identify unused code
The idea is to find out what can be changed in the code without failing compilation or without failing tests.

For many of those we may expect that it's not explicitly tested but for some it may show that we are not testing something that we believe we are testing.

sveitser:
Close #2227 

- Update nixpkgs to get latest cargo-mutants.
- Remove nixWithFlakes from flake.nix because it would have to be changed to work with the latest version of nixpkgs and I think we no longer need it.
- Create a non-flaky test profile.
- Run with `cargo mutants`.
- Inspect output and `./mutants.out` directory

For example `./mutants.out/missed.txt` looks like this

    contract-bindings/src/plonk_verifier_2.rs:463:13: replace plonk_verifier_2::<impl AbiEncode for PlonkVerifier2Calls>::encode -> Vec<u8> with vec![]
    hotshot-state-prover/src/mock_ledger.rs:496:57: replace + with * in gen_circuit_for_test
    hotshot-state-prover/src/mock_ledger.rs:496:53: replace * with / in gen_circuit_for_test
    hotshot-state-prover/src/mock_ledger.rs:149:9: replace MockLedger::elapse_without_block with ()
    contract-bindings/src/light_client_mock.rs:1928:21: replace == with != in light_client_mock::<impl ContractRevert for LightClientMockErrors>::valid_selector
    contract-bindings/src/erc1967_proxy.rs:319:13: replace erc1967_proxy::<impl AbiEncode for ERC1967ProxyErrors>::encode -> ::std::vec::Vec<u8> with vec![0]
    hotshot-state-prover/src/circuit.rs:236:29: replace % with / in build

To make it really useful further configuration is likely necessary, such as marking some functions to not mutate them.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,7 +1,15 @@
+# cargo-mutants does not seem to support specifying the nextest profile? :(
 [profile.default]
 slow-timeout = "2m"
-default-filter = 'not (test(slow_) | package(tests))'
-retries = 2
+default-filter = """not (test(slow_) | package(tests) \
+  | test(test_fee_upgrade_time_based) \
+  | test(test_fee_upgrade_view_based) \
+  | test(test_marketplace_upgrade_time_based) \
+  | test(test_marketplace_upgrade_view_based) \
+  | test(test_process_client_handling_stream_subscribe_voters) \
+  | test(test_process_client_handling_stream_subscribe_node_identity) \
+  )"""
+retries = 0
 
 # The restart tests run an entire sequencing network, and so are quite resource intensive.
 [[profile.default.overrides]]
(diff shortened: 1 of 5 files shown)


tbro:
Can you comment about how we are meant to read the output? What does this mean:

```rust
contract-bindings/src/plonk_verifier_2.rs:463:13: replace plonk_verifier_2::<impl AbiEncode for PlonkVerifier2Calls>::encode -> Vec<u8> with vec![]
```

`cargo mutants` made that replacement? What was the outcome? What was the expectation?

.config/nextest.toml
```diff
@@ -1,7 +1,15 @@
+# cargo-mutants does not seem to support specifying the nextest profile? :(
```
sveitser, line 1: I think we can use the env var `NEXTEST_PROFILE`

sveitser:
> cargo mutants made that replacement? What was the outcome? What was the expectation?

@jbearer this is still WIP and I will document it better once I think it's useful and I understand it well enough.

The expectation is that for every change that cargo mutants makes there is at least one test that fails as a result. If not it's considered a mutant that our tests "missed.

However to make it effective we need to annotate some code so that cargo mutants doesn't mutate it (for example it's likely not useful if some change in mock_ledger is "missed"). There is also a bug I need to sort out where I get a warning from cargo mutants that the test run exited with an unexpected status code. I think it currently leads to false positive "missed" results.

PR #2238, opened 2024-10-30
remove V1 version variant from the binary

imabdulbasit:
addresses https://github.com/EspressoSystems/espresso-sequencer/issues/2211

This PR removes the V1 version variants from the binaries, as they are no longer in use. The mainnet will launch with fees as the initial version, which is V2. This change speeds up binary compilation. 

It also updates the integration CI to only run smoke test and not the upgrade test. We can remove this but I think the smoke test has more validation like the transactions submission that demo-native CI does not have


.github/workflows/test.yml
```diff
@@ -77,8 +77,7 @@ jobs:
       - name: Test Integration
         env:
           NEXTEST_PROFILE: integration
-          INTEGRATION_TEST_SEQUENCER_VERSION: 02
-        run: cargo nextest run --locked --release --all-features --verbose --nocapture
+        run: cargo nextest run --locked --release --all-features --verbose --nocapture --test integration smoke
```
sveitser, line 80: What exactly does this do now?

imabdulbasit: so the there are two tests inside integration: smoke and upgrade test. This one just runs smoke test for V2 version.

imabdulbasit: as we are not doing any upgrade for the  native demo now

tbro: test.yaml changed a lot last week. In the new version you probably only need to remove the `02` from `matrix.version`

.github/workflows/test.yml
```diff
@@ -183,10 +184,10 @@ jobs:
       - name: Test Integration
         env:
           NEXTEST_PROFILE: integration
-          INTEGRATION_TEST_SEQUENCER_VERSION: ${{matrix.version}}
+          INTEGRATION_TEST_SEQUENCER_VERSION: ${{ matrix.version }}
         run: |
-          cargo nextest run --archive-file nextest-archive.tar.zst  --verbose --no-fail-fast \
-          --workspace-remap $PWD
+          cargo nextest run --archive-file nextest-archive.tar.zst --verbose --no-fail-fast \
+          --workspace-remap $PWD $(if [ "${{ matrix.version }}" == "2" ]; then echo " smoke"; fi)
```
tbro, line 190: I guess this is ok for now, but if we are going to run different tests for different targets matrix may not make much sense here anymore. If we are always going to have 2, then we can just have `base` job and `upgrade` job. They will still run in parallel. We can leave it like this for now though, I guess, and leave it for a later reorg PR.

tbro (approved):

.github/workflows/test.yml
imabdulbasit: I think we will run upgrade test for every version other than initial (v2)

PR #2240, opened 2024-10-31
Make L1 client async again

jbearer:
Make L1 client update asynchronously in a background task. This allows us to subscribe to updates from the L1, rather than polling every time we have a new HotShot block. This vastly reduces the number of RPC calls we make. This is in response to an increase in the number of L1 RPC calls being made, due to the new logic that validates the L1 head and finalized in each proposal.

### This PR:
* Rewrites the L1 client to work based on an in-memory state that is updated by an async task
* Implements the various `wait_for_*` functions using an event stream populated by the async task
* Supports either HTTP or WebSockets as a backend for the L1 client
* Adds an LRU cache for L1 blocks
* Adds more configurability for L1 client

### Key places to review:
* `types/src/v0/impls/l1.rs`


tbro:
I don't see that any calls to the `L1Client` are updated here to now use the websocket. Do you have a specific use case in mind?

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
+impl PubsubClient for RpcClient {
+    type NotificationStream = <Ws as PubsubClient>::NotificationStream;
+
+    fn subscribe<T>(&self, id: T) -> Result<Self::NotificationStream, Self::Error>
+    where
+        T: Into<U256>,
+    {
+        match self {
+            Self::Http(_) => Err(ProviderError::CustomError(
+                "subscriptions not supportedg with HTTP client".into(),
```
sveitser, line 104:
```suggestion
                "subscriptions not supported with HTTP client".into(),
```

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
+        }
+    }
+
+    fn unsubscribe<T>(&self, id: T) -> Result<(), Self::Error>
+    where
+        T: Into<U256>,
+    {
+        match self {
+            Self::Http(_) => Err(ProviderError::CustomError(
+                "subscriptions not supportedg with HTTP client".into(),
```
sveitser, line 116:
```suggestion
                "subscriptions not supported with HTTP client".into(),
```

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
 
-        if finalized.number == number {
-            return finalized;
-        }
+            // Wait for the block.
+            while let Some(event) = events.next().await {
+                let L1Event::NewFinalized { finalized } = event else {
+                    continue;
+                };
+                if finalized.number < number {
```
sveitser, line 405: nit: for the latest block we do the comparison the other way around above: `if head >= number {` I think we can do it the same way both times. Then we also don't need this `continue`, I think.

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
-        let interval = self.provider.get_interval();
-
-        // Sleep until approximately the right time for the desired block to appear.
-        let now = U256::from(OffsetDateTime::now_utc().unix_timestamp() as u64);
-        if timestamp > now {
-            let dur = (timestamp - now).as_u64();
-            tracing::warn!("sleeping for {dur:?}, until {timestamp}");
-            sleep(Duration::from_secs(dur)).await;
-        }
+        // Wait until the finalized block has timestamp >= `timestamp`.
```
sveitser, line 424: Wondering if we actually need this function. It's only used on startup if the genesis block is not finalized.

https://github.com/EspressoSystems/espresso-sequencer/blob/b712b84c502d3fe74a530808e40bd2f870eb1d94/sequencer/src/lib.rs#L483

I'm wondering if it would be okay to just panic in that case.

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
+    async fn get_finalized_block<'a>(
+        &'a self,
+        mut state: MutexGuard<'a, L1State>,
+        number: u64,
+    ) -> (MutexGuard<'a, L1State>, L1BlockInfo) {
+        // Try to get the block from the finalized block cache.
+        assert!(
+            state.snapshot.finalized.is_some()
+                && number <= state.snapshot.finalized.unwrap().number,
+            "requesting a finalized block {number} that isn't finalized; snapshot: {:?}",
```
sveitser, line 497: Wonder why we don't make this a fallible function instead of one that might panic.

types/src/v0/v0_1/l1.rs
```diff
@@ -33,12 +38,88 @@ pub struct L1Snapshot {
...
+    Http(Http),
+    Ws(Ws),
+}
+
+/// In-memory view of the L1 state, updated asynchronously.
+#[derive(Debug)]
+pub(crate) struct L1State {
+    pub(crate) snapshot: L1Snapshot,
+    pub(crate) finalized: LruCache<u64, L1BlockInfo>,
+    pub(crate) finalized_by_timestamp: BTreeMap<U256, u64>,
```
sveitser, line 115: Similar to my last comment, IIUC since is only used on startup and in-memory only I think it might be simpler if we didn't have it at all or at least didn't keep a cache for it.

sveitser (approved):
Overall LGTM, just non-urgent suggestions.

types/src/v0/impls/l1.rs
```diff
@@ -64,180 +75,470 @@ impl L1BlockInfo {
...
+        }
+    }
+
+    fn update_loop(&self) -> impl Future<Output = ()> {
+        let rpc = self.provider.clone();
+        let retry_delay = self.retry_delay;
+        let state = self.state.clone();
+        let sender = self.sender.clone();
+
+        let span = tracing::warn_span!("L1 client update");
```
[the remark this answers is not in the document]
jbearer: It applies to anything but the span will only show up on WARN or higher. But now that I think about it since this span is just a tag with no extra data, it's quite cheap to log this all the time, so this could easily be `info_span!`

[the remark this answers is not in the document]
jbearer: er wait I had it backwards. This appears any time the log level is set to WARN or _lower_, just like a normal `tracing::warn!`. So we actually want to keep it this way, we want to see this info even if the node is running with `RUST_LOG=warn`

types/src/v0/impls/l1.rs
jbearer: No, we can't just panic there. The intended use case is you set the genesis block to something in the future, then you can start up all the nodes ahead of time, and they will just wait until the genesis block is ready and then start running

types/src/v0/impls/l1.rs
jbearer: The idea is this can't panic based solely on the internal behavior of this module, i.e. we only ever call this private helper function with a finalized block number. There is no input from callers in any other modules that can cause this function to panic, so we shouldn't bubble the error up to the caller.

And because the eventual interface is infallible (e.g. `validate_apply_header` calls `wait_for_finalized_block` and blocks until we either get it or time out) any errors we return would have to be handled with retry loops which IMO makes the calling code more complicated for little reason, when it's an error that should never happen to begin with

types/src/v0/v0_1/l1.rs
jbearer: Good call, removing this cache makes things considerably simpler at very little cost

PR #2252, opened 2024-11-04
Reduce risk of losing data due to garbage collection or restarts

jbearer:
Partially addresses an issue we saw on `water`, where all DA nodes restarting at the same time can cause loss of data. The problematic sequence of events is:
* DA nodes receive, store, and sign a DA proposal A, causing a DAC to be formed
* DA nodes shut down
* The corresponding proposal is decided
* DA nodes start back up, and receive a decide event for a subsequent proposal B
* DA nodes garbage collect all views <= B, including proposal A, which they never decided and thus never moved to archival storage

This change addresses this potential problem in two ways:
1. All nodes (including non-DA nodes) garbage collect consensus storage only for views they have definitively processed (i.e. moved to archival storage). That is, instead of garbage collecting all views <= v upon seeing a decide event for leaf v, we *only* collect views which have been part of a decide event we've seen (ie whose view numbers fall within a range of view numbers defined by the first and last view numbers in a chain of leaves with consecutive heights). This means view which are not encompassed by a decide event are conservatively not garbage collected (at least, not immediately). This change prevents unintentional loss of data.
2. We implement the query service fetching provider traits by retrieving data from consensus storage. This change leverages the previous change preventing loss of data, allowing the database to automatically recover from missing data caused by missed decide events.

### This PR:

* When processing a decide event in persistent storage, look for chains of consecutive leaves, and only garbage collect views which fall within those chains
* Separately, garbage collect any views over a certain age, to clean up any undecided views which weren't used to recover archival storage
  * GC more aggressively when storage used by consensus storage exceeds a threshold (just like in the query service)
* Implement query service provider traits based on consensus storage

### This PR does not:

Fully solve the restart issue. Specifically: the protocol guarantees that at least one DA node will store every DA proposal, and this change ensures every stored DA proposal will eventually make it to archival storage. We do not have the same guarantee for quorum proposals and certificates, because DA nodes are not required to form a QC. Thus, we can still have a situation where every DA node misses a quorum proposal or certificate, and furthermore, the query service needs the quorum proposal for the block header before it can even try to fetch a missing DA proposal. This problem is solved by having non-DA nodes run a version of the availability service which stores leaves but not payloads. That will be facilitated by sqlite support.


rob-maron (approved):
I didn't see any problems with it, and it seems like a good improvement

PR #2253, opened 2024-11-04
reuse build artifacts in slowtest.yml

tbro:
Reuse build artifacts for slow tests instead of building twice.


tbro:
Closing since workflow is never triggered. There are a lot of caveats: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

PR #2254, opened 2024-11-05
Accept WebSockets provider for fee contract proxy check

jbearer:
Fixes an issue where the new WebSockets functionality for the L1 client is unusable, because the sequencer takes the same `ESPRESSO_SEQUENCER_L1_PROVIDER` URL and passes it to a client that only accepts HTTP, for the fee contract proxy check.

### This PR:
* Changes the fee contract proxy check to use `L1Client` (which accepts HTTP or WS) instead of `Provider<Http>1
* Changes the local demo to run some nodes with WS L1 providers and others with HTTP


imabdulbasit (approved):

.env
```diff
@@ -37,6 +37,7 @@ ESPRESSO_SEQUENCER_GENESIS_FILE=/genesis/demo.toml
 ESPRESSO_SEQUENCER_L1_PORT=8545
 ESPRESSO_SEQUENCER_L1_WS_PORT=8546
 ESPRESSO_SEQUENCER_L1_PROVIDER=http://demo-l1-network:${ESPRESSO_SEQUENCER_L1_PORT}
+ESPRESSO_SEQUENCER_L1_WS_PROVIDER=http://demo-l1-network:${ESPRESSO_SEQUENCER_L1_WS_PORT}
```
sveitser, line 40: @jbearer is this `http://` on purpose?

Issue #2256, opened 2024-11-05
Compilation broken without Cargo.lock file


sveitser:
This means crates can't compile if they include crates from this repo as dependency.

To reproduce 

```
rm Cargo.lock
cargo check
```

will output

```
error[E0277]: the trait bound `async_lock::RwLock<OrchestratorState<KEY>>: WriteState` is not satisfied
   --> /home/lulu/.cargo-nix/git/checkouts/HotShot-de3b585b288dcaaa/14b460a/crates/orchestrator/src/lib.rs:874:9
    |
874 |         define_api().map_err(|_e| io::Error::new(ErrorKind::Other, "Failed to define api"));
    |         ^^^^^^^^^^^^ the trait `WriteState` is not implemented for `async_lock::RwLock<OrchestratorState<KEY>>`
    |
    = help: the following other types implement trait `WriteState`:
              ()
              Arc<W>
              async_lock::mutex::Mutex<State>
              async_lock::rwlock::RwLock<State>
note: required by a bound in `define_api`
   --> /home/lulu/.cargo-nix/git/checkouts/HotShot-de3b585b288dcaaa/14b460a/crates/orchestrator/src/lib.rs:664:48
    |
662 | fn define_api<KEY, State, VER>() -> Result<Api<State, ServerError, VER>, ApiError>
    |    ---------- required by a bound in this function
663 | where
664 |     State: 'static + Send + Sync + ReadState + WriteState,
    |                                                ^^^^^^^^^^ required by this bound in `define_api`

    Checking hotshot-task-impls v0.5.79 (https://www.github.com/EspressoSystems/HotShot.git?tag=0.5.79-patch1#14b460a1)
For more information about this error, try `rustc --explain E0277`.
error: could not compile `hotshot-orchestrator` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

PR #2257, opened 2024-11-05
CI: add job to check compilation without lock file

sveitser:
Avoid situations like #2256 going unnoticed for so long.

This is expected to fail until #2256 is addressed.


.github/workflows/build-without-lockfile.yml
```diff
@@ -0,0 +1,40 @@
...
+    - cron: "0 0 * * 1"
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  RUSTFLAGS: '--cfg async_executor_impl="async-std" --cfg async_channel_impl="async-std"'
```
sveitser, line 21: should remove this

imabdulbasit, line 21: not needed anymore

imabdulbasit (approved):

PR #2259, opened 2024-11-05
Arity is already available as an associated constant.

tbro:
I think it is slightly better if we always access the associated constant instead of duplicating it into a stand-alone constant in addition to the associated one.

### DIFF
--- a/types/src/v0/v0_1/state.rs
+++ b/types/src/v0/v0_1/state.rs
@@ -15,11 +15,12 @@ use std::collections::HashSet;
 pub type BlockMerkleTree = LightWeightSHA3MerkleTree<Commitment<Header>>;
 pub type BlockMerkleCommitment = <BlockMerkleTree as MerkleTreeScheme>::Commitment;
 
-pub type FeeMerkleTree = UniversalMerkleTree<FeeAmount, Sha3Digest, FeeAccount, 256, Sha3Node>;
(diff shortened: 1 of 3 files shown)


types/src/v0/v0_1/state.rs
```diff
@@ -15,12 +15,11 @@ use std::collections::HashSet;
 pub type BlockMerkleTree = LightWeightSHA3MerkleTree<Commitment<Header>>;
 pub type BlockMerkleCommitment = <BlockMerkleTree as MerkleTreeScheme>::Commitment;
 
-pub type FeeMerkleTree = UniversalMerkleTree<FeeAmount, Sha3Digest, FeeAccount, FEE_MERKLE_TREE_ARITY, Sha3Node>;
+pub type FeeMerkleTree = UniversalMerkleTree<FeeAmount, Sha3Digest, FeeAccount, 256, Sha3Node>;
```
sveitser, line 18: I think now it's harder to know what the 256 is for if you just look the code outside an editor. I think it would be fine to keep the constant just for that. Could make it not put though, or just put a comment /* arity */

sveitser (approved):

types/src/v0/v0_1/state.rs
tbro: 4a157c2c2497b8177a3c53cfeba089428e414332

imabdulbasit (approved):

PR #2261, opened 2024-11-06
Add pipefail to docker demo

tbro:
docker demo is succeeding on main, even though services never become ready. The error in the logs is: 

```bash
dependency failed to start: container espresso-sequencer-sequencer1-1 exited (101)
error: Recipe `demo` failed on line 8 with exit code 1
```
example run: https://github.com/EspressoSystems/espresso-sequencer/actions/runs/11706340826/job/32604796112

In addition, !pull_request filter has been temporarily removed so we can reproduce in this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -478,7 +478,7 @@ jobs:
           labels: ${{ steps.dev-rollup.outputs.labels }}
 
   test-demo:
-    # if: ${{ github.event_name != 'pull_request' }}
+    if: ${{ github.event_name != 'pull_request' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -478,7 +478,7 @@ jobs:
           labels: ${{ steps.dev-rollup.outputs.labels }}
 
   test-demo:
-    if: ${{ github.event_name != 'pull_request' }}
```
sveitser, line 481: I think you can also use workflow dispatch to trigger it even though it has this if statement. Not 100% sure though.

sveitser (approved):

PR #2262, opened 2024-11-06
Tb/marketplace/infinate version

### LINKED ISSUE #2251: Adjust MarketplaceVersion in Versions trait implementation  for SeqType
Since we have other upgrades planned in the near future and the marketplace will launch next year, the marketplace version in the Versions trait for SeqType should be temporarily set to a large value, such as u64::MAX

tbro:
Closes #2251 

Moves `Marketplace` version out of the way and creates place holder `v0_3` to hold ProofOfStake upgrade.

### This PR:
Marketplace is delayed and we need some other upgrades sooner. This will allow us to do that.

### This PR does not:
The new `v3` doesn't really exist yet. So no logic has been added. It just change the version number of `Marketplace` and a place holder for `ProofOfStake` to be filled in later.

### Key places to review:
There are a lot of very small changes spread across many files.

### How to test this PR:
I think any errors should be caught by the compiler. That, combined w/ our tests should be good enough IMHO. So if CI is green its probably OK.


marketplace-builder/src/builder.rs
```diff
@@ -526,6 +526,8 @@ mod test {
         parent_view_number: u64,
         parent_commitment: VidCommitment,
     ) -> Bundle<SeqTypes> {
+        // FIXME error here:
```
tbro, line 529: link

imabdulbasit: it is coming from here https://github.com/EspressoSystems/HotShot/blob/main/crates/builder-api/src/v0_3/mod.rs#L6

https://github.com/EspressoSystems/HotShot/blob/main/crates/builder-api/src/v0_3/builder.rs#L12C31-L12C38

PR #2264, opened 2024-11-07
Fix WS provider URL

sveitser:
This caused calls to the L1 provider to fail.

- Ignore `.env` for typos.
- Log fee contract address during proxy check for easier debugging.

To test run `just demo` and in another terminal `scripts/smoke-test-demo`.


tbro (approved):

tbro:
To me it looks like `just demo` runs the  docker images w/ `main` tag ... the ones from the PR would have a different tag right? So I'm not sure of these test steps.

sveitser:
@tbro for locally fixing it only the change in the .env file is needed, so using the `main` images is fine. I don't know if this PR fixes everything but at least it help for running `just demo` locally, so I'll merge.

PR #2265, opened 2024-11-07
Clean up logging

jbearer:
Demotes some logging and cleans up some instrumentation to avoid overly noisy output that isn't cause for concern.

### This PR:

* In catchup, demote retryable failures from WARN to INFO. We now only WARN if the overall operation fails (failed to fetch from any provider) or if we fail to fetch from a single provider in a way that indicates that provider is malicious (e.g. invalid Merkle proof) instead of just unreachable (e.g. HTTP error)
* Add instance state to `skip` list for `instrument`, since it generates a very large debug output with each log
* Remove `Debug` requirement from `StateCatchup` trait. This was generating a huge dump of useless data for the SQL provider in particular. Instead, we have a function `name() -> String` which is meant to be a short, readable identifier of the provider. We now use this when logging information about a specific catchup provider
* Update `derive_more`, so we could customize the `Debug` impl of `NodeState` to use the `name()` function of the nested `Box<dyn StateCatchup>`, instead of the `Debug` impl. This update was a breaking change and thus required minor tweaks to a lot of other uses of `derive_more`.
* Demote `timed out fetching proposal` from WARN to INFO. This happens normally when we are fetching an old proposal that the rest of the network has already GCed, and the task will eventually just exit on its own
* Demote `background task exited` from INFO to DEBUG for short-lived tasks (proposal fetching) only
* Demote L1 logging that happens every L1 block from INFO to DEBUG

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2578,9 +2578,11 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb7330aeadfbe296029522e6c40f315320aba36fc43a5b3632f3795348f3bd22"
 dependencies = [
+ "convert_case 0.6.0",
  "proc-macro2",
  "quote",
  "syn 2.0.77",
+ "unicode-xid",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


sveitser (approved):
LGTM

PR #2272, opened 2024-11-11
Add Audits folder and audit reports

0xkato:
Adding an Audits folder, containing the external audits and internal reviews which can also be found in [espresso-audits](https://github.com/EspressoSystems/espresso-audits).


philippecamacho:
Suggestion: add a README.md file in the folder containing the list of the reports in a table.

Also, don't we have a report about HotShot as well?

philippecamacho (approved):
LGTM

PR #2273, opened 2024-11-11
`HotShot/rc-0.5.81`

rob-maron:


sequencer/src/api/data_source.rs
```diff
@@ -216,7 +217,6 @@ pub struct PublicHotShotConfig {
     num_nodes_with_stake: NonZeroUsize,
     known_nodes_with_stake: Vec<PeerConfig<PubKey>>,
     known_da_nodes: Vec<PeerConfig<PubKey>>,
-    my_own_validator_config: PublicValidatorConfig,
```
jbearer, line 219: Have we removed private stuff from the `HotShotConfig`? If so we can delete this whole struct and just publish the `HotShotConfig` directly from API endpoints. Although maybe it's still good to have this explicit so we don't shoot ourselves in the foot later by adding something private to `HotShotConfig`...

sequencer/src/api/options.rs
```diff
@@ -267,7 +267,7 @@ impl Options {
         N: ConnectedNetwork<PubKey>,
         P: SequencerPersistence,
         D: SequencerDataSource + CatchupStorage + Send + Sync + 'static,
-        for<'a> D::Transaction<'a>: UpdateDataSource<SeqTypes>,
+        for<'a> D::Transaction<'a>: UpdateAvailabilityStorage<SeqTypes>,
```
jbearer, line 270: Probably shouldn't need this anymore

sequencer/src/options.rs
```diff
@@ -197,10 +199,18 @@ pub struct Options {
     /// The maximum number of bytes we will send in a single Libp2p gossip message
     #[clap(
         long,
-        env = "ESPRESSO_SEQUENCER_LIBP2P_MAX_TRANSMIT_SIZE",
+        env = "ESPRESSO_SEQUENCER_LIBP2P_MAX_GOSSIP_TRANSMIT_SIZE",
```
jbearer, line 202: Can you make sure all these libp2p vars we've added recently are in `public-env-vars.toml`? Might come in handy one day to look up a node's p2p config via it's HTTP server

types/src/v0/impls/header.rs
```diff
@@ -440,8 +440,11 @@ impl Header {
                 );
             } else {
                 ensure!(
-                    fee_account
-                        .validate_sequencing_fee_signature_marketplace(fee_signature, *fee_amount,),
+                    fee_account.validate_sequencing_fee_signature_marketplace(
+                        fee_signature,
+                        *fee_amount,
+                        height,
```
jbearer, line 446: I don't 100% know what this is but I thought we added the view number to a builder signature, not the height, so I want to double check this is correct

types/src/v0/impls/state.rs
```diff
@@ -663,6 +663,7 @@ fn validate_builder_fee(proposed_header: &Header) -> Result<(), BuilderValidatio
                 .validate_sequencing_fee_signature_marketplace(
                     &signature,
                     fee_info.amount().as_u64().unwrap(),
+                    proposed_header.height(),
```
jbearer, line 666: Same here

sequencer/src/api/data_source.rs
rob-maron: Yeah we did remove private stuff. Let me see if I can get rid of this, I don't think we should put anything private in the "network" config

rob-maron: Oh, right now we still have some sensitive info in there (`manual_start_password`)? So I will have a separate PR to merge everything into one once we remove that from the "network" config

jbearer (approved):

PR #2282, opened 2024-11-14
update comment to match implementation

### LINKED ISSUE #2278: CP audit: update comment to match logic on LightClient.sol

alysiahuggins:
Closes #2278 

### This PR:
Updates comment to match implementation

### This PR does not:
Doesn't change any functionality

### Key places to review:
The comment


contracts/src/LightClient.sol
```diff
@@ -426,8 +426,8 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
             revert InvalidHotShotBlockForCommitmentCheck();
         }
         for (uint256 i = stateHistoryFirstIndex; i < commitmentsHeight; i++) {
-            // The first commitment greater than the provided height is the root of the tree
-            // that leaf at that HotShot height
+            // The first HotShot commitment found that is greater than or equal to the provided
```
philippecamacho, line 429: I would add "which corresponding height" after "HotShot commitment found". Otherwise it seems we are comparing a commitment to some number. 

Also the last part of the sentence is hard to read: "height is the root of the tree that leaf at that HotShot height".

alysiahuggins: yes it was confusing, updated here https://github.com/EspressoSystems/espresso-sequencer/commit/5a295bdc233406ffd908952b82bd644433111dc6

zacshowa (approved):
Looks good to me!

PR #2283, opened 2024-11-14
Use espresso-types without testing feature

### LINKED ISSUE #1491: Don't turn on `testing` feature for non-test builds
Need to at least remove `features = ["testing"]` from `sequencer = { path = "../sequencer", features = ["testing"] }` from the non-dev dependencies of builder. Then there are some compilation errors to fix because we have accidentally started to rely on the accidentally turned on "testing" feature.

Should be easier to fix when 
- https://github.com/EspressoSystems/espresso-sequencer/pull/1467

Is done because it will remove the call to `NodeState::mock()` which requires the "testing" feature.

sveitser:
We would like to be able to use the types crate in other crates but currently it won't compile without the testing feature enabled. This PR fixes that.

- Use `NullStateCatchup` instead of `MockStateCatchup` where we don't need catchup.
- Feature gate `MockStateCatchup` behind testing in types crate.
- Remove the unused `InstanceState` argument from `from_transactions_sync`.
- Add CI job to check compilation of each feature in each crate with `cargo hack`.
- Remove the libp2p feature.

After this change I can add the `types` crate as a dependency in a rust project without turning on the `testing` feature and run `cargo check` successfully.

Closes #1491


sequencer/Cargo.toml
```diff
@@ -54,7 +54,7 @@ derivative = "2.2"
 derive_more = { workspace = true }
 dotenvy = { workspace = true }
 dyn-clone = { workspace = true }
-espresso-types = { path = "../types", features = ["testing"] }
+espresso-types = { path = "../types" }
```
sveitser, line 57: Maybe we shouldn't do this.

sveitser: fixed in https://github.com/EspressoSystems/espresso-sequencer/pull/2283/commits/56dcb1e57f5875c4259831ad83d241865d4997e7 i think

imabdulbasit (approved):
LGTM!

sequencer/src/lib.rs
```diff
@@ -488,15 +482,6 @@ pub async fn init_node<P: PersistenceOptions, V: Versions>(
         ))
     };
 
-    // Wait for the CDN network to be ready if we're not using the P2P network
```
sveitser, line 491: @rob-maron I removed this wait and think that should be fine because we always have the libp2p enabled. Let me know if you see an issue with this.

PR #2284, opened 2024-11-14
Ab/sqlite

imabdulbasit:
This PR adds support for sqlite
   - Updates the hotshot-query-service which includes the sqlite support.
   - Adds separate migrations for PostgreSQL and SQLite. 
   - Some data types in the state table columns have been updated to JSONB, and a separate migration updates those data types. This was necessary because SQLite does not support BYTEA or arrays, so a BLOB column is used instead. This change unifies the queries across databases.
  -  Introduces a feature flag, `embedded-db,` which activates SQLite (embedded-db). If not provided, postgres is used by default.
  -  Updates workflows to include a job for testing PostgreSQL, as all-features includes embedded-db and tests with SQLite.
    -Updates the just recipes to enable local testing with both SQLite and PostgreSQL.

Note: SQLite sometimes lacks support for basic SQL syntax that one might expect, such as ORDER BY (col1, col2...) DESC. Instead, we use ORDER BY col1 DESC, col2 DESC to ensure compatibility with both PostgreSQL and SQLite.


sequencer/api/migrations/postgres/V43__update_state_tables_types.sql
```diff
@@ -0,0 +1,18 @@
+
+ALTER TABLE fee_merkle_tree
+  ALTER COLUMN path TYPE JSONB USING array_to_json(path);
+
+ALTER TABLE fee_merkle_tree
+  ALTER COLUMN children TYPE JSONB USING array_to_json(children);
+
+ALTER TABLE fee_merkle_tree
+  RENAME COLUMN index TO idx;
```
jbearer, line 9: Not sure if it matters but you can put more than one directive in an `ALTER TABLE` statement, so these 3 maybe could all be done in one statement

sequencer/api/migrations/postgres/V43__update_state_tables_types.sql
jbearer, line 1: The latest deployed migration is going to be 300 I think so this will have to come after, e.g. `V301__...`

sequencer/api/migrations/sqlite/V31__drop_merklized_state_created_index.sql
jbearer, line 1: Since we're going to be deploying these migrations for the first time for sqlite nodes, we could delete this migration and just not create these indices in the first place, if we wanted to

sequencer/src/persistence/sql.rs
```diff
@@ -277,13 +404,16 @@ impl From<PruningOptions> for PrunerCfg {
 impl PersistenceOptions for Options {
     type Persistence = Persistence;
 
-    async fn create(self) -> anyhow::Result<Persistence> {
+    async fn create(mut self) -> anyhow::Result<(Self, Self::Persistence)> {
```
jbearer, line 407: Would it be cleaner to have this take `&mut self` and only return `Self::Persistence`?

sequencer/Cargo.toml
```diff
@@ -75,31 +76,31 @@ hotshot-orchestrator = { workspace = true }
...
 libp2p = { workspace = true }
+libp2p-networking = { workspace = true }
 marketplace-solver = { path = "../marketplace-solver" }
 num_enum = "0.7"
 portpicker = { workspace = true }
 rand = { workspace = true }
 rand_chacha = { workspace = true }
 rand_distr = { workspace = true }
-sequencer-utils = { path = "../utils" }
+sequencer-utils = { path = "../utils", features = ["testing"] }
```
sveitser, line 103: Can we do without the "testing" feature in the build dependencies?

justfile
```diff
@@ -38,22 +42,34 @@ anvil *args:
...
 
 test-integration:
 	@echo 'NOTE that demo-native must be running for this test to succeed.'
 	cargo nextest run --all-features --nocapture --profile integration
 
 clippy:
+    @echo 'features: "all"'
     cargo clippy --workspace --all-features --all-targets -- -D warnings
+    @echo 'features: "testing"'
+    cargo clippy --workspace --all-features --features testing -- -D warnings
```
sveitser, line 72: this has all-features and testing

sequencer/api/migrations/postgres/V43__update_state_tables_types.sql
imabdulbasit: RENAME requires a separate statement

process-compose-sqlite.yaml
```diff
@@ -0,0 +1,531 @@
...
+        scheme: http
+        host: localhost
+        port: $ESPRESSO_SEQUENCER1_API_PORT
+        path: /healthcheck
+      failure_threshold: 100
+    availability:
+      exit_on_skipped: true
+
+  sequencer2:
+    command: sequencer -- http -- catchup -- status
```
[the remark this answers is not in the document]
imabdulbasit: The Cargo features are additive, so the `embedded-db` feature is automatically enabled for ` sequencer-postgres`  when running `cargo build`.

https://github.com/rust-lang/cargo/issues/674#issuecomment-904050069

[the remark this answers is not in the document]
imabdulbasit: Two options come to mind: use the cfg directive instead of a feature, or remove these two crates from the workspace.

sveitser:
The error 

> Error response from daemon: Head "https://ghcr.io/v2/espressosystems/espresso-sequencer/sequencer-sqlite/manifests/main": denied

on CI is expected because we don't have that image yet. I don't know why we pull docker images though. I don't think we need all of them.

sequencer-sqlite/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+# As a workaround for feature unification by cargo this separate crate that is
+# not part of the workspace that allows us to compile a sequencer binary without
+# the embedded-db feature turned on.
+[package]
+name = "sequencer-sqlite"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+sequencer = { path = "../sequencer" , features = ["embedded-db"] }
```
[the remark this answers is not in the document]
imabdulbasit: No, it doesn't work. I already tried it

sequencer/src/persistence/sql.rs
```diff
@@ -153,41 +182,157 @@ pub struct Options {
...
-    type Error = anyhow::Error;
+    // Keep the database connection pool when persistence is created,
+    // allowing it to be reused across multiple instances instead of creating
+    // a new pool each time such as for API, consensus storage etc
+    // This also ensures all storage instances adhere to the MAX_CONNECTIONS limit if set
+    //
+    // Note: Cloning the `Pool` is lightweight and efficient because it simply
+    // creates a new reference-counted handle to the underlying pool state.
+    #[clap(skip)]
+    pub(crate) pool: Option<sqlx::Pool<Db>>,
```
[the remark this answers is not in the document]
imabdulbasit: I am passing the options to init_with_query_module_sql() while everything else takes the persistence. The persistence in init_with_query_module_sql() is then created by reusing the pool

sequencer/src/persistence/sql.rs
```diff
@@ -861,23 +1025,31 @@ mod testing {
 
     #[async_trait]
     impl TestablePersistence for Persistence {
-        type Storage = TmpDb;
+        type Storage = Arc<TmpDb>;
 
         async fn tmp_storage() -> Self::Storage {
-            TmpDb::init().await
+            Arc::new(TmpDb::persistent().await)
```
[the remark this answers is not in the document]
imabdulbasit: Ah, I used it for debugging the sqlite and forgot to revert it

scripts/demo-native
```diff
@@ -18,13 +18,13 @@ ESPRESSO_BASE_STORAGE_PATH=$(mktemp -d -t espresso-XXXXXXXX)
...
 
-# If keeping the storage path, is desired for debugging, comment out the line
-# below or set a custom storage path in process-compose.yaml.
-trap "exit" INT TERM
-trap cleanup EXIT
-cleanup(){
-    echo "Cleaning up sequencer storage path: $ESPRESSO_BASE_STORAGE_PATH"
-    rm -rv "$ESPRESSO_BASE_STORAGE_PATH"
-}
+# # If keeping the storage path, is desired for debugging, comment out the line
```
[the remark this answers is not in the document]
sveitser: https://github.com/EspressoSystems/espresso-sequencer/pull/2284/commits/8db8f967bb468e06947548bac4a93dcdd1cab7e9

.github/workflows/test.yml
```diff
@@ -108,67 +140,102 @@ jobs:
...
+        
       - name: Test Integration
         env:
           NEXTEST_PROFILE: integration
           INTEGRATION_TEST_SEQUENCER_VERSION: ${{ matrix.version }}
         run: |
-          cargo nextest run --archive-file nextest-archive.tar.zst --verbose --no-fail-fast \
-          --workspace-remap $PWD $(if [ "${{ matrix.version }}" == "2" ]; then echo " smoke"; fi)
-        timeout-minutes: 40
+          cargo nextest run --no-run
```
sveitser, line 227: I think we should figure out of this was actually the problem. Since the tests in the `./tests` crate shouldn't change with the embedded-db feature. Any of the two nextest archives should be able to run the tests. And if the tests actually run I think it's very unlikely that the archive is the issue.

imabdulbasit: yeah lets revert it

justfile
```diff
@@ -38,22 +46,30 @@ anvil *args:
     docker run -p 127.0.0.1:8545:8545 ghcr.io/foundry-rs/foundry:latest "anvil {{args}}"
 
 test *args:
-	@echo 'Omitting slow tests. Use `test-slow` for those. Or `test-all` for all tests.'
-	cargo nextest run --locked --workspace --all-features --verbose {{args}}
+    @echo 'Omitting slow tests. Use `test-slow` for those. Or `test-all` for all tests.'
+    @echo 'features: "all"'
+    cargo nextest run --locked --workspace --all-features --verbose {{args}}
+    cargo nextest run --locked --workspace --verbose {{args}}
```
tbro, line 52: Why are we running all the tests twice now?

.github/workflows/build.yml
```diff
@@ -46,11 +46,14 @@ jobs:
...
           cargo build --locked --release --workspace
+      
+      - name: Build sequencer-sqlite 
+        run: cargo build --locked --release --manifest-path ./sequencer-sqlite/Cargo.toml --target-dir ./target
 
       - name: Build Espresso Dev Node
         # Espresso Dev Node currently requires testing feature, so it is built separately.
         run: |
-          cargo build --locked --release --features testing --bin espresso-dev-node
+          cargo build --locked --release --features "testing embedded-db" --bin espresso-dev-node
```
tbro, line 56: Can we explicitly specify the `embedded-db` features everywhere? I think in other places we are just specifying `--all-features` which isn't clear. Especially since we were using `--all-features` previously (*I think*).

.github/workflows/lint.yml
```diff
@@ -32,5 +32,8 @@ jobs:
       - name: Format Check
         run: cargo fmt -- --check
 
-      - name: Check
+      - name: Check (embedded-db)
```
tbro, line 35: Confused here too. Won't `--all-features` + `--all-targets` lint everything? If so, why run clippy a second time?

imabdulbasit: all-features would enable the "embedded-db" feature and so this would disable any postgres code. This is because we have #[cfg(not(feature = "embedded-db")) for the postgres code. We are doing lint check second time to see that postgres lint doesn't have any error. So one clippy run would check that sqlite lint and other one checks the postgres. The features here don't add the functionality but rather deactivate one and activate the other.

justfile
imabdulbasit: one uses sqlite for tests using database and the other uses postgres

.github/workflows/test.yml
```diff
@@ -40,13 +42,40 @@ jobs:
...
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+
+      - name: Enable Rust Caching
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-all-crates: true
+          cache-provider: buildjet
+
+      - name: Build and archive tests
```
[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2284/commits/09e08788d0af92129c96c235ebb498053a9aa16a

.github/workflows/build.yml
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2284/commits/09e08788d0af92129c96c235ebb498053a9aa16a

.github/workflows/test.yml
```diff
@@ -144,31 +206,32 @@ jobs:
...
       - name: Test Integration
         env:
           NEXTEST_PROFILE: integration
           INTEGRATION_TEST_SEQUENCER_VERSION: ${{ matrix.version }}
         run: |
-          cargo nextest run --archive-file nextest-archive.tar.zst --verbose --no-fail-fast \
+          cargo nextest run --archive-file nextest-archive-postgres.tar.zst --verbose --no-fail-fast --nocapture \
           --workspace-remap $PWD $(if [ "${{ matrix.version }}" == "2" ]; then echo " smoke"; fi)
-        timeout-minutes: 40
+        timeout-minutes: 20
```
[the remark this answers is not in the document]
imabdulbasit: changed it back to 10m

docker-compose.yaml
```diff
@@ -431,14 +435,16 @@ services:
     image: ghcr.io/espressosystems/espresso-sequencer/sequencer:main
     ports:
       - "$ESPRESSO_SEQUENCER4_API_PORT:$ESPRESSO_SEQUENCER_API_PORT"
-    command: sequencer -- http -- catchup -- status
+    command: sequencer-sqlite -- storage-sql -- http -- query -- catchup -- status
```
jbearer, line 438: Shouldn't this work without `sequencer-sqlite` and `storage-sql` if we set `ESPRESSO_SEQUENCER_EMBEDDED_DB=true`?

justfile
```diff
@@ -9,11 +9,19 @@ demo *args:
 
 demo-native *args:
     cargo build --profile test
+    cargo build --profile test --manifest-path ./sequencer-sqlite/Cargo.toml
```
jbearer, line 12: Why don't we use the `build` target here like in `demo-native-mp`?

process-compose.yaml
```diff
@@ -225,7 +225,7 @@ processes:
       - ESPRESSO_SEQUENCER_API_PORT=$ESPRESSO_SEQUENCER2_API_PORT
       - ESPRESSO_SEQUENCER_LIBP2P_BIND_ADDRESS=0.0.0.0:$ESPRESSO_DEMO_SEQUENCER_LIBP2P_PORT_2
       - ESPRESSO_SEQUENCER_LIBP2P_ADVERTISE_ADDRESS=localhost:$ESPRESSO_DEMO_SEQUENCER_LIBP2P_PORT_2
-      - ESPRESSO_SEQUENCER_STATE_PEERS=http://localhost:$ESPRESSO_SEQUENCER3_API_PORT
+      - ESPRESSO_SEQUENCER_STATE_PEERS=http://localhost:$ESPRESSO_SEQUENCER_API_PORT, http://localhost:$ESPRESSO_SEQUENCER3_API_PORT
```
jbearer, line 228: Why this change? Does this mean we aren't able to use the sqlite node for state catchup?

process-compose.yaml
```diff
@@ -332,6 +333,8 @@ processes:
       - ESPRESSO_SEQUENCER_IDENTITY_LATITUDE=-25.2744
       - ESPRESSO_SEQUENCER_IDENTITY_LONGITUDE=133.7751
       - ESPRESSO_SEQUENCER_PUBLIC_API_URL=http://localhost:$ESPRESSO_SEQUENCER4_API_PORT/
+      - ESPRESSO_SEQUENCER_SQLITE_PATH=$ESPRESSO_BASE_STORAGE_PATH/seq4-sqlite.db
```
jbearer, line 336: IMO we shouldn't have a separate variable for this. It should just be some fixed sub-directory of `ESPRESSO_SEQUENCER_STORAGE_PATH`, so that this is backwards compatible with the current FS-based storage (eventually we will ask operators to switch)

sequencer/src/persistence/sql.rs
```diff
@@ -132,17 +161,20 @@ pub struct Options {
     /// (and, if needed, replaced), even if it is otherwise healthy. It is good practice to refresh
     /// even healthy connections once in a while (e.g. daily) in case of resource leaks in the
     /// server implementation.
-    #[clap(long, env = "ESPRESSO_SEQUENCER_POSTGRES_CONNECTION_TIMEOUT", value_parser = parse_duration, default_value = "30m")]
+    #[clap(long, env = "ESPRESSO_SEQUENCER_DATABASE_CONNECTION_TIMEOUT", value_parser = parse_duration, default_value = "30m")]
     pub(crate) connection_timeout: Duration,
 
+    #[clap(long, env = "ESPRESSO_SEQUENCER_DATABASE_SLOW_STATEMENT_THRESHOLD", value_parser = parse_duration, default_value = "1m")]
```
jbearer, line 167: The default for this is currently 1s and I think we should keep that; 1s is slow for a single statement and we generally want to know about those. 1m is ridiculous. Maybe we could have a different default for sqlite if 1s makes it too verbose

jbearer (commented):
Code looks good, some comments about configuration.

In terms of releasing, unfortunately we are a bit backed up atm due to recent patches regarding the L1 client. Ideally I think we should wait for the current weekly release to merge, then merge this right after. That way we can cut a release branch which includes the fixes for the L1 client but excludes sqlite, so we can QA and deploy one big change at a time. This would then go into the following release.

docker-compose.yaml
imabdulbasit: I think we still have to pass storage-sql module as it just loads a different binary?

imabdulbasit: I can update the script to pass the --storage-sql module

process-compose.yaml
imabdulbasit: No, was debugging integration CI. reverted

.github/workflows/lint.yml
tbro: Humm is this not addressed by having sqlite in a separate package? `cargo lint --workspace` would hit both the package w/ `embedded-db` and the other sequencer binary that does not enable that feature.

tbro (approved):
I'm still looking over some things, but since this is already approved and tests are passing I'm approving also. We can work out details in future PRs.

.github/workflows/lint.yml
imabdulbasit: its not the binary we are checking. We are checking the sequencer library with embedded-db code activated and postgres disabled and vice versa

PR #2291, opened 2024-11-15
Decrease load on DB

jbearer:
We have seen that in some cases, when storage is slow or the node is way behind on its event stream, it can trigger many fetches of old proposals, which other nodes have all garbage collected. These fetches accumulate over time, leading to a worsening problem.

While debugging this, I also noticed that we don't "unfill" the leaf payload when storing anchor leaves, so DB performance gets worse when blocks are very large, because we are storing/loading these large payloads when we think we are only dealing with leaves.

### This PR:
Limits the number of parallelism in proposal fetching. Instead of spawning a new task dynamically each time we need to fetch a proposal, we spawn a fixed number of worker tasks, each of which will only fetch one proposal at a time. A scanner task follows the event stream and detects when a proposal needs to be fetched, then broadcasts the request to fetch it to the worker tasks. It will be picked up when a worker is free.

Unfills leaf payloads before storing decided leaves.


sequencer/src/context.rs
```diff
@@ -533,20 +569,25 @@ async fn fetch_proposal_chain<N, P, V>(
...
+            Ok(())
+        }
+        .instrument(span)
+        .await;
+        if let Err(err) = res {
+            tracing::warn!("failed to fetch proposal: {err:#}");
+
+            // If we fail fetching the proposal, don't let it clog up the fetching task. Just push
+            // it back onto the queue and move onto the next proposal.
+            sender.broadcast_direct((view, leaf)).await.ok();
```
sveitser, line 589: Does this create a busy loop if the task is failing consistently?

jbearer: I guess it kind of does. I was thinking we would be rate limited by the finite number of workers and time spent waiting on I/O for each failure. But I guess that's not guaranteed, depending on what the failure is. I'll add a sleep here

rob-maron (approved):

PR #2294, opened 2024-11-18
fix: dtype for gasPrice in BlockNative Oracle

### LINKED ISSUE #2293: Fix price oracle error
We have recieved Gas oracle error in the mainnet log, reported by @Ancient123 

```
"!! BlockNative Price Oracle failed: error decoding response body: invalid type: floating point 41.0, expected u64 at line 1 column 67"
```

after investigation, the data type for `.maxPrice` is float instead of u64, (maybe this changed, but in any case we should add a unit-test to constantly monitor).

alxiong:
Closes #2293


utils/src/blocknative.rs
```diff
@@ -148,3 +147,10 @@ fn gas_category_to_confidence(gas_category: &GasCategory) -> u64 {
         GasCategory::Fastest => 99,
     }
 }
+
+#[tokio::test]
+async fn test_blocknative_gas_oracle() -> Result<()> {
+    let gas_oracle = BlockNative::new(None).category(GasCategory::SafeLow);
+    let (_max_fee, _priority_fee) = gas_oracle.estimate_eip1559_fees().await?;
```
sveitser, line 154: Not a big fan of making requests to third party APIs from unittests because our tests/CI would fail if their API has issues, or we get rate limited which is always a bit tricky to anticipate.

sveitser (approved):

utils/src/blocknative.rs
alxiong: so either we catch it in CI time, or we setup a cron job, or we detect only on runtime failure? 
the rate limit is once per second, which should be fine? 

I'm happy to remove it.

sveitser: Yes, but even if we run this test in the CI it will still fail in production first unless we run the CI very often.

alxiong: okay, remove this test in [c16366d](https://github.com/EspressoSystems/espresso-sequencer/pull/2294/commits/c16366d2c47ae94c2d7f61592fb5cb9ee280bdf8)

PR #2300, opened 2024-11-19
ci: remove cruft

### LINKED ISSUE #2289: CI: remove submodule checkout for jobs that don't need them
I think we only need it in jobs where we compile the contracts. This will make CI builds a bit faster.

sveitser:
- Only checkout submodules when neeeded (building contracts)
- Only install forge when needed (build contracts, docs), edit: anvil is needed to **run** the tests
- ~~Nextest is already installed on ubuntu-latest~~ (edit: I have seen it installed, however things failed here when I removed it)

Some jobs might fail, will fix them after CI runs.

Closes #2289


contracts/.github/workflows/test.yml
```diff
@@ -1,34 +0,0 @@
-name: test
```
sveitser, line 1: This file is not used probably created by foundry when initializing the project.

tbro (approved):
:+1:

PR #2302, opened 2024-11-19
Add stake table endpoint

rob-maron:
### This PR:

Adds a stake table endpoint to sequencer nodes at `/node` to both:
- Get the stake table at the current epoch (`/node/stake-table/current`)
- Get the stake table at a specific epoch (`/node/stake-table/:epoch_number`)

We need this for components like the CDN and state prover to get the stake table without needing to run a lightweight node

### Key places to review:

`api.rs`


jbearer (approved):
LGTM

PR #2325, opened 2024-11-25
Permissioned L1 stake table

sveitser:
Some info: https://www.notion.so/espressosys/Permissioned-stake-table-contract-for-Dec-2024-release-1462431b68e980a0a4abcd6bb65d7e43?pvs=4

### TODO
- [x] Pass stake table loaded from file to contract during deployment. This requires conversion from the PeerConfigKeys to the solidity struct `NodeInfo`.
- [x] Clean up hotshot/jellyfish to and from solidity types conversion code: should all be moved to the contract adapter crate.
- [x] Add deployment of permissoned stake table to demos.
- Fetching:  See #2361 
- E2E test for stake table upgrade. (-> separate issue)
- [x] Tests for crypto key conversions between solidity and rust

### Follow up work
- #2361 
- #2384


contracts/src/PermissionedStakeTable.sol
```diff
@@ -0,0 +1,66 @@
...
+        /// The consensus signing key. Only used for storage in this contract.
+        EdOnBN254.EdOnBN254Point schnorrVK;
+        /// Is the Node DA Node? Only used for storage in this contract.
+        bool isDA;
+    }
+
+    // State mapping from staker IDs to their staking status
+    mapping(bytes32 nodeID => bool isStaker) private stakers;
+
+    constructor(address initialOwner, NodeInfo[] memory initialStakers) Ownable(initialOwner) {
```
alysiahuggins, line 31: in what scenarios do we imagine the initialOwner being different from the deployer?

contracts/src/PermissionedStakeTable.sol
```diff
@@ -0,0 +1,66 @@
...
+        insert(initialStakers);
+    }
+
+    function insert(NodeInfo[] memory newStakers) public onlyOwner {
+        // TODO: revert if array empty
+        for (uint256 i = 0; i < newStakers.length; i++) {
+            bytes32 stakerID = _hashBlsKey(newStakers[i].blsVK);
```
alysiahuggins, line 38: should there be a check to validate this blsVK? E.g. ensuring it's non zero. Or perhaps there are some crypto libraries that can be used for validity checks.

contracts/src/PermissionedStakeTable.sol
```diff
@@ -0,0 +1,66 @@
...
+ * @dev An stake table mapping with owner-only access control.
+ */
+contract PermissionedStakeTable is Ownable {
+    event Added(NodeInfo[]);
+    event Removed(NodeInfo[]);
+
+    error StakerAlreadyExists(BN254.G2Point);
+    error StakerNotFound(BN254.G2Point);
+
+    struct NodeInfo {
```
alysiahuggins, line 19: if a validator loses their key and needs to be readded is the workflow that we remove them from the stakers mapping and re-add them with their new info?

contracts/src/PermissionedStakeTable.sol
```diff
@@ -0,0 +1,66 @@
...
+        /// The consensus signing key
+        BN254.G2Point blsVK;
+        /// The consensus signing key. Only used for storage in this contract.
+        EdOnBN254.EdOnBN254Point schnorrVK;
+        /// Is the Node DA Node? Only used for storage in this contract.
+        bool isDA;
+    }
+
+    // State mapping from staker IDs to their staking status
+    mapping(bytes32 nodeID => bool isStaker) private stakers;
```
alysiahuggins, line 29: should we also add a validator threshold to this contract to ensure that there are no more than 100 stakers as per this version of our network?

contracts/src/PermissionedStakeTable.sol
alysiahuggins, line 19: nit: Just thinking about naming, we've been using node and staker interchangeably? Could be good to just use one term to refer to this entity

alysiahuggins: this struct also isn't used below except for parameters and event emissions with those parameters, do we intend to use the other fields in this struct?

justfile
```diff
@@ -79,7 +79,7 @@ build-docker-images:
     scripts/build-docker-images-native
 
 # generate rust bindings for contracts
-REGEXP := "^LightClient$|^LightClientStateUpdateVK$|^FeeContract$|PlonkVerifier$|^ERC1967Proxy$|^LightClientMock$|^LightClientStateUpdateVKMock$|^PlonkVerifier2$"
+REGEXP := "^LightClient$|^LightClientStateUpdateVK$|^FeeContract$|PlonkVerifier$|^ERC1967Proxy$|^LightClientMock$|^LightClientStateUpdateVKMock$|^PlonkVerifier2$|^PermissionedStakeTable$"
```
alysiahuggins, line 82: hey @alxiong, do we still need PlonkVerifier2?

utils/src/deployer.rs
```diff
@@ -426,6 +440,27 @@ pub async fn deploy(
...
+            )
+            .await?;
+        let stake_table = PermissionedStakeTable::new(stake_table_address, l1.clone());
+
+        // Transfer ownership to the multisig wallet if provided.
+        if let Some(owner) = multisig_address {
+            tracing::info!(
+                %stake_table_address,
+                %owner,
+                "transferring fee contract proxy ownership to multisig",
```
alysiahuggins, line 458:
```suggestion
                "transferring PermissionedStakeTable ownership to multisig",
```

docker-compose.yaml
```diff
@@ -17,7 +17,7 @@ services:
 
   deploy-sequencer-contracts:
     image: ghcr.io/espressosystems/espresso-sequencer/deploy:main
-    command: deploy --only fee-contract
```
alysiahuggins, line 20: 👍

justfile
alysiahuggins: also we no longer need the fee contract i don't think? @sveitser

contracts/src/PermissionedStakeTable.sol
sveitser: I was thinking if we want to set the multisig as owner during deployment. But yes, we should probably just do the same as with our other contracts.

contracts/src/PermissionedStakeTable.sol
sveitser: I'm thinking we might want to add a function that allows to remove some and add some stakers in one transaction so that it's easy to make an atomic change.

We only really care about the Events for building the stake table here. We expect this to be called very rarely so gas isn't much of a concern.

justfile
alxiong: re: PV2
we don't. we can remove them (can always fish out of the git history if needed)

contracts/src/PermissionedStakeTable.sol
alysiahuggins: yes that's valid but agree to keep it inline with other deployments. Confirmed with Mat and Tobias that there's no need for a different initialOwner other than the deployer since we immediately change ownership to the multisig.

justfile
tbro: > also we no longer need the fee contract i don't think? @sveitser

We don't need fee contract anymore?

sveitser: @alysiahuggins how do the builders pay fees if we don't have the fee contract anymore?

alysiahuggins:
Ah didn’t realize it was actively being used in this version.

docker-compose.yaml
```diff
@@ -17,7 +17,7 @@ services:
...
+    command: deploy --only sequencer,permissioned-stake-table
```
imabdulbasit, line 20: what is the `sequencer` contract here? I only see fee, LC, and stake table contracts

sequencer/src/bin/deploy.rs
```diff
@@ -111,6 +112,20 @@ struct Options {
...
 
+    /// A toml file with the initial stake table.
+    ///
+    /// Schema:
+    ///
+    /// public_keys = [
+    ///   {
+    ///     stake_table_key = "BLS_VER_KEY~...",
+    ///     state_ver_key = "SCHNORR_VER_KEY~...",
+    ///     da = true,
```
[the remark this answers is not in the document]
sveitser: Good point. I think in the sequencer we shouldn't really use PeerConfigKeys but have our own type (and also not be generic over the pubkey). However for loading the initial stake table I think it's fine to use PeerConfigKeys because this is the type that I think the orchestrator loads from file on startup and initially we will put the same stake table that was used by the orchestrator into the contract. However for this version we don't need a stake amount field so I chose not to add it. 

We can add it to the initial stake table file or somehow make the rust code work without it. I think both are fine.

docker-compose.yaml
sveitser: I might have edited this by accident because I was messing with contract groups some time ago.

justfile
sveitser: We are using the fee contract.

contracts/src/PermissionedStakeTable.sol
sveitser: So I think I fixed this

contracts/src/PermissionedStakeTable.sol
sveitser: https://github.com/EspressoSystems/espresso-sequencer/issues/2384

contracts/src/PermissionedStakeTable.sol
sveitser: https://github.com/EspressoSystems/espresso-sequencer/issues/2384

contracts/src/PermissionedStakeTable.sol
alysiahuggins: i'm guessing this is not needed because it's trusted? @sveitser @tbro

sveitser: https://github.com/EspressoSystems/espresso-sequencer/issues/2384

alysiahuggins (approved):
Approving so that work can progress but a question around event logs. 

So theres's an event when the stakers list is updated but not an event when one is inserted or removed. 
Is that because inserts only happen at the beginning when the existing list of stakers are added to the contract? 
Though I presume a removal can only happen in the future. 

Assuming there's a service that depends entirely on this contract to know the full active set of stakers, shouldn't there be emit logs for those methods too? 

*EDITED*

only the update is public so nevermind

contracts/src/PermissionedStakeTable.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    constructor(NodeInfo[] memory initialStakers) Ownable(msg.sender) {
+        insert(initialStakers);
+    }
+
+    // public methods
+
+    function update(NodeInfo[] memory stakersToRemove, NodeInfo[] memory newStakers)
```
sveitser, line 36: @alysiahuggins only the `update` method is public.

alysiahuggins: ah missed that, thanks

PR #2335, opened 2024-11-26
Add `:height` param to balance endpoint and use in test

tbro:
Adds `:height` param to balance endpoint. Adds client method, and use it in tests.


tbro:
Currently getting error: 
```bash
called `Result::unwrap()` on an `Err` value: ServerError { status: StatusCode(404), message: "No route matches /fee-balance/19/0x23618e81e3f5cdf7f54c3d65f7fbc0abf5b21e8f<br></br>\n 
```

PR #2344, opened 2024-12-02
[WEEKLY RELEASE] Bump HotShot

QuentinI:
- Updates HotShot with epoch related changes, incleading new Leaf2 and QuorumProposal2 types.
- Moves the Marketplace upgrade to version 0.99 to make room for the Epoch upgrade which will be carried out first.


sequencer/src/api.rs
```diff
@@ -318,23 +317,10 @@ impl<
...
-            view,
-            View {
-                view_inner: ViewInner::Leaf {
-                    state,
-                    delta,
-                    leaf: leaf_commit,
-                },
-            },
-        ) {
+        if let Err(err) = consensus.update_leaf(leaf, Arc::clone(&state), delta) {
```
QuentinI, line 323: `update_saved_leaves` and `update_validated_state_map` are now private. I figured `update_leaf` should cover the need for both of those

sequencer/src/persistence/sql.rs
```diff
@@ -363,15 +363,16 @@ impl SequencerPersistence for Persistence {
         let json = serde_json::to_value(cfg)?;
 
         let mut tx = self.db.write().await?;
-        tx.execute_one_with_retries("INSERT INTO network_config (config) VALUES ($1)", (json,))
```
QuentinI, line 366: @imabdulbasit not sure what to do here. Would just lifting the old code for `execute_one_with_retries` here be sufficient?

types/src/v0/traits.rs
```diff
@@ -498,6 +503,7 @@ pub trait SequencerPersistence: Sized + Send + Sync + 'static {
         // starting in a view in which we had already voted before the restart, and prevents
         // unnecessary catchup from starting in a view earlier than the anchor leaf.
         let view = max(highest_voted_view, leaf.view_number());
+        let epoch = todo!();
```
QuentinI, line 506: No idea where this should come from. I'd expect there to be `Leaf2::epoch_number()`, but that's not a thing.

sequencer/src/api/sql.rs
```diff
@@ -442,8 +442,8 @@ where
     .bind(param)
     .fetch_one(tx.as_mut())
     .await?;
-    let proposal: Proposal<SeqTypes, QuorumProposal<SeqTypes>> = bincode::deserialize(&data)?;
-    Ok(Leaf::from_quorum_proposal(&proposal.data))
+    let proposal: Proposal<SeqTypes, QuorumProposal2<SeqTypes>> = bincode::deserialize(&data)?;
```
imabdulbasit, line 445: are we inserting QuorumProposal1 or QuorumProposal2 in database? If we are storing QuorumProposal2 then we should have a migration to make the existing proposals (q1) to q2?

sequencer/src/persistence/fs.rs
```diff
@@ -623,8 +635,10 @@ impl SequencerPersistence for Persistence {
     }
     async fn append_quorum_proposal(
         &self,
-        proposal: &Proposal<SeqTypes, QuorumProposal<SeqTypes>>,
+        proposal: &Proposal<SeqTypes, QuorumProposal2<SeqTypes>>,
```
imabdulbasit, line 638: ok we are storing QuorumProposal2. I think that existing QuorumProposal1 stored will not be deserialized into QuorumProposal2 so we need to think about that

sequencer/src/persistence/sql.rs
```diff
@@ -743,9 +764,9 @@ async fn collect_garbage(
         .map(|row| {
             let view: i64 = row.get("view");
             let leaf_data: Vec<u8> = row.get("leaf");
-            let leaf = bincode::deserialize::<Leaf>(&leaf_data)?;
+            let leaf = bincode::deserialize::<Leaf2>(&leaf_data)?;
```
imabdulbasit, line 767: so we stored Leaf1 in anchor_leaf table before. We should deserialize it as Leaf1 and then convert it to Leaf2

sequencer/src/persistence.rs
```diff
@@ -20,6 +22,42 @@ pub trait ChainConfigPersistence: Sized + Send + Sync {
     async fn insert_chain_config(&mut self, chain_config: ChainConfig) -> anyhow::Result<()>;
 }
 
+fn downgrade_leaf(leaf2: Leaf2) -> Leaf {
+    if leaf2.drb_seed != [0; 96] && leaf2.drb_result != [0; 32] {
+        tracing::error!("Losing DRB information!");
```
sveitser, line 27: Should this be a panic. Is continuing here ever the right thing to do?

sequencer/src/persistence/fs.rs
```diff
@@ -753,6 +769,16 @@ impl SequencerPersistence for Persistence {
...
     }
+
+    async fn migrate_consensus(
+        &self,
+        _migrate_leaf: fn(Leaf) -> Leaf2,
+        _migrate_proposal: fn(
+            Proposal<SeqTypes, QuorumProposal<SeqTypes>>,
+        ) -> Proposal<SeqTypes, QuorumProposal2<SeqTypes>>,
+    ) -> anyhow::Result<()> {
+        Ok(())
```
sveitser, line 780: Should add a comment on why it's okay to do nothing here and/or link to issue where it will be implemented.

sequencer/src/persistence.rs
sveitser: 4e2e3398946f36908a76f77a0f007e76c30d15f1

sequencer/src/persistence/sql.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2344/commits/18c838e47b2d92d8c8870f8da393b16d486122bc

sequencer/src/persistence/sql.rs
```diff
@@ -743,9 +764,9 @@ async fn collect_garbage(
...
             let qc_data: Vec<u8> = row.get("qc");
-            let qc = bincode::deserialize::<QuorumCertificate<SeqTypes>>(&qc_data)?;
+            let qc = bincode::deserialize::<QuorumCertificate2<SeqTypes>>(&qc_data)?;
```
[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2344/commits/18c838e47b2d92d8c8870f8da393b16d486122bc

sveitser (approved):
I think this looks okay now. Since many collaborated should have more than a single approval though.

Cargo.toml
```diff
@@ -58,57 +58,59 @@ dyn-clone = "1.0"
...
-hotshot-builder-api = { git = "https://github.com/EspressoSystems/HotShot.git", tag = "0.5.81" }
-hotshot-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.1.56" }
-marketplace-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.1.56" }
-marketplace-builder-shared = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.1.56" }
-hotshot-events-service = { git = "https://github.com/EspressoSystems/hotshot-events-service.git", tag = "0.1.54" }
-hotshot-orchestrator = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.81" }
-hotshot-query-service = { git = "https://github.com/EspressoSystems/hotshot-query-service", tag = "0.1.74" }
-hotshot-stake-table = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.81" }
+hotshot-builder-api = { git = "https://github.com/EspressoSystems/hotshot", tag = "rc-0.5.82" }
+hotshot-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", branch = "hotshot/0.5.82" }
```
imabdulbasit, line 70: need to change it to tags?

imabdulbasit, line 70: point to tag?

sequencer/src/persistence/sql.rs
```diff
@@ -486,9 +488,9 @@ impl Persistence {
             if hash.is_none() {
                 let view: i64 = row.try_get("view")?;
                 let data: Vec<u8> = row.try_get("data")?;
-                let proposal: Proposal<SeqTypes, QuorumProposal<SeqTypes>> =
+                let proposal: Proposal<SeqTypes, QuorumProposal2<SeqTypes>> =
```
QuentinI, line 491: @imabdulbasit sorry, why this change? Aren't we storing V1 types?

sequencer/src/persistence/sql.rs
QuentinI: Good catch,  thanks

sequencer/src/persistence/sql.rs
imabdulbasit: tests were failing, and this change made them pass, but this needs to be reverted. 
reverted and fixed test_quorum_proposals_leaf_hash_migration()

https://github.com/EspressoSystems/espresso-sequencer/pull/2344/commits/a84d23244ee926618a464d2428ef20fec7f2e838

sequencer/src/persistence.rs
```diff
@@ -176,16 +216,19 @@ mod persistence_tests {
...
                     &ValidatedState::default(),
                     &NodeState::mock(),
                 )
-                .await,
+                .await
+                .to_qc2(),
                 upgrade_certificate: None,
-                proposal_certificate: None,
+                view_change_evidence: None,
+                drb_seed: [0; 32],
```
ss-es, line 230: can we use `hotshot_types::drb::INITIAL_DRB_SEED_INPUT` and `hotshot_types::drb::INITIAL_DRB_RESULT` for the constants?

they should be the same thing but we may still change the "default" type/number

ss-es (commented):
looks good on the whole, just left some comments

sequencer/src/api/sql.rs
```diff
@@ -433,7 +433,7 @@ async fn get_leaf_from_proposal<Mode, P>(
     tx: &mut Transaction<Mode>,
     where_clause: &str,
     param: P,
-) -> anyhow::Result<Leaf>
+) -> anyhow::Result<Leaf2>
```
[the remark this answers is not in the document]
imabdulbasit: Salman is right, we need to deserialize it to QuorumProposal1 in get_leaf_from_proposal()

sequencer/src/api.rs
```diff
@@ -1265,22 +1249,26 @@ mod api_tests {
...
+            drb_seed: [0; 32],
+            drb_result: [0; 32],
         };
         let mut qc = QuorumCertificate::genesis::<MockSequencerVersions>(
             &ValidatedState::default(),
             &NodeState::mock(),
         )
-        .await;
+        .await
+        .to_qc2();
```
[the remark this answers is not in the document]
QuentinI: Not available yet

sequencer/src/api/sql.rs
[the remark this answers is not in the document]
QuentinI: Oof, thank you for the catch!

> return `Leaf1`
@imabdulbasit is there a reason we don't want to return `Leaf2` by conversion?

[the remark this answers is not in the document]
imabdulbasit: no good reason, you can return leaf2/leaf1 whatever you prefer

imabdulbasit (approved):

sequencer/src/persistence.rs
```diff
@@ -20,6 +22,42 @@ pub trait ChainConfigPersistence: Sized + Send + Sync {
     async fn insert_chain_config(&mut self, chain_config: ChainConfig) -> anyhow::Result<()>;
 }
 
+fn downgrade_leaf(leaf2: Leaf2) -> Leaf {
+    if leaf2.drb_seed != [0; 32] && leaf2.drb_result != [0; 32] {
+        panic!("Downgrade of Leaf2 to Leaf will lose DRB information!");
```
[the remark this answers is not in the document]
sveitser: I was thinking if we are calling this on a leaf2 with DRB info then it probably indicates that our assumptions (e. g. this should not happen) are no longer correct at which point I think it's reasonable to stop running before potentially doing more damage, for example by throwing away important information. That's why I changed it to a panic.

ss-es (approved):

PR #2351, opened 2024-12-04
Fix Missing Block Data For Node Validator

Ayiga:
### This PR:

- Fixes missing Block Size Data by sourcing the block data from a separate Block Stream
- Modifies the way logic for how these data streams are connected, and how they get reconnected when closed
- Modifies the processing logic to operate on a pair of Leaf and Block data, as the Block data can no longer be reconstructed from a serialized Leaf
- Removed an early return that may have potentially caused the WebSocket to no longer process incoming connections


node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
```diff
@@ -411,39 +464,63 @@ mod test {
...
-                    .unwrap(),
-            ),
-            leaf_sender,
+        let client = surf_disco::Client::new(
+            "https://query.main.net.espresso.network/v0/"
+                .parse()
+                .unwrap(),
         );
 
+        // Let's get the current starting block height.
```
tbro, line 473: I think there is identical code in `lib.rs` (line 247). Can we de-duplicate?

node-metrics/src/lib.rs
```diff
@@ -240,20 +244,43 @@ pub async fn run_standalone_service(options: Options) {
...
+    let client = surf_disco::Client::new(options.leaf_stream_base_url().clone());
+
+    // Let's get the current starting block height.
+    let block_height = {
+        let block_height_result = client.get("status/block-height").send().await;
+        let block_height: u64 = match block_height_result {
+            Ok(block_height) => block_height,
+            Err(err) => {
+                tracing::warn!("retrieve block height request failed: {}", err);
+                panic!("error retrieving block height request failed: {}", err);
```
tbro, line 258: I'm unsure what depends on this library, but panicking after a single request seems drastic. Should we retry a few times?

node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
```diff
@@ -260,6 +263,53 @@ impl ProcessExternalMessageHandlingTask {
...
+        }
+
+        let mut url_sender = url_sender;
+        tracing::debug!("sending initial urls to url sender to process node identity");
+        loop {
+            for url in urls.iter() {
+                let send_result = url_sender.send(url.clone()).await;
+                if let Err(err) = send_result {
+                    tracing::error!("url sender closed: {}", err);
+                    panic!("SubmitPublicUrlsToScrapeTask url sender is closed, unrecoverable, the node state will stagnate.");
```
tbro, line 303: I don't think we want to panic here. If we get an error we want to still submit the other urls, no? And we also want to send again in five minutes, right?

node-metrics/src/lib.rs
tbro: So looking closer, I guess we can't proceed if this fails, so panicking may be appropriate. But shouldn't we retry a few times?

node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
Ayiga: This one is a little odd.  This specific test case isn't really meant to test the code, hence it has #[ignore] applied to it.  Instead, I utilize this test to run again the live server to make sure everything is working without issue.  It's more a convenience thing and an example for how to setup the service, and make sure it's working with little effort.  AS a result, it is being fed hard-coded values rather than deriving them from a `Config`.

node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
Ayiga: The reason we panic here is because the only conceivable error we are able to receive here is a `SendError`.  I believe the `SendError` indicates that the other side of the channel is closed, so there is no longer anyone to receive the `url`.  If that's the case, then we will be unable to make any additional progress either now or in the future.

I suppose I could just return out of this loop here instead of `panic`.  However, if this is a one-off issue that only affects the `url_sender`, then it means that we'd simply continue running with potentially stale data.  This isn't ideal, but it **could** still make progress.

What do you think?

node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
Ayiga: This has been modified to remove the duplicated logic in 159631aa321620607c84e008dc94b13fec259045.

node-metrics/src/lib.rs
Ayiga: The only thing that depends on this library at the moment is the `main` file that launches it within the same directory:

`node-metrics/src/main.rs`.

I will look into retrying this initial request.  Though this only happens on service launch at the moment.  So if the service isn't able to secure this value, it's better to fail fast and allow for the service to be restarted by the underlying orchestration services, which will effectively retry them anyway.

tbro: I had imagined that the intent was for the orchestration layer to restart the service. So I think it is *workable* but not ideal. Network issues are frequent and I think a 1 to 1 relationship between network caos and service crashes is problematic. Please consider adding retries in a future update.

tbro: Oh now I see you already did this in [1f8fb0e](https://github.com/EspressoSystems/espresso-network/pull/2351/commits/1f8fb0ecc02fb93d30e2c4d60c031f46aad8f8c4)

node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
tbro: Thanks for the response. I don't have a great solution off the top of my head. Feels like there may be a more graceful way of going about this. But your right, this should be good enough for the time being.

PR #2361, opened 2024-12-05
Hotshot 0.5.83

tbro:
Mostly Proof of Stake updates. Includes Hotshot w/ preliminary epochs and `Leaf2`  and related types (currently no-op).

Also:
  * `Membership` trait update
  * `Membership` implementation (stake table cache).


types/src/v0/impls/l1.rs
```diff
@@ -1246,4 +1261,39 @@ mod test {
...
+        let provider =
+            Provider::<Http>::try_from(anvil.endpoint())?.interval(Duration::from_millis(10u64));
+        let client =
+            SignerMiddleware::new(provider.clone(), wallet.with_chain_id(anvil.chain_id()));
+        let client = Arc::new(client);
+
+        // Initialize a contract with some deposits
+
+        // deploy the fee contract
+        let stake_table_contract =
```
tbro, line 1288: link

types/src/v0/impls/stake_table.rs
```diff
@@ -0,0 +1,331 @@
...
+#[derive(Clone, Debug)]
+pub struct StaticCommittee<T: NodeType> {
+    /// The nodes eligible for leadership.
+    /// NOTE: This is currently a hack because the DA leader needs to be the quorum
+    /// leader but without voting rights.
+    eligible_leaders: Vec<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
+
+    /// The nodes on the committee and their stake
+    // TODO I feel like this should be `HashSet` instead of `Vec`
+    stake_table: HashSet<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
```
tbro, line 28: link

types/src/v0/impls/stake_table.rs
```diff
@@ -0,0 +1,347 @@
...
+            .await;
+
+        let added: Vec<NodeInfoJf> = updates
+            .into_iter()
+            .flat_map(|e: StakersUpdatedFilter| e.added.into_iter().map(NodeInfoJf::from))
+            .collect();
+        for node in added {
+            if !node.da {
+                let entry: <TYPES::SignatureKey as SignatureKey>::StakeTableEntry =
+                    StakeTableEntry {
```
tbro, line 73: Need to match up types here.

types/src/v0/impls/stake_table.rs
```diff
@@ -0,0 +1,347 @@
...
+    cmp::max,
+    collections::{BTreeMap, HashSet},
+    num::NonZeroU64,
+    str::FromStr,
+};
+use thiserror::Error;
+use url::Url;
+
+#[derive(Clone, Debug)]
+pub struct StaticCommittee<T: NodeType> {
```
imabdulbasit, line 23: Can we avoid using generics here since we already have `SeqTypes` in this crate?

tbro: Yes

tbro: ef90df412caf7a0757dfc7b6387a22471136f8ff

types/src/v0/impls/stake_table.rs
tbro: https://github.com/EspressoSystems/espresso-sequencer/commit/ef90df412caf7a0757dfc7b6387a22471136f8ff

contracts/src/StakeTable.sol
```diff
@@ -21,6 +21,9 @@ contract StakeTable is AbstractStakeTable {
     /// account.
     error NodeAlreadyRegistered();
 
+    /// Error raised when a user tries to withdraw funds from a node that is not registered.
+    error NodeNotRegistered();
```
imabdulbasit, line 25: Why are we changing this contract?. We have new `PermissionedStakeStable` contract for stake table

types/src/v0/impls/stake_table.rs
imabdulbasit: removed some more generics here https://github.com/EspressoSystems/espresso-sequencer/pull/2383/files

types/src/v0/impls/stake_table.rs
```diff
@@ -20,27 +22,94 @@ use thiserror::Error;
...
+                            .into_iter()
+                            .map(|node_info| StakeTableDelta::add(node_info.into())),
+                    )
+            })
+            .group_by(|delta| delta.node_info.stake_table_key);
+
+        // If the last event for a stakers is `Added` the staker is currently
+        // staking, if the last event is removed or (or the staker is not present)
+        // they are not staking.
+        let currently_staking = changes_per_node
```
imabdulbasit, line 75: can a node be added and then removed or vice versa?

imabdulbasit: so all the deltas of a node will have removed events first and then added? What happens if it is added and then removed?

sveitser: Yes, the removed events are processed first.

So you can remove a node and in the same transaction add it again (to update the keys or DA status of that node).

Adding and removing is possible but not inside the same Ethereum transaction because the removals are processed first.

imabdulbasit: so `Vec<StakersUpdatedFilter>` is all the events from the start till the current block right? so we may do that in different transactions but always process the removed one first?

contracts/src/StakeTable.sol
tbro: The branches are just out of sync.

types/src/v0/impls/stake_table.rs
sveitser: A transaction sent to the permissioned stake table contract will emit one event which will give us one `StakersUpdatedFilter`. That type has two Vecs of NodeInfo for removing and adding nodes respectively.

When doing the query to ethers one can set the "from block" and the "to block" to fetch a subset of events. Currenly here we would assume that we set "from block" to zero but "to block" we need to set correctly. So that we exclude changes during the current epoch until the next epoch.

types/src/v0/impls/stake_table.rs
```diff
@@ -0,0 +1,492 @@
...
+    str::FromStr,
+};
+use thiserror::Error;
+use url::Url;
+
+type Epoch = <SeqTypes as NodeType>::Epoch;
+
+// NewTypes for two types of stake tables to avoid confusion
+#[derive(Clone, Debug, From, Into)]
+pub struct DAStakeTable(pub Vec<StakeTableEntry<PubKey>>);
```
tbro, line 28: link

ss-es (dismissed):
only had a quick look at a few things, seems fine to me overall

I think the compat test changes should be reverted though

sequencer/src/message_compat_tests.rs
```diff
@@ -83,20 +93,24 @@ async fn test_message_compat<Ver: StaticVersionType>(_ver: Ver) {
     )
     .await
     .unwrap();
-    let view_sync_pre_commit_data = ViewSyncPreCommitData {
+    let view_sync_pre_commit_data = ViewSyncPreCommitData2 {
```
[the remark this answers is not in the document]
tbro: reverted (13a3260416da67914fed45fd47f74c462b452317) and made these changes in 8f2d3a426b38ecd80591dcc7429a5b959dc7f999. However there are caveats. See above comment.

data/v1/messages.json
```diff
@@ -107,8 +108,9 @@
     "kind": {
       "Consensus": {
         "General": {
-          "ViewSyncPreCommitVote": {
+          "ViewSyncPreCommitVote2": {
             "data": {
+              "epoch": 0,
```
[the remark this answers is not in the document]
tbro: Most of these can be removed, but `epoch` is now required on `DaProposal`. See: https://github.com/EspressoSystems/HotShot/blob/main/crates/types/src/data.rs#L142-L151

Same for  `UpgradeProposal` I believe.

ss-es (approved):

types/src/v0/mod.rs
```diff
@@ -182,6 +181,9 @@ pub type FeeVersion = StaticVersion<0, 2>;
 pub type MarketplaceVersion = StaticVersion<0, 99>;
 pub type EpochVersion = StaticVersion<0, 100>;
 
+// TODO: >>>>>???
+pub const EPOCH_HEIGHT: u64 = 10000;
```
imabdulbasit, line 185: I think this can be removed now?

tbro: Probably. For some reason compile times have shot up, and I its causing pains testing small changes like this. So I'm going to merge and we can remove this in a Followup.

Issue #2304, opened 2024-11-19
Stake Table: Registration


alysiahuggins:
Create a registration system for validators with a fixed stake amount.

alysiahuggins:
Current plan
- use the existing staketable and write tests for the registration function
- remove epoch logic as it will be handled in the future
- ensure tests still pass
- write tests for the registration function as per the modified spec (e.g. validator keys, fixed stake
- override/replace the registration function as per the modified spec
- add tests and implementation for the validator threshold and registration queues

alysiahuggins:
sub tasks closed

PR #2373, opened 2024-12-06
2370 staketable update validator identification keys

### LINKED ISSUE #2370: StakeTable: Update Validator identification keys
subtask of #2304 
- Introduce additional staking key as per spec
- add function(s) to update blsVK and schnorrVK

alysiahuggins:
Closes #2370  which is a sub issue of #2304

### This PR:
- validator is identified by their ethereum account (msg.sender), blsVK and schnorrVK 
- validator can update consensus key once they have access to their ethereum account 

### This PR does not:
- consider epochs

### Key places to review:
- Updates to the AbstractStakeTable interface, StakeTable.sol and StakeTable.t.sol

### How to test this PR:
`forge test --match-contract StakeTable`


contracts/src/StakeTable.sol
```diff
@@ -97,18 +110,11 @@ contract StakeTable is AbstractStakeTable {
...
-    function totalStake() external view override returns (uint256, uint256) {
-        return (totalNativeStake, totalRestakedStake);
-    }
-
     /// @notice Look up the balance of `blsVK`
     /// @param blsVK BLS public key controlled by the user.
     /// @return Current balance owned by the user.
-    function lookupStake(BN254.G2Point memory blsVK) external view override returns (uint64) {
+    /// TODO modify this according to the current spec
+    function lookupStake(BN254.G2Point memory blsVK) external view override returns (uint256) {
```
alysiahuggins, line 117: now that a validator is identified by the `stakingAddr`, `blsKey` and `schnorrKey`. Do we want functions like `lookupStake()`, `lookupNode()` and `withdrawFunds()` to use the `stakingAddr` as a parameter? @sveitser @tbro

contracts/src/StakeTable.sol
```diff
@@ -374,60 +397,134 @@ contract StakeTable is AbstractStakeTable {
...
 
         return balance;
     }
 
+    /// @notice Update the consensus keys for a validator
+    /// @dev This function is used to update the consensus keys for a validator
+    /// @dev This function can only be called by the validator itself when it's not in the exit
+    /// queue
+    /// @dev The validator will need to give up either its old BLS key and/or old Schnorr key
+    /// @dev The validator will need to provide a BLS signature over the new BLS key
```
philippecamacho, line 449: The signature is over the validator ethereum's address. The comment is wrong, yet the code is right.

contracts/src/interfaces/AbstractStakeTable.sol
```diff
@@ -108,28 +115,32 @@ abstract contract AbstractStakeTable {
...
-        virtual
-        returns (uint256);
+    function withdrawFunds() external virtual returns (uint256);
+
+    /// @notice Update the consensus keys for a validator
+    /// @dev This function is used to update the consensus keys for a validator
+    /// @dev This function can only be called by the validator itself when it's not in the exit
+    /// queue
+    /// @dev The validator will need to give up either its old BLS key and/or old Schnorr key
+    /// @dev The validator will need to provide a BLS signature over the new BLS key
```
philippecamacho, line 137: See comment above. The message signed is the address of the validator.

contracts/test/StakeTable.t.sol
```diff
@@ -85,9 +83,10 @@ contract StakeTable_register_Test is Test {
     function testFuzz_RevertWhen_InvalidBLSSig(uint256 scalar) external {
         uint64 depositAmount = 10 ether;
         uint64 validUntilEpoch = 5;
+        string memory seed = "123";
 
         (BN254.G2Point memory blsVK, EdOnBN254.EdOnBN254Point memory schnorrVK,) =
-            genClientWallet(exampleTokenCreator);
+            genClientWallet(exampleTokenCreator, seed);
```
philippecamacho, line 89: What is the purpose of this change?

philippecamacho: Got it.

contracts/test/StakeTable.t.sol
```diff
@@ -242,21 +245,352 @@ contract StakeTable_register_Test is Test {
...
+        assertEq(node.balance, depositAmount); //same balance
+        assertEq(node.account, exampleTokenCreator); //same account
+
+        vm.stopPrank();
+    }
+
+    function test_lookupNodeAndLookupStake_fails() public {
+        address randomUser = makeAddr("randomUser");
+
+        // lookup the stake for an address that is not registered and verify the data is empty
```
philippecamacho, line 552: The comment should refer to the stake amount not the the data.

contracts/test/StakeTable.t.sol
```diff
@@ -242,21 +245,352 @@ contract StakeTable_register_Test is Test {
...
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0)
+                )
+            )
+        );
+        assertTrue(EdOnBN254.isEqual(node.schnorrVK, EdOnBN254.EdOnBN254Point(0, 0)));
+
+        // look up the stake for the zero address and verify the data is empty
```
philippecamacho, line 574: Same as above. The comment should refer to the stake amount not the the data.

contracts/src/StakeTable.sol
```diff
@@ -338,23 +358,26 @@ contract StakeTable is AbstractStakeTable {
...
     ///
     /// @dev TODO modify this according to the current spec
-    /// @param blsVK The BLS verification key to exit
-    function requestExit(BN254.G2Point memory blsVK) external override {
-        bytes32 key = _hashBlsKey(blsVK);
-        Node memory node = nodes[key];
+    function requestExit() external override {
+        Node memory node = nodes[msg.sender];
+
+        // if the node is not registered, revert
```
philippecamacho, line 377:
```suggestion
        // TODO test this behaviour when re-implementing the logic for handling epochs
        // if the node is not registered, revert
```

contracts/src/interfaces/AbstractStakeTable.sol
```diff
@@ -24,20 +24,24 @@ abstract contract AbstractStakeTable {
     uint256 public totalVotingStake;
 
     /// @notice Signals a registration of a BLS public key.
-    /// @param blsVKhash hash of the BLS public key that is registered.
+    /// @param account the address of the validator
     /// @param registerEpoch epoch when the registration becomes effective.
```
tbro, line 28: This is HotShot epoch, correct?

contracts/src/StakeTable.sol
```diff
@@ -374,60 +397,134 @@ contract StakeTable is AbstractStakeTable {
...
+                        )
+                    ) && EdOnBN254.isEqual(newSchnorrVK, EdOnBN254.EdOnBN254Point(0, 0))
+                )
+        ) {
+            revert NoKeyChange();
+        }
+
+        // Update the node's schnorr key once it's not the same as the old one and it's nonzero
+        if (
+            !EdOnBN254.isEqual(newSchnorrVK, node.schnorrVK)
```
sveitser, line 494: For the `EdOnBn254` and `BN254` types, can we use `using ... for ...` to make the code a bit more readable?

contracts/src/StakeTable.sol
```diff
@@ -299,28 +319,28 @@ contract StakeTable is AbstractStakeTable {
         // Create an entry for the node.
         node.account = msg.sender;
         node.balance = fixedStakeAmount;
+        node.blsVK = blsVK;
```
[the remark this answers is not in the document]
sveitser: I guess we don't need to check the blsVk because we verify a signature but what about the schnorrVk?

contracts/src/StakeTable.sol
```diff
@@ -374,60 +397,134 @@ contract StakeTable is AbstractStakeTable {
...
+        }
+
+        // The staker does not provide a key change
+        if (
+            (
+                _isEqualBlsKey(newBlsVK, node.blsVK)
+                    && EdOnBN254.isEqual(newSchnorrVK, node.schnorrVK)
+            )
+                || (
+                    _isEqualBlsKey(
```
sveitser, line 478: Since we verify the BLS sig, is it necessary to check that the key is non-zero?

sveitser: Ah I see it's allowed to be zero if only the schnorr key is updated.

contracts/src/StakeTable.sol
```diff
@@ -374,60 +397,134 @@ contract StakeTable is AbstractStakeTable {
...
+    /// @dev This function is used to update the consensus keys for a validator
+    /// @dev This function can only be called by the validator itself when it's not in the exit
+    /// queue
+    /// @dev The validator will need to give up either its old BLS key and/or old Schnorr key
+    /// @dev The validator will need to provide a BLS signature over the new BLS key
+    /// @param newBlsVK The new BLS verification key
+    /// @param newSchnorrVK The new Schnorr verification key
+    /// @param newBlsSig The BLS signature that the account owns the new BLS key
+    /// TODO: consider emitting the changed keys
+    function updateConsensusKeys(
```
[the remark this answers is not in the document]
alysiahuggins: That's a fair point. From what I understand, the key generator creates both schnorr and bls keys at the same time. However, the validator may only want to change one of those keys. If we ever wanted to wait for an epoch to apply changes, then separating the key updates (in the instance where they want to update both) might be tricky. Perhaps we make it mandatory to update both at the same time. 

Any thoughts on this from the crypto side @philippecamacho and from the validator side @rob-maron ?

contracts/src/interfaces/AbstractStakeTable.sol
alysiahuggins: good q. The epoch side of things haven't been fully implemented yet (issue [here](https://github.com/EspressoSystems/espresso-sequencer/issues/2316)). Ideally it's the hotshot epoch but yet to determine how the contract would be aware of hotshot's epochs

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: yes you're right, there should be checks on the vk(s) here too!

contracts/src/StakeTable.sol
alysiahuggins: oh is my understanding correct that they have to send the bls sig for the new bls key to ensure that they own the bls key?

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
rob-maron: Correct - they are generated at the same time for a new node. From my understanding as a user I'd want to use this function primarily to roll new validator keys without having to unstake and stake (and waiting the time required)

From that perspective it makes sense to do both at once. If possible, perhaps we can treat "no new key change" as just specifying the same key as their old one as opposed to a zero address?

contracts/src/StakeTable.sol
philippecamacho: Imagine the old bls key has been compromised. Using it to authenticate the new one would not be secure in this case. The reason why we want the bls key to be used for signing when registering/updating keys is to avoid [the rogue key attack](https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html). The fact that the bls key belongs to the validator is implicit as the validator calls the contract with its own ethereum account.

philippecamacho: Actually we need to be more careful regarding the message that is signed. I created a new ticket for that: https://github.com/EspressoSystems/espresso-sequencer/issues/2441

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: perfect, thanks Rob, yes that's a good alternative!

contracts/src/StakeTable.sol
alysiahuggins: In this case I'm not using the bls key to validate the owner, just using the fact that they signed the transaction with the same ethereum address related to this account. Instead, I expect them to provide a new blsSig for the new bls key that they're trying to update to.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: https://github.com/EspressoSystems/espresso-sequencer/commit/5f3a988137098b285f0f397495d29541cd568c63

if only one key is a zero key, it's ignored and we just update the supplied key
if both keys are the same as the old one then we revert with NoKeyChange()

From a UX pov, i wonder if we want to revert even if one key is a zero key

contracts/src/StakeTable.sol
philippecamacho: Yes, I agree with this line of code (taking into account that the message should be slightly different as described in #2441).

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
philippecamacho: Like Rob, I would prefer the caller of this function to pass two valid keys even though only one of the key is new.  I see little extra work to make this call explicit and it definitely might avoid troubles (I can imagine horrible things like the zero key is later interpreted as a test key everybody controls...)

> From a UX pov, i wonder if we want to revert even if one key is a zero key

Yes, it would be better to revert.

contracts/src/StakeTable.sol
alysiahuggins: ok got it! thanks!

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: Thanks! https://github.com/EspressoSystems/espresso-sequencer/commit/638ed78b27900d7b354d2b9212c8f8f45077f6f0 any zero point VKs cause a revert

PR #2378, opened 2024-12-09
Speed up demo-native

QuentinI:
Closes #<ISSUE_NUMBER>


sequencer/src/bin/espresso-bridge.rs
```diff
@@ -195,10 +220,16 @@ async fn deposit(opt: Deposit) -> anyhow::Result<()> {
...
+            if let Some(l1_finalized) = header.l1_finalized() {
+                l1_finalized.number
+            } else {
+                continue;
+            }
+        } else {
+            header.l1_head()
         };
-        if l1_finalized.number >= l1_block {
+        if l1_finalized >= l1_block {
```
jbearer, line 232: Does this actually do anything? It makes the bridge tool exit sooner. But those funds still won't be available on Espresso until Espresso has seen the deposit finalized on L1. So we aren't able to start building blocks any earlier. If anything this might increase the chance we start the builder too soon and start building invalid blocks (which will stall consensus) since the builder isn't funded

jbearer (changes requested):

sequencer/src/bin/espresso-bridge.rs
QuentinI: Right, I didn't consider that, will rollback this part.

jbearer (approved):

PR #2379, opened 2024-12-09
Fix deadlock and race condition in proposal fetching

jbearer:
### This PR:
* Change from broadcast channel to multi-consumer channel. This means only one fetcher task will
  receive each proposal to be fetched, which is the actual behavior we want. Before, with broadcast,
  we had multiple fetchers always fetching the same proposal, which is why we saw race conditions
  causing database serialization errors. It should now be possible to reenable multiple workers.
* Use an unbounded channel. This prevents a deadlock where a consumer sends back into the channel
  (e.g. to recursively fetch the parent of the proposal it had just fetched), but the channel is
  full, blocking the consumer, the very task responsible for clearing the blockage.
* Add metrics for monitoring proposal fetcher
* Move proposal fetcher to its own file, as its accumulating more and more private types and helper functions


rob-maron (approved):
Looks good to me

PR #2387, opened 2024-12-10
[Audit fix] Cargo lock to v3

rob-maron:


sveitser (approved):

sveitser:
Cargo audit should finally be fixed with the next hotshot update that will upgade `libp2p` which will update `idna` that has the current critical vulnerability IIUC.

PR #2390, opened 2024-12-11
Fetch Memberships from l1

imabdulbasit:
Builds on #2361 reworks many things for realistic use in our current code-base.


types/src/v0/impls/stake_table.rs
```diff
@@ -257,36 +304,39 @@ impl Membership<SeqTypes> for StaticCommittee {
...
             _epoch_size: 12, // TODO get the real number from config (I think)
-            _provider: L1Client::http(Url::from_str("http:://ab.b").unwrap()),
+            l1_client: L1Client::http(Url::from_str("http:://ab.b").unwrap()),
             _contract_address: None,
-        }
+        };
+
+        let committee = Arc::new(RwLock::new(committee));
+
+        let _handle = task::spawn(StaticCommittee::update_loop(committee.clone()));
```
tbro, line 321: This is never spawned, since `new` is never called. See comment at the fn signature.

tbro: Also I think the handle will get dropped and close, so we won't get any updates. We should hold on to the handle.

types/src/v0/mod.rs
```diff
@@ -140,7 +141,7 @@ impl NodeType for SeqTypes {
     type Transaction = Transaction;
     type InstanceState = NodeState;
     type ValidatedState = ValidatedState;
-    type Membership = StaticCommittee;
+    type Membership = Arc<RwLock<StaticCommittee>>;
```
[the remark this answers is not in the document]
imabdulbasit: because we are updating state in the update_loop() spawned. Not sure if we can get away with removing RwLock

types/src/v0/impls/stake_table.rs
imabdulbasit: Yeah, I just wanted to push the rough draft of what I plan to do. Also, I don't think dropping the handle will cancel the task. But yeah, we should hold on to the handle

types/src/v0/impls/stake_table.rs
```diff
@@ -216,17 +223,57 @@ impl StaticCommittee {
...
+                    } = event
+                    else {
+                        continue;
+                    };
+
+                    loop {
+                        {
+                            let mut committee_lock = committee.write().await;
+                            if let Err(err) =
+                                committee_lock.update_stake_table(l1_block_number).await
```
sveitser, line 253: If we do need a lock I think it will be a lot safer to take it out only when we change the in memory stake table but not while we fetch it from the RPC.

types/src/v0/impls/stake_table.rs
```diff
@@ -154,17 +152,19 @@ impl StaticCommittee {
     /// to be called before calling `self.stake()` so that
     /// `Self.stake_table` only needs to be updated once in a given
     /// life-cycle but may be read from many times.
-    async fn _update_stake_table(&mut self, l1_block_height: u64) {
+    async fn update_stake_table(&mut self, l1_block_height: u64) -> anyhow::Result<()> {
         let updates: StakeTables = self
-            ._provider
+            .l1_client
             .get_stake_table(l1_block_height, self._contract_address.unwrap())
```
sveitser, line 158: In order to not require a lock during network requests we should probably not combine the network request and the pure function to compute the stake table from the events here into a single function

https://github.com/EspressoSystems/espresso-sequencer/blob/816f8194ff3ca745e8bdd8cf8a0ffcfd3cc42dd3/types/src/v0/impls/l1.rs#L796-L803

types/src/v0/impls/stake_table.rs
```diff
@@ -216,17 +223,57 @@ impl StaticCommittee {
...
+        let retry_delay = l1_client.retry_delay;
+
+        let span = tracing::warn_span!("updating committee stake tables");
+        async move {
+            loop {
+                let mut events = l1_client.receiver.activate_cloned();
+                while let Some(event) = events.next().await {
+                    let L1Event::NewEpoch {
+                        epoch,
+                        l1_block_number,
```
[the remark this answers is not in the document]
imabdulbasit: for stakers, we would need to make a call to l1 in validate_and_apply_header()?

types/src/v0/impls/stake_table.rs
tbro: https://docs.rs/async-broadcast/latest/async_broadcast/struct.Receiver.html

imabdulbasit: but we are not dropping the receiver? sorry if I am misunderstanding. Here the handle is dropped but the spawn task will continue to run and that task does not drop the receiver because of infinite loop.

tbro: Oh yea sorry my mistake

types/src/v0/impls/stake_table.rs
```diff
@@ -93,32 +87,16 @@ impl StakeTables {
...
-
-    /// The nodes on the committee and their stake, indexed by public key
-    indexed_stake_table: BTreeMap<PubKey, StakeTableEntry<PubKey>>,
-
-    /// The nodes on the committee and their stake, indexed by public key
-    indexed_da_stake_table: BTreeMap<PubKey, StakeTableEntry<PubKey>>,
+/// Type to describe DA and Stake memberships
+pub struct MembershipCommittee {
+    /// Holds Stake table and da stake
+    state: Arc<RwLock<HashMap<Epoch, StaticCommittee>>>,
```
sveitser, line 93: This looks good to me.

types/src/v0/impls/stake_table.rs
```diff
@@ -147,24 +125,62 @@ impl StakeTableDelta {
...
+/// Holds Stake table and da stake
+#[derive(Clone, Debug)]
+struct StaticCommittee {
+    /// The nodes eligible for leadership.
+    /// NOTE: This is currently a hack because the DA leader needs to be the quorum
+    /// leader but without voting rights.
+    eligible_leaders: Vec<StakeTableEntry<PubKey>>,
+
+    /// TODO: add comment
+    quorum: HashMap<PubKey, StakeTableEntry<PubKey>>,
```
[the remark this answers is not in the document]
tbro: Why does order matter here?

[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2390/commits/d585efa2f7fb5a939ef0098d13d41a2f2f214819

types/src/v0/impls/l1.rs
```diff
@@ -530,6 +562,19 @@ impl L1Client {
         self.state.lock().await.snapshot
     }
 
+    pub fn stake_table(&self, epoch: &EpochNumber) -> StakeTables {
```
tbro, line 565: link

types/src/v0/impls/stake_table.rs
```diff
@@ -264,128 +287,222 @@ impl Membership<SeqTypes> for StaticCommittee {
...
     /// Get the stake table for the current view
-    fn stake_table(&self, _epoch: Epoch) -> Vec<StakeTableEntry<PubKey>> {
-        self.stake_table.clone().into_iter().collect()
+    fn stake_table(&self, epoch: Epoch) -> Vec<StakeTableEntry<PubKey>> {
+        let state = self.state.read_blocking();
+
+        match state.get(&epoch) {
+            Some(st) => st.indexed_stake_table.clone().into_values().collect(),
+            None => {
+                let stake_tables = self.l1_client.stake_table(&epoch);
```
tbro, line 318: I don't think we are waiting are doing any long waits here, but as best practice it may be nice to release the lock at the beginning of the `None` branch. Is this possible?

types/src/v0/impls/stake_table.rs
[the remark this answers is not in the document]
sveitser: We found out that all hotshot nodes need to always compute the same stake table. Since the order of HashMap is not preserved for clones, etc. that won't work.

types/src/v0/impls/stake_table.rs
tbro: I think 79c1617cf4c96f7440b8aa419dd035e1ac21666d will do this. Depending on if NLL does what I think it does.

types/src/v0/impls/state.rs
```diff
@@ -694,6 +695,38 @@ fn validate_builder_fee(
...
+) -> anyhow::Result<()> {
+    let Some(contract_addr) = chain_config.stake_table_contract else {
+        bail!("stake table contract not found");
+    };
+
+    let prev_epoch = epoch_from_block_number(header.block_number().saturating_sub(1), EPOCH_HEIGHT);
+    let epoch = epoch_from_block_number(header.block_number(), EPOCH_HEIGHT);
+
+    if epoch <= prev_epoch {
+        return Ok(());
```
tbro, line 711: Should this be an error... Or, if we don't want this to propagate up to `BlockError` we should at least have an `info!` or `warn!`.

tbro: 5685090edc97c99366c45b6fd45b762dc82d5e77

types/src/v0/v0_3/stake_table.rs
```diff
@@ -10,3 +10,17 @@ pub struct PermissionedStakeTableEntry(NodeInfoJf);
...
+#[derive(Clone, Debug, From, Into)]
+/// NewType to disambiguate DA Membership
+pub struct DAMembers(pub Vec<StakeTableEntry<PubKey>>);
+
+#[derive(Clone, Debug, From, Into)]
+/// NewType to disambiguate StakeTable
+pub struct StakeTable(pub Vec<StakeTableEntry<PubKey>>);
+
+#[derive(Clone, Debug)]
+pub struct StakeTables {
```
tbro, line 23: I'm thinking we still have some unneeded complexity here to to maintaining to separate vectors..

imabdulbasit: I think its fine to have one vec. The only downside would be to filter the entries for da_stake_table() which is fine IMO

tests/smoke.rs
```diff
@@ -6,7 +6,7 @@ use std::time::Instant;
 /// We allow for no change in state across this many consecutive iterations.
 const MAX_STATE_NOT_INCREMENTING: u8 = 1;
 /// We allow for no new transactions across this many consecutive iterations.
-const MAX_TXNS_NOT_INCREMENTING: u8 = 3;
+const MAX_TXNS_NOT_INCREMENTING: u8 = 5;
```
tbro, line 9: Any idea why transactions are now incrementing at a slower rate?

.env
```diff
@@ -58,7 +58,7 @@ ESPRESSO_BUILDER_ETH_ACCOUNT_INDEX=8
 ESPRESSO_DEPLOYER_ACCOUNT_INDEX=9
 
 # Contracts
-ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0xf7cd8fa9b94db2aa972023b379c7f72c65e4de9d
+ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0xed1db453c3156ff3155a97ad217b3087d5dc5f6e
```
tbro, line 61: Why did this change?

data/genesis/demo-epoch.toml
```diff
@@ -0,0 +1,33 @@
...
+fee_contract = '0xa15bb66138824a1c7167f5e85b957d04dd34e468'
+
+[header]
+timestamp = "1970-01-01T00:00:00Z"
+
+[l1_finalized]
+number = 0
+
+[[upgrade]]
+version = "0.99"
```
sveitser, line 21: @imabdulbasit Can you remind me why this is 0.99 and not 0.3?

imabdulbasit: fixed this locally but this should be v3

.env
imabdulbasit: so we are deploying contracts locally for demos where the sequence of transactions matter to get same addresses. This sequence has changed because of stake table contract deployment

tests/smoke.rs
imabdulbasit: Not sure but I don't think that it is because of our changes. We can try this smoke test for main multiple times

Issue #2316, opened 2024-11-20
Stake Table: Epochs


alysiahuggins:
Ensure epochs are followed correctly for 
- exits
- registrations
- stake deposits
- consensus key updates
- rewards withdrawal

alysiahuggins:
subtasks completed

PR #2394, opened 2024-12-12
2372 stake table add tests to the withdrawfunds function

### LINKED ISSUE #2372: Stake Table: Add tests to the withdrawFunds function
See more #2365

alysiahuggins:
Closes #2372 

### This PR:
- add withdrawFunds tests, 
- removed epoch logic until from withdrawFunds function until epochs are implemented, 
- removed unnecessary checks from the withdrawFunds function

### This PR does not:
- test exit functionality as it depends on epochs #2316 

### Key places to review:
- `withdrawFunds()` function and its related tests

 ### How to test this PR:  
`forge test --match-contract StakeTable`


contracts/test/StakeTable.t.sol
```diff
@@ -242,21 +245,420 @@ contract StakeTable_register_Test is Test {
...
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0)
+                )
+            )
+        );
+        assertTrue(EdOnBN254.isEqual(node.schnorrVK, EdOnBN254.EdOnBN254Point(0, 0)));
+    }
+
+    function test_WithdrawFunds_succeeds() public {
```
philippecamacho, line 597: I would add another test where a user tries to withdraw twice. The first time it works, and then it reverts. I think it is ok to add this case in this same function.

contracts/src/StakeTable.sol
```diff
@@ -410,24 +424,105 @@ contract StakeTable is AbstractStakeTable {
...
-        }
+        // // Verify that the exit escrow period is over.
+        // if (currentEpoch() < node.exitEpoch + exitEscrowPeriod(node)) {
+        //     revert PrematureWithdrawal();
+        // }
+        totalStake -= balance;
 
         // Delete the node from the stake table.
-        delete nodes[key];
+        delete nodes[msg.sender];
```
philippecamacho, line 434: Not something to change in this PR, but I think it is not right to unregister a user after a withdrawal. I would prefer to add a function for depositing more funds for an already existing user.  Let us discuss, especially in the context of stake delegation.

contracts/test/StakeTable.t.sol
alysiahuggins: good one to add, done here https://github.com/EspressoSystems/espresso-sequencer/commit/f9bfe6dc4d821070d77aea3aecd7337b95f107f5

contracts/src/StakeTable.sol
alysiahuggins: i agree with you

PR #2395, opened 2024-12-12
Update the PATH environment variable to be able to run solidity tests…

philippecamacho:
I made this PR because I was not able to run the solidity tests using nix. The problem was that the diff-test executable was not in the PATH environment variable.

### This PR:
Update the `PATH` environment variable in `flake.nix` so that the `diff-test` executable can be found.

### Key places to review:
`flake.nix` file.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -233,7 +233,7 @@
             export CARGO_HOME=$HOME/.cargo-nix
 
             # Add rust binaries to PATH for native demo
-            export PATH="$PWD/$CARGO_TARGET_DIR/debug:$PATH"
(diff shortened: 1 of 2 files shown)


alysiahuggins (approved):

sveitser:
I think we should avoid this because it leads to cases where one may compile with "release" expecting to run these binaries but due to PATH settings one will still run the debug binaries. 

I think we don't need to compile diff test with release or is it too slow? We now have some optimizations for the "test" profile too and I think it should be fine?

sveitser (changes requested):
See other comment

sveitser:
I think we can change the just file recipes to build the diff test binary to `cargo build --profile test --bin diff-test`

sveitser (approved):

PR #2410, opened 2024-12-17
Add rust binary to update permissioned stake table

### LINKED ISSUE #2408: Rust binary to interact with the permissioned stake table.
Read a toml file containing an upgrade and apply it to the contract. This will be used for testing, for a real release we would need a script that uses the multisig.

sveitser:
Close #2408 

Compiles, but there's no test yet.


sveitser:
I think the binary might be useful for testing: for example we want to carry out a change to the stake table in our testnet.

imabdulbasit:
should we merge this PR into ab/st?

PR #2421, opened 2024-12-20
integration test for epoch

imabdulbasit:


tests/common/mod.rs
```diff
@@ -30,6 +31,55 @@ pub struct TestConfig {
     pub initial_txns: u64,
 }
 
+#[derive(Debug, Clone, Parser)]
+pub struct StakeTableUpdateOpts {
```
sveitser, line 35: Don't quite understand why we use Clap for this type.

PR #2422, opened 2024-12-20
store next epoch qc in persistence

imabdulbasit:
Hotshot now has a new function update_next_epoch_high_qc2() in its storage trait to update the next epoch qc in persistence. This PR adds storage for this quorum certificate and also loads it when loading the consensus data,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4249,7 +4249,7 @@ dependencies = [
 [[package]]
 name = "hotshot-query-service"
 version = "0.1.76"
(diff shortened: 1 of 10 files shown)


sveitser (approved):
Changes look good to me. I guess there's some more work needed to make the CI happy. Feel free to merge into [ma/hotshot-0.5.83](https://github.com/EspressoSystems/espresso-sequencer/tree/ma/hotshot-0.5.83) and fix there though.

PR #2423, opened 2024-12-22
Add LightClientArbitrum

Sneh1999:
Replacing block.number in arbitrum because it returns the block number of eth instead of arbitrum 
Discussion [here](https://espresso.zulipchat.com/#narrow/channel/409868-eng-integrations-arbitrum/topic/Prover.20Bug.20on.20Arb.20sepolia)


contracts/src/LightClientArbitrum.sol
```diff
@@ -0,0 +1,15 @@
...
+pragma solidity ^0.8.0;
+
+import { LightClient } from "./LightClient.sol";
+
+interface ArbSys {
+    function arbBlockNumber() external view returns (uint256);
+}
+
+contract LightClientArbitrum is LightClient {
+    function currentBlockNumber() public view virtual override returns (uint256) {
```
alysiahuggins, line 12: lovely! would it be possible to add some tests for this function?
Examples:
1.  show that block.number!=currentBlockNumber() in LightClientArbitrum 
2. currentBlockNumber() == Arbitrum's block number in LightClientArbitrum
3. calls to LightClient currentBlock is the L1 blocknumber

contracts/src/LightClient.sol
```diff
@@ -160,6 +160,11 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         _initializeState(_genesis, _genesisStakeTableState, _stateHistoryRetentionPeriod);
     }
 
+    /// @notice returns the current block number
+    function currentBlockNumber() public view virtual returns (uint256) {
```
alysiahuggins, line 164: 👍

contracts/src/LightClientArbitrum.sol
```diff
@@ -0,0 +1,15 @@
...
+        return ArbSys(address(uint160(100))).arbBlockNumber();
```
alysiahuggins, line 13: just highlighting that we're depending on the fact that the ArbSys address for Arbitrum is always the same, which it should be. However, if it changes we can upgrade the contract. Another option is to make the ArbSys address configurable but that shouldn't be necessary. Unless we're already aware that it's possible/likely for arbitrum (or orbit chains) to change their precompile addresses.

alysiahuggins:
> additionally, we'll need to create another issue to deal with adding upgradeable deployment scripts for LightClientArbitrum.sol and upgrading LightClient.sol. In that issue, we should consider versioning as mentioned here https://github.com/EspressoSystems/espresso-sequencer/blob/main/doc/smart-contract-upgrades.md

https://github.com/EspressoSystems/espresso-sequencer/issues/2436

contracts/src/LightClientArbitrum.sol
Sneh1999: Sounds good, will add these tests!

contracts/test/LightClientArbitrum.t.sol
```diff
@@ -0,0 +1,29 @@
...
+}
+
+contract LightClientArbitrumTest is Test {
+    LightClientArbitrum public lc;
+    MockArbSys mockArbsys;
+
+    function setUp() public {
+        vm.createSelectFork("https://arb1.arbitrum.io/rpc");
+        mockArbsys = new MockArbSys();
+        vm.etch(address(100), address(mockArbsys).code); // Replace address(100) with mock
```
Sneh1999, line 20: Note: this is important because foundry doesnt support arbitrum pre-compiles so we have to mock them in tests

contracts/src/LightClientArbitrum.sol
Sneh1999: https://github.com/EspressoSystems/espresso-sequencer/pull/2423/commits/56ab328fe9068131a6edac67658b28d8b9bc48b9

alysiahuggins (approved):
lgtm!

we'd need to add the `LightClientArbitrum` contract to the `just gen-bindings` command but that can be handled in the next github issue https://github.com/EspressoSystems/espresso-sequencer/issues/2436

PR #2433, opened 2025-01-03
Add `payload_size` to `ExplorerTransaction` impl

### LINKED ISSUE #2432: Correct the transaction size that is reported on the Explorer
The Explorer is able to show various pieces of information about the Blocks and Transactions.  However, when inspecting individual transactions it seems to report the size of the entire block rather than the size of just that individual transaction.

A change to the `hotshot-query-service` allows for this size to be specified at the Transaction level:
https://github.com/EspressoSystems/hotshot-query-service/pull/742

By implementing this update, the transaction size will be fixed.

Ayiga:
Closes #2432

### This PR:

Updates the dependency on the `hotshot-query-service` to one that has some error handling fixes for the Block Explorer, and one that has an update to the `ExplorerTransaction` trait.

Updates the implementation of the `ExplorerTransaction` trait to implement the new `payload_size` method.

The updated `hotshot-query-service` dependency should also correct how many transactions are pulled when querying a block with more than the requested limit of transactions.

### This PR does not:

Target a new explicit version tag of the `hotshot-query-service`, but instead targets a subtag:
`v0.1.75-explorer-fixes`.


Ayiga:
The key point of this PR was already addressed #2361 , and all of the other adjustments / improvements have been moved to  #2488 instead.

PR #2434, opened 2025-01-03
Jb/l1 client

### LINKED ISSUE #1238: Allow to setup multiple RPCs in the .env for fallback

### LINKED ISSUE #2360: L1 client improvements
- [ ] Always use HTTP for normal requests, use WS for subscriptions only, as [recommended](https://docs.alchemy.com/reference/best-practices-for-using-websockets-in-web3#reasons-to-use-https-instead-of-websockets-for-json-rpc-node-requests). This means that instead of having an enum with HTTP _or_ WS connection, our RPC client will be a struct with HTTP _and_ optional WS connection
- [ ] Handle 429 rate limit HTTP errors: sleep for a short time
- [ ] Failover: allow for multiple URIs. If a provider gives us 2 request errors (other than 429 errors) in a configurable time window _or_ if a WS stream ends/disconnects once, failover to the next provider in the list

jbearer:
Closes #2360
Closes #1238 

### This PR:
* Uses HTTP for non-subscription RPC requests, even if a WebSockets provider is also given
* Tracks RPC request failures and fails over to another provider if possible when the current provider is failing too often
* Handles rate limit (429) errors from the provider


types/src/v0/v0_1/l1.rs
```diff
@@ -88,6 +91,49 @@ pub struct L1ClientOptions {
...
+        env = "ESPRESSO_SEQUENCER_L1_RATE_LIMIT_DELAY",
+        value_parser = parse_duration,
+    )]
+    pub l1_rate_limit_delay: Option<Duration>,
+
+    /// Separate provider to use for subscription feeds.
+    ///
+    /// Typically this would be a WebSockets endpoint while the main provider uses HTTP.
+    #[clap(long, env = "ESPRESSO_SEQUENCER_L1_WS_PROVIDER", value_delimiter = ',')]
+    pub l1_ws_provider: Option<Vec<Url>>,
```
rob-maron, line 135: Why the separate WS-focused provider? Is it to compel users to use WS for subscriptions?

types/Cargo.toml
```diff
@@ -11,6 +11,7 @@ testing = []
 anyhow = { workspace = true }
 ark-serialize = { workspace = true }
 async-broadcast = { workspace = true }
+async-lock = { workspace = true }
```
rob-maron, line 14: Incredibly minor but you could use the tokio `RwLock` instead of this one. I think this one is in the workspace due to the fact we need it for `tide-disco` compat

rob-maron (approved):
Looks good, one minor q

types/src/v0/v0_1/l1.rs
jbearer: No it's to give them the option of using WS for subscriptions, but fall back to using the HTTP provider if they don't give a WS provider. Since WS support is not uniform across providers

PR #2443, opened 2025-01-09
fix healthcheck for builder service in integration test

imabdulbasit:


Ayiga (approved):
All of my questions have been answered via zulip.

PR #2449, opened 2025-01-13
2436 deployment scripts for lightclientarbitrum upgrade scripts lightclient

### LINKED ISSUE #2436: Deployment scripts for LightClientArbitrum & Upgrade scripts LightClient
block.number returns the block number of the parent chain for contracts deployed on Arbitrum.
As such the lightClient has been updated to introduce a new function and an additional contract was for Arbitrum chains was added here. https://github.com/EspressoSystems/espresso-sequencer/pull/2423

Tasks
- [] add deployment script for LightClientArbitrum
- [] upgrade scripts for LightClient

https://github.com/EspressoSystems/espresso-sequencer/blob/main/doc/smart-contract-upgrades.md

alysiahuggins:
Closes #2436 

### This PR:
- creates foundry forge scripts for deploying the LightClientArbitrum contract
- makes changes to the `initialize()` method on `LightClient.sol`, it no longer updates the state history upon initialization as precompiles can't be called when deploying to networks

### This PR does not:
- update the rust scripts for deployment

### Key places to review:
- README.md
- change to LightClient.sol


contracts/script/LightClientArb.s.sol
```diff
@@ -0,0 +1,85 @@
...
+        string[] memory cmds = new string[](3);
+        cmds[0] = "diff-test";
+        cmds[1] = "mock-genesis";
+        cmds[2] = vm.toString(uint256(numInitValidators));
+
+        bytes memory result = vm.ffi(cmds);
+        (LC.LightClientState memory state, LC.StakeTableState memory stakeState) =
+            abi.decode(result, (LC.LightClientState, LC.StakeTableState));
+
+        proxyAddress = Upgrades.deployUUPSProxy(
```
Sneh1999, line 47: why is this a proxy address? why cant we just deploy an implementation contract and then upgrade our existing proxy?

contracts/src/LightClientArbitrum.sol
```diff
@@ -9,6 +9,11 @@ interface ArbSys {
 }
 
 contract LightClientArbitrum is LightClient {
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
```
Sneh1999, line 13: why do we need to do this? it should inherit the constructor from LightClient

alysiahuggins: that's right, this code lingered here because i originally planned to override the initialize function and wanted to be explicit. will remove

contracts/script/LightClientArb.s.sol
alysiahuggins: opted for a deploy instead of an upgrade whilst dealing with the precompile issue. once this quick review is fine, i can continue with the upgrade steps

Sneh1999 (approved):
LGTM!

contracts/script/LightClientArb.s.sol
alysiahuggins: created a ticket for the upgrade script as the safe transaction service is not supported in arbitrum sepolia. this service is needed for proposing transactions to the safe multisig wallet, created another ticket https://github.com/EspressoSystems/espresso-sequencer/issues/2457

PR #2478, opened 2025-01-20
types migration

imabdulbasit:
This PR migrates existing types, such as leaf1 and VID, to the new V2 types. There are separate migrations for both query service storage and consensus storage types.

To ensure data safety, the migration creates new files for filesystem storage and new tables for SQL storage. This approach allows us to retain the old data as a fallback in case of any issues. The old data can be removed in a future release once this migration has been successfully deployed on decaf/mainnet.

Query Service Storage:

    Migrates leaf1 → leaf2 and  vid → wrapped vid (now an enum)

Consensus Storage:
```
    -  leaves
    -  DA proposals
    -  VID shares
    -  undecided state
    -  quorum proposals
    -  quorum certificates
```

followup work:
- test that checks that the migrated data is expected


types/src/v0/traits.rs
```diff
@@ -686,11 +684,39 @@ pub trait SequencerPersistence: Sized + Send + Sync + Clone + 'static {
...
-    ) -> anyhow::Result<()>;
+    ) -> anyhow::Result<()> {
+        tracing::info!("migrating consensus data");
+
+        self.migrate_anchor_leaf().await?;
+        self.migrate_da_proposals().await?;
+        self.migrate_vid_shares().await?;
+        self.migrate_undecided_state().await?;
+        self.migrate_quorum_proposals().await?;
+        self.migrate_quorum_certificates().await
```
bfish713, line 699: Can we gate this by versions(and the other parts we start storing the new types) so we can merge this and not worry about deploying it

types/src/v0/traits.rs
```diff
@@ -686,11 +684,39 @@ pub trait SequencerPersistence: Sized + Send + Sync + Clone + 'static {
...
+    }
+    async fn migrate_undecided_state(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
+    async fn migrate_quorum_proposals(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
+    async fn migrate_quorum_certificates(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
```
ss-es, line 719: I think all of these should be fully implemented in the same PR

imabdulbasit: this is default implementation. I have it implemented for SQL. I can remove this default implmentation to avoid confusion

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -1495,4 +1620,123 @@ mod test {
...
+
+        storage
+            .migrate_types::<MockTypes>()
+            .await
+            .expect("failed to migrate");
+        let mut tx = storage.read().await.unwrap();
+        let (count,) = query_as::<(i64,)>("SELECT COUNT(*) from leaf2")
+            .fetch_one(tx.as_mut())
+            .await
+            .unwrap();
```
sveitser, line 1738: Instead of checking the number of rows it would be better to check that the data we migrated is as expected.

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -796,6 +802,114 @@ impl VersionedDataSource for SqlStorage {
...
+    block_hash: String,
+    leaf2: Value,
+    qc2: Value,
+}
+impl SqlStorage {
+    pub async fn migrate_types<Types: NodeType>(&self) -> anyhow::Result<()> {
+        let mut offset = 0;
+        let limit = 10000;
+
+        loop {
```
sveitser, line 817: Lets' add some logs

1. Inform about start of migration
2. Inform about progress, unless it
3. Inform about completion.

sequencer/api/migrations/postgres/V501__epoch_tables.sql
```diff
@@ -0,0 +1,54 @@
...
+
+CREATE TABLE vid_share2 (
+    view BIGINT PRIMARY KEY,
+    payload_hash VARCHAR,
+    data BYTEA
+);
+
+
+CREATE TABLE undecided_state2 (
+    -- The ID is always set to 0. Setting it explicitly allows us to enforce with every insert or
```
sveitser, line 22: Did you consider adding a constraint for this?

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -796,6 +802,114 @@ impl VersionedDataSource for SqlStorage {
...
+                    "SELECT leaf, qc FROM leaf ORDER BY height LIMIT {} OFFSET {}",
+                    limit, offset
+                ))
+                .fetch_all(tx.as_mut())
+                .await?;
+
+            drop(tx);
+
+            if rows.is_empty() {
+                tracing::info!("no leaf1 rows found");
```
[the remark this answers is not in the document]
imabdulbasit: yeah it should break here like we do in consensus migration

hotshot-query-service/src/data_source/storage/sql.rs
imabdulbasit: in my todo list

hotshot-query-service/src/availability.rs
```diff
@@ -161,9 +166,97 @@ impl Error {
...
+    pub(crate) leaf: Leaf<Types>,
+    pub(crate) qc: QuorumCertificate<Types>,
+}
+
+fn downgrade_leaf<Types: NodeType>(leaf2: Leaf2<Types>) -> Leaf<Types> {
+    // TODO do we still need some check here?
+    // `drb_seed` no longer exists on `Leaf2`
+    // if leaf2.drb_seed != [0; 32] && leaf2.drb_result != [0; 32] {
+    //     panic!("Downgrade of Leaf2 to Leaf will lose DRB information!");
+    // }
```
sveitser, line 181: @ss-es @lukaszrzasik maybe you can comment on this? I think the downgrade is mainly used in the query service now when paths to API v0 are invoked?

On a related note, I think we should **not** implement `From` when the conversion is lossy. This is the second bullet point here: https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from

I think if the conversion is lossy it's better to have a function or method, this then also makes it easier to search the codebase for these lossy conversions.

https://github.com/EspressoSystems/espresso-sequencer/blob/e15b2b8f6a2d8eb542505f36809cfcc07945a1b1/hotshot-query-service/src/availability.rs#L277-L301

cc @imabdulbasit

hotshot-query-service/src/availability.rs
```diff
@@ -161,9 +166,97 @@ impl Error {
...
+    };
+    let mut leaf = Leaf::from_quorum_proposal(&quorum_proposal);
+    if let Some(payload) = leaf2.block_payload() {
+        leaf.fill_block_payload_unchecked(payload);
+    }
+    leaf
+}
+
+impl<Types: NodeType> From<LeafQueryData<Types>> for Leaf1QueryData<Types> {
+    fn from(value: LeafQueryData<Types>) -> Self {
```
sveitser, line 197: I think we should **not** implement `From` when the conversion is lossy. This is the second bullet point here: https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from

I think if the conversion is lossy it's better to have a function or method, this then also makes it easier to search the codebase for these lossy conversions.

cc @imabdulbasit

imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2478/commits/0d272a30b9db0d86399188d83db4c0624b88a070

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -810,6 +816,156 @@ impl VersionedDataSource for SqlStorage {
...
+            }
+
+            offset += limit;
+        }
+
+        let mut tx = self.write().await.map_err(|err| QueryError::Error {
+            message: err.to_string(),
+        })?;
+
+        tracing::info!("query service types migration is completed!");
```
sveitser, line 954: Shouldn't we make this `warn` too?

hotshot-query-service/src/data_source/storage/sql.rs
sveitser: https://github.com/EspressoSystems/espresso-sequencer/issues/2710

hotshot-query-service/src/data_source/storage/sql.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-sequencer/pull/2478/commits/82155b510a1cf1a18531af54fada01fa0130344f

sequencer/src/persistence/fs.rs
```diff
@@ -893,14 +934,322 @@ impl SequencerPersistence for Persistence {
...
+            return Ok(());
+        }
+
+        let decided_leaf2_path = inner.decided_leaf2_path();
+
+        fs::create_dir_all(decided_leaf2_path.clone())
+            .context("failed to create anchor leaf 2  dir")?;
+
+        let decided_leaf_path = inner.decided_leaf_path();
+        if !decided_leaf_path.is_dir() {
```
sveitser, line 986: @imabdulbasit When would this not be a directory?

sequencer/src/persistence/fs.rs
```diff
@@ -893,14 +934,322 @@ impl SequencerPersistence for Persistence {
...
+            return Ok(());
+        }
+
+        tracing::warn!("migrating decided leaves..");
+        for entry in fs::read_dir(decided_leaf_path)? {
```
sveitser, line 991: Do I understand correctly that if the migration is interrupted it will just migrate everything again the next time? I don't think that's a problem, as long as things eventually get migrated and we don't end up in a state where the node can't start up because the migration fails.

sequencer/src/persistence/fs.rs
```diff
@@ -893,14 +934,322 @@ impl SequencerPersistence for Persistence {
...
+            let Ok(view) = file.parse::<u64>() else {
+                continue;
+            };
+
+            let bytes =
+                fs::read(&path).context(format!("reading da proposal {}", path.display()))?;
+            let proposal = bincode::deserialize::<Proposal<SeqTypes, DaProposal<SeqTypes>>>(&bytes)
+                .context(format!("parsing da proposal {}", path.display()))?;
+
+            let file_path = da2_path.join(view.to_string()).with_extension("txt");
```
sveitser, line 1068: I think in these functions we could be a bit more explicit with naming to make it easier to read. As in for example `old_path` `new_path`, `old_file`, `new_file` or something like that, instead of `path`, `file`, `file_path`, `file`.

sequencer/src/persistence/fs.rs
```diff
@@ -893,14 +934,322 @@ impl SequencerPersistence for Persistence {
...
+        fs::create_dir_all(da2_path.clone()).context("failed to create da proposals 2 dir")?;
+
+        let da_dir = inner.da_dir_path();
+        if !da_dir.is_dir() {
+            return Ok(());
+        }
+
+        tracing::warn!("migrating da proposals..");
+
+        for entry in fs::read_dir(da_dir)? {
```
sveitser, line 1052: How long do we expect this migration to take? Should we log something about progress?

sequencer/src/persistence/sql.rs
```diff
@@ -1343,14 +1317,478 @@ impl SequencerPersistence for Persistence {
...
-    async fn migrate_consensus(
-        &self,
-        _migrate_leaf: fn(Leaf) -> Leaf2,
-        _migrate_proposal: fn(
-            Proposal<SeqTypes, QuorumProposal<SeqTypes>>,
-        ) -> Proposal<SeqTypes, QuorumProposal2<SeqTypes>>,
-    ) -> anyhow::Result<()> {
-        // TODO: https://github.com/EspressoSystems/espresso-sequencer/issues/2357
+    async fn migrate_anchor_leaf(&self) -> anyhow::Result<()> {
+        let batch_size: i64 = 1000;
```
sveitser, line 1321: just curious, why are these `i64`?

sequencer/src/persistence/sql.rs
```diff
@@ -1343,14 +1317,478 @@ impl SequencerPersistence for Persistence {
...
+            }
+
+            let mut query_builder: sqlx::QueryBuilder<Db> =
+                sqlx::QueryBuilder::new("INSERT INTO vid_share2 (view, payload_hash, data) ");
+
+            query_builder.push_values(values.into_iter(), |mut b, (view, payload_hash, data)| {
+                b.push_bind(view).push_bind(payload_hash).push_bind(data);
+            });
+
+            offset += batch_size;
```
sveitser, line 1544: I think most of the functions here still don't log any progress. I think it's fine if it's reasonably fast but if it takes more than say 10 seconds or a minute I prefer logging something about the progress.

sequencer/src/persistence/sql.rs
```diff
@@ -1757,52 +2256,55 @@ mod test {
...
 
         // Add an extra quorum proposal so we have a QC pointing back at `leaf`.
         storage
-            .append_quorum_proposal(&next_quorum_proposal)
+            .append_quorum_proposal2(&next_quorum_proposal)
             .await
             .unwrap();
 
         // Fetch it as if we were rebuilding an archive.
+        // TODO (abdul):
```
sveitser, line 2283: @imabdulbasit another TODO comment

types/src/v0/impls/block/full_payload/payload.rs
```diff
@@ -64,7 +64,7 @@ impl Payload {
         self.read_ns_payload(&ns_payload_range)
     }
 
-    pub(crate) fn byte_len(&self) -> PayloadByteLen {
+    pub fn byte_len(&self) -> PayloadByteLen {
```
sveitser, line 67: I wonder why we need to make this `pub`.

@mrain are we using the new API correctly? I had a brief look at `parse_ns_table` and it seems that due to the types used and the return value it may be rather easy to use it wrongly in the code. We have an `NsTable` type so I'm wondering why we now also use `Vec<Range<usize>>` to represent an NsTable.

types/src/v0/impls/header.rs
```diff
@@ -2,18 +2,16 @@ use anyhow::{ensure, Context};
...
+    data::VidCommitment,
     traits::{
         block_contents::{BlockHeader, BuilderFee},
         node_implementation::NodeType,
         signature_key::BuilderSignatureKey,
         BlockPayload, ValidatedState as _,
     },
-    utils::BuilderCommitment,
-    // vid::advz::{ADVZCommon, ADVZScheme},
+    utils::BuilderCommitment, // vid::advz::{ADVZCommon, ADVZScheme},
```
sveitser, line 14: Remove the comment instead?

types/src/v0/traits.rs
```diff
@@ -804,16 +837,30 @@ impl<P: SequencerPersistence> Storage<SeqTypes> for Arc<P> {
...
-            .append_quorum_proposal(&convert_proposal(proposal.clone()))
+            .append_quorum_proposal2(&convert_proposal(proposal.clone()))
             .await
     }
 
+    async fn append_proposal2(
+        &self,
+        proposal: &Proposal<SeqTypes, QuorumProposal2<SeqTypes>>,
+    ) -> anyhow::Result<()> {
+        // TODO: this is a bug in hotshot with makes with_epoch = true
```
sveitser, line 853: Is this comment still valid? It's not clear to me what we need to do. Fix a bug in hotshot? Do some changes when a bug in hotshot is fixed? Or?

sveitser (approved):
LGTM. But some remaining questions. I don't think any of that is critical though. What I guess would be nice if we had some kind of data dump from an existing node somewhere that we can try the migration on. I think the tests you added are good and I think it should work but it's always nice to also test big migrations on a real databases. I know we don't currently have good infrastructure for doing that but I think it's something we should consider building.

sequencer/src/persistence/sql.rs
imabdulbasit: related to vid common field. I'll remove it

sequencer/src/persistence/sql.rs
```diff
@@ -1343,14 +1317,478 @@ impl SequencerPersistence for Persistence {
...
+            });
+
+            offset += batch_size;
+
+            let query = query_builder.build();
+
+            let mut tx = self.db.write().await?;
+            query.execute(tx.as_mut()).await?;
+            tx.commit().await?;
+            if rows.len() < batch_size as usize {
```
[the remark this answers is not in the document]
imabdulbasit: So I added this to avoid an extra query but I guess that doesn't really makes any performance difference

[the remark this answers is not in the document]
imabdulbasit: This statement should be there for all SQL migrations. I think it is missing for undecided_state but that is fine because there can be only one entry for undecided_state

sequencer/src/persistence/fs.rs
imabdulbasit: if I spin up a new node with PoS version

sequencer/src/persistence/fs.rs
imabdulbasit: Yeah, we can keep the offsets after each batch upsert and resume from there too. I didn't think about that but can add it after doing some rewards work

sequencer/src/persistence/sql.rs
imabdulbasit: because postgres BIGINT is of i64 type

sequencer/src/persistence/fs.rs
imabdulbasit: done in https://github.com/EspressoSystems/espresso-sequencer/pull/2478/commits/0ea3d2b4678dfb32f53dc1255ee83eb78e9aab52

sequencer/src/persistence/fs.rs
imabdulbasit: Haven't tested it locally. added more logs https://github.com/EspressoSystems/espresso-sequencer/pull/2478/commits/0ea3d2b4678dfb32f53dc1255ee83eb78e9aab52

sequencer/src/persistence/sql.rs
```diff
@@ -1343,14 +1317,478 @@ impl SequencerPersistence for Persistence {
...
+                b.push_bind(view).push_bind(leaf).push_bind(qc);
+            });
+
+            let query = query_builder.build();
+
+            let mut tx = self.db.write().await?;
+            query.execute(tx.as_mut()).await?;
+            tx.commit().await?;
+
+            tracing::info!("anchor leaf: migrated {} rows", offset);
```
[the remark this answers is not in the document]
imabdulbasit: done in https://github.com/EspressoSystems/espresso-sequencer/pull/2478/commits/0ea3d2b4678dfb32f53dc1255ee83eb78e9aab52

PR #2488, opened 2025-01-20
Address some smoke test issues

### LINKED ISSUE #2486: The `wait_for_service` function should be improved
The `wait_for_service` function defined [here](https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L239) is meant to check that a service is available and ready before resolving and continuing.

However this function may not fully return when a service is actually ready.

## Healthcheck URL construction

The primary way this function is meant to check if a service is ready to go is by inspecting the health check url of the service in question.  The service is assumed to be a service that has a health check URL defined at the `/healthcheck` path.  However, the way the URL is constructed just adds `/healthcheck` to the end of an existing full URL:

https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L242

Example:
If provided the URL `https://example.com/sample`, the resulting URL would be `https://example/sample/healthcheck`, which seems not to be intended.

## Healthcheck is ready behavior

The way a usual HTTP health check URL tends to work is by ensuring that the resulting status code from the hit endpoint returns a valid status code.  However in the current code the status code is not checked or even considered:

https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L243

Instead, we just check that the body can return a textual representation of the response.  I don't believe this textual check is strictly necessary, however, I don't see that it harms anything to have it either.  But in any case, the status code of the response **should** be checked to in order to ensure that the service is actually ready.

### LINKED ISSUE #2487: Replace hard-coded version number checks for smoke tests
In the file `tests/common/mod.rs` there remain some hard-coded versions that are now out of date:
 
- https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L84
- https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L87
- https://github.com/EspressoSystems/espresso-sequencer/blob/018fbc50b268e87828e1f9da50dbd4c242c66672/tests/common/mod.rs#L96

The version 2 should refer to the Fee version of the sequencer, and version 3 referred to the Marketplace version.  However, the marketplace version is no longer version 3, and is now version 99.  Additionally, it is very likely that the Marketplace version will not remain version 99 in the future.

The recommended change is to not reference hard-coded numbers, and instead modify the logic to be based on the software defined constants instead.

Ayiga:
Closes #2486
Closes #2487 

### This PR:

Is a rebase of the original PR #2433, to eliminate the commit history and dependency of attempting to implement the `ExplorerTransaction` changes.

Instead, this PR focuses entirely on adjustments made to the smoke tests that were failing under the previous PR.

This PR aims to address some potential implementation issues and details with the `wait_for_service` function.  Specifically, to fix the URL construction of the healthcheck endpoint, and to add http status code check to the response.

Additionally, we remove the hard-coded version numbers in favor of the programmatically stored values of the versions instead.

### This PR does not:

Modify the other aspects of evaluation.

### DIFF
--- a/justfile
+++ b/justfile
@@ -70,7 +70,10 @@ test-all:
 
 test-integration:
 	@echo 'NOTE that demo-native must be running for this test to succeed.'
-	cargo nextest run --all-features --nocapture --profile integration
+	INTEGRATION_TEST_SEQUENCER_VERSION=2 cargo nextest run --all-features --nocapture --profile integration smoke
+test-integration-mp:
+    @echo 'NOTE that demo-native-mp must be running for this test to succeed.'
+    INTEGRATION_TEST_SEQUENCER_VERSION=99 cargo nextest run --all-features --nocapture --profile integration
 
 clippy:
     @echo 'features: "embedded-db"'
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, Result};
 use client::SequencerClient;
-use espresso_types::{FeeAmount, MarketplaceVersion};
+use espresso_types::{FeeAmount, FeeVersion, MarketplaceVersion};
 use ethers::prelude::*;
 use futures::future::join_all;
 use std::{fmt, str::FromStr, time::Duration};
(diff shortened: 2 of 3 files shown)


zacshowa (approved):
Lgtm!

PR #2504, opened 2025-01-29
max block size env for dev node

imabdulbasit:


Sneh1999 (approved):
LGTM! Thanks @imabdulbasit

PR #2507, opened 2025-01-29
[Libp2p] Finish generic DHT persistence

rob-maron:
## This PR:
Finishes implementing generic DHT persistence now that HotShot is updated

## What to look at:
- @jbearer @imabdulbasit do the migration scripts look okay?


sequencer/src/persistence/sql.rs
```diff
@@ -1346,6 +1349,65 @@ impl SequencerPersistence for Persistence {
...
+    /// # Errors
+    /// - If we fail to serialize the records
+    /// - If we fail to write the serialized records to the DB
+    async fn save(&self, records: Vec<SerializableRecord>) -> anyhow::Result<()> {
+        // Bincode-serialize the records
+        let to_save =
+            bincode::serialize(&records).with_context(|| "failed to serialize records")?;
+
+        // Prepare the statement
+        let stmt = "INSERT OR REPLACE INTO libp2p_dht (id, serialized_records) VALUES (0, $1)";
```
imabdulbasit, line 1365: I don't think postgres supports` INSERT OR REPLACE`. You can use `ON CONFLICT (id) DO UPDATE`
or there is a `upsert() `on transaction which can do it for you

sequencer/src/persistence/sql.rs
```diff
@@ -1346,6 +1349,65 @@ impl SequencerPersistence for Persistence {
...
+    /// - If we fail to read from the DB
+    /// - If we fail to deserialize the records
+    async fn load(&self) -> anyhow::Result<Vec<SerializableRecord>> {
+        // Fetch the results from the DB
+        let Some(result) = self
+            .db
+            .read()
+            .await
+            .with_context(|| "failed to start a DB read transaction")?
+            .fetch_optional("SELECT * FROM libp2p_dht where id = 0")
```
imabdulbasit, line 1393: can use fetch_one() here

sequencer/src/persistence/fs.rs
```diff
@@ -894,6 +901,42 @@ impl SequencerPersistence for Persistence {
...
+    /// # Errors
+    /// - If we fail to serialize the records
+    /// - If we fail to write the serialized records to the file
+    async fn save(&self, records: Vec<SerializableRecord>) -> anyhow::Result<()> {
+        // Bincode-serialize the records
+        let to_save =
+            bincode::serialize(&records).with_context(|| "Failed to serialize records")?;
+
+        // Write the serialized records to the file
+        std::fs::write(self.inner.read().await.libp2p_dht_path(), to_save)
```
imabdulbasit, line 917: This would fail if directory does not exist. We can use fs::create_dir_all() to create it first.
Also, we are using inner.replace() everywhere to wwrite the files

sequencer/src/persistence/sql.rs
```diff
@@ -1389,6 +1392,62 @@ impl SequencerPersistence for Persistence {
...
+    /// # Errors
+    /// - If we fail to serialize the records
+    /// - If we fail to write the serialized records to the DB
+    async fn save(&self, records: Vec<SerializableRecord>) -> anyhow::Result<()> {
+        // Bincode-serialize the records
+        let to_save =
+            bincode::serialize(&records).with_context(|| "failed to serialize records")?;
+
+        // Prepare the statement
+        let stmt = "UPSERT INTO libp2p_dht (id, serialized_records) VALUES (0, $1)";
```
imabdulbasit, line 1408: ah I meant upsert function on transaction. 
something like this [here](https://github.com/EspressoSystems/espresso-sequencer/blob/main/sequencer/src/persistence/sql.rs#L1178)

imabdulbasit (approved):

PR #2509, opened 2025-01-30
Fix light client bug

zacshowa:
### This PR:
Fixes some bugs in the light client that would return invalid snapshots when the snapshot of the height requested was in the last index of `stateHistoryCommitments`

### Key places to review:
`contracts/src/LightClient.sol`

### Things tested:
Consumers of the light client are no longer affected by this bug


zacshowa:
The light client test is still failing locally.

I need to figure out how to add one additional entry to its state to give us a valid target as the "max height" to read from. When we only have one state we can't read it with this change, and we can't read from an index less than zero.

I also want to add a case where we expect a revert when trying to read from the most recent state

jbearer (commented):
Change looks good to me, pending the additional testing you suggested. It would also be good to have @alysiahuggins @alxiong @0xkato look at this

alxiong (commented):
The code comment on getHotshotCommitment is wrong. Please update that as well.
Would be nice to include the reason for this boundary condition  in the code comment. 

Otherwise looks good to me.

contracts/src/LightClient.sol
0xkato, line 1: Like Alex mentioned update comments [here](https://github.com/EspressoSystems/espresso-sequencer/pull/2509/files#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292L416-L418) and [here](https://github.com/EspressoSystems/espresso-sequencer/pull/2509/files#diff-d85261c27bf48fe0ec00e8cc9bb3ca101d708ec68c85968531b2e2cf4f9ed292L435-L436) otherwise it looks good!

contracts/test/LightClient.t.sol
```diff
@@ -1064,10 +1064,11 @@ contract LightClient_HotShotCommUpdatesTest is LightClientCommonTest {
         assertEqBN254(hotShotBlockComm, newState.blockCommRoot);
         assertEq(hotShotBlockHeight, newState.blockHeight);
 
-        // Test for max hotShotBlockHeight stored on contract
+        // Test for max hotShotBlockHeight stored on contract that is available (max
+        // hotshotBlockHeight -1)
         // Get the highest HotShot blockheight recorded
         uint256 numCommitments = lc.getStateHistoryCount();
-        (,, hotShotBlockHeight, hotShotBlockComm) = lc.stateHistoryCommitments(numCommitments - 1);
+        (,, hotShotBlockHeight, hotShotBlockComm) = lc.stateHistoryCommitments(numCommitments - 2);
```
alysiahuggins, line 1071: Hi there, There's an integer underflow here because at this point there's only one state history commitment that would have been set so `numCommitments` = 1. What was the reason for subtracting 2?

alysiahuggins (commented):
lgtm except for the failing test `test_GetHotShotCommitmentValid` where I've left a comment.

contracts/test/LightClient.t.sol
zacshowa: I hadn't fully fixed this test yet. It should be fixed in this commit. We don't actually need to subtract by 2, but now the function semantic is different in that it will return 1 index in the list past what it previously did. Because of this I had to adjust the height passed into the next function.

Sneh1999 (approved):
LGTM!

PR #2510, opened 2025-01-30
enable stake table epochs

### LINKED ISSUE #2505: set configurable blocks per epoch

alysiahuggins:
Closes #2505

### This PR:
- re-adds epoch functionality so that validator registrations and exits can be handled on a epoch basis
- ensures that the currentEpoch is a based on the latest hotshot block height and the `hotShotBlocksPerEpoch`
- the `hotShotBlocksPerEpoch` is set at constructor time
- consolidates `nextRegistraionEpoch()` and `nextExitEpoch()` into one function, `nextAvailableEpoch(...)` since the logic was the same
- adds tests for the `currentEpoch()` and `nextAvailableEpoch(...)` functions

### This PR does not:
- provide a way to update hotshot blocks per epoch as it's assumed that this will not change in the near future, when it does change the function can be updated
- handle integer overflows (e.g. currentEpoch()+1) - this is a discussion to have
- modify epoch logic for functions such as `withdrawFunds()` which depends on the epoch functionality, this will be handled in other github issues.


sveitser:
> provide a way to update hotshot blocks per epoch as it's assumed that this will not change in the near future, when it does change the function can be updated

If the blocks per epoch changes then the computation of current epoch becomes more complicated and we need to store the original value of blocks per epoch and the new value to compute epochs correctly as a function of the hotshot block height. I wonder if we should make the blockPerEpoch an array of `(hotshotBlockNumber, blocksPerEpoch)` to enable this computation.

alysiahuggins:
> > provide a way to update hotshot blocks per epoch as it's assumed that this will not change in the near future, when it does change the function can be updated
> 
> If the blocks per epoch changes then the computation of current epoch becomes more complicated and we need to store the original value of blocks per epoch and the new value to compute epochs correctly as a function of the hotshot block height. I wonder if we should make the blockPerEpoch an array of `(hotshotBlockNumber, blocksPerEpoch)` to enable this computation.

great idea, we discussed this and concluded that we didn't need to support changing hotshot blocks per epoch at this moment. but we could do, @jbearer @bfish713 what do you think?

contracts/src/StakeTable.sol
```diff
@@ -129,11 +153,45 @@ contract StakeTable is AbstractStakeTable {
...
+    function nextAvailableEpoch(uint64 firstAvailableEpoch, uint64 numPending)
+        external
+        view
+        override
+        returns (uint64, uint64)
+    {
+        uint64 epoch;
+        uint64 queueSize;
+
+        if (firstAvailableEpoch < currentEpoch() + 1) {
```
sveitser, line 184: Add a comment as to why we don't need to worry about overflow?

contracts/test/StakeTable.t.sol
```diff
@@ -745,4 +755,193 @@ contract StakeTable_register_Test is Test {
...
+
+    // TESTS FOR CURRENT EPOCH
+    function test_initialEpoch_isZero() public {
+        // assert the current block height is initialBlockHeight
+        uint64 initialBlockHeight = 0;
+        (, uint64 currentBlockHeight,) = lcMock.finalizedState();
+        assertEq(currentBlockHeight, initialBlockHeight);
+
+        // Calculate the expected epoch
+        uint64 expectedEpoch = initialBlockHeight / hotShotBlocksPerEpoch;
```
sveitser, line 767: I think the test would be easier to read if it just hardcoded zeroes, wherever zero is expected.

contracts/src/StakeTable.sol
```diff
@@ -323,7 +343,8 @@ contract StakeTable is AbstractStakeTable {
         // currentEpoch() + 1 (the start of the next full epoch), but in periods of high churn the
         // queue may fill up and it may be later. If the queue is so full that the wait time exceeds
         // the caller's desired maximum wait, abort.
-        (uint64 registerEpoch, uint64 queueSize) = this.nextRegistrationEpoch();
+        (uint64 registerEpoch, uint64 queueSize) =
+            this.nextAvailableEpoch(firstAvailableRegistrationEpoch, _numPendingRegistrations);
```
sveitser, line 347: There is the comment above how `_numPendingRegistrations` is not the total queue size.

https://github.com/EspressoSystems/espresso-sequencer/blob/072806816ed349315e47f695c689d55ea3537d5d/contracts/src/StakeTable.sol#L89-L91

But later we set it as such, although there is also a TODO in that comment

https://github.com/EspressoSystems/espresso-sequencer/blob/072806816ed349315e47f695c689d55ea3537d5d/contracts/src/StakeTable.sol#L213-L219

I think it would be nice if we named things a bit more explicitly. Since there are many epochs, queue sizes, etc. especially for those variables it's useful to be explicit.

contracts/src/StakeTable.sol
alysiahuggins: i added it as a todo in line 174, i want to start a discussion around this, will do so now

contracts/test/StakeTable.t.sol
alysiahuggins: makes sense, updated here https://github.com/EspressoSystems/espresso-sequencer/commit/af8841c5fb0b0985a495b8fd82e103775786f6ae

contracts/src/StakeTable.sol
alysiahuggins: so they don't handle overflow in the offchain code, so i think a revert if it overflows can work? though it's a check every time this is invoked. what do you think?

contracts/test/StakeTable.t.sol
sveitser: I meant more generally l if we use asserts in a test it's usually implied that we are asserting that some value is equal to an expected value. So instead of `assert(epoch, expectedEpoch)` we can write `assert(epoch, 0)` and end up with fewer indirections.

However, if the left and right hand size are some complicated expressions then defining some extra variables is helpful for reading.

sveitser:
@alysiahuggins I think many of the tests in this PR currently only assert how the state of the contract state changes. That is fine, but what is IMO more important and less of an implementation detail is if the events with correct epochs are emitted because this is what the application will actually consume (not the state of the contract). I think we should add some asserts to check that the epochs in the emitted events evolve as expected.

alysiahuggins:
> @alysiahuggins I think many of the tests in this PR currently only assert how the contract state changes. That is fine, but what is IMO more important and less of an implementation detail is if the events with correct epochs are emitted because this is what the application will actually consume (not the state of the contract). I think we should add some asserts to check that the epochs in the emitted events evolve as expected.

agreed and will do

contracts/src/StakeTable.sol
alysiahuggins: added a test to show that it panics/reverts when currentEpoch() == max(uint64) https://github.com/EspressoSystems/espresso-sequencer/commit/b215c7f80cad672413ed0e29c7b47d85aaac94c4

sveitser (approved):
I will approve an merge this now to start working on delegation since we have work that's blocked by it. I created a new issue for improving the tests.

Issue #2310, opened 2024-11-19
Stake Table: Variable Stake Amounts


alysiahuggins:
Allow validators to stake variable amounts above a minimum threshold.

- [x] Modify registration and staking functions to accept variable stake amounts
- [x] Implement validation to ensure stakes meet the minimum required amount

alysiahuggins:
closed by #2512

PR #2512, opened 2025-01-30
add admin only functions

### LINKED ISSUE #2309: Stake Table: Admin methods
Implement admin functions for the following.

- [ ] update validator threshold
- [x] update stake amount

alysiahuggins:
Closes #2309 & #2310

The admin of the contract will be a multisig and will have the ability to update
- churn rate
- min stake amount 
- light client address
- the admin

### This PR:
- adds an admin
- allows the admin to update variables above

### This PR does not:
- allow the admin to update validator threshold as that logic hasn't been implemented yet 
- determine a max churn rate or max minStakeAmount allowed, but would like to discuss tihs
- create an emergencyExit functions that only validators can call
- add a function to withdraw unstaked funds to the contract (e.g. funds that persons sent by mistake)

### Key places to review:
- admin logic and discussion topics mentioned 

### How to test this PR: 
`forge test match-contract StakeTable`

### DIFF
--- a/contracts/src/StakeTable.sol
+++ b/contracts/src/StakeTable.sol
@@ -6,11 +6,12 @@ import { BLSSig } from "./libraries/BLSSig.sol";
 import { AbstractStakeTable } from "./interfaces/AbstractStakeTable.sol";
 import { LightClient } from "../src/LightClient.sol";
 import { EdOnBN254 } from "./libraries/EdOnBn254.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 using EdOnBN254 for EdOnBN254.EdOnBN254Point;
 
 /// @title Implementation of the Stake Table interface
-contract StakeTable is AbstractStakeTable {
+contract StakeTable is AbstractStakeTable, Ownable {
     /// Error to notify restaking is not implemented yet.
     error RestakingNotImplemented();
 
@@ -64,7 +65,7 @@ contract StakeTable is AbstractStakeTable {
     // Error raised when zero point keys are provided
     error NoKeyChange();
 
-    /// Error raised when the caller is not the admin
+    /// Error raised when the caller is not the owner
     error Unauthorized();
 
     /// Error raised when the light client address is invalid
@@ -103,15 +104,14 @@ contract StakeTable is AbstractStakeTable {
 
     uint256 public minStakeAmount;
 
-    address public admin;
-
     /// TODO change constructor to initialize function when we make the contract upgradeable
     constructor(
         address _tokenAddress,
         address _lightClientAddress,
         uint64 churnRate,
-        uint256 _minStakeAmount
-    ) {
+        uint256 _minStakeAmount,
+        address initialOwner
+    ) Ownable(initialOwner) {
         tokenAddress = _tokenAddress;
         lightClient = LightClient(_lightClientAddress);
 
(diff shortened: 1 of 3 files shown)


contracts/src/StakeTable.sol
```diff
@@ -524,10 +549,43 @@ contract StakeTable is AbstractStakeTable {
...
-    /// @notice Minimum stake amount
-    /// @return Minimum stake amount
-    /// TODO: This value should be a variable modifiable by admin
-    function minStakeAmount() public pure returns (uint256) {
-        return 10 ether;
+    /// @notice Update the admin
+    /// @dev The admin cannot be set to the zero address
+    /// @param _admin The new admin
+    function updateAdmin(address _admin) external {
+        if (msg.sender != admin) revert Unauthorized();
```
sveitser, line 556: Is it a conscientious decision not to inherit from an existing contract with access control?

contracts/src/StakeTable.sol
```diff
@@ -269,7 +292,7 @@ contract StakeTable is AbstractStakeTable {
         BN254.G1Point memory blsSig,
         uint64 validUntilEpoch
     ) external override {
-        uint256 fixedStakeAmount = minStakeAmount();
+        uint256 fixedStakeAmount = minStakeAmount;
 
         // Verify that the sender amount is the minStakeAmount
         if (amount < fixedStakeAmount) {
```
0xkato, line 298: Why do we have this `amount` variable if we are not using it? Everyone puts in the same amount (`minStakeAmount`) when calling `register()`.

contracts/src/StakeTable.sol
alysiahuggins: you know what, i was thinking through it, wondered if this was more gas efficient

contracts/src/StakeTable.sol
alysiahuggins: thanks for pointing this out, confirming with @veilgets. Are we doing a fixedAmount and thus the amount they stake has to be that fixedAmount or are we doing a variable amount that must be >= some minAmount?

contracts/src/StakeTable.sol
alysiahuggins: switched to Ownable here https://github.com/EspressoSystems/espresso-sequencer/commit/dbcfeb3144509b493b104fd69fedb823d964cf16

sveitser:
I think this PR shouldn't close #2309 because it does not address all the items in it (e.g. validator threshold).

alysiahuggins:
> I think this PR shouldn't close #2309 because it does not address all the items in it (e.g. validator threshold).

yea i'll create another issue for the validator thresholds, we didn't have consensus whether we were using thresholds until a few days ago from @veilgets

contracts/src/StakeTable.sol
alysiahuggins: we're choosing to go with a variable stake, so updated here https://github.com/EspressoSystems/espresso-sequencer/commit/2feda042ff490bac38de6566e0880bbf16993705

contracts/test/StakeTable.t.sol
```diff
@@ -207,7 +207,7 @@ contract StakeTable_register_Test is Test {
     }
 
     function test_RevertWhen_WrongStakeAmount() external {
-        uint64 depositAmount = 5 ether;
+        uint64 depositAmount = uint64(stakeTable.minStakeAmount()) - 1;
```
sveitser, line 210: Rename test function? Insufficent stake amount?

sveitser (approved):

contracts/test/StakeTable.t.sol
alysiahuggins: true, updated https://github.com/EspressoSystems/espresso-sequencer/pull/2512/commits/5dcb69738312f818275b7b76487787f5c04feb61

PR #2514, opened 2025-01-30
Fix `EpochCommittees` returning stake tables out of order

ss-es:
### This PR:
Fixes an issue where `EpochCommittees` was previously returning stake tables out of order. We now store (and prefer) the vector passed to `EpochCommittees::new_stake` without modification, and only retain a `BTreeMap` for lookup.

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -126,16 +126,16 @@ struct Committee {
     /// leader but without voting rights.
     eligible_leaders: Vec<StakeTableEntry<PubKey>>,
 
-    /// Stake table
+    /// Keys for nodes participating in the network
     stake_table: Vec<StakeTableEntry<PubKey>>,
 
-    /// DA members
+    /// Keys for DA members
     da_members: Vec<StakeTableEntry<PubKey>>,
 
-    /// Stake table indexed by public key, for efficient lookup.
+    /// Stake entries indexed by public key, for efficient lookup.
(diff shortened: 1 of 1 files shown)


imabdulbasit (approved):

types/src/v0/impls/stake_table.rs
```diff
@@ -126,10 +126,16 @@ struct Committee {
     /// leader but without voting rights.
     eligible_leaders: Vec<StakeTableEntry<PubKey>>,
 
-    /// TODO: add comment
+    /// Stake table
```
imabdulbasit, line 129: maybe add a comment explaining why we need these vectors

types/src/v0/impls/stake_table.rs
```diff
@@ -126,10 +126,16 @@ struct Committee {
...
+
+    /// DA members
+    da_members: Vec<StakeTableEntry<PubKey>>,
+
+    /// Stake table indexed by public key, for efficient lookup.
     indexed_stake_table: BTreeMap<PubKey, StakeTableEntry<PubKey>>,
 
-    /// TODO: comment
+    /// DA members indexed by public key, for efficient lookup.
     indexed_da_members: BTreeMap<PubKey, StakeTableEntry<PubKey>>,
```
jbearer, line 139: Nit: maybe make the value here `usize` index into `stake_table` and `da_members`, so we only have a single source of truth for the actual contents of the stake table entry?

jbearer: Also consider `BTreeMap` -> `HashMap` so it is more clear that nothing is actually ordered by keys?

PR #2526, opened 2025-02-03
Stake Table  Integration test

### LINKED ISSUE #2511: Stake Table Epoch Integration Tests
Create a test which states nodes with the Epoch version enbabled.  The test should test that a quorum of nodes can make progress through multiple epochs with the stake table changing.  The tests should test the following scenarios

- [ ] A single node is added to the stake table
- [ ] A node is removed from the stake table
- [ ] A node is taken down for an epoch that is in and started back up in the next epoch, which it is also a part of

tbro:
Closes #2511

Updates integration tests for POS.


types/src/v0/impls/state.rs
```diff
@@ -667,7 +667,7 @@ fn validate_builder_fee(
         // TODO Marketplace signatures are placeholders for now. In
         // finished Marketplace signatures will cover the full
         // transaction.
-        if version.minor >= 3 {
+        if version.minor >= 99 {
```
imabdulbasit, line 670: ah I think we should replace this hardcoded version with the type everywhere

tbro: 37307e47d4ea8a00de8311963a115bd15f77f3ff

types/src/v0/impls/stake_table.rs
```diff
@@ -526,6 +482,31 @@ impl Membership<SeqTypes> for EpochCommittees {
             })
     }
 }
+
+pub(crate) async fn get_chain_config(
```
tbro, line 486: Great fix. But I think `Header` should be responsible for returning it's `ChainConfig`. Since header already has `get_chain_config` maybe this can be `Header.resolve_chain_config()`. Or possibly this can be added as a case to `Header.get_chain_config`... haven't thought about it too much...

imabdulbasit: Header does have get_chain_config() but it takes in validated_state

imabdulbasit: yeah, maybe I can re-use it by changing the function signature

tests/smoke.rs
```diff
@@ -80,6 +80,14 @@ pub async fn assert_native_demo_works() -> Result<()> {
 
         last = new;
     }
+
+    if let Some(epoch) = testing.espresso.current_epoch().await? {
```
tbro, line 84: This is the only place we do epoch related check, but it if returns `None` we don't test anything at all.

imabdulbasit: None means it is not running epoch version

tbro: I don't believe that is necessarily true. I could be that we are running epoch version, but it is misbehaving.

tbro:
obsolete

PR #2528, opened 2025-02-04
Ignore unrecognized files when loading quorum proposals

jbearer:
Fixes an issue discovered in production, where the presence of unrecognized or unparseable files (e.g. swap files) in quorum proposal storage can cause quorum proposal loading to fail and thus prevent the node from starting, even if other well-formed files are also present.

### This PR:
* Modifies file system version of `load_quorum_proposals` to ignore unrecognized files instead of failing
* Adds two regression tests

### DIFF
--- a/sequencer/src/persistence/fs.rs
+++ b/sequencer/src/persistence/fs.rs
@@ -754,43 +754,66 @@ impl SequencerPersistence for Persistence {
 
         // Then, we want to get the entries in this directory since they'll be the
         // key/value pairs for our map.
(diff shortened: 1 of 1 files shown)


tbro (approved):
LGTM

sveitser (approved):
LGTM. @jbearer should this have a backport label?

jbearer:
Not sure. I think we abandoned the `release-labrador` branch and have been working off main again. I'll ask @babdor

PR #2529, opened 2025-02-04
light weight node

imabdulbasit:
This PR
- adds a flag leaf_only to the FetchingDataSource. This only stores leaves, headers and VID. It does not store/fetch block payloads and supports only relevant endpoints.
- supports merklized pruning by deleting all the nodes with created < h and only if it is not the newest node from the fee_merkle_tree state and block_merkle_tree_state
- makes query module required
- `state`, `catchup` and `status` module are enabled by default for `SQL` data source
- includes a test for merklized state pruning in `query-service` crate
- test for basic light weight functionality in `sequencer` crate

Note:
This PR does not remove the file system storage as we weren't able to sucessfully test the sqlite nodes on staging. After we have tested sqlite feature, we can completely remove the file system storage and enable  `state`,` catchup` and `status` modules by default for all nodes


hotshot-query-service/src/availability/data_source.rs
```diff
@@ -131,15 +131,19 @@ where
     where
         ID: Into<LeafId<Types>> + Send + Sync;
 
-    async fn get_block<ID>(&self, id: ID) -> Fetch<BlockQueryData<Types>>
+    async fn get_header<ID>(&self, id: ID) -> QueryResult<Fetch<Header<Types>>>
```
jbearer, line 134: I don't like the change to make this return a result. It's pretty contrary to the philosophy of this module outlined [above](https://github.com/EspressoSystems/espresso-sequencer/pull/2529/files#diff-80a83d00b88375a80aeb798e273cae755a90f9902328805783d974ecbe2f34f8R107), and it substantially complicates any code downstream of this for dubious benefit.

I think instead, it's fine if the methods where the lightweight node is not fetching return a `Fetch` that never resolves. At the API level, this will just timeout and ultimately return 404, so the UX is not really disrupted. And we shouldn't really be calling these endpoints on a lightweight node anyways so I think it's fine. Anyways, if the main change here is to add a flag in the fetching data source that simply limits what type of data we spawn an active fetch for, we should get this behavior (of returning an always-pending fetch sometimes) automatically

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -400,28 +411,34 @@ where
...
+    _types: PhantomData<(Types, S)>,
 }
 
-impl<Types, S, P> Pruner<Types, S, P>
+impl<Types, S, P> Fetcher<Types, S, P>
+where
+    Types: NodeType,
+    S: PruneStorage + Sync,
+{
+}
```
jbearer, line 432: What's this?

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -400,28 +411,34 @@ where
...
     aggregator: Option<BackgroundTask>,
-    pruner: Pruner<Types, S, P>,
+    pruner: Pruner<Types, S>,
 }
 
 #[derive(Derivative)]
-#[derivative(Clone(bound = ""), Debug(bound = "S: Debug, P: Debug"))]
-pub struct Pruner<Types, S, P>
+#[derivative(Clone(bound = ""), Debug(bound = "S: Debug,   "))]
+pub struct Pruner<Types, S>
```
jbearer, line 419: This is a nice simplification

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -568,25 +622,57 @@ where
...
     {
-        self.fetcher.get(id.into()).await
+        if self.fetcher.leaf_only {
+            let mut tx = self.read().await.map_err(|err| QueryError::Error {
+                message: err.to_string(),
+            })?;
+            return tx.get_header(id.into()).await.map(Fetch::Ready);
+        }
+
+        let block: Fetch<BlockQueryData<Types>> = self.fetcher.get(id.into()).await;
```
jbearer, line 636: This could get the leaf instead of the block, which would not only be more efficient but would also work with the leaf-only version, so we could get rid of the conditional logic here

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -568,25 +622,57 @@ where
...
+        Ok(block.map(|b| b.header().clone()))
     }
 
-    async fn get_payload<ID>(&self, id: ID) -> Fetch<PayloadQueryData<Types>>
+    async fn get_block<ID>(&self, id: ID) -> QueryResult<Fetch<BlockQueryData<Types>>>
     where
         ID: Into<BlockId<Types>> + Send + Sync,
     {
-        self.fetcher.get(id.into()).await
+        if self.fetcher.leaf_only {
```
jbearer, line 644: As per my comment in `availability/data_source.rs`...to me it would make more sense if we keep these `leaf_only` checks as low level as possible. So rather than checking on the receiving side (ie this query), we would check on the "sending" side...ie right before we fetch a non-leaf object or right before we store one in the database. That means all these high-level functions don't have to worry about it. The same high-level logic will return a fetch that always resolves (for the full data source) or one that sometimes pends forever (for the leaf-only version). Note that even in the leaf-only version such fetches might still resolve in some cases if we receive the resource passively (i.e. via the consensus event stream).

This has another benefit: to me, the code becomes easier to reason about and maintain, because we don't have to remember to add this check in every high level function, or worry about adding a new high-level function that mistakenly doesn't do the check and ends up fetching or storing a resource that it shouldn't. By guarding just the locations in the code that we really care about (don't want to store or fetch non-leaf objects) we can be less worried about everything built on top of that.

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -823,26 +963,36 @@ where
...
+
         Ok(Self {
-            storage: builder.storage,
+            storage: Arc::new(builder.storage),
             notifiers: Default::default(),
             provider: Arc::new(builder.provider),
-            payload_fetcher: Arc::new(payload_fetcher),
+            payload_fetcher,
             leaf_fetcher: Arc::new(leaf_fetcher),
             vid_common_fetcher: Arc::new(vid_common_fetcher),
```
jbearer, line 989: How come `payload_fetcher` was made optional but not `vid_common_fetcher`?

hotshot-query-service/src/data_source/fetching/block.rs
```diff
@@ -96,6 +96,11 @@ where
             AvailabilityStorage<Types> + NodeStorage<Types> + PrunedHeightStorage,
         P: AvailabilityProvider<Types>,
     {
+        // Do not fetch if we are in light weight mode
+        if fetcher.leaf_only {
```
jbearer, line 100: Yep, this seems like the appropriate place for this check to me

jbearer: And also in `Storable::store` below

jbearer (changes requested):

hotshot-query-service/src/data_source/fetching/block.rs
imabdulbasit: ah forgot about that since I had diffferent implementation for that leaf only ds

hotshot-query-service/src/data_source/fetching.rs
jbearer: Ohhh that's why :( We really should index leaves by hash and such as well, all the information is there...

In the meantime I guess this hack is ok

imabdulbasit: I have added `HeaderQueryData` type which is just a wrapper but implements `Fetchable`. We can just subscribe to notifications for header request by block hash and payload hash. If we get a block or leaf then we take header from there and notify the subscribers

https://github.com/EspressoSystems/espresso-sequencer/pull/2529/commits/732b15af5adb8b9a59d704609fe17e054cbc03e1

hotshot-query-service/src/availability/data_source.rs
imabdulbasit: ok reverted all the result changes back

jbearer (commented):
Bunch of minor comments but this looks broadly correct 👍

hotshot-query-service/src/data_source/fetching.rs
jbearer: Beautiful!

hotshot-query-service/src/data_source/storage/sql/transaction.rs
```diff
@@ -409,9 +409,32 @@ impl Transaction<Write> {
...
+        // is not the newest node with its position
+        for state_table in state_tables {
+            self.execute(
+                query(&format!(
+                    "
+                DELETE FROM {state_table} WHERE (path, created) IN
+                (SELECT path, created FROM 
+                (SELECT path, created, 
+                ROW_NUMBER() OVER (PARTITION BY path ORDER BY created DESC) as rank 
+                FROM {state_table} WHERE created < $1) ranked_nodes WHERE rank != 1)"
```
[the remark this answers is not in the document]
imabdulbasit: So, I may have misunderstood this, but let's say we deleted all the data where created < 150. If we want a snapshot for created = 170, and there are no nodes between 150 and 170,  so the query will fail?

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -1619,39 +1704,14 @@ where
...
-}
-
 #[derive(Debug)]
 struct Notifiers<Types>
 where
     Types: NodeType,
 {
     block: Notifier<BlockQueryData<Types>>,
     leaf: Notifier<LeafQueryData<Types>>,
+    header: Notifier<HeaderQueryData<Types>>,
```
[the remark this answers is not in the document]
imabdulbasit: This works whether we get a block or a leaf. I thought it would be nice to have a separate notifier, but if we don’t want to add one, the header can always rely on the leaf. I guess yeah that is fine because blocks can not exist without the corresponding leaf

docker-compose.yaml
```diff
@@ -344,7 +344,7 @@ services:
     image: ghcr.io/espressosystems/espresso-sequencer/sequencer:main
     ports:
       - "$ESPRESSO_SEQUENCER2_API_PORT:$ESPRESSO_SEQUENCER_API_PORT"
-    command: sequencer -- http -- catchup -- status
+    command: sequencer -- http -- query -- storage-fs
```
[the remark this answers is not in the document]
imabdulbasit: something weird going on with sqlite nodes in upgrade test on CI. Might be related with issue that we saw on deployment.

hotshot-query-service/src/data_source/storage/sql/transaction.rs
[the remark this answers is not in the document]
imabdulbasit: So, I may have misunderstood this, but let's say we deleted some of the paths with created < 150. The newer version for some of the paths has created = 200. If we want a snapshot for created = 170, and those nodes have no path between 150 and 170, so the snapshot query will fail?

hotshot-query-service/src/data_source/fetching.rs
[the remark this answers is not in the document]
jbearer: Yeah any time we get a block we would have already notified for the leaf, or we would have already had the leaf in which case `get_header` would have succeeded in the first place. I think it's best to keep the number of notifiers to a minimum because the fetching logic is tricky, and as much as possible just have wrapper types that are derived from other fetchable types

docker-compose.yaml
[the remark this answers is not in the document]
jbearer: Ok we should figure this out but I guess it doesn't have to be a blocker since sqlite is already broken anyways lol

hotshot-query-service/src/data_source/storage/sql/transaction.rs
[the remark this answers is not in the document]
jbearer: Oh yeah you are entirely correct. Never mind this thread then

jbearer (approved):
Loosk good!

PR #2531, opened 2025-02-04
add timeout to provider::connect() in l1 update loop

imabdulbasit:
This PR adds a timeout to `provider::connect()` in the L1 update state loop to prevent hangs due to DNS resolution issues.


sveitser:
If this works to fix the test for you I think that's great. But please check the logs if it actually does since myself and the CI don't seem to hit this problem.

PR #2541, opened 2025-02-05
[Monorepo] `marketplace-builder-core`

rob-maron:


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.82"
```
tbro, line 2: There was a PR to explicitly move this back to stable: 
https://github.com/EspressoSystems/espresso-sequencer/pull/2522/files

Is there a reason for pinning to version number again?

Cargo.toml
```diff
@@ -79,13 +83,18 @@ tokio = { version = "1", default-features = false, features = [
...
+tracing-test = "0.1"
+num_cpus = "1"
 
 hotshot = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.84" }
 # Hotshot imports
 hotshot-builder-api = { git = "https://github.com/EspressoSystems/hotshot", tag = "0.5.84" }
-hotshot-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.5.84" }
-marketplace-builder-core = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.5.84" }
-marketplace-builder-shared = { git = "https://github.com/EspressoSystems/marketplace-builder-core", tag = "0.5.84" }
+hotshot-builder-core = { path = "hotshot-builder-core" }
```
tbro, line 93: I think we should preserve the order of all the `builder-` things together.

hotshot-builder-core-refactored/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
```
tbro, line 1: I don't see this crate referenced in `Cargo.toml`. Is it included by mistake?

tbro (commented):
Just a few small questions.

rust-toolchain.toml
rob-maron: Oh nice, I changed it because I was hitting it locally. Let me revert it and `rustup`, ty

tbro (approved):

PR #2544, opened 2025-02-05
fix process compose logs

imabdulbasit:
process compose logs uploaded is actually the demo native script and not the actual logs.
This PR defines the path in process-compose.yml where the logs are stored and changes the workflow to upload the logs.


sveitser (approved):
I think for locally running maybe not writing to /tmp is better?

Issue #2535, opened 2025-02-05
Migrate all contract bindings to alloy


sveitser:
After #2456 is merged we will have bindings for all contracts. We can remove the ethers bindings one by one until we have migrated all the contracts to alloy.

- [ ] LightClient contracts
- [ ] PlonkVerifier contracts
- [ ] FeeContract contract
- [ ] PermissionedStakeTable contract

PR #2547, opened 2025-02-06
Deploy fee contract with alloy

sveitser:
Adds a few functions to deploy contracts with alloy and instantiate alloy bindings. These are generic and should also work for the other contracts. The functions are copies of the functions we had with minimal changes to work with alloy. Added the `_alloy` suffix, and types and lifetimes fixed to please compiler.

Note: the fee contract ethers bindings are still used in Espresso bridge, to migrate the bridge code we need to also create a version of EthSignatureKey that uses alloy types.

Part of #2535


sveitser:
Demo consistency CI failure is expected because something needed fixing in the process-compose but since docker compose doesn't load env vars by default it's not affected.

rob-maron (approved):

PR #2548, opened 2025-02-06
Simplify data flow for epoch queues

sveitser:
Note this PR is from reviewing #2316 and into that branch.

- Some renames to make it easier to read.
- Mutate the state of the queue in place relying on EVM reverts to restore the state of the contract when the transactions revert.

Would have to still update abstract contract and tests, so mainly pushing this for discussion. We would also have a new function for clients to determine available epochs, if that's actually required.

I think for clients it might be fine if we just return the max of the registrationEpoch variable in the contract and currentEpoch() + 1. If the registration epoch gets bumped by one by the time their transaction is included that's just a race condition we can't avoid anyway. They can set a reasonable `validUntilEpoch` to avoid having to wait too long.


contracts/src/StakeTable.sol
```diff
@@ -166,32 +164,37 @@ contract StakeTable is AbstractStakeTable {
...
-        uint64 epoch;
-        uint64 queueSize;
-
-        if (firstAvailableEpoch < currentEpoch() + 1) {
-            epoch = currentEpoch() + 1;
-            queueSize = 0;
-        } else if (numPending >= maxChurnRate) {
-            epoch = firstAvailableEpoch + 1;
-            queueSize = 0;
+    function pushToRegistrationQueue() internal override returns (uint64, uint64) {
```
alysiahuggins, line 169: previously i merged these into one function `nextAvailableEpoch(...)` since the logic being performed was the same - just different queues (register and exit). pros of that are: smaller contract size, reusability and consistency, cons: reduced readability which can lead to errors if the function usage isn't clear.

Which do you think is best?

sveitser: In order for them to be general to work for both queues we can't directly operate on the state I think, so I prefer to have them separate.

PR #2565, opened 2025-02-09
fix: replace panic! with assert! in builder_state.rs tests for better error reporting

avorylli:
Description:
- Replaced panic! calls with assert! in test functions for more descriptive error messages
- Added builder_state_id to error messages for better debugging
- Improved test failure diagnostics while maintaining the same test functionality
### No functional changes, only test improvements


jparr721:
Thanks for your contribution. Since assert!(false) is just an arguably less-readable version of panic!, we'll have to close this PR.

PR #2574, opened 2025-02-10
[Monorepo] Put all crates in a separate folder

rob-maron:


jparr721 (approved):

rob-maron:
Drafting so I don't merge before a couple of the big ported PRs

Issue #2458, opened 2025-01-15
[VID] update the logic for namespace proof verification


mrain:
new namespace proof will contain a proof of correct encoding or a proof of incorrect encoding.
namespace verification should decide the case and check the corresponding proof.

Other items
- https://github.com/EspressoSystems/espresso-sequencer/pull/2579#discussion_r1969965954
- https://github.com/EspressoSystems/espresso-sequencer/pull/2579#discussion_r1969621453 (and https://github.com/EspressoSystems/espresso-sequencer/pull/2579#discussion_r1969901340)

Update:
proof of incorrect encoding will be handled in #2865

PR #2579, opened 2025-02-11
[VID] AvidM integration

### LINKED ISSUE #2600: [VID Upgrade] new VID scheme integration

mrain:
Closes #2600 

### This PR:
Integrate the efficient AvidM scheme.
- New VID scheme ships with Epoch upgrade.
- Underlying it's currently using `keccak256` hash function. Should also consider `poseidon2`.

Temporarily disable hotshot test_1 b/c #2664 

### This PR does not:
- Update the namespace proof logic. Will deal with it in this issue #2458


hotshot-builder-core-refactored/src/service.rs
```diff
@@ -402,8 +403,10 @@ where
         let num_nodes = self.num_nodes.load(Ordering::Relaxed);
 
         let fut = async move {
-            let join_handle =
-                tokio::task::spawn_blocking(move || advz_commitment(&encoded_txns, num_nodes));
+            let join_handle = tokio::task::spawn_blocking(move || {
+                let encoded_tx_len = encoded_txns.len();
+                advz_scheme(num_nodes).commit_only(encoded_txns).map(VidCommitment::V0).unwrap_or_else(|err| panic!("VidScheme::commit_only failure:(num_storage_nodes,payload_byte_len)=({num_nodes},{encoded_tx_len}) error: {err}"))
```
ss-es, line 408: I don't think it matters either way, but is this keeping the same logic?

the new version seems much less readable to me

hotshot-example-types/src/block_types.rs
```diff
@@ -400,10 +398,8 @@ impl Committable for TestBlockHeader {
                 <TestBlockHeader as BlockHeader<TestTypes>>::block_number(self),
             )
             .constant_str("payload commitment")
-            .fixed_size_bytes(
-                <TestBlockHeader as BlockHeader<TestTypes>>::payload_commitment(self)
-                    .as_ref()
-                    .as_ref(),
+            .var_size_bytes(
+                <TestBlockHeader as BlockHeader<TestTypes>>::payload_commitment(self).as_ref(),
```
ss-es, line 402: no objection here, but wondering: is this just a cleanup for tests or was the commitment required to be changed?

hotshot-testing/tests/tests_1.rs
```diff
@@ -5,5 +5,6 @@
 // along with the HotShot repository. If not, see <https://mit-license.org/>.
 
 mod tests_1 {
-    automod::dir!("tests/tests_1");
+    // These tests are temporarily disabled for the following issue: <https://github.com/EspressoSystems/espresso-sequencer/issues/2664>
+    // automod::dir!("tests/tests_1");
```
ss-es, line 9: I think we should revert this here and just add an `#[ignore]` for the specific test that was failing (`test_da_task`?)

hotshot-types/src/consensus.rs
```diff
@@ -332,6 +331,12 @@ pub struct Consensus<TYPES: NodeType> {
     pub drb_seeds_and_results: DrbSeedsAndResults<TYPES>,
 }
 
+/// Type alias to avoid complexity
+pub type PayloadWithMetadata<TYPES> = (
+    <TYPES as NodeType>::BlockPayload,
+    <<TYPES as NodeType>::BlockPayload as BlockPayload<TYPES>>::Metadata,
+);
```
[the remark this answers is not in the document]
ss-es: +1 to Lucas' suggestion. This seems like it should be a struct with `payload` and `metadata` fields

hotshot-types/src/data/ns_table.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use std::ops::Range;
+
+/// Byte lengths for the different items that could appear in a namespace table.
+const NUM_NSS_BYTE_LEN: usize = 4;
+const NS_OFFSET_BYTE_LEN: usize = 4;
+const NS_ID_BYTE_LEN: usize = 4;
+
+/// Helper function for AvidM scheme to parse a namespace table.
+/// If the namespace table is invalid, it returns a default single entry namespace table.
```
ss-es, line 17: Can you add even a rough description of what a valid namespace table is and/or what the layout should be?

e.g. I assume that `bytes` is an array with namespace tables at the start and `payload_byte_len` bytes at the end, but it would be nice to have this written down here

types/src/v0/impls/header.rs
```diff
@@ -813,7 +815,7 @@ impl BlockHeader<SeqTypes> for Header {
             height = parent_leaf.block_header().block_number() + 1,
             parent_view = ?parent_leaf.view_number(),
             payload_commitment,
-            payload_size = VidSchemeType::get_payload_byte_len(&_vid_common),
+            // payload_size = ADVZScheme::get_payload_byte_len(&_vid_common),
```
ss-es, line 818: should this be deleted? (not just commented out)

I think there's a few lines that were commented out in this file

hotshot-types/src/consensus.rs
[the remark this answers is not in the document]
mrain: Now wrapped as a struct: [015aa10](https://github.com/EspressoSystems/espresso-sequencer/pull/2579/commits/015aa101732eac1e28076399c16159c87d863d6a)

hotshot-builder-core-refactored/src/service.rs
mrain: Yes this is annoying. Good that VID is removed in the recent merge. [a2657f8](https://github.com/EspressoSystems/espresso-sequencer/pull/2579/commits/a2657f8c035da3cf27998c4bec8466af3f6b550d)

hotshot-example-types/src/block_types.rs
mrain: Ahh this shall be a `fixed_size_bytes`. [015aa10](https://github.com/EspressoSystems/espresso-sequencer/pull/2579/commits/015aa101732eac1e28076399c16159c87d863d6a)
Commitments won't change, as those actual types in the sequencer.

hotshot-types/src/data/ns_table.rs
mrain: This is mostly `copied` from the sequencer repo. Add a link to it [015aa10](https://github.com/EspressoSystems/espresso-sequencer/pull/2579/commits/015aa101732eac1e28076399c16159c87d863d6a)

vid/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "vid"
+version = "0.1.0"
+edition = "2021"
+authors = ["Espresso Systems <hello@espressosys.com>"]
+license = "MIT"
```
alxiong, line 6: use workspace=true instead?

vid/rustfmt.toml
alxiong, line 1: Q: is it a good idea to have sub-crate rustfmt? or should we just forgo this. (personally I like this config but feels intrusive to impose on the entire monorepo)

vid/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn from(err: Poseidon2Error) -> Self {
+        VidError::Internal(err.into())
+    }
+}
+
+/// Alias
+type VidResult<T> = Result<T, VidError>;
+
+/// Trait definition for a Verifiable Information Dispersal (VID) scheme.
+pub trait VidScheme {
```
alxiong, line 43: Q: we already have `trait jf_vid::VidScheme` (and used everywhere), would it be confusing to migrate? or can we try to unify the two and stablize this trait to avoid future disruption?

vid/src/utils/bytes_to_field.rs
alxiong, line 1: this file contains lots of duplicated functions from https://github.com/EspressoSystems/jellyfish/blob/main/vid/src/advz/bytes_to_field.rs, should we dedup or just completely makes the `jf-vid` obsolete.

hotshot-types/src/vid/advz.rs
```diff
@@ -0,0 +1,259 @@
...
+
+/// Convert a [`VidDisperse<Advz>`] to a [`VidDisperse<VidSchemeType>`].
+///
+/// Foreign type rules prevent us from doing:
+/// - `impl From<VidDisperse<VidSchemeType>> for VidDisperse<Advz>`
+/// - `impl VidDisperse<VidSchemeType> {...}`
+///
+/// and similarly for `Statement`.
+/// Thus, we accomplish type conversion via functions.
+fn vid_disperse_conversion(vid_disperse: VidDisperse<Advz>) -> VidDisperse<ADVZScheme> {
```
alxiong, line 243: these feels more like `From<A> for B`?

hotshot-types/src/data/ns_table.rs
alxiong: I think we should use [`NsTable::ns_range()` ](https://github.com/EspressoSystems/espresso-sequencer/blob/993c22cb3ed9897fab091dc19ee289fe35a077e4/sequencer/src/block/full_payload/ns_table.rs#L257-L261)which does exactly this already.

alxiong (commented):
@mrain crate `vid` is directly copied over without new addition, right? anything I should pay special attention to?

vid/src/utils/bytes_to_field.rs
mrain: I don't want the new crate to depend on the old one. Let's plan to merge these two crate in the future.

vid/rustfmt.toml
mrain: cc @rob-maron do you have any suggestion here?

sequencer/src/persistence/sql.rs
```diff
@@ -1184,6 +1184,25 @@ impl SequencerPersistence for Persistence {
         tx.commit().await
     }
 
+    async fn append_vid2(
```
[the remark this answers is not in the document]
imabdulbasit: this would break things because hotshot calls `append_vid2()`  which converts VID1 to VID0 and calls `append_vid()`. With this change, we are going to insert new VID into same tables/files.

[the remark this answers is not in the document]
imabdulbasit: this would require a migration https://github.com/EspressoSystems/espresso-sequencer/pull/2478

hotshot-types/src/vid/advz.rs
mrain: Yes and the comment explains the reason.

mrain: For `NsTable::ns_range()` comment, I don't know why I cannot reply there.
That's within sequencer scope and I don't want to introduce a circular dependency. So I duplicated the logic.

sequencer/src/persistence/sql.rs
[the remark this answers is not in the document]
mrain: Another solution is to treat the input `proposal` as bytes and takes additional inputs `view_number` and `payload_commitment`. WDYT?
cc @ss-es @rob-maron

hotshot-query-service/src/data_source/update.rs
```diff
@@ -139,16 +142,13 @@ where
...
-                        Some(share.share.clone()),
+                        Some(VidShare::V0(share.share.clone())),
                     ),
                     Some(VidDisperseShare::V1(share)) => (
-                        Some(VidCommonQueryData::new(
-                            leaf.block_header().clone(),
-                            share.common.clone(),
-                        )),
-                        Some(share.share.clone()),
+                        Some(VidCommonQueryData::new(leaf.block_header().clone(), None)),
```
imabdulbasit, line 150: so new VID variant does not have common(). We do store common() data, which would be None in this case. I don't know how this is being used but need to think about this.

hotshot-query-service/src/data_source/update.rs
```diff
@@ -139,16 +142,13 @@ where
...
+                        Some(VidShare::V1(share.share.clone())),
```
imabdulbasit, line 151: so this is also a breaking change, and requires a migration

hotshot-types/src/vid/advz.rs
alxiong: > That's within sequencer scope and I don't want to introduce a circular dependency. So I duplicated the logic.

i see, maybe it's an indication that we should relocate those code to here then.

vid/src/avid_m/namespaced.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// The list of all namespace commitments
+    ns_commits: Vec<AvidMCommit>,
+    /// The size of each namespace
+    ns_lens: Vec<usize>,
+    /// Actual share content
+    content: Vec<RawAvidMShare>,
+}
+
+impl NsAvidMShare {
+    fn ns_share(&self, ns_id: usize) -> AvidMShare {
```
alxiong, line 34: nit: shall we call it `inner()`? to indicate the fact that it's a internal detail

vid/src/avid_m/namespaced.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /// Commit to a payload given namespace table.
+    pub fn commit(
+        param: &NsAvidMParam,
+        payload: &[u8],
+        ns_table: impl IntoIterator<Item = Range<usize>>,
+    ) -> VidResult<NsAvidMCommit> {
+        let ns_commits = ns_table
+            .into_iter()
```
alxiong, line 61: should we validate `ns_table` before committing them? better yet, can we rely on type safety, i remember `NsTable` struct can be constructed safely from an iter of ranges with validation, and we can just accept a `NsTable` type here.

vid/src/utils/bytes_to_field.rs
alxiong: sure, let's discuss the plan in the corresponding issue of this PR?

vid/src/avid_m/namespaced.rs
mrain: Again `NsTable` currently resides in `sequencer` scope.
Let's handle it together in #2458

vid/src/avid_m/namespaced.rs
mrain: Done in 25c2965

vid/Cargo.toml
mrain: Done in 25c2965

vid/src/utils/bytes_to_field.rs
mrain: Issue created here: #2679

hotshot-types/src/vid/advz.rs
mrain: Let's discuss it here: https://github.com/EspressoSystems/espresso-sequencer/issues/2458

vid/rustfmt.toml
mrain: Locally on my side it works well and only applies to the `vid` crate.

bfish713 (approved):

sequencer/src/persistence/sql.rs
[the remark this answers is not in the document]
imabdulbasit: think its fine to merge this, I can do followup work related to migration in my types PR

PR #2613, opened 2025-02-13
Fix builder API verisoning

rob-maron:


ss-es (approved):
can you rerun CI just to make sure the builder test that's failing is just flakiness?

ss-es (changes requested):
actually, it looks like the test failed 6 times in a row -- I'm guessing it's not just flakiness

PR #2620, opened 2025-02-14
Separate HotShot CI

ss-es:
### This PR:
Splits HotShot integration tests into their own CI workflow, with nextest settings appropriate for HotShot.


tbro (commented):
I think if you update the rust-doc workflow to use github action to install just (as you do in the test workflow) instead of `apt install` it will get a newer version and turn green.

tbro (approved):

PR #2623, opened 2025-02-14
Stake Table Catchup in Hotshot

bfish713:
Closes #<ISSUE_NUMBER>

### This PR:

Recreates: https://github.com/EspressoSystems/HotShot/pull/4109


hotshot-task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -126,11 +126,14 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
-                let membership_reader = self.membership.read().await;
-                let membership_stake_table = membership_reader.stake_table(qc.data.epoch);
-                let membership_success_threshold =
-                    membership_reader.success_threshold(qc.data.epoch);
-                drop(membership_reader);
+                let prev_epoch = qc.data.epoch;
+                let mem = if prev_epoch < self.membership.epoch() {
+                    &self.membership.prev_epoch().await.ok()?
+                } else {
+                    &self.membership
```
ss-es, line 133: this seems suspicious, is there not a better way to deal with this? e.g. I feel like we might rather just pass in the `EpochMembershipCoordinator` so we can specifically get the `EpochMembership` for `prev_epoch`

hotshot-task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -307,16 +310,15 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             self.epoch_height,
         );
 
+        let mem = if self.membership.epoch() != epoch {
+            self.membership.next_epoch().await?
+        } else {
+            self.membership.clone()
+        };
```
ss-es, line 317: same comment here

since there's a specific epoch we need the `Membership` for shouldn't we be asking for that

hotshot-task-impls/src/quorum_proposal/handlers.rs
ss-es: I don't really know what a realistic attack might be but you can imagine e.g. someone puts in a fake `epoch` into `qc.data.epoch` but tricks you into only validating against `prev_epoch`

hotshot-task-impls/src/quorum_proposal_recv/mod.rs
```diff
@@ -139,18 +143,30 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
     ) {
         match event.as_ref() {
             HotShotEvent::QuorumProposalRecv(proposal, sender) => {
+                // tracing::error!("proposal recv {:?}", proposal.data);
```
ss-es, line 146: meant to be deleted?

ss-es (commented):
just leaving some initial comments about the swap in the `Membership` object internally in hotshot. I think overall it makes sense, but I'm really concerned about some of the uses of `next_epoch` and `prev_epoch` and would feel much better if we explicitly passed the full `EpochMembershipCoordinator` to query for a specific epoch in those places (though in some other places, like where we e.g. check to see if we're the leader for the next epoch I think `next_epoch` makes complete sense)

I haven't yet looked into the actual catchup logic, I'll try to do that tomorrow

hotshot-task-impls/src/quorum_proposal/handlers.rs
bfish713: Yeah agree with this point, I had some debugging issues with other places with this pattern as well.  I think we should maybe just eplicitly checking that the epoch is for the next or prev epoch rather than `<` or `>` the one in the membership would work?

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,440 @@
...
+            let mut map = coordinator.catchup_map.lock().await;
+            if map.contains_key(&epoch) {
+                return;
+            }
+            let (tx, rx) = broadcast(1);
+            map.insert(epoch, rx.deactivate());
+            tx
+        };
+        // do catchup
+        let ret = coordinator.catchup(epoch).await;
```
ss-es, line 177: oh, I see! so `catchup` is a blocking call, we're just `spawn`ing it?

are there no issues with calling `spawn` in mutually recursive functions like this? it seems like `catchup` calls `spawn_catchup`, which spawns a `catchup` task itself

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    /// Get the prior epoch
+    pub async fn prev_epoch(&self) -> Result<Self> {
+        let Some(epoch) = self.epoch else {
+            return Ok(self.clone());
+        };
+        if *epoch == 0 {
+            return Ok(self.clone());
+        }
```
ss-es, line 223: I feel like this should return `Err("there is no preceding epoch")` or something

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// Get the epoch this membership is good for
+    pub fn epoch(&self) -> Option<TYPES::Epoch> {
+        self.epoch
+    }
+
+    /// Get a membership for the next epoch
+    pub async fn next_epoch(&self) -> Result<Self> {
+        if self.epoch.is_none() {
+            Ok(self.clone())
+        } else {
```
ss-es, line 209: I feel like this should return an `Err("missing next epoch")`; returning the current one feels very unexpected here and I feel like I'd be very likely to use this incorrectly

ss-es (commented):
some more comments on the actual catchup logic

hotshot-types/src/epoch_membership.rs
bfish713: I don't think there is any problem with it, basically we'll spawn a task for each epoch that needs to be completed, each task will block until the task for previous epoch completes.  I think the only potential issue is it would create a lot of tasks for very deep catchups but I think the most we can get is one task per epoch (no matter how many times we call catchup)

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// If there is another catchup in progress this will not duplicate efforts
+    /// e.g. if we start with only epoch 0 stake table and call catchup for epoch 10, then call catchup for epoch 20
+    /// the first caller will actually do the work for to catchup to epoch 10 then the second caller will continue
+    /// catching up to epoch 20
+    async fn catchup(self, epoch: TYPES::Epoch) -> Option<EpochMembership<TYPES>> {
+        // recursively catchup until we have a stake table for the epoch containing our root
+        assert!(
+            *epoch != 0,
+            "We are trying to catchup to epoch 0! This means the initial stake table is missing!"
+        );
```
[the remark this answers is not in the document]
ss-es: I think I agree that it shouldn't happen, I'm just a bit nervous about putting a panic in code that could theoretically get called from anywhere when there's an alternative like just returning `None`

though it probably doesn't matter a lot either way

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,440 @@
...
+            .add_epoch_root(next_epoch, header)
+            .await?;
+        updater(&mut *(self.membership.write().await));
+
+        let sync = self.membership.read().await.sync_l1().await?;
+        sync(&mut *(self.membership.write().await));
+        Some(EpochMembership {
+            epoch: Some(epoch),
+            coordinator: self.clone(),
+        })
```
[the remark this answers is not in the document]
bfish713: Yeah I think just calling wait_for_catchup is all that's needed

hotshot-types/src/epoch_membership.rs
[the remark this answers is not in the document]
bfish713: I changed to retrun an error for epoch 1 and 0

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    /// Get the prior epoch
+    pub async fn prev_epoch(&self) -> Result<Self> {
+        let Some(epoch) = self.epoch else {
+            return Ok(self.clone());
+        };
+        ensure!(*epoch > 0, "There is no previous epoch");
+
+        self.coordinator.membership_for_epoch(Some(epoch - 1)).await
+    }
```
ss-es, line 219: I'm worried about the asymmetry. `next_epoch` will return `None` if we are not using epochs, but `prev_epoch` will return the current epoch. I feel like either is fine, and we should just pick one and stick with it.


I think the fact that we're doing both though suggests to me we should scrap the methods and just call `membership.coordinator.membership_for_epoch(...)` wherever we need it

ss-es (approved):
I think this looks fine to me overall, though I admittedly didn't check the details of every membership replacement

I still have some concerns about a few things, but they could be addressed separately. maybe they're not a big deal either

hotshot-types/src/epoch_membership.rs
bfish713: I'll get rid of `prev_epoch`, I think next epoch is pretty useful for epoch transtion but the prev epoch is a little bit trickier and better to just explicitly get the EpochMembership

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,425 @@
...
+            coordinator: self.clone(),
+        };
+        let Some(epoch) = maybe_epoch else {
+            return Ok(ret_val);
+        };
+        if self.membership.read().await.has_epoch(epoch) {
+            return Ok(ret_val);
+        }
+        if self.catchup_map.lock().await.contains_key(&epoch) {
+            return Err(warn!(
```
imabdulbasit, line 83: Shouldn't we wait for the catch up request to complete and notify instead of returning an error?

hotshot-types/src/epoch_membership.rs
```diff
@@ -0,0 +1,425 @@
...
+                "Stake table for Epoch {:?} Unavailable. Catch up already in Progress",
+                epoch
+            ));
+        }
+        let coordinator = self.clone();
+        spawn_catchup(coordinator, epoch);
+
+        Err(warn!(
```
imabdulbasit, line 91: wait here too?

hotshot-types/src/epoch_membership.rs
imabdulbasit: I haven't looked at the whole PR, but I think everything goes through this function. If we return an error, won't that prevent consensus from making progress?

ss-es: yes, this is intentional -- we wanted to make sure we weren't potentially blocking consensus on catchup

blocking consensus on this call can be very dangerous, because e.g. imagine a malicious node sends us a proposal for an epoch well in the future. we would attempt to perform catchup (which will always fail), but in the meantime consensus gets blocked while catchup is happening

with the current implementation, we may fail to vote on a proposal if we don't have the stake table but this should be a much less serious failure. we will request information from other nodes in the background, and should be able to start participating again within a few views at worst

bfish713: Yeah exactly, and to add, it's not really worth waiting for catchup, if catchup is fast we'll be caught up by the next proposal, if it's slow by the time we're caught up the message triggered the membership to catchup will be quite old and we could end up getting stuck behind everyone else trying to churn through all the old messages that queued up waiting for catchup

bfish713: Also there is `wait_for_catchup` if you want a blocking call

imabdulbasit: Ok that makes sense. Thank you for the explanation

PR #2624, opened 2025-02-14
Fix handling of swap files

jbearer:
Cleans up some mis-handled swap files that I missed in the last fix (#2528). Thanks to NodeGuardians for identifying this.

### DIFF
--- a/sequencer/src/persistence/fs.rs
+++ b/sequencer/src/persistence/fs.rs
@@ -221,44 +221,56 @@ impl Inner {
 
     fn collect_garbage(
         &mut self,
-        view: ViewNumber,
-        intervals: &[RangeInclusive<ViewNumber>],
+        decided_view: ViewNumber,
+        prune_intervals: &[RangeInclusive<ViewNumber>],
     ) -> anyhow::Result<()> {
-        let view_number = view;
-        let prune_view = ViewNumber::new(view.saturating_sub(self.view_retention));
+        let prune_view = ViewNumber::new(decided_view.saturating_sub(self.view_retention));
 
-        let delete_files = |intervals: &[RangeInclusive<ViewNumber>],
-                            keep,
-                            dir_path: PathBuf|
-         -> anyhow::Result<()> {
-            if !dir_path.is_dir() {
-                return Ok(());
-            }
-
-            for (v, path) in view_files(dir_path)? {
-                // If the view is the anchor view, keep it no matter what.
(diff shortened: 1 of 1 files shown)


sequencer/src/persistence/fs.rs
```diff
@@ -222,41 +222,36 @@ impl Inner {
     fn collect_garbage(
         &mut self,
         view: ViewNumber,
-        intervals: &[RangeInclusive<u64>],
+        intervals: &[RangeInclusive<ViewNumber>],
```
sveitser, line 225: I can't work out why this function takes a mutable reference to self. Is it to avoid being called in two places at once?

sequencer/src/persistence/fs.rs
```diff
@@ -222,41 +222,36 @@ impl Inner {
...
-
-        let delete_files =
-            |intervals: &[RangeInclusive<u64>], keep, dir_path: PathBuf| -> anyhow::Result<()> {
-                if !dir_path.is_dir() {
-                    return Ok(());
-                }
+        let view_number = view;
+        let prune_view = ViewNumber::new(view.saturating_sub(self.view_retention));
+
+        let delete_files = |intervals: &[RangeInclusive<ViewNumber>],
```
sveitser, line 230: I think it would be a bit more straightforward to make `delete_files` a method on `Inner`. IIUC the only thing it takes from the scope is `prune_view`.

sequencer/src/persistence/fs.rs
```diff
@@ -276,22 +271,13 @@ impl Inner {
...
-            let path = entry.path();
-
-            let Some(file) = path.file_stem().and_then(|n| n.to_str()) else {
-                continue;
-            };
-            let Ok(v) = file.parse::<u64>() else {
-                continue;
-            };
-            if v > view.u64() {
+        for (v, path) in view_files(self.decided_leaf_path())? {
```
sveitser, line 279: More descriptive name than `v`? Looking at this PR diff, or file in general, I think it would be nice to be a bit more explicit with names that describe different kinds of view number.

sveitser (approved):
LGTM just some nits.

sequencer/src/persistence/fs.rs
jbearer: Yeah, I think that's the idea. It's not mutating the in-memory state that Rust cares about but it's still conceptually a mutable/exclusive function since it's updating shared files and directoreis

PR #2625, opened 2025-02-14
Push DRB result to Membership

pls148:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-testing/src/view_generator.rs
+++ b/hotshot-testing/src/view_generator.rs
@@ -37,6 +37,7 @@ use hotshot_types::{
     },
     traits::{
         consensus_api::ConsensusApi,
+        election::Membership,
         node_implementation::{ConsensusTime, NodeType, Versions},
         BlockPayload,
     },
@@ -46,7 +47,7 @@ use rand::{thread_rng, Rng};
 use sha2::{Digest, Sha256};
 
 use crate::helpers::{
-    build_cert, build_da_certificate, build_vid_proposal, da_payload_commitment, key_pair_for_id,
+    build_cert, build_da_certificate, build_vid_proposal, da_payload_commitment, TestNodeKeyMap,
 };
 
 #[derive(Clone)]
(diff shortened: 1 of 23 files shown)


hotshot-types/src/drb.rs
```diff
@@ -81,18 +81,22 @@ pub fn compute_drb_result<TYPES: NodeType>(drb_seed_input: DrbSeedInput) -> DrbR
 /// The DRB result is the output of a spawned `compute_drb_result` call.
 #[must_use]
 pub fn leader<TYPES: NodeType>(
```
ss-es, line 83: is this function actually used anywhere?

I feel like this one should be removed; it'll happen in the `Membership` implementation and I believe the logic should happen via that. I think I also have this function re-implemented in #2638.

hotshot-types/src/traits/election.rs
```diff
@@ -152,4 +152,8 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
     async fn sync_l1(&self) -> Option<Box<dyn FnOnce(&mut Self) + Send>> {
         None
     }
+
+    /// Called to notify the Membership when a new DRB result has been calculated.
+    /// Observes the same semantics as add_epoch_root
+    async fn add_drb_result(&mut self, _epoch: TYPES::Epoch, _drb_result: DrbResult) {}
```
ss-es, line 158: Can we remove the default no-op implementation here (also for `add_epoch_root`)?

hotshot-testing/src/view_generator.rs
```diff
@@ -73,8 +75,31 @@ pub struct TestView {
...
+            .expect("expected Membership::leader to succeed");
+        drop(membership_reader);
+
+        let sk = node_key_map
+            .get(&leader)
+            .expect("expected Membership::leader public key to be in node_key_map");
+
+        (sk.clone(), leader)
+    }
+
```
ss-es, line 99: do we need this extra logic? I think previously `key_pair_for_id` used the fact that all keys were generated from pre-determined seeds (which made sense in the context of a test)

I'm just a bit worried about complicating the logic here any more, since I don't think we need to pass around the keys like this in the integration tests

ss-es (approved):
looks fine to me

PR #2631, opened 2025-02-17
Optional VID compute field

rob-maron:
Makes new nodes backwards compatible with the old builder by accepting a field as optional that is no longer sent


ss-es (approved):
looks good to me!

I guess we're relying on the fact that any `None` for an `Option<_>` serializes the same way?

PR #2638, opened 2025-02-18
Variable stake and randomized leader selection

ss-es:
### This PR:
Adds variable stake and randomized leader selection

### DIFF
--- a/hotshot-types/src/traits/election.rs
+++ b/hotshot-types/src/traits/election.rs
@@ -7,8 +7,12 @@
 //! The election trait, used to decide which node is the leader and determine if a vote is valid.
 use std::{collections::BTreeSet, fmt::Debug, num::NonZeroU64};
 
+use sha2::{Digest, Sha256, Sha512};
+
+use crate::traits::signature_key::StakeTableEntryType;
 use async_trait::async_trait;
 use hotshot_utils::anytrace::Result;
+use primitive_types::{U256, U512};
 
 use super::node_implementation::NodeType;
 use crate::{traits::signature_key::SignatureKey, PeerConfig};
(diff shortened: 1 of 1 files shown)


hotshot-types/src/traits/election.rs
```diff
@@ -153,3 +157,78 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
...
+    hasher.update(view.to_le_bytes());
+    hasher.update(stake_table_hash);
+    let raw_breakpoint: [u8; 64] = hasher.finalize().into();
+
+    // then calculate the remainder modulo the total stake as a U512
+    let remainder: U512 =
+        U512::from_little_endian(&raw_breakpoint) % U512::from(cdf.last().unwrap().1);
+
+    // and drop the top 32 bytes, downcasting to a U256
+    let breakpoint: U256 = {
```
mrain, line 221: Should we use this instead? 
https://docs.rs/primitive-types/latest/primitive_types/struct.U256.html#impl-TryFrom%3CU512%3E-for-U256

hotshot-types/src/traits/election.rs
```diff
@@ -153,3 +157,75 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
...
+/// Panics if `cdf` is empty. Results in undefined behaviour if `cdf` is not ordered.
+///
+/// Note that we try to downcast a U512 to a U256,
+/// but this should never panic because the U512 should be strictly smaller than U256::MAX by construction.
+pub fn select_randomized_leader<SignatureKey, Entry: StakeTableEntryType<SignatureKey> + Clone>(
+    cdf: Vec<(Entry, U256)>,
+    stake_table_hash: [u8; 32],
+    drb: [u8; 32],
+    view: u64,
+) -> Entry {
```
alxiong, line 211: shouldn't this be a method of stake table struct or trait, something like `StakeTable.next_leader()` ?

alxiong (approved):
The logic looks correct. 
Do you mind adding what you posted on Zulip about the design in the code doc? (even though the code itself is self-explanatory, sometimes reading natural language in code doc is nice.

hotshot-types/src/traits/election.rs
ss-es: yes, it will be! I wanted to separate the raw logic into its own function, but the trait that handles the stake table will implement leader selection by calling this.

we just have multiple implementations that may or may not use this leader selection for tests etc

ss-es: I ended up just creating a wrapper type for all the DRB-related information we want to store between views, and the implementation now uses this opaquely!

bfish713 (approved):

PR #2644, opened 2025-02-19
Add a default port to ESPRESSO_SEQUENCER_API_PORT

Ancient123:
To reduce the number of required options to launch a basic sequencer docker, we want to set `ESPRESSO_SEQUENCER_API_PORT` to a sane default. Port 80 chosen since the API communicates over http.


jparr721:
Won't this require sudo to bind on startup?

tbro:
Slight preference for 8080 over 80.

imabdulbasit (approved):

PR #2647, opened 2025-02-20
Expose deployment block number of stake table contracts to clients

sveitser:
See comments in `InitializedAt.sol` for reasons for this change.

Some more background. When querying infura or alchemy for logs there are limits, for infura its

> To prevent queries from consuming too many resources, eth_getLogs requests are currently limited by three constraints:
> * A maximum of 5,000 parameters in a single request
> * A maximum of 10,000 results can be returned by a single query
> * Query duration must not exceed 10 seconds

for alchemy its

> You can make eth_getLogs requests on any block range with a cap of 10K logs in the response OR a 2K block range with no cap on logs in the response and 150MB limit on the response size

The way to limit queries is to specify a `from` and a `to` block. However, absent any other information there is no good value for the `from` block.

Clients can now read `initializedAtBlock` and start fetching events from that block.


contracts/src/InitializedAt.sol
```diff
@@ -0,0 +1,28 @@
...
+contract InitializedAt is Initializable {
+    // @notice The block number the contract was initialized at.
+    uint256 public initializedAtBlock;
+
+    // @dev If this contract is used as part of a non-upgradable contract the constructor is
+    // expected to be called.
+    constructor() {
+        initialize();
+    }
+
```
alxiong, line 20: wait, shouldn't the upgradable's constructor be:

```solidity
constructor() {
    _disableInitializers();
}
```

contracts/src/StakeTable.sol
```diff
@@ -8,10 +8,13 @@ import { LightClient } from "../src/LightClient.sol";
 import { EdOnBN254 } from "./libraries/EdOnBn254.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
+// TODO: remove this when contract is made upgradable
```
alxiong, line 11: remove this?

alxiong (approved):

contracts/src/StakeTable.sol
sveitser: https://github.com/EspressoSystems/espresso-sequencer/pull/2647/commits/c9b2501c2ddf0366f7c2b140b20f0c41a4586513

PR #2649, opened 2025-02-20
Fix builder compatibility with Decaf version

rob-maron:
### This PR:

Fixes builder compatibility introduced by removing `vid_precompute_data`. We do this by:
- Adding a `v2` endpoint that supports the new type (without the field)
- Having new nodes hit the `v2` endpoint
- Adding back `vid_precompute` to the old endpoint so that older nodes work with the new builder

### This PR does not:

- Do any type-fu to have one type work for both. This seemed like the best way since the `precompute` field was completely unused; but `bincode` does not support optional fields and deserialization interop with `surf_disco` was proving complex 

This was tested locally and nodes on both the old and new versions are able to build nonempty blocks

### DIFF
--- a/hotshot-builder-api/src/v0_1/builder.rs
+++ b/hotshot-builder-api/src/v0_1/builder.rs
@@ -18,6 +18,7 @@ use tide_disco::{api::ApiError, method::ReadState, Api, RequestError, RequestPar
 use vbs::version::StaticVersionType;
 
 use super::{
+    block_info::AvailableBlockHeaderInputV2,
     data_source::{AcceptsTxnSubmits, BuilderDataSource},
     Version,
 };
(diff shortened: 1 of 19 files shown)


hotshot-types/src/traits/block_contents.rs
```diff
@@ -251,3 +252,19 @@ pub trait BlockHeader<TYPES: NodeType>:
     /// Get the results of the auction for this Header. Only used in post-marketplace versions
     fn get_auction_results(&self) -> Option<TYPES::AuctionResult>;
 }
+
```
QuentinI, line 255: For my understanding: this is forward-ported from previously removed HotShot code, right?

hotshot-builder-api/src/v0_1/builder.rs
```diff
@@ -209,6 +210,29 @@ where
             }
             .boxed()
         })?
+        .get("claim_header_input_v2", |req, state| {
```
QuentinI, line 213: Correct me if I'm wrong, but HotShot doesn't use `vid_commitment` from the V2 struct either, does it? Thus it can be removed, and furthermore, if we stop delegating to `claim_block_header_input` here we can skip computing VID on the builder for V2 altogether and save on latency. Ideally, we could make V2 of `claim_block` that returns all the data hotshot needs to save on network; the endpoints were separated only because of VID computing costs in the first place, so that HotShot proposer didn't have to wait for VID to do things that don't require it (in practice HotShot still waited for both before proceeding, but oh well)

hotshot-types/src/traits/block_contents.rs
rob-maron: Yeah 👍

hotshot-builder-api/src/v0_1/builder.rs
rob-maron: Alright, so @shenkeyao is working on removing the VID computation. After that PR is merged I can come back to this and piggyback the V2 API by removing that too (if a live version is not deployed before then)

QuentinI (approved):

PR #2650, opened 2025-02-20
Fetch Epoch Root via Leaves for Stake Table Catchup

bfish713:
This PR implements Leaf fetching to get the epoch root for each epoch we need for catchup.  

This PR adds to the CatchupDataSource the ability to fetch leafs from others for use in catchup.


sequencer/src/api.rs
```diff
@@ -463,6 +477,48 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> CatchupD
...
+            }
+            if leaf.view_number() == last_leaf.view_number() + 1 {
+                // one away from decide
+                last_leaf = leaf;
+                break;
+            }
+            last_leaf = leaf;
+        }
+        // Make sure we got one more leaf to confirm the decide
+        for leaf in leaves
```
imabdulbasit, line 511: can we not use the first loop to create the chain? the only extra check in the first loop is `if leaf.view_number() == last_leaf.view_number() + 1 ` which I don't understand why is not needed for the 3rd leaf in this loop.

sequencer/src/api.rs
```diff
@@ -1068,6 +1124,15 @@ pub mod test_helpers {
                 })
             }
             .boxed()
+        })?
+        .get("leafchain", |_req, _state| {
+            async move {
+                Result::<Vec<Leaf2>, _>::Err(hotshot_query_service::Error::catch_all(
+                    StatusCode::BAD_REQUEST,
+                    "no block found".to_string(),
```
imabdulbasit, line 1132:
```suggestion
                    "No leafchain found".to_string(),
```

sequencer/src/api/sql.rs
```diff
@@ -159,6 +159,35 @@ impl CatchupStorage for SqlStorage {
...
+            .await
+            .context(format!("leaf {height} not available"))?;
+        let last_leaf: Leaf2 = leaf.leaf().clone().into();
+        let mut chain = vec![last_leaf.clone()];
+        let mut h = height + 1;
+        while let Ok(leaf) = tx.get_leaf((h as usize).into()).await {
+            let leaf: Leaf2 = leaf.leaf().clone().into();
+            chain.push(leaf.clone());
+            // just one away from deciding
+            if leaf.view_number() == last_leaf.view_number() + 1 {
```
imabdulbasit, line 179: The query service only has decided leaves in storage, so wouldn't this check always pass? Also, let's say this fails.

Suppose we insert Leaf (bh = 100) into the chain before this loop. If in this loop, leaf (bh= 104) is inserted, then how is this a valid chain?

imabdulbasit: We can use the `get_leaf_range()` if we want the h, h+1, h+2 chain.

`tx.get_leaf_range(height as usize..=(height as usize + 2)).await`

types/src/v0/traits.rs
```diff
@@ -315,6 +370,21 @@ impl<T: StateCatchup + ?Sized> StateCatchup for Arc<T> {
...
 #[async_trait]
 impl<T: StateCatchup> StateCatchup for Vec<T> {
+    async fn try_fetch_leaves(&self, retry: usize, height: u64) -> anyhow::Result<Vec<Leaf2>> {
+        for provider in self {
+            match provider.try_fetch_leaves(retry, height).await {
+                Ok(leaves) => return Ok(leaves),
+                Err(err) => {
+                    tracing::info!(
+                        provider = provider.name(),
+                        "failed to fetch frontier: {err:#}"
```
imabdulbasit, line 380:
```suggestion
                        "failed to fetch leaves: {err:#}"
```

types/src/v0/traits.rs
```diff
@@ -315,6 +370,21 @@ impl<T: StateCatchup + ?Sized> StateCatchup for Arc<T> {
...
+                    );
+                }
+            }
+        }
+
+        bail!("could not fetch account from any provider");
```
imabdulbasit, line 386:
```suggestion
        bail!("could not fetch leaves from any provider");
```

imabdulbasit (changes requested):
Overall it LGTM

hotshot-types/src/utils.rs
```diff
@@ -98,6 +102,88 @@ pub type StateAndDelta<TYPES> = (
...
+        || parent.justify_qc().view_number() != grand_parent.view_number()
+    {
+        return Err(anyhow!("Leaf views do not chain"));
+    }
+    if newest_leaf.justify_qc().data.leaf_commit != parent.commit()
+        || parent.justify_qc().data().leaf_commit != grand_parent.commit()
+    {
+        return Err(anyhow!("Leaf commits do not chain"));
+    }
+    if parent.view_number() != grand_parent.view_number() + 1 {
```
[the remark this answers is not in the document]
bfish713: It's the way the decide rule for Hotstuff is, you need 2 consecutive QCs then any other chained QC causes the decide, so 1 <- 2 <- 4 decides 1, but 1 <- 3 <- 4 does not decide 1

sequencer/src/api.rs
bfish713: Explained above why we need a different rule

hotshot-types/src/utils.rs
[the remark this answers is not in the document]
imabdulbasit: gotcha

imabdulbasit (approved):

PR #2654, opened 2025-02-21
2652 Membership::set_first_epoch

### LINKED ISSUE #2652: Membership::set_first_epoch

pls148:
Closes #2652

### This PR:

* Adds a fn to Membership called set_first_epoch
* Adds impl of that fn in both hotshot and sequencer side
* Adds testing to StaticCommittee to ensure set_first_epoch is called appropriately

### DIFF
--- a/hotshot-task-impls/src/helpers.rs
+++ b/hotshot-task-impls/src/helpers.rs
@@ -12,7 +12,6 @@ use hotshot_task::dependency::{Dependency, EventDependency};
 use hotshot_types::{
     consensus::OuterConsensus,
     data::{Leaf2, QuorumProposalWrapper, ViewChangeEvidence2},
-    drb::INITIAL_DRB_RESULT,
     event::{Event, EventType, LeafInfo},
     message::{Proposal, UpgradeLock},
     request_response::ProposalRequestPayload,
@@ -40,7 +39,6 @@ use std::{
 };
 use tokio::time::timeout;
 use tracing::instrument;
-use vbs::version::StaticVersionType;
 
 use crate::{events::HotShotEvent, quorum_proposal_recv::ValidationInfo, request::REQUEST_TIMEOUT};
 
@@ -188,6 +186,7 @@ async fn decide_epoch_root<TYPES: NodeType>(
             TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height) + 1);
 
         let write_callback = {
+            tracing::debug!("Calling add_epoch_root for epoch {:?}", next_epoch_number);
             let membership_reader = membership.read().await;
             membership_reader
                 .add_epoch_root(next_epoch_number, decided_leaf.block_header().clone())
@@ -245,7 +244,7 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
 /// # Panics
 /// If the leaf chain contains no decided leaf while reaching a decided view, which should be
 /// impossible.
-pub async fn decide_from_proposal_2<TYPES: NodeType, V: Versions>(
+pub async fn decide_from_proposal_2<TYPES: NodeType>(
     proposal: &QuorumProposalWrapper<TYPES>,
     consensus: OuterConsensus<TYPES>,
     existing_upgrade_cert: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
@@ -487,20 +486,6 @@ pub async fn decide_from_proposal<TYPES: NodeType, V: Versions>(
             decide_epoch_root(&decided_leaf_info.leaf, epoch_height, membership).await;
         } else {
             tracing::info!("No decided leaf while a view has been decided.");
-        }
-    }
-
-    if let (Some(decided_upgrade_cert), Some(decided_leaf_info)) =
-        (&res.decided_upgrade_cert, res.leaf_views.last())
-    {
-        if decided_upgrade_cert.data.new_version == V::Epochs::VERSION {
-            let decided_block_number = decided_leaf_info.leaf.block_header().block_number();
-            let first_epoch_number =
-                TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height));
-            membership
-                .write()
-                .await
-                .set_first_epoch(first_epoch_number, INITIAL_DRB_RESULT);
         }
     }
 
--- a/hotshot-task-impls/src/quorum_vote/handlers.rs
+++ b/hotshot-task-impls/src/quorum_vote/handlers.rs
@@ -13,7 +13,7 @@ use committable::Committable;
 use hotshot_types::{
     consensus::OuterConsensus,
     data::{Leaf2, QuorumProposalWrapper, VidDisperseShare},
-    drb::{compute_drb_result, DrbResult},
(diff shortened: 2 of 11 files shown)


hotshot/src/lib.rs
```diff
@@ -297,6 +298,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
             config.epoch_height,
         );
 
+        if let Some(epoch) = epoch {
+            // #2652 REVIEW NOTE: This epoch can't be right; what if a node starts up between when we upgrade to epochs
+            // and when we're at epoch+2? This would cause the current node to propagate the non-epochs stake table
```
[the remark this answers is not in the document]
ss-es: if we start up with epochs enabled for genesis, the `Membership` trait needs to know that epochs 1 & 2 have no epoch root/drb

I agree we should not use the anchor leaf, but I think we have to specifically do it for genesis unless I'm missing something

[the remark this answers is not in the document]
bfish713: I think this needs to be handled in a general way, if epochs are enabled when we start, we need to load with an epoch root or roots.  I think this should be included in the `HotshotInitializer`,  We should also make sure we can load up again with the stored epoch, like how we do with the anchor leaf.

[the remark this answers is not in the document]
bfish713: I agree we can't just ignore at startup, I was mistaken. I was only considering the mainnet upgrade

hotshot/src/lib.rs
```diff
@@ -243,6 +244,53 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
         .await
     }
 
+    async fn load_start_epoch_info(
```
ss-es, line 247: is there a reason this is a method on `SystemContext`?

it seems like it should either be a method on `Membership` or a stand-alone function

hotshot/src/lib.rs
```diff
@@ -243,6 +244,53 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
...
+        // If the first element in start_epoch_info does not have a block header,
+        // use that to set_first_epoch instead
+        if let Some(epoch_info) = start_epoch_info.first() {
+            if epoch_info.block_header.is_none() {
+                membership
+                    .write()
+                    .await
+                    .set_first_epoch(epoch_info.epoch, INITIAL_DRB_RESULT);
+            }
+        }
```
ss-es, line 260: I would actually suggest just calling `set_first_epoch` in the loop, as part of the `else` on `if let Some(block_header)`

I think doing it the way it's done here is more likely to result in unexpected behaviour, because we can silently discard parts of the initializer. and I don't think there's much to gain from second-guessing the initializer: if it's misconfigured and results in `set_first_epoch` being called twice, it shouldn't be `load_start_epoch_info`'s job to second-guess it

ss-es: I do think we should log by the way, maybe at `tracing::debug!()`, every call to `set_first_epoch` or `add_drb_result`

ss-es (commented):
left some comments. I think notably there aren't any methods storing the initializer info, will that be a separate PR?

otherwise I think it looks fine overall. I'll leave to @bfish713 or @lukaszrzasik to approve (since I accidentally merged the upgrade logic into this branch and I don't want to approve my own code)

hotshot-task-impls/src/helpers.rs
```diff
@@ -489,17 +492,32 @@ pub async fn decide_from_proposal<TYPES: NodeType>(
...
         }
     }
 
+    if let (Some(decided_upgrade_cert), Some(decided_leaf_info)) =
+        (&res.decided_upgrade_cert, res.leaf_views.last())
+    {
+        if decided_upgrade_cert.data.new_version == V::Epochs::VERSION {
+            let decided_block_number = decided_leaf_info.leaf.block_header().block_number();
+            let first_epoch_number =
+                TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height));
```
bfish713, line 512: I thought the plan was to have the block height configured for the first epoch, but his code is getting the epoch from when we decide the upgrade.  Is that change for a different PR or am I misunderstanding/misremembering?

hotshot-task-impls/src/helpers.rs
```diff
@@ -489,17 +492,32 @@ pub async fn decide_from_proposal<TYPES: NodeType>(
         tracing::debug!("Leaf ascension failed; error={e}");
     }
 
-    if with_epochs && res.new_decided_view_number.is_some() {
-        let epoch_height = consensus_reader.epoch_height;
-        drop(consensus_reader);
+    let epoch_height = consensus_reader.epoch_height;
+    drop(consensus_reader);
 
+    if with_epochs && res.new_decided_view_number.is_some() {
```
bfish713, line 498: FYI this function stops being called when we upgrade, decide_from_proposal_2 gets activated, any logic specific to epochs would need to be in there too

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -390,21 +391,44 @@ pub(crate) async fn handle_quorum_proposal_validated<
         .await
     };
 
+    if let Some(cert) = &task_state.staged_epoch_upgrade_certificate {
+        if leaf_views.last().unwrap().leaf.height() >= task_state.epoch_upgrade_block_height {
```
bfish713, line 395: So here we are waiting to apply the upgrade until the configured height, but we already set the first epoch, I think that needs to happen here

bfish713 (changes requested):

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -390,21 +391,55 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+            let epoch_height = task_state.consensus.read().await.epoch_height;
+            let first_epoch_number = TYPES::Epoch::new(epoch_from_block_number(
+                task_state.epoch_upgrade_block_height,
+                epoch_height,
+            ));
+            tracing::debug!("Calling set_first_epoch for epoch {:?}", first_epoch_number);
+            membership
+                .write()
+                .await
+                .set_first_epoch(first_epoch_number, INITIAL_DRB_RESULT);
```
bfish713, line 420: Set this in the block below 
```rust
    if let Some(cert) = decided_upgrade_cert.clone() {
        if cert.data.new_version == V::Epochs::VERSION {
            task_state.staged_epoch_upgrade_certificate = Some(cert);
```

hotshot-task-impls/src/helpers.rs
bfish713: you are correct once we in epochs we don't have to worry about epoch upgrade stuff

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -390,21 +391,56 @@ pub(crate) async fn handle_quorum_proposal_validated<
         .await
     };
 
+    if let Some(cert) = &task_state.staged_epoch_upgrade_certificate {
```
bfish713, line 394: clear this option after we apply the upgrade so we don't hit this on loop

bfish713 (approved):

PR #2655, opened 2025-02-21
Remove VID computation from builder

shenkeyao:
Closes #<ISSUE_NUMBER>

### This PR:
* Removes VID computation from builder code.
* Updates `json` and `bin` files with the new signature due to the removal.

### This PR does not:
* Fix the precompute code.


hotshot-testing/src/block_builder/simple.rs
```diff
@@ -249,15 +248,8 @@ where
         }
 
         // Let new VID scheme ships with Epochs upgrade
```
mrain, line 250: remove this comment?

data/v1/header.json
```diff
@@ -2,8 +2,8 @@
   "block_merkle_tree_root": "MERKLE_COMM~AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAQA",
   "builder_commitment": "BUILDER_COMMITMENT~jlEvJoHPETCSwXF6UKcD22zOjfoHGuyVFTVkP_BNc-no",
   "builder_signature": {
-    "r": "0xa1c3795850b7b490e616b60fead89753841fbc9fffe1a939d483f1d959ad1c45",
-    "s": "0x20228f5b63b14792d371dce479978e45020f19602189ef6d325b73029a2848ac",
+    "r": "0xa66e49fb371b83d36a6f326b95eb6146c397f3b037fd04b5f3c973d0e5624105",
+    "s": "0x43c6d12317c58dd8f831d34e2fa64bba04f51374ef7fce826121ba5dc5df64cf",
```
rob-maron, line 6: Are these changes for the reference header tests? If so, does it break backwards compatibility with the current version?

hotshot-builder-core/src/lib.rs
```diff
@@ -31,45 +30,3 @@ pub enum WaitAndKeep<T> {
...
-            WaitAndKeepGetError::FailedToResolvedVidCommitmentFromChannel => {
-                BuildError::Error("failed to resolve VidCommitment from channel".to_string())
-            }
-        }
-    }
-}
-
-impl<T: Clone> WaitAndKeep<T> {
-    /// get will return a clone of the value that is already stored within the
-    /// value of `WaitAndKeep::Keep` if the value is already resolved.  Otherwise
```
rob-maron, line 52: Should we remove the entire `WaitAndKeep` structure if it no longer has an implementation? Or does it get used somewhere else?

shenkeyao: Yeah, it looks like `BlockInfo` still uses it!

data/v1/header.json
shenkeyao: Yeah, they are for the reference header tests. Not sure about the backward compatibility though. Asked this on Zulip: https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/new.20VID.20.3C.3E.20builder.20interface/near/501616156.

rob-maron (approved):

PR #2662, opened 2025-02-21
WIP: draft for delegation events

sveitser:
This isn't useful at the moment. I'm trying to work out the best storage layout and event schema for adding delegations to the smart contract.

cc @tbro @alysiahuggins @bfish713

### DIFF
--- a/contracts/src/StakeTable.sol
+++ b/contracts/src/StakeTable.sol
@@ -63,6 +63,9 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
     // Error raised when the staker does not provide a new blsVK
     error InvalidBlsVK();
 
+    // Pool fee is invalid
+    error InvalidFee();
+
     // Error raised when zero point keys are provided
     error NoKeyChange();
 
@@ -80,6 +83,12 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
 
     /// Mapping from a hash of a BLS key to a node struct defined in the abstract contract.
     mapping(address account => Node node) public nodes;
+
+    // TODO book-keeping for pools
+    // Need to keep track of total amount deposited.
+    // Need to keep track of amounts marked for withdrawal.
+    mapping(address validator => mapping (address delegator => uint256 amount)) public poolBalances;
+    mapping(address delegator => RequestedWithdrawal) public poolWithdrawals;
 
     /// Total stake locked;
     uint256 public totalStake;
@@ -350,7 +359,7 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
 
         nodes[msg.sender] = node;
 
-        emit Registered(msg.sender, registrationEpoch, amount);
+        emit Registered(node);
     }
 
     /// @notice Deposit more stakes to registered keys
@@ -514,6 +523,82 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
 
         emit UpdatedConsensusKeys(msg.sender, node.blsVK, node.schnorrVK);
     }
+
+    // Delegation functions
+
+    // @notice creates a pool and configure the initial fee
+    // TODO: docs
+    function createPool(
+        BN254.G2Point memory blsVK,
+        EdOnBN254.EdOnBN254Point memory schnorrVK,
+        BN254.G1Point memory blsSig,
+        uint8 fee
+    ) public {
+        if (fee > 100) {
+            revert InvalidFee();
+        }
+
+        Node memory node = nodes[msg.sender];
+
+        if (node.account != address(0x0)) {
+            revert NodeAlreadyRegistered();
+        }
+
+        if (
+            _isEqualBlsKey(
+                blsVK,
+                BN254.G2Point(
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0),
+                    BN254.BaseField.wrap(0)
+                )
+            )
+        ) {
+            revert InvalidBlsVK();
+        }
+
+        // Verify that the validator can sign for that blsVK by ensuring that the message that has
+        // been signed is the sender's address
+        // This is to prevent "rogue public-key attack"
+        bytes memory message = abi.encode(msg.sender);
+        BLSSig.verifyBlsSig(message, blsSig, blsVK);
+
+        if (schnorrVK.isEqual(EdOnBN254.EdOnBN254Point(0, 0))) {
+            revert InvalidSchnorrVK();
+        }
+
+        node.account = msg.sender;
+        node.blsVK = blsVK;
+        node.schnorrVK = schnorrVK;
+        node.registerEpoch = registrationEpoch;
+
+        nodes[msg.sender] = node;
+
+        // TODO: what events to emit?
+        // TODO: figure out bookkeeping for pools vs normal validators
+        // emit Registered(node);
+        emit PoolRegistered(node, fee);
+    }
+
+    // TODO, this is a skeleton
+    function depositPool(address validatorAddress, uint256 amount) external {
+        // TODO check that pool exists and is live
+        SafeTransferLib.safeTransferFrom(ERC20(tokenAddress), msg.sender, address(this), amount);
+        emit PoolDeposit(validatorAddress, msg.sender, amount);
+    }
+
+    // TODO, this is a skeleton
+    function requestPoolWithdrawal(address validatorAddress, uint256 amount) external {
+        // TODO check exit queue (if we decide to keep that)
+        // TODO check balance
+        // TODO mark funds withdrawable at the right epoch
+        emit PoolWithdrawalRequested(validatorAddress, msg.sender, amount, exitEpoch);
+    }
+
+    // TODO handle pool withdrawal, ideally the same way we handle other withdrwals
+
+    /// Admin functions
 
     /// @notice Update the min stake amount
     /// @dev The min stake amount cannot be set to zero
--- a/contract-bindings-alloy/src/lib.rs
+++ b/contract-bindings-alloy/src/lib.rs
@@ -12,3 +12,4 @@ pub mod lightclientmock;
 pub mod permissionedstaketable;
 pub mod plonkverifier;
 pub mod plonkverifier2;
+pub mod staketable;
--- a/contract-bindings-ethers/src/lib.rs
+++ b/contract-bindings-ethers/src/lib.rs
@@ -13,3 +13,4 @@ pub mod permissioned_stake_table;
 pub mod plonk_verifier;
 pub mod plonk_verifier_2;
 pub mod shared_types;
+pub mod stake_table;
--- a/contract-bindings-ethers/src/permissioned_stake_table.rs
+++ b/contract-bindings-ethers/src/permissioned_stake_table.rs
@@ -10,6 +10,7 @@ pub use permissioned_stake_table::*;
     non_camel_case_types
 )]
 pub mod permissioned_stake_table {
+    pub use super::super::shared_types::*;
     #[allow(deprecated)]
     fn __abi() -> ::ethers::core::abi::Abi {
         ::ethers::core::abi::ethabi::Contract {
@@ -1297,42 +1298,6 @@ pub mod permissioned_stake_table {
         Hash,
     )]
     pub struct OwnerReturn(pub ::ethers::core::types::Address);
-    ///`G2Point(uint256,uint256,uint256,uint256)`
-    #[derive(
-        Clone,
-        ::ethers::contract::EthAbiType,
-        ::ethers::contract::EthAbiCodec,
-        serde::Serialize,
-        serde::Deserialize,
-        Default,
-        Debug,
-        PartialEq,
-        Eq,
-        Hash,
-    )]
-    pub struct G2Point {
-        pub x_0: ::ethers::core::types::U256,
-        pub x_1: ::ethers::core::types::U256,
-        pub y_0: ::ethers::core::types::U256,
-        pub y_1: ::ethers::core::types::U256,
-    }
-    ///`EdOnBN254Point(uint256,uint256)`
-    #[derive(
-        Clone,
-        ::ethers::contract::EthAbiType,
-        ::ethers::contract::EthAbiCodec,
-        serde::Serialize,
-        serde::Deserialize,
-        Default,
-        Debug,
-        PartialEq,
-        Eq,
-        Hash,
-    )]
-    pub struct EdOnBN254Point {
-        pub x: ::ethers::core::types::U256,
-        pub y: ::ethers::core::types::U256,
-    }
     ///`NodeInfo((uint256,uint256,uint256,uint256),(uint256,uint256),bool)`
     #[derive(
         Clone,
--- a/contract-bindings-ethers/src/shared_types.rs
+++ b/contract-bindings-ethers/src/shared_types.rs
@@ -15,6 +15,42 @@ pub struct G1Point {
     pub x: ::ethers::core::types::U256,
     pub y: ::ethers::core::types::U256,
 }
+///`G2Point(uint256,uint256,uint256,uint256)`
+#[derive(
+    Clone,
+    ::ethers::contract::EthAbiType,
+    ::ethers::contract::EthAbiCodec,
+    serde::Serialize,
+    serde::Deserialize,
+    Default,
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+)]
+pub struct G2Point {
+    pub x_0: ::ethers::core::types::U256,
+    pub x_1: ::ethers::core::types::U256,
+    pub y_0: ::ethers::core::types::U256,
+    pub y_1: ::ethers::core::types::U256,
+}
+///`EdOnBN254Point(uint256,uint256)`
+#[derive(
+    Clone,
+    ::ethers::contract::EthAbiType,
+    ::ethers::contract::EthAbiCodec,
+    serde::Serialize,
+    serde::Deserialize,
+    Default,
+    Debug,
+    PartialEq,
+    Eq,
+    Hash,
+)]
+pub struct EdOnBN254Point {
+    pub x: ::ethers::core::types::U256,
+    pub y: ::ethers::core::types::U256,
+}
 ///`PlonkProof((uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),(uint256,uint256),uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256)`
 #[derive(
     Clone,
--- a/contracts/src/interfaces/AbstractStakeTable.sol
+++ b/contracts/src/interfaces/AbstractStakeTable.sol
@@ -23,21 +23,32 @@ abstract contract AbstractStakeTable {
     /// (LastEpochStart)
     uint256 public totalVotingStake;
 
-    /// @notice Signals a registration of a BLS public key.
-    /// @param account the address of the validator
-    /// @param registerEpoch epoch when the registration becomes effective.
-    /// @param amountDeposited amount deposited when registering the new node.
-    event Registered(address account, uint64 registerEpoch, uint256 amountDeposited);
+    /// @notice Signals a registration of a new Validator.
+    event Registered(Node node);
+
+    // TODO docs
+    event PoolRegistered(Node node, uint8 fee);
 
     /// @notice Signals an exit request has been granted.
     /// @param account the address of the validator
     /// @param exitEpoch epoch when the user will be allowed to withdraw its funds.
+    // TODO amount should be customizable
     event Exit(address account, uint64 exitEpoch);
 
     /// @notice Signals a deposit to a BLS public key.
     /// @param account the address of the validator
     /// @param amount amount of the deposit
     event Deposit(address account, uint256 amount);
+
+    // TODO docs
+    /// @notice Delegator deposits into pool.
+    event PoolDeposit(address validatorAddress, address delegatorAddress, uint256 amount);
+
+    // TODO docs
+    /// @notice Delegator requests withdrawal from a pool.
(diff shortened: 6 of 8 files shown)


contracts/src/StakeTable.sol
```diff
@@ -515,6 +524,82 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
         emit UpdatedConsensusKeys(msg.sender, node.blsVK, node.schnorrVK);
     }
 
+    // Delegation functions
+
+    // @notice creates a pool and configure the initial fee
+    // TODO: docs
+    function createPool(
+        BN254.G2Point memory blsVK,
```
alysiahuggins, line 532: curious if hotshot is aware of the ethereum address of a validator and if so if it should be added here (since the address is meant to be the unique identifier and the bls key can change)

sveitser: It's currently assumed that it's `msg.sender`. I expect the confirmation layer to get it from the emitted event. Do you think we should do something different?

contracts/src/EspToken.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /// @notice This contract is called by the proxy when you deploy this contract
+    function initialize(address _owner, address _initialGrantRecipient) public initializer {
+        __Ownable_init(_owner);
+        __UUPSUpgradeable_init();
+        _mint(_initialGrantRecipient, 1000_000_000 ether);
+    }
+
+    /// @notice only the owner can authorize an upgrade
+    function _authorizeUpgrade(address newImplementation) internal override onlyOwner {
```
alysiahuggins, line 34: it could be interesting to add some validation checks here to ensure that the new implementation is valid e.g. a check that the version on the new implementation is greater than the current version

contracts/src/StakeTable.sol
```diff
@@ -149,396 +163,208 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-    /// @dev The function will revert if the sender does not have the correct balance.
-    /// @dev The function will revert if the sender does not have the correct BLS signature.
-    /// `blsSig` field is necessary to prevent "rogue public-key attack".
-    /// The signature is over the caller address of the function to ensure that each message is
-    /// unique.
     /// @dev No validity check on `schnorrVK`, as it's assumed to be sender's responsibility,
-    /// the contract only treat it as auxiliary info submitted by `blsVK`.
-    /// @dev The function will revert if the sender does not have the correct registration epoch.
-    function register(
+    function registerValidator(
```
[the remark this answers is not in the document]
sveitser: Everything is bound by L1 throughput and gas cost. So I think no unless we somehow process these events in such a manner that we cannot keep up with L1 events then no.

contracts/src/EspToken.sol
sveitser: Actually it sounds like we need to do more here and actually check msg.sender?

https://docs.openzeppelin.com/contracts/5.x/api/proxy#UUPSUpgradeable-_authorizeUpgrade-address-

sveitser: ah no it has the `onlyOwner` modifier, maybe adding a test wouldn't be a bad idea then.

sveitser: On a related note. UUPSUpgradable already emits an `Upgraded(address newImpl)` event: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fda6b85f2c65d146b86d513a604554d15abd6679/contracts/proxy/ERC1967/ERC1967Utils.sol#L69

staking-cli/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+To load the stake table for the demo, run the demo first
+
+    just demo-native
+
+and once that is started up
+
+    cargo run --bin staking-cli -p staking-cli -- stake-for-demo
+
+it should that validators are being registered
```
tbro, line 17:
```suggestion
it should show that validators are being registered
```

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: that's fair @ block size

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
 
-        nodes[msg.sender].balance += amount;
-        SafeTransferLib.safeTransferFrom(ERC20(tokenAddress), msg.sender, address(this), amount);
-
-        emit Deposit(msg.sender, uint256(amount));
+        validators[validator].delegatedAmount += amount;
+        delegations[validator][msg.sender] += amount;
 
-        uint64 effectiveEpoch = _currentEpoch + 1;
+        SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);
```
alysiahuggins, line 349: very low chance that the token contract could be a malicious contract - but if it were, then it could re-enter the delegation method here and increase the amount delegated to itself without transferring the tokens over. It has to be a sophisticated attack though because they'd have to update the token address to a malicious contract, perform the re-entrancy attack and then update the token to the correct token address. 

To be extra safe, here are potential solutions:
-  we could add the Openzeppelin modifier `nonReentrant` 
- transfer the tokens first before updating storage (since for the attacker it's more beneficial to have delegations double counted) 
- or have another way to verify that only value token addresses would ever would used

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-        Node memory node = nodes[msg.sender];
+    /// @notice Undelegate from a validator
+    /// @param validator The validator to undelegate from
+    /// @param amount The amount to undelegate
+    function undelegate(address validator, uint256 amount) external virtual {
+        ensureValidatorActive(validator);
 
-        if (node.account == address(0)) {
-            revert NodeNotRegistered();
+        if (validators[msg.sender].status == ValidatorStatus.Exited) {
```
alysiahuggins, line 360: hi hi, why can't a user undelegate? when a validator exits - are the delegated amounts already sent to the delegatos? if not, they should be able to undelegate at any point? based on my current understanding of the flow

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
+            revert ValidatorAlreadyExited();
         }
 
-        if (node.account != msg.sender) {
-            revert Unauthenticated();
+        uint256 balance = delegations[validator][msg.sender];
```
alysiahuggins, line 364: leaving this comment to remind myself to check whether there's a way for a delegator to know how many tokens they delegated

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-        }
+        emit Withdrawal(msg.sender, amount);
+    }
 
-        if (node.account != msg.sender) {
-            revert Unauthenticated();
+    /// @notice Withdraw previously delegated funds after a validator has exited
+    /// @param validator The validator to withdraw from
+    function claimValidatorExit(address validator) external virtual {
+        uint256 unlocksAt = validatorExits[msg.sender];
```
alysiahuggins, line 400: `validatorExits` is only updated when the `deregisterValidator()` function is called so does that mean this function is only to be called by validators? if so, it's not immediately obvious with the naming

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,356 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-        return balance;
+        emit Withdrawal(msg.sender, amount);
     }
 
     /// @notice Update the consensus keys for a validator
     /// @dev This function is used to update the consensus keys for a validator
-    /// @dev This function can only be called by the validator itself when it's not in the exit
-    /// queue
+    /// @dev This function can only be called by the validator itself when it hasn't exited
+    ///      TODO: MA: is this a good idea? Why should key rotation be blocked for an exiting
```
alysiahuggins, line 425: hmm, why would a validator rotate their keys if they are exiting and no longer intend to vote in consensus - unless they lost their keys and that's why they're exiting? but then it's either you exit or you update your keys? any thoughts on this @rob-maron

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,356 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
+    ///      validator?
     /// @dev The validator will need to give up either its old BLS key and/or old Schnorr key
     /// @dev The validator will need to provide a BLS signature to prove that the account owns the
     /// new BLS key
     /// @param newBlsVK The new BLS verification key
     /// @param newSchnorrVK The new Schnorr verification key
     /// @param newBlsSig The BLS signature that the account owns the new BLS key
+    ///
+    /// TODO: MA: I think this function should be reworked. Is it fine to always force updating both
```
alysiahuggins, line 434: i believe the key generation tool generates both a schnorr and blskey at the same time - so you might as well update both? but i could see reasons for not doing so
cc @rob-maron

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,356 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-
-        // Verify that the validator can sign for that newBlsVK, otherwise it inner reverts with
-        // BLSSigVerificationFailed
+        // Verify that the validator can sign for that blsVK. This prevents rogue public-key
+        // attacks.
         bytes memory message = abi.encode(msg.sender);
         BLSSig.verifyBlsSig(message, newBlsSig, newBlsVK);
 
-        node.blsVK = newBlsVK;
+        blsKeys[_hashBlsKey(newBlsVK)] = true;
```
alysiahuggins, line 453: wondering how are bls keys currently associated with ethereum accounts in the contract? i believe there should be an association as delegations are based on the validator ethereum account address. 

Maybe this is handled in the confirmation layer?

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
+    }
 
-    /// Mapping from a hash of a BLS key to a node struct defined in the abstract contract.
-    mapping(address account => Node node) public nodes;
+    /// @notice The status of a validator.
+    ///
+    /// By default a validator is in the `Unknown` state. This means it has never registered. Upon
+    /// registration the status will become `Active` and if the validator deregisters its status
+    /// becomes `Exited`.
+    enum ValidatorStatus {
```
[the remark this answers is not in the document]
sveitser: I prefer it this way with the more general concepts first.

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
+    /// @param validator The validator to delegate to
+    /// @param amount The amount to delegate
+    function delegate(address validator, uint256 amount) external virtual {
+        ensureValidatorActive(validator);
 
-        // A node cannot deposit more tokens if an exit request is in progress.
-        if (node.exitEpoch != 0) {
-            revert ExitRequestInProgress();
+        uint256 allowance = token.allowance(msg.sender, address(this));
+        if (allowance < amount) {
```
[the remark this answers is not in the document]
sveitser: You mean not allowing to delegate a zero amount?

contracts/src/StakeTable.sol
sveitser: I thought about this. Should have left a comment. Yes, we are calling our own contract so I think it's mostly an unnecessary complication of the contract if we defend against a malicious token contract. The token contract address isn't modifiable and if an attacker was able to do that then I think it's likely game over for all the funds anyway.

I can move the transfer before the state change, sure.

@0xkato any opinions on this?

contracts/src/StakeTable.sol
sveitser: In was thinking in theory they could discover a key leak after registering and exit and want to rotate their keys. However I realize now that wouldn't take effect until they exited the stake table anyway so that key change would be useless.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
alysiahuggins: sorry yes

contracts/src/StakeTable.sol
sveitser: Via the event

```
    event ValidatorRegistered(
        address indexed account,
        BN254.G2Point blsVk,
        EdOnBN254.EdOnBN254Point schnorrVk,
        uint16 commission
    );
```

the contract doesn't currently associate them in the contract storage.

contracts/src/StakeTable.sol
sveitser: this is indeed wrong, fixed in https://github.com/EspressoSystems/espresso-network/pull/2662/commits/fc3b31c9ff8c5d6616d816f6b6fd6176ba076331

contracts/src/StakeTable.sol
sveitser: there's no way to do so directly from from the state of the contract, but it's possible via events

contracts/src/StakeTable.sol
sveitser: The validator exit will undelegate all the funds of the delegators.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
sveitser: sure we can add that

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-    /// @param blsVK BLS verification key in G2
-    /// @return keccak256(blsVK)
-    function _hashBlsKey(BN254.G2Point memory blsVK) public pure returns (bytes32) {
-        return keccak256(abi.encode(blsVK.x0, blsVK.x1, blsVK.y0, blsVK.y1));
+    function initializeState(
+        address _tokenAddress,
+        address _lightClientAddress,
+        uint256 _exitEscrowPeriod
+    ) internal {
+        // TODO ensure address not zero
```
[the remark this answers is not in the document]
sveitser: https://github.com/EspressoSystems/espresso-sequencer/commit/5f22187ca1d0bc43259cf954c6c376db8f913a4e

contracts/src/StakeTable.sol
0xkato: Its a good call out! Sounds like it’s a non-issue. It would only really become a concern in cases where we allow adding new tokens permissionlessly, if our token were upgradable, or if we were using ERC777 or things like this none of which I believe apply here.

I’m not the biggest supporter of “just slapping a nonReentrant on functions.” moving the transfer before the state change should be fine.

contracts/src/StakeTable.sol
alysiahuggins: i think this is a good feature for analytics and also so that a user can determine how much they want to `undelegate` - @veilgets thoughts on adding the ability for a delegator to know the total number of tokens delegated? it's currently not an option

contracts/src/StakeTable.sol
alysiahuggins: ok i understand that now, thanks!

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,352 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
+    // === Structs ===
 
-    // Error raised when the hotShotBlocksPerEpoch is zero
-    error InvalidHotShotBlocksPerEpoch();
+    /// @notice Represents an Espresso validator and tracks funds currently delegated to them.
+    ///
+    /// @notice The `delegatedAmount` excludes funds that are currently marked for withdrawal via
+    /// undelegation or validator exit.
+    struct Validator {
+        ValidatorStatus status;
```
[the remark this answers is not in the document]
sveitser: https://github.com/EspressoSystems/espresso-sequencer/commit/b6db36bbc

contracts/src/StakeTable.sol
sveitser: So the register and key update events tell the confirmation layer what consensus keys to use for which Ethereum account. The users don't really have to worry about the keys of validators although we will have APIs for this in the confirmation layer so that User interfaces can show all the information. The contract is kind of minimal now in terms of what it stores and computes, which the exception of checking that keys are new and the BLS signature check but we might move all of this to the confirmation layer too.

For slashing we will ensure that the address to slash can be authenticated via merkle proof against the LC state root.

Does this answer your concerns?

contracts/src/StakeTable.sol
alysiahuggins: yea i also don't love slapping `nonReentrant` but it can also be avoided if we put the transfer before the storage updates. ps, the token is upgradeable. Another scenario this can be an issue is if the admin of this contract (a multisig) is somehow compromised - but i agree they would do other bad things. 

But what do you think about changing the order to this. @sveitser @0xkato 

```solidity
SafeTransferLib.safeTransferFrom(token, msg.sender, address(this), amount);

validators[validator].delegatedAmount += amount;
delegations[validator][msg.sender] += amount;
```

(it's not the recommended CEI pattern but it makes sense in this case i think)

contracts/src/StakeTable.sol
```diff
@@ -52,493 +102,356 @@ contract StakeTable is AbstractStakeTable, Ownable, InitializedAt {
...
-        return (nodes[msg.sender].balance, effectiveEpoch);
+        emit Delegated(msg.sender, validator, amount);
     }
 
-    /// @notice Request to exit from the stake table, not immediately withdrawable!
-    ///
-    /// @dev TODO modify this according to the current spec
-    function requestExit() external virtual override {
-        Node memory node = nodes[msg.sender];
+    /// @notice Undelegate from a validator
```
alysiahuggins, line 354:
```suggestion
    /// @notice Undelegate from a validator
    /// @dev if the validator is in the Exited state, the delegator can attempt to withdraw their delegation via the `claimValidatorExit(...)` function
    
```

contracts/src/StakeTable.sol
alysiahuggins: gotcha - and in hotshot they would be keeping track of it so even though the bls key changed, they know it's the same ethereum accoun that owns it? 

that's fair since the user would never care about bls keys

contracts/src/EspToken.sol
sveitser: I thought the purpose of this function was mainly to check if someone has the right to upgrade but not if the upgrade is "good". I think we need to test our upgrades well but I'm not sure this function can do much for us in this regard.

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
sveitser: tracked in #2785

contracts/src/StakeTable.sol
alysiahuggins: it does! thank yout

contracts/src/StakeTable.sol
sveitser: I'm fine with that. I think most importantly a comment why we're not so worried since it's our own contract may put some readers at ease.

contracts/src/StakeTable.sol
sveitser: If we need it for on-chain voting it's a different story but that's a separate discussion.

staking-cli/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The CLI is currently intended for testing/demo purposes only.
+
+To load the stake table for the demo, run the demo first
+
+    just demo-native
+
+and once that is started up
+
+    cargo run --bin staking-cli -p staking-cli -- stake-for-demo
```
tbro, line 15:
```suggestion
    RUST_LOG=info cargo run --bin staking-cli -p staking-cli -- stake-for-demo
```

staking-cli/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+it should show that validators are being registered
+
+    2025-03-14T16:10:06.635922Z  INFO staking_cli::demo: Deploying validator 0 with commission 0.00 %
+    2025-03-14T16:10:10.665231Z  INFO staking_cli::demo: Deploying validator 1 with commission 1.00 %
+    2025-03-14T16:10:14.692189Z  INFO staking_cli::demo: Deploying validator 2 with commission 2.00 %
+    2025-03-14T16:10:18.720833Z  INFO staking_cli::demo: Deploying validator 3 with commission 3.00 %
+    2025-03-14T16:10:22.560015Z  INFO staking_cli::demo: Deploying validator 4 with commission 4.00 %
```
tbro, line 23: Can we add how to verify in the logs of the long running services? Maybe we need to update our event processing before this is possible.

tbro (approved):
Approving in spite of a few comments. I think this is in good shape and leave it to the author to address the commented items here or in follow ups. Personally I place higher value on  unblocking subsequent work. We can address the small issues here as we go.

staking-cli/README.md
sveitser: Can you elaborate. I don't understand. Our long running services will only consume the new stake table events after #2772 is merged.

contracts/src/StakeTable.sol
sveitser: @0xkato you mention 

> if our token were upgradable

it actually will be initially. But what can we really do? Or what do we defend against? Let's say the token is taken over. Do we care, or can we even protect the tokens that are staked in the smart contract? If the attacker can modify the storage of the token contract they can mint and move tokens at will. The token balance of the stake table contract is just a field in the balances mapping of the token contract.

0xkato: Hmm, why is the token contract upgradable?

A reentrancy guard is the better option then. Sure, if our contract gets hijacked, there’s not much defense anyway. But since it’s open source, we should limit the footguns others need to watch out for.

contracts/src/StakeTable.sol
alysiahuggins: we're actually having a conversation today about on-chain voting so may need to come back to this and create a github issue if necessary

alysiahuggins (approved):
thanks a lot for the great work @sveitser! A lot of work covered here and we can address the comments in the newly created github issues.

PR #2670, opened 2025-02-24
Broadcast eQC

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* Adds a new message `ExtendedQc` which includes both the current epoch quorum certificate and the next epoch quorum certificate. This message is broadcast to everyone.
* Adds corresponding two new events: `ExtendedQcRecv` and `ExtendedQcSend`
* Modifies the `send_high_qc` function so that if the current high QC is an eQC, the new `ExtendedQc` is broadcast to everyone.
* Upon receiving `ExtendedQc`, a node makes sure that:
* * it really is an eQC,
* * the received current and next epoch high QCs actually correspond to each other,
* * signatures on both QCs are valid
* Adds a helper function `validate_qc_and_next_epoch_qc` which is used for the above checks and is also used in `handle_quorum_proposal_recv`
* if the received `ExtendedQc` is valid, moves the node to the new epoch
* Moves the `get_next_epoch_qc` function to the helpers file to be easily reused.
* Adds an additional check: if a node just formed a current epoch eQC, it makes sure that it also has the corresponding next epoch eQC.

### DIFF
--- a/hotshot-task-impls/src/helpers.rs
+++ b/hotshot-task-impls/src/helpers.rs
@@ -9,14 +9,13 @@ use async_lock::RwLock;
 use committable::{Commitment, Committable};
 use either::Either;
 use hotshot_task::dependency::{Dependency, EventDependency};
-use hotshot_types::simple_certificate::NextEpochQuorumCertificate2;
 use hotshot_types::{
     consensus::OuterConsensus,
     data::{Leaf2, QuorumProposalWrapper, ViewChangeEvidence2},
     event::{Event, EventType, LeafInfo},
     message::{Proposal, UpgradeLock},
     request_response::ProposalRequestPayload,
-    simple_certificate::{QuorumCertificate2, UpgradeCertificate},
+    simple_certificate::{NextEpochQuorumCertificate2, QuorumCertificate2, UpgradeCertificate},
     simple_vote::HasEpoch,
     traits::{
         block_contents::BlockHeader,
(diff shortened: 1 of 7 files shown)


hotshot-task-impls/src/helpers.rs
```diff
@@ -863,3 +865,122 @@ pub async fn broadcast_event<E: Clone + std::fmt::Debug>(event: E, sender: &Send
...
+
+            warn!("Invalid certificate: {}", e)
+        })?;
+    }
+
+    if let Some(next_epoch_qc) = maybe_next_epoch_qc {
+        // If the next epoch qc exists, make sure it's equal to the qc
+        if qc.view_number() != next_epoch_qc.view_number()
+            || qc.data.epoch != next_epoch_qc.data.epoch
+            || qc.data.leaf_commit != next_epoch_qc.data.leaf_commit
```
ss-es, line 963: can this just be `qc.data != next_epoch_qc.data`? (we can derive `Eq` if needed)

hotshot-task-impls/src/helpers.rs
```diff
@@ -863,3 +865,122 @@ pub async fn broadcast_event<E: Clone + std::fmt::Debug>(event: E, sender: &Send
         }
     }
 }
+
+/// Gets the next epoch QC corresponding to this epoch QC, times out if it takes too long.
```
ss-es, line 869: I'm a bit confused from the description here

we aren't actually getting anything right? the code seems like it's just waiting for the event where we form the QC with a timeout, but based on the description + name I was assuming it would be fetching something from other nodes (which I guess obviously doesn't make sense)

can we clarify the comment + function name if so? e.g. `wait_for_next_epoch_qc` or similar

ss-es (approved):
I think this looks good to me, just some minor comments

PR #2673, opened 2025-02-24
Remove VID precomputation from builder and WaitAndKeep

shenkeyao:
Closes #<ISSUE_NUMBER>

### This PR:
* Removes VID precomputation from builder.
* Removes `WaitAndKeep` as it's no longer used.


mrain (approved):
LGTM

PR #2677, opened 2025-02-25
refinery: 0.8.15 -> 0.8.16

sveitser:
Undoes changes from f20af066e225fa2a65ed578cd2a4bbc609f8f690

- 0.8.15 of refinery was semver breaking
- 0.8.16 undoes the semver breaking change.
- 0.9 hasn't been released and we don't know when it will come.

For new projects using us a dependency resolve to 0.8.16 with 0.8 specified in Cargo.toml so I think this is all we need.

Also updates Cargo.lock files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.62"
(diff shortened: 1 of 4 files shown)


imabdulbasit (approved):
thank you!

PR #2686, opened 2025-02-26
Next epoch payload commit

### LINKED ISSUE #2669: Add next epoch payload commitment to DAC
Currently next epoch VID shares contain the previous epoch payload commitment so it can be compared with the DAC.
This approach doesn't guarantee that the next epoch VID share actually corresponds to the DAC.
Change it so that the DAC contains the next epoch payload commitment so that the next epoch VID share can be compared to it.

lukaszrzasik:
Closes #2669

### This PR:

* DA nodes calculate a payload commitment for the current and next epoch when they receive DA proposal
* DA nodes attach both payload commitments to DA vote and so DAC includes both commits as well.
* Changes logic how the vote dependency task handles payload commitment comparison. Quorum proposal's payload commit is always compared to DAC's current epoch payload commit. VID's commit is compared to the current or next epoch commit accordingly, depending on which epoch the VID recipient belongs to.
* Removes an obsolete `data_epoch_payload_commitment` field from VID

### This PR does not:

* Optimize the next epoch payload calculation
* Change the `append_da` methods to store both commitments.


bfish713 (commented):
This makes sense to me, one thing we also want to probably handle is for validators in both committees, to verify that they got both shares that's a bit trickier and we can handle it in a follow up

bfish713 (approved):

PR #2688, opened 2025-02-26
Implement request/response protocol

rob-maron:
Closes https://github.com/EspressoSystems/HotShot/issues/3871

### This PR:

Finishes the plumbing for the request/response protocol for use in the confirmation layer. It does this by making use of the preexisting `external` messaging functionality so that the application layer can both send and receive messages.

As a reminder, this protocol was introduced in https://github.com/EspressoSystems/HotShot/pull/4071

### This PR does not:

Make use of the protocol. These will be introduced in subsequent PRs. I decided not to do a `NetworkConfig` example, since the protocol has a dependency on HotShot (HotShot is not initialized when we need the config)

### Key places to review:

- `sequencer/src/context.rs` for protocol defaults (I think these are sane, but others would know more)
- All the `RollCall` removal - cc @Ayiga you previously said we didn't need this, and it doesn't look like it breaks anything

I tested that this works and each node is able to get a response on startup. I didn't include this since it felt unnecessary in the PR. Perhaps we can add better testing, but unsure if it would be easy to do with the existing integration tests


sequencer/src/external_event_handler.rs
```diff
@@ -3,94 +3,67 @@
...
+#[allow(dead_code)]
 pub enum OutboundMessage {
-    Direct(Vec<u8>, PubKey),
-    Broadcast(Vec<u8>),
+    Direct(MessageKind<SeqTypes>, PubKey),
+    Broadcast(MessageKind<SeqTypes>),
 }
 
 impl<V: Versions> ExternalEventHandler<V> {
     /// Creates a new `ExternalEventHandler` with the given network and roll call info
```
sveitser, line 44:
```suggestion
    /// Creates a new `ExternalEventHandler` with the given network
```

sequencer/src/external_event_handler.rs
```diff
@@ -105,73 +78,63 @@ impl<V: Versions> ExternalEventHandler<V> {
...
+                    // Wrap it in the real message type
+                    let message_inner = Message {
+                        sender: public_key,
+                        kind: message,
+                    };
+
+                    // Serialize it
+                    let Ok(message_bytes) = hotshot_upgrade_lock.serialize(&message_inner).await
+                    else {
+                        tracing::error!("Failed to serialize direct message");
```
sveitser, line 111: It looks like we are swallowing the error here but logging an error to indicate there's something really wrong. I think if this is always benign we probably shouldn't log an error and if it isn't we should log the error.

sequencer/src/external_event_handler.rs
```diff
@@ -105,73 +78,63 @@ impl<V: Versions> ExternalEventHandler<V> {
...
+                    // Wrap it in the real message type
+                    let message_inner = Message {
+                        sender: public_key,
+                        kind: message,
+                    };
+
+                    // Serialize it
+                    let Ok(message_bytes) = hotshot_upgrade_lock.serialize(&message_inner).await
+                    else {
+                        tracing::error!("Failed to serialize broadcast message");
```
sveitser, line 131: Similarly here.

request-response/src/recipient_source.rs
```diff
@@ -7,7 +7,7 @@ use super::request::Request;
 /// expect responses from. In `HotShot` this would go on top of the [`Membership`] trait and determine
 /// which nodes are able (quorum/DA) to respond to which requests
 #[async_trait]
-pub trait RecipientSource<K: SignatureKey + 'static>: Send + Sync + 'static {
+pub trait RecipientSource<R: Request, K: SignatureKey + 'static>: Send + Sync + 'static {
     /// Get all the recipients that the specific request should expect responses from
-    async fn get_recipients_for<R: Request>(&self, request: &R) -> Vec<K>;
+    async fn get_recipients_for(&self, request: &R) -> Vec<K>;
```
sveitser, line 12: Based on the doc comment how about `expected_responders`? I think this labels the role more explicitly because there are recipients of requests, but also recipients of responses. I know the `Request` type are there so it's still deducible but anyway.

sequencer/src/request_response/request.rs
```diff
@@ -0,0 +1,94 @@
...
+        // Match the type of the response and request
+        match (self, request) {
+            (Response::Example, Request::Example) => Ok(()),
+        }
+    }
+}
+
+/// Implement the `Serializable` trait for the `Request` type. This tells the request response
+/// protocol how to serialize and deserialize the request
+impl Serializable for Request {
```
sveitser, line 50: Out of curiosity may I ask why we have our own `Serialize` trait? I see that we use the first byte to label something as a request or a response but wouldn't that be taken care of if we directly serialized the `Message` enum?

sveitser (approved):

sveitser:
LGTM. Just some questions.

sequencer/src/external_event_handler.rs
sveitser: I just mean logging the error message, not returning it.

sequencer/src/request_response/request.rs
rob-maron: We can definitely just derive w/ serde. I did this initially as a reaction to seeing breaking API changes on `main` and thought doing this explicitly may also ease the upgrade paths. I have no real preference though

sequencer/src/external_event_handler.rs
rob-maron: Ah okay, yeah in that case I will log it. I also think that it should probably be a warning

sequencer/src/request_response/request.rs
sveitser: If it will make things easier let's keep it.

PR #2697, opened 2025-02-27
vacuum sqlite after pruning data

imabdulbasit:


rob-maron (approved):

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -731,6 +731,13 @@ impl PruneStorage for SqlStorage {
             }
         }
 
+        #[cfg(feature = "embedded-db")]
+        {
+            let mut conn = self.pool().acquire().await?;
+            query("VACUUM").execute(conn.as_mut()).await?;
+            conn.close().await?;
```
imabdulbasit, line 738: think this get closed after drop but I am not sure

PR #2700, opened 2025-02-28
Pull in nightly rust-fmt, use vid rustfmt.toml

pls148:
Old HS repo used to have nightly rustfmt. After some discussion in the monorepo channel, I figured we'd try nightly rustfmt to get better grouping of dependencies. This gets rid of the mess of redundant "use hotshot_types::a::b::some thing".

Tested locally with nix/direnv. Not sure if I got other peoples' use cases.

The main changes are to the following files:
- flake.nix
- rust-toolchain.toml
- rustfmt.toml

The rest of the changes were done by rustfmt picking up the new config.

### DIFF
--- a/builder/src/bin/permissionless-builder.rs
+++ b/builder/src/bin/permissionless-builder.rs
@@ -122,11 +122,11 @@ async fn main() -> anyhow::Result<()> {
     match (base, upgrade) {
         (FeeVersion::VERSION, MarketplaceVersion::VERSION) => {
             run::<SequencerVersions<FeeVersion, MarketplaceVersion>>(genesis, opt).await
-        }
+        },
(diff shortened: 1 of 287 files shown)


sveitser (approved):
Looks great thanks.

For next time, making a commit first with the changes and a separate commit for the formatting makes it easier to review because the reviewer can check out the first commit and run `just fmt` then diff against the branch HEAD. I did this manually now which (still) seemed a lot easier than reading at the diff.

One question, hotshot had `group_imports = "StdExternalCrate"` which we don't have here. I think that's quite nice. Any particular reason for excluding it?

sveitser:
@pls148 this commit reverted changes in flake.nix from #2736 . I wonder how that happened.

PR #2702, opened 2025-02-28
Use PeerConfig for stake table committees

imabdulbasit:
This PR:
- changes StakeTableEntry type to PeerConfig
- changes return type of membership get functions to return Result

### DIFF
--- a/hotshot-task-impls/src/helpers.rs
+++ b/hotshot-task-impls/src/helpers.rs
@@ -29,6 +29,7 @@ use hotshot_types::{
         option_epoch_from_block_number, Terminator, View, ViewInner,
     },
     vote::{Certificate, HasViewNumber},
+    StakeTableEntries,
 };
 use hotshot_utils::anytrace::*;
 use std::{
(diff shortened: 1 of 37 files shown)


hotshot-task-impls/src/helpers.rs
```diff
@@ -133,10 +133,27 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+        })?;
     drop(membership_reader);
 
+    let membership_stake_table = membership_stake_table
+        .into_iter()
+        .map(|config| config.stake_table_entry)
+        .collect::<Vec<_>>();
+
     justify_qc
         .is_valid_cert(
```
imabdulbasit, line 158: might be better to change this function signature but I'll leave it for now

hotshot-types/src/traits/election.rs
```diff
@@ -31,56 +29,64 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
     fn stake_table(
         &self,
         epoch: Option<TYPES::Epoch>,
-    ) -> Vec<<TYPES::SignatureKey as SignatureKey>::StakeTableEntry>;
+    ) -> Result<Vec<PeerConfig<TYPES::SignatureKey>>, Self::Error>;
```
mrain, line 32: Is it necessary to wrap these returns in `Result`? All the impls seem infallible and the uses simply do something like `unwrap_or_default()`.

imabdulbasit: This was part of tech-debt that we discussed on zulip few weeks ago. Basically, we wanted this to return Option or Result. I agree that this doesn't do much because of  unwrap_or_default() but that requires more work.

mrain (approved):
LGTM

PR #2705, opened 2025-03-03
Validators in two epochs should verify that they got both VID shares

### LINKED ISSUE #2692: Validators in two epochs should verify that they got both VID shares
Validators in both committees should verify that they got both VID shares.
Additionally, they should store both VID shares in shared consensus state and in a persistent storage.

lukaszrzasik:
Closes #2692

### This PR:

* Changes `type VidShares<TYPES>` to include an optional epoch. This is needed to accommodate the fact a node can have two VID shares in a given view: one for the current epoch and one for the next epoch.
* Adds a new requirement while voting for the last block in the epoch: if a node belongs to both epochs, it waits until it receives VID shares for both epochs. There's also a check that makes sure that VID commits correspond to DAC commits.
* the VID response task now sends two VID shares if the requester belongs to both epochs and it's the last block in the epoch
* the decide event includes the next epoch VID share if it's the last block in the epoch and the node belongs to the next epoch
* Stores the next epoch VID share if it's the last block in the epoch and the node belongs to the next epoch

### This PR does not:

~~Optimize VID response. Both VID shares are only strictly required for the last block in the epoch. Currently VID response sends both VID shares whenever the request sender belongs to both epochs.~~
This has been optimized. VID responder sends the next epoch VID share only if it's the last block in the epoch.

### DIFF
--- a/crates/hotshot/task-impls/src/quorum_vote/mod.rs
+++ b/crates/hotshot/task-impls/src/quorum_vote/mod.rs
@@ -246,7 +246,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
 
             let Ok(current_epoch_membership) = self
                 .membership_coordinator
-                .membership_for_epoch(current_epoch)
+                .stake_table_for_epoch(current_epoch)
                 .await
             else {
                 tracing::warn!("Couldn't acquire current epoch membership. Do not vote!");
@@ -254,7 +254,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
             };
             let Ok(next_epoch_membership) = self
                 .membership_coordinator
-                .membership_for_epoch(next_epoch)
+                .stake_table_for_epoch(next_epoch)
                 .await
             else {
                 tracing::warn!("Couldn't acquire next epoch membership. Do not vote!");
--- a/crates/hotshot/types/src/consensus.rs
+++ b/crates/hotshot/types/src/consensus.rs
@@ -643,7 +643,7 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         let is_epoch_transition = is_epoch_transition(parent_leaf.height(), self.epoch_height);
         let target_epoch = if is_epoch_transition
             && membership
-                .membership_for_epoch(next_epoch)
+                .stake_table_for_epoch(next_epoch)
                 .await
                 .ok()?
                 .has_stake(public_key)


hotshot-task-impls/src/quorum_vote/mod.rs
```diff
@@ -232,6 +241,52 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
+                leaf.with_epoch,
+                leaf.block_header().block_number(),
+                self.epoch_height,
+            );
+
+            let membership_reader = self.membership_coordinator.membership().read().await;
+            let committee_member_in_current_epoch =
+                membership_reader.has_stake(&self.public_key, current_epoch);
+            let committee_member_in_next_epoch =
+                membership_reader.has_stake(&self.public_key, current_epoch.map(|e| e + 1));
```
bfish713, line 266: use `self.membership_coordinator.membership_for_epoch(epoch)`

hotshot-task-impls/src/helpers.rs
```diff
@@ -996,3 +1005,107 @@ pub async fn validate_qc_and_next_epoch_qc<TYPES: NodeType, V: Versions>(
...
+    timeout: u64,
+    view_start_time: Instant,
+    receiver: &Receiver<Arc<HotShotEvent<TYPES>>>,
+) -> Result<Proposal<TYPES, VidDisperseShare<TYPES>>> {
+    // If the VID share that we already have is for the current epoch, get the next epoch VID.
+    // And the other way around.
+    let target_epoch = if vid_share.data.epoch() == vid_share.data.target_epoch() {
+        vid_share.data.target_epoch().map(|e| e + 1)
+    } else {
+        vid_share.data.target_epoch().map(|e| e - 1)
```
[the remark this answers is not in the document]
bfish713: I feel like this logic isn't necessary, at this point we know that we need a VID share for both of two epochs, A and B.  If the share is from A we need B, if the share is for B we need A.  Can't this function just take the epoch we need a share for?

hotshot-task-impls/src/helpers.rs
```diff
@@ -996,3 +1005,107 @@ pub async fn validate_qc_and_next_epoch_qc<TYPES: NodeType, V: Versions>(
...
+                return Ok(second_vid_share);
+            }
+        }
+        return Err(warn!("Error while waiting for the second VID share."));
+    };
+    let HotShotEvent::VidShareValidated(second_vid_share) = event.as_ref() else {
+        // this shouldn't happen
+        return Err(warn!("Received event is not VidShareValidated but we checked it earlier. Shouldn't be possible."));
+    };
+    Ok(second_vid_share.clone())
```
bfish713, line 1110: This kind of logic is why `EventDependency` exists.  Checking shared state over and over is not a good pattern.  I think we should rewrite this to just wait for a VidShareValidated event where the epoch number matches what we want

ss-es (commented):
I didn't review the full PR thoroughly yet, but had a question

mrain (approved):

crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -216,6 +224,80 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
+                .membership_coordinator
+                .membership_for_epoch(current_epoch)
+                .await
+            else {
+                tracing::warn!("Couldn't acquire current epoch membership. Do not vote!");
+                return;
+            };
+            let Ok(next_epoch_membership) = self
+                .membership_coordinator
+                .membership_for_epoch(next_epoch)
```
ss-es, line 257: I think this should be `stake_table_for_epoch`

crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -216,6 +224,80 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
+            let current_epoch = option_epoch_from_block_number::<TYPES>(
+                leaf.with_epoch,
+                leaf.block_header().block_number(),
+                self.epoch_height,
+            );
+            let next_epoch = current_epoch.map(|e| e + 1);
+
+            let Ok(current_epoch_membership) = self
+                .membership_coordinator
+                .membership_for_epoch(current_epoch)
```
ss-es, line 249: this too

unless we need the leader for the epoch, we should use `stake_table_for_epoch` rather than `membership_for_epoch`

crates/hotshot/types/src/consensus.rs
```diff
@@ -619,23 +623,41 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
             .get(&leaf.justify_qc().data().leaf_commit)?;
         let parent_state_and_delta = self.state_and_delta(parent_view_number);
         let (Some(state), delta) = parent_state_and_delta else {
             return None;
         };
+
+        let is_epoch_transition = is_epoch_transition(parent_leaf.height(), self.epoch_height);
+        let target_epoch = if is_epoch_transition
+            && membership
+                .membership_for_epoch(next_epoch)
```
ss-es, line 646: here too

PR #2706, opened 2025-03-03
feat: Epoch change to Light Client contract and prover service

### LINKED ISSUE #2684: [Light Client] Update light client contract to support epoch
Epoch logic is removed from light client contract in #1939 
Now the contract sticks with the genesis stake table.
We should change the existing API or add a new API to update the stake table when the epoch advances.

alxiong:
Closes #2684

### This PR:

Apologize for such a large PR, but many components are so interconnected that I have to fix them all to even compile, let alone pass all the tests. 

locally just `just sol-test` and `cargo test -p hotshot-state-prover` should show all tests passing. 

This PR is **best reviewed commit by commit** -- I have tried my best to make each one self-contained and limited in scope. 

### This PR does not:

Fetch the actual Stake table for the next block from Sequencer's URL, which is handled in #2702

in future follow-up, We need to utilize this API return to parse out the input to `newFinalizedState()`

### Key places to review:

Contracts: `LightClient.sol`, `PlonkVerifier.sol`, `deployer.rs`
Prover: `cirucit.rs`, `service.rs`


mrain (approved):
LGTM with a minor comment

contracts/src/LightClient.sol
```diff
@@ -38,17 +38,27 @@ contract LightClient is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
 
     // === System Parameters ===
     //
+    /// @notice number of blocks per epoch
+    uint64 public BLOCKS_PER_EPOCH;
+
     // === Storage ===
     //
     /// @notice genesis stake commitment
     StakeTableState public genesisStakeTableState;
```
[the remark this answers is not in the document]
sveitser: I think it's because we can't remove any members due to this being an upgradable contract: https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable#modifying-your-contracts

@alxiong from that PoV shouldn't BLOCKS_PER_EPOCH be added at the end?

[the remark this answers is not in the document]
alxiong: yah, that's a good point. maybe I need to relocate this.
I should have add a "fork test" to make sure we can upgrade existing LC contract to this one.

sveitser (changes requested):
Just putting request for changes here based on the discussion so we don't accidentally merge early because there's already an approval.

contracts/src/LightClient.sol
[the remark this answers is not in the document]
sveitser: I think we definitely need some kind of test to show that the contract still works as expected after upgrading.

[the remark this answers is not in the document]
alxiong: > I decide to use a child contract to contain all the diff instead

While this is definitely clearer and desirable, I realized that introducing a `LightClientV2.sol` would require a copy of `PlonkVerifier2.sol` and `PolynomialEval2.sol` with minimal changes, thus lots of code duplication.

I think for the time being, I'd rather keep it in the same file, rename the file to `LightClientV2.sol` to indicate the version instead of having a separate such file. Retarget all tests to this and add a forge test to ensure successful upgrade from current light client contract. 

cc @sveitser @alysiahuggins

alxiong:
I didn't plan the change well as I forgot the upgradability of LightClient. This is a costly oversight that waste lots of engineering energy.

Here's the engineering plan for myself: 

- Introduce V2 for `LightClient.sol`, `PlonkVerifier.sol` and `PolynomialEval.sol` (the last two will be turned into contracts from libraries, and some functions marked as `virtual` for future inheritance and override) 
- Revert changes made to V1 versions of contracts, and their solidity tests
- Add fork test on the upgradability from LCV1 to LCV2
- Test all existing V1 contract's test on V2 versions to ensure functional equivalence
- inside `deployer.rs`, incorporate code for `upgrade.rs`

contracts/src/LightClient.sol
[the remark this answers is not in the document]
alxiong: done, please see the fork test in  [4131205](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/4131205b3902f0ceb9f55a5eb6cf3c8dc4326656)

alxiong:
This PR is currently being blocked by `forge bind` bug that wouldn't produce rust bindings, (without it, I can't further modify our `deployer.rs` or rather `upgrade.rs`) and also the prover service code.

Meanwhile, the contract code contains all the necessary changes (and tests pass locally), thus ready for review. Here's a summary of **contract related changes**.

- Introduce `PolynomialEvalV2.sol` and `PlonkVerifierV2.sol` that changes public inputs from 7 to 11. Both contracts are now `contract` instead of `library` since we don't have much saving from making them as library, and linking library makes our deployment code more complicated.
  - tests are now renamed to `test/xxV2.t.sol`, i.e. only testing the latest version. since the old version are already deployed and can always be digged out from git history
- Introduce `LightClientV2.sol` that extends `LightClient.sol` and add epoch-related logic
  -  some functions are moved over from `StakeTable.sol`
  - similarly, tests are only in `test/LightClientV2.t.sol`. 
    - in `init()`, `LCV2` are deployed in a way that reflects how we would deploy it in practice, 
    - I rewrote the tests for HotShotCommitment history and `lagOver` liveness detection in [851048d](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/851048d47c15e9414a8adb933ec7df322f630155). These tests should be testing the functionality of these function rather than specific to our implementations. (cc @alysiahuggins) These tests also don't have to generate bunch of proofs (each taking ~5 sec), instead we mock-set state, thus the total testing time cuts down from a few minutes to 30 sec. 
    - added a fork tests to ensure our existing deployment on sepolia can be upgraded to our V2 in [4131205](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/4131205b3902f0ceb9f55a5eb6cf3c8dc4326656)
- Introduce `LCArbitrumV2.sol` and retarget its test.

## Key places to review

- Old test logic but modified: `contract LightClient_LivenessDetectionTest`, `contract LightClient_StateHistoryTest`
- Epoch-aware `newFinalizedState()` in `LightClientV2.sol` including its `verifyProof()` choosing the right public inputs, and how we update the stake table commitments at the last block in every epoch.
- The way we upgrade to LCV2.

cc @philippecamacho @sveitser @alysiahuggins

contracts/src/LightClientV2.sol
```diff
@@ -0,0 +1,198 @@
...
+        IPlonkVerifier.PlonkProof memory proof
+    ) external virtual {
+        //revert if we're in permissionedProver mode and the permissioned prover has not been set
+        if (isPermissionedProverEnabled() && msg.sender != permissionedProver) {
+            revert ProverNotPermissioned();
+        }
+
+        if (
+            newState.viewNum <= finalizedState.viewNum
+                || newState.blockHeight <= finalizedState.blockHeight
```
alxiong, line 82: > After epoch upgrade, current transition (we'll have a new transition logic soon) will produce three consecutive views with the same block height.

pointed out by @mrain, I will fix this later in this PR

ss-es: just to note that this is the current logic on main, but this will change very soon with #2853. after that merges (and I believe we are treating it as urgent), hotshot will no longer emit decides for the same block in multiple views

alxiong: oh i see. so we don't need to change contract logic then. Thanks for pointing out!

contracts/lib/forge-std
sveitser, line 1: This change looks unexpected. Is this an accidental revert of the submodule?

alxiong: i'm not sure, i remember when I sync with main and resolving conflict, i just pick the one from `main`. possibly due to the `just gen-bindings` which will git pull the submodules, and we are pointing to `master` in the `.gitmodule` file?

alxiong: I searched `forge-std`, yes I accidentally revert to an older commit. I will revert back. Thanks for catching this.

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -0,0 +1,207 @@
...
+//! thus treated by the rust type systems as distinct types.
+//!
+//! Another usage is in the differential testing in Solidity tests. In those cases, the actual types don't matter, since they will all `abi_encode()`
+//! into the exact same bytes before being communicated over to contract via FFI. Thus using any one of them is fine.
+
+use alloy::sol;
+
+/// # What to re-export, what to hide?
+/// - export contract struct itself, but try to avoid export instance type (instead, use ::new() to get a handle)
+/// - avoid exporting `xxCall` and `xxReturn` types, they usually can be converted/transmuted from existing struct
```
sveitser, line 21: It would be better to say when types can be safely transmuted. If we didn't want to use transmute for this, how much extra code would it be?

contracts/src/LightClientV2.sol
```diff
@@ -0,0 +1,208 @@
...
+import { PlonkVerifierV2 as PV } from "./libraries/PlonkVerifierV2.sol";
+import { LightClientStateUpdateVKV2 as VkLib } from "./libraries/LightClientStateUpdateVKV2.sol";
+
+/// @notice LightClient V2: with stake table snapshot update during epoch change.
+contract LightClientV2 is LightClient {
+    /// @notice When entering a new epoch and a new stake table snapshot.
+    event NewEpoch(uint64 epoch);
+
+    /// @notice PlonkVerifier contract address
+    address public _verifier;
```
sveitser, line 17: @alxiong can you remind why we are making the verifier a contract and not an external library?

contracts/rust/adapter/src/sol_types.rs
alxiong: it's all just a matter of placating the type system, type conversion is needed. Of course everything can be down without `transmute`, but the code will be extremely verbose. you have to map field-by-field between 4 different types of `BN254::G1Point`.

contracts/src/LightClientV2.sol
alxiong: just to reduce contract size. we don't inline the whole `verify()` logic into `LightClientV2`. 
Is this okay? do you still prefer the old way?

sveitser: External libraries are also separately deployed, and linked by injecting the address to jump to in the caller contract bytecode. So the argument about contract size IMO does not apply. I think this is a change without any benefit and therefore worse than keeping it the same but let me know if you disagree.

contracts/lib/forge-std
alxiong: done in [ccd9e62](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/ccd9e6246e640c32b753e3e4b7c91fde4b591123)

contracts/src/LightClientV2.sol
alxiong: okay, even if the contract size argument doesn't apply, avoiding "injecting the linked address in caller's bytecode" is a minor plus. 
as doing so requires:

https://github.com/EspressoSystems/espresso-network/blob/ccd9e6246e640c32b753e3e4b7c91fde4b591123/utils/src/deployer.rs#L217-L227

okay, so are we reverting them back to:

- `PlonkVerifierV2` as `library`
- linked `PV2` to `LightClientV2`?

alxiong: done in [1350b70](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/1350b70696c6f3625d5859897e7217dce8d08c94)

alxiong:
Let's divide and conquer on reviewing this PR. 

- Contracts: @alysiahuggins, @sveitser changes were described before: https://github.com/EspressoSystems/espresso-network/pull/2706#issuecomment-2716254210
- alloy migration: @rob-maron primarily look at `alloy-compat` crate, changes are in #2848 (feel free to leave review there as well since it has slightly restricted diff)
- state signer and relay server: @mrain (who have reviewed #2903, but pls skim through the later changes I made to pass the tests in that PR)
- updated state prover circuit: @mrain 
- `deployer.rs` and rust deployment refactor, and `espresso-dev-node.rs`: @jbearer, @sveitser 
- everything under `types/`: @rob-maron @imabdulbasit (see if I have anything that violates your expectation or incompatible with epoch logic

Thanks everyone! 🙏 (again, sry for the monstrous PR)

hotshot-stake-table/src/mt_based.rs
```diff
@@ -9,10 +9,10 @@
 mod config;
 mod internal;
 
+use alloy::primitives::{U256, U512};
```
mrain, line 12: nit: I feel like we should deprecate this module.

hotshot-stake-table/src/utils.rs
```diff
@@ -20,7 +20,12 @@ pub trait ToFields<F: Field> {
...
 pub(crate) fn u256_to_field<F: PrimeField>(v: &U256) -> F {
-    let mut bytes = vec![0u8; 32];
-    v.to_little_endian(&mut bytes);
+    let bytes: [u8; 32] = v.to_le_bytes();
     F::from_le_bytes_mod_order(&bytes)
 }
+
+#[inline]
+/// A helper function to compute the quorum threshold given a total amount of stake.
+pub fn one_honest_threshold(total_stake: U256) -> U256 {
```
mrain, line 29: Are we supposed to use the quorum threshold fn provided by membership?

hotshot-state-prover/src/service.rs
```diff
@@ -382,19 +354,34 @@ pub async fn sync_state<ApiVer: StaticVersionType>(
     tracing::debug!("Old state: {old_state:?}");
     tracing::debug!("New state: {:?}", bundle.state);
 
+    let blocks_per_epoch = config
```
mrain, line 357: Is the `epoch_height` changing? If not we should initialize it only once.
cc @bfish713 @ss-es @jbearer

hotshot-state-prover/src/service.rs
```diff
@@ -382,19 +354,34 @@ pub async fn sync_state<ApiVer: StaticVersionType>(
...
+        .blocks_per_epoch()
+        .await
+        .map_err(ProverError::NetworkError)?;
+    let next_stake = if is_last_block(bundle.state.block_height as u64, blocks_per_epoch) {
```
mrain, line 361: nit: The last block of an epoch should be epoch root. Because, only epoch root is persisted in the consensus and the stake table update is bundled with it. 
I can also fix this in my later PR to fetch epoch root information from the hotshot query service.

hotshot-state-prover/src/service.rs
```diff
@@ -466,6 +454,7 @@ pub async fn run_prover_service<ApiVer: StaticVersionType + 'static>(
     let stake_table_capacity = config.stake_table_capacity;
     tracing::info!("Stake table capacity: {}", stake_table_capacity);
     // TODO(#1022): maintain the following stake table
```
mrain, line 456: time to remove this comment

sequencer/src/state_signature.rs
```diff
@@ -84,13 +104,39 @@ impl<ApiVer: StaticVersionType> StateSigner<ApiVer> {
...
         {
             Ok(state) => {
-                let signature = self.sign_new_state(&state).await;
                 tracing::debug!("New leaves decided. Latest block height: {}", leaf.height(),);
 
+                let consensus = consensus_state.read().await;
+                let cur_block_height = state.block_height;
+                let blocks_per_epoch = consensus.epoch_height;
+
+                let next_stake_table = if is_last_block(cur_block_height, blocks_per_epoch) {
```
mrain, line 113: nit: we shouldn't vote after epoch root.
I can fix it in my later PR.

mrain (commented):
state-prover and relay-server generally LGTM.
Left some minor comments.

hotshot-stake-table/src/mt_based.rs
alxiong: sure, maybe in future PR.

hotshot-stake-table/src/utils.rs
alxiong: can you point me to it? I move this from some utils.rs i remember.

hotshot-state-prover/src/service.rs
alxiong: yup, i asked the exact question on zulip, and it sounds like this won't change and will only be used as genesis setup. iiuc.

hotshot-stake-table/src/utils.rs
mrain: Like here: https://github.com/EspressoSystems/espresso-network/blob/861a30e83490f2f1fa8a951c5275a738a06764a8/hotshot-types/src/traits/election.rs#L126

We can also leave it as a tech debt and discuss later.

hotshot-state-prover/src/service.rs
alxiong: so you want to put it inside `StateProverConfig`? that's the only place to persist. 
imo, it's not a big deal to query it on demand. shouldn't be too costly. but yeah, open to suggestions here.

hotshot-state-prover/src/service.rs
mrain: issue: https://github.com/EspressoSystems/espresso-network/issues/2924

hotshot-stake-table/src/utils.rs
alxiong: i took a look, so `success_threshold()` here means `2/3f+1`, but here we only need a single honest signature, which is a different threshold. 

In future PRs, we could add an additional method in the `trait Membership` ? but definitely future work.

alxiong:
Regarding "slow test: sqlite" failed, while postgres counterpart always succeed. So i compared logs, and found that all 19 tests are passed in both runs. The only difference is that `sqlite` version we have a deadlock (I assume, which is why it leads to timeout) in `fn slow_dev_node_multiple_lc_providers_test()` which I'm debugging now.

alxiong:
Frustratingly, there's more bugs and failure. even previously `test-demo` is running non-stop (i have to kill it) 
It seems that `prover-service` is not up while the rest of the services are running normally. Investigating why.

alxiong:
So locally, i tested the "slow-test-sqlite" that timeout on [271c022](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/271c02247f8defabcaecfcf5edea6c72a6025d01), tests are actually passing in 26 min (sans the build time)

```
  TRY 1 SLOW [>120.000s] sequencer api::test::test_fee_upgrade_view_based
  TRY 1 SLOW [>120.000s] staking-cli::cli test_cli_approve
Summary [1596.239s] 488 tests run: 488 passed (3 slow), 34 skipped
```

not entirely sure why it ran for 40min+ on CI, but i think the crazy amount of logs are definitely slowing it down as bandwidth is limited on CI machines. so I'd turn them off. honestly, with a 20min run of logs, it's quite hard to debug from github's limited search tool IMO. i think the better workflow is just noticing the tests that failed, reproduce locally with log turns on for relevant crates only (e.g if I'm testing state prover, then I'm not gonna worry about hotshot's internal voting), and then fix it.
^^ folks could disagree with this tho (in which case I can revert)

utils/src/deployer.rs
```diff
@@ -1,46 +1,39 @@
...
-use ethers_conv::{ToAlloy as _, ToEthers};
-use futures::future::{BoxFuture, FutureExt};
-use hotshot_contract_adapter::light_client::{
-    LightClientConstructorArgs, ParsedLightClientState, ParsedStakeTableState,
-};
-use url::Url;
+use hotshot_contract_adapter::sol_types::*;
+
+// We pass this during `forge bind --libraries` as a placeholder for the actual deployed library address
+const LIBRARY_PLACEHOLDER_ADDRESS: &str = "ffffffffffffffffffffffffffffffffffffffff";
```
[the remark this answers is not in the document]
alxiong: you mean we accidentally have a value that equals this value and mistakenly swapped out? 
how is a random value help?

sveitser:
@alxiong I think the deployer code looks good but please open an issue regarding follow up work to track all of that. For the LC deployment we need to add some code to deploy the V2 and the peform the upgrade. For mainnet we need to make it possible to deploy the v2 without running upgradeToAndCall (which will go through the multisig).

contracts/src/LightClientV2.sol
```diff
@@ -0,0 +1,205 @@
...
+            revert OutdatedState();
+        }
+        // format validity check
+        BN254.validateScalarField(newState.blockCommRoot);
+
+        // epoch-related checks
+        uint64 lastUpdateEpoch = currentEpoch();
+        uint64 newEpoch = epochFromBlockNumber(newState.blockHeight, _blocksPerEpoch);
+        if (
+            newEpoch == lastUpdateEpoch + 1 && !isLastBlockInEpoch(finalizedState.blockHeight)
```
sveitser, line 90: I think this is a bit cryptic now. I feel like this could be  more readable with separating the case where `lastUpdateEpoch == 0` and then adding some require statements, but not sure.

.github/workflows/slowtest.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         run: just nextest --features embedded-db --no-run
 
       - name: Run slow test
-        run: just nextest --features embedded-db --no-fail-fast --nocapture
+        run: RUST_LOG=off just nextest --features embedded-db --no-fail-fast --nocapture
```
sveitser, line 50: Remove `--nocapture` instead?

contracts/lib/openzeppelin-foundry-upgrades
sveitser, line 1: This shows lines were removed from the changelog. Is this an accidental revert?

contracts/lib/solidity-bytes-utils
sveitser, line 1: same  here

contracts/lib/solmate
sveitser, line 1: this as well

imabdulbasit (approved):
the `types` crate changes LGTM

contracts/src/LightClientV2.sol
alxiong: hmm.. i thought about this when I wrote the code. the decision needs to take into consideration of the next `if` as well, if we want `lastUpdateEpoch==0` as a separate case, then most likely we end up with `if else if else else` 3 clauses at least, and I find it harder to read than the current 2 `if` flatten out.

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -0,0 +1,207 @@
+//! Solidity types for interacting with contracts
+//! Re-export types that are used, sometimes renamed to avoid collision.
+//!
+//! NOTE: Due to <https://github.com/foundry-rs/foundry/issues/10153>,
```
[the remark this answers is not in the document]
alxiong: done in [09f5d73](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/09f5d73835a99389c5bb67612c5a26840db74605)

rob-maron (approved):
`alloy-compat` changes look good to me

utils/src/deployer.rs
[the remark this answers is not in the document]
sveitser: I mean the chances that a randomly generated value ends up in the contract bytecode is much lower than the chances that 0xffff... is in the contract bytecode.

I think I have a better idea. 

We can use `replacen(..., /*count*/ 1)` then I think we will get an error if there are multiple matches (an we replaced only one). What I want to avoid is weird to debug errors where for some reason a contract doesn't weird stuff.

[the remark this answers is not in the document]
alxiong: i think there are two occurences of the placeholder

.github/workflows/slowtest.yaml
alxiong: done in [ac3644d](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/ac3644d9740673668f46034164f9d77162e18b53)

utils/src/deployer.rs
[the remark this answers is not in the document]
alxiong: done in [ac3644d](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/ac3644d9740673668f46034164f9d77162e18b53)

alxiong (commented):
anyone feel free to merge this PR after @sveitser gives his approval.

contracts/test/LightClient.t.sol
sveitser, line 1: @alxiong we delete a lot of tests here. Can you argue briefly why these tests are not needed, or where the equivalent tests are now?

contracts/src/LightClientArbitrumV2.sol
sveitser, line 1: Do we currently have any code to deploy this upgrade? If not, please add it to the follow up work.

contracts/test/LightClient.t.sol
alxiong: i didn't really "delete" them, they are now only testing against the latest `PlonkVeriferV2.sol` now, and i rewrote tests for `lagsOver` and `getCommitment` history tests as they are overly complicated and less readable before. 


see here: https://github.com/EspressoSystems/espresso-network/pull/2706#issuecomment-2716254210

other than the rewrite, i keep the rest of the tests.

contracts/src/LightClientArbitrumV2.sol
alxiong: it should be exactly the same as upgrading the LightClient.sol

process-compose.yaml
```diff
@@ -104,6 +102,9 @@ processes:
 
   state-relay-server:
     command: state-relay-server
+    environment:
```
sveitser, line 105: just curious, is this needed? for process compose all env vars should be passed automatically

utils/src/deployer.rs
```diff
@@ -332,9 +333,11 @@ pub async fn upgrade_light_client_v2(
...
                 LightClientV2::BYTECODE.encode_hex()
             };
-            let lcv2_linked_bytecode = Bytes::from_hex(
-                target_lcv2_bytecode.replace(LIBRARY_PLACEHOLDER_ADDRESS, &pv2_addr.encode_hex()),
-            )?;
+            let lcv2_linked_bytecode = Bytes::from_hex(target_lcv2_bytecode.replacen(
+                LIBRARY_PLACEHOLDER_ADDRESS,
+                &pv2_addr.encode_hex(),
+                1,
+            ))?;
```
sveitser, line 340: I guess to actually make this error we need to check if the string is still there after the first replacement.

sveitser (approved):
Thanks @alxiong. Next time multiple PRs >.<

utils/src/deployer.rs
alxiong: you are right, i'll fix that

process-compose.yaml
alxiong: frankly i don't know, so `.env` are all automatically loaded, if so, i can remove them

utils/src/deployer.rs
alxiong: done in [dae9d5c](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/dae9d5cf3d9ec241696693e70f2e10e8fd0f02b9)

process-compose.yaml
alxiong: done in [dae9d5c](https://github.com/EspressoSystems/espresso-network/pull/2706/commits/dae9d5cf3d9ec241696693e70f2e10e8fd0f02b9)

PR #2709, opened 2025-03-04
move public config structs to types crate

imabdulbasit:
This PR moves the public HotShot config structs to the types crate. This is useful for other crates, such as the client or prover when calling config endpoint


alxiong (approved):
LGTM. as long as this is purely a cherry-pick from another PR that already got review from system team.

PR #2716, opened 2025-03-05
Prevent overwriting existing payload with empty Payload 

imabdulbasit:
details here https://www.notion.so/espressosys/Payload-Inconsistency-in-Epoch-Dev-Deploy-1ac2431b68e980bdaaf5eb06324347ea

### DIFF
--- a/hotshot-query-service/src/data_source/storage/sql/transaction.rs
+++ b/hotshot-query-service/src/data_source/storage/sql/transaction.rs
@@ -499,7 +499,7 @@ where
         let query = query("INSERT INTO payload (height) VALUES ($1) ON CONFLICT DO NOTHING")
             .bind(height as i64);
         query.execute(self.as_mut()).await?;
-        
+
         // Finally, we insert the leaf itself, which references the header row we created.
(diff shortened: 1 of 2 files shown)


hotshot-query-service/src/data_source/storage/sql/transaction.rs
```diff
@@ -496,8 +496,9 @@ where
         // Similarly, we can initialize the payload table with a null payload, which can help us
         // distinguish between blocks that haven't been produced yet and blocks we haven't received
         // yet when answering queries.
-        self.upsert("payload", ["height"], ["height"], [(height as i64,)])
-            .await?;
+        let query = query("INSERT INTO payload (height) VALUES ($1) ON CONFLICT DO NOTHING")
```
sveitser, line 499: I feel like this needs a comment. Ideally also a regression test that fails on `main` and passes with this change.

imabdulbasit: the issue comes up only in deployment for PoS, so unfortunately, its not easy to add a test

PR #2728, opened 2025-03-06
Fix decide logic in response to audit feedback

ss-es:
Fixes an issue identified during an audit where we might skip an `add_epoch_root` call depending on where in the chain a leaf was decided


jparr721 (approved):
Thanks! cc @0xkato

PR #2732, opened 2025-03-07
fix genesis vid for Proof of stake version

imabdulbasit:


mrain (approved):
LGTM

hotshot-query-service/src/data_source/update.rs
```diff
@@ -168,19 +171,45 @@ fn genesis_vid<Types: NodeType>(
...
+
+            ensure!(
+                calculated_commit == commit,
+                "computed VID commit {} for genesis block does not match header commit {}",
+                calculated_commit,
+                commit
+            );
+
+            Ok((
+                VidCommonQueryData::new(leaf.block_header().clone(), None),
```
sveitser, line 208: So the new VID scheme does not use VidCommon anymore? Should we make this a bit more clear by having an enum Type for V1 and V2 instead?

imabdulbasit: Chengyu is going to add it

sveitser (approved):
LGTM. Just a question because I think it's a bit non-trivial if we define type aliases as `Option<OldType>`.

PR #2734, opened 2025-03-07
Store epochs and drbs from HotShot into Storage

pls148:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-task-impls/src/helpers.rs
+++ b/hotshot-task-impls/src/helpers.rs
@@ -9,6 +9,7 @@ use async_lock::RwLock;
 use committable::{Commitment, Committable};
 use either::Either;
 use hotshot_task::dependency::{Dependency, EventDependency};
+use hotshot_types::traits::storage::Storage;
 use hotshot_types::{
     consensus::OuterConsensus,
     data::{Leaf2, QuorumProposalWrapper, ViewChangeEvidence2},
@@ -180,10 +181,11 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
 }
 
 /// Handles calling add_epoch_root and sync_l1 on Membership if necessary.
-async fn decide_epoch_root<TYPES: NodeType>(
+async fn decide_epoch_root<TYPES: NodeType, I: NodeImplementation<TYPES>>(
     decided_leaf: &Leaf2<TYPES>,
     epoch_height: u64,
     membership: &Arc<RwLock<TYPES::Membership>>,
+    storage: &Arc<RwLock<I::Storage>>,
 ) {
     let decided_block_number = decided_leaf.block_header().block_number();
 
@@ -191,6 +193,19 @@ async fn decide_epoch_root<TYPES: NodeType>(
     if epoch_height != 0 && is_epoch_root(decided_block_number, epoch_height) {
         let next_epoch_number =
             TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height) + 2);
+
+        if let Err(e) = storage
+            .write()
+            .await
+            .add_epoch_root(next_epoch_number, decided_leaf.block_header().clone())
+            .await
+        {
+            tracing::error!(
+                "Failed to store epoch root for epoch {:?}: {}",
+                next_epoch_number,
+                e
+            );
+        }
 
         let write_callback = {
             tracing::debug!("Calling add_epoch_root for epoch {:?}", next_epoch_number);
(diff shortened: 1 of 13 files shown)


hotshot-task-impls/src/helpers.rs
```diff
@@ -185,6 +187,13 @@ async fn decide_epoch_root<TYPES: NodeType>(
         let next_epoch_number =
             TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height) + 2);
 
+        // REVIEW NOTE: What should we do if storage write fails?
```
ss-es, line 190: good question, I think just log the failure and carry on. So `if let Err(e) = ... { tracing::error!("Failed to store...: {}", e); }`

cc @bfish713 for his opinion

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -46,12 +46,19 @@ use crate::{
...
 
-async fn notify_membership_of_drb_result<TYPES: NodeType>(
+async fn notify_add_drb_result<TYPES: NodeType, I: NodeImplementation<TYPES>>(
     membership: &Arc<RwLock<TYPES::Membership>>,
+    storage: &Arc<RwLock<I::Storage>>,
     epoch: <TYPES as NodeType>::Epoch,
     drb_result: DrbResult,
 ) {
     tracing::debug!("Calling add_drb_result for epoch {:?}", epoch);
+    // REVIEW NOTE: What should we do if storage write fails?
```
ss-es, line 56: same comment

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -46,12 +46,19 @@ use crate::{
     quorum_vote::Versions,
 };
 
-async fn notify_membership_of_drb_result<TYPES: NodeType>(
+async fn notify_add_drb_result<TYPES: NodeType, I: NodeImplementation<TYPES>>(
```
ss-es, line 49: I think I dislike this name (but also the old name)

can we maybe call it `handle_drb_result`? I don't necessarily think that's good but it's at least in line with the rest of our naming

ss-es (commented):
Left some comments, but looks good to me! I think @imabdulbasit would be a better person to review the changes under `sequencer/`

sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE epoch_drb_and_root (
```
imabdulbasit, line 1: I don't see the migration for postgres

types/src/v0/traits.rs
```diff
@@ -595,7 +603,7 @@ pub trait SequencerPersistence: Sized + Send + Sync + Clone + 'static {
                     .collect(),
                 undecided_state,
                 saved_vid_shares: Default::default(), // TODO: implement saved_vid_shares
-                start_epoch_info: Default::default(), // TODO: implement start_epoch_info
+                start_epoch_info,
```
imabdulbasit, line 606: need to update start_epoch field as well. Currently, it passes genesis epoch

sequencer/api/migrations/postgres/V502__epoch_drb_and_root.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE epoch_drb_and_root (
+    epoch BIGINT PRIMARY KEY,
+    drb_result BYTEA,
+    block_header BYTEA,
```
imabdulbasit, line 4: think migration will fail with this extra comma at the end
```suggestion
    block_header BYTEA
```

sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
```diff
@@ -0,0 +1,5 @@
...
+    epoch BIGINT PRIMARY KEY,
+    drb_result BLOB,
+    block_header BLOB,
```
imabdulbasit, line 4: same here

imabdulbasit (approved):
LGTM, but would be nice to have simple test to store and load. You can include the test inside the module persistence_tests with  function signature [like this](https://github.com/EspressoSystems/espresso-network/blob/main/sequencer/src/persistence.rs#L117) (takes P: TestablePersistence). This would test both FS and SQL

sequencer/src/persistence/sql.rs
```diff
@@ -1873,6 +1874,80 @@ impl SequencerPersistence for Persistence {
...
+        &self,
+        epoch: EpochNumber,
+        drb_result: DrbResult,
+    ) -> anyhow::Result<()> {
+        let mut tx = self.db.write().await?;
+        tx.upsert(
+            "epoch_drb_and_root",
+            ["epoch", "drb_result"],
+            ["epoch"],
+            [(epoch.u64() as i64, drb_result)],
```
imabdulbasit, line 1888: you may have to convert drb_result bytes to a vec

PR #2736, opened 2025-03-07
Remove foundry patch

sveitser:
The patch in alloy is now included in foundry.

- Create a separate dev shell with the legacy version of foundry that we need to generate ethers bindings.
- Switch to foundry.nix overlay because the nixpkgs version is likely to lag a lot because foundry doesn't properly version their code.
- Enable the optimizer. This used to be the default but this was changed in foundry recently.


jparr721 (approved):

sveitser:
Tests previously passed and the contracts CI passed now, so I'm bypassing.

PR #2737, opened 2025-03-07
upgrade geth to v1.15

sveitser:
I can't test locally. One layer of the block explorer image takes ages to pull. So just putting this CI job out to see if it breaks anything.

If this works we can finally upgrade solc because it should support all the latest and greatest opcode.

cc and thanks @babdor


sveitser:
@babdor everything passed. The flaky CI jobs that failed a few times are also flaky on main.

PR #2740, opened 2025-03-07
Add VidCommon enum to support both VID schemes

bfish713:
### This PR:
This change introduces a new VidCommon enum to support both ADVZ and AVIDM VID schemes. Key modifications include:
- Create AvidMCommon type which is just the the parameter `AvidMParam`
- Replacing `Option<ADVZCommon>` with the `VidCommon` enum
- Adding support for V0 (ADVZ) and V1 (AVIDM) VID common data
- Modifying fetching, persistence, and query service components to work with the new enum
### Key places to review:

The changes to the the Query service to add in the common info.  in query_service.rs where we use the common info to verify the V1 share.

Some of the tests were updated to show we are returning a V1 common info when we should.
We should also observe the elimination of "Failed to get VIDCommon" Logs

Just what is in unit tests


mrain (approved):
LGTM

sequencer/src/api/endpoints.rs
```diff
@@ -135,10 +135,12 @@ where
                         })
                 }
             )?;
-            let common = &common.common().clone().context(CustomSnafu {
-                message: format!("failed to make proof for namespace {ns_id}"),
-                status: StatusCode::NOT_FOUND,
-            })?;
+            let VidCommon::V0(common) = &common.common().clone() else {
```
imabdulbasit, line 138: I guess we will need to handle namespace proof verification at some point using V1 VID.

hotshot-query-service/src/lib.rs
```diff
@@ -447,7 +447,11 @@ use task::BackgroundTask;
 use tide_disco::{method::ReadState, App, StatusCode};
 use vbs::version::StaticVersionType;
 
-pub type VidCommon = Option<hotshot_types::vid::advz::ADVZCommon>;
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+pub enum VidCommon {
+    V0(hotshot_types::vid::advz::ADVZCommon),
```
imabdulbasit, line 452: so we bincode serialize the `VidCommon` when inserting data. This would mean, we require a migration for vid as well in query service

imabdulbasit (approved):

hotshot-query-service/src/lib.rs
imabdulbasit: basically, just changing VIDCommon serialization in the vid migration that we already have

bfish713: Yes I wasn't sure, should I add this to the migration in this PR?

imabdulbasit: should be a few lines change. I can do it in a separate PR

PR #2749, opened 2025-03-10
fix: enable retries for integration tests

sveitser:
Currently the retries don't work because we start the native demo once then retry the asserts. However if the tests fail it usually means the native demo is not working so we would have to restart it. This PR moves starts the native demo as part of the integration tests, thereby making the tests standalone.

Unfortunately the upgrade test currently never works for me locally, even on `main`. edit: it does work on CI though so it's probably me.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -212,10 +212,11 @@ jobs:
         version: [02,99]
         include:
           - version: 02
-            compose: "-f process-compose.yaml"
+            test-name: test_native_demo
 
           - version: 99
-            compose: "-f process-compose.yaml -f process-compose-mp.yml"
+            test-name: test_native_demo_upgrade
+
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


jparr721 (approved):

.github/workflows/test.yml
```diff
@@ -212,10 +212,11 @@ jobs:
         version: [02,99]
         include:
           - version: 02
-            compose: "-f process-compose.yaml"
+            test-name: test_native_demo_basic
```
imabdulbasit, line 215: so previously this was running smoke and upgrade test both. Now its just upgrade test?

tests/upgrades.rs
```diff
@@ -3,12 +3,21 @@ use espresso_types::{FeeVersion, MarketplaceVersion};
...
+    smoke::assert_native_demo_works,
+};
 
 const SEQUENCER_BLOCKS_TIMEOUT: u64 = 200;
 
 #[tokio::test(flavor = "multi_thread")]
-async fn test_upgrade() -> Result<()> {
+async fn test_native_demo_upgrade() -> Result<()> {
+    let _demo = NativeDemo::run(Some(
+        "-f process-compose.yaml -f process-compose-mp.yml".to_string(),
```
imabdulbasit, line 16: this would only run marketplace upgrade, so have to maybe do it through an env variable

.github/workflows/test.yml
imabdulbasit: nvm I see assert_native_demo_works() being called

PR #2752, opened 2025-03-10
Calculate and use weights for new VID

ss-es:

### DIFF
--- a/hotshot-types/src/data/vid_disperse.rs
+++ b/hotshot-types/src/data/vid_disperse.rs
@@ -10,6 +10,7 @@ use std::{collections::BTreeMap, fmt::Debug, hash::Hash, marker::PhantomData};
 
 use hotshot_utils::anytrace::*;
 use jf_vid::{VidDisperse as JfVidDisperse, VidScheme};
+use primitive_types::U256;
 use serde::{Deserialize, Serialize};
 use tokio::task::spawn_blocking;
 
@@ -20,7 +21,9 @@ use crate::{
     message::Proposal,
     simple_vote::HasEpoch,
     traits::{
-        block_contents::EncodeBytes, node_implementation::NodeType, signature_key::SignatureKey,
+        block_contents::EncodeBytes,
+        node_implementation::NodeType,
+        signature_key::{SignatureKey, StakeTableEntryType},
         BlockPayload,
     },
     vid::{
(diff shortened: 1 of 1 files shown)


hotshot-types/src/data/vid_disperse.rs
```diff
@@ -274,6 +279,39 @@ impl<TYPES: NodeType> HasViewNumber<TYPES> for AvidMDisperse<TYPES> {
     }
 }
 
+/// The target total stake to scale to for VID.
+pub const VID_TARGET_TOTAL_STAKE: u32 = 10000;
```
mrain, line 283: Just tested, using 4 threads the dispersal of 32MB payload to storage nodes with 10k total weights costs 6.5s on my local laptop. Do you think it's acceptable?

ss-es: @bfish713 is this a reasonable time for VID? I guess 32MB is on the large side for a payload?

hotshot-types/src/data/vid_disperse.rs
```diff
@@ -274,6 +279,66 @@ impl<TYPES: NodeType> HasViewNumber<TYPES> for AvidMDisperse<TYPES> {
...
+    // weights, in stake table order
+    weights: Vec<u32>,
+
+    // total weight
+    total_weight: usize,
+}
+
+fn approximate_weights<TYPES: NodeType>(
+    stake_table: Vec<PeerConfig<TYPES::SignatureKey>>,
+) -> Weights {
```
mrain, line 296: Should we also have a simple function just to calculate the total_weights? Share verification and commitment computation also need this information.

- https://github.com/EspressoSystems/espresso-network/blob/c7deb08a3bba9997d8842a892d99ca40d36e0ec3/hotshot-types/src/data.rs#L334
- https://github.com/EspressoSystems/espresso-network/blob/c7deb08a3bba9997d8842a892d99ca40d36e0ec3/hotshot-types/src/data/vid_disperse.rs#L489

Also there are other calls from query-service and sequencer.

ss-es: added `vid_total_weight`: https://github.com/EspressoSystems/espresso-network/pull/2752/commits/f2d06f57c326941af9e3d4576c2542d5e7bf1b30

mrain (approved):
LGTM!

PR #2754, opened 2025-03-11
fix: geth genesis: set difficult to zero

sveitser:
We updated geth and the difficulty must now be zero.


imabdulbasit (approved):

sveitser:
This shouldn't affect the hotshot integration tests. So I will bypass.

PR #2756, opened 2025-03-11
Add stake store and load methods to `SequencerPersistance`

tbro:
Closes #<ISSUE_NUMBER>

Add methods to `SequencerPersistance` to store and load stake table.

### DIFF
--- a/sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
+++ b/sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
@@ -2,4 +2,5 @@ CREATE TABLE epoch_drb_and_root (
     epoch BIGINT PRIMARY KEY,
     drb_result BLOB,
     block_header BLOB
+    stake BLOB
 );
--- a/sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
+++ b/sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
@@ -1,6 +1,6 @@
 CREATE TABLE epoch_drb_and_root (
     epoch BIGINT PRIMARY KEY,
     drb_result BLOB,
-    block_header BLOB
+    block_header BLOB,
     stake BLOB
 );
--- a/sequencer/src/persistence/fs.rs
+++ b/sequencer/src/persistence/fs.rs
@@ -14,6 +14,7 @@ use clap::Parser;
 use espresso_types::{
     upgrade_commitment_map,
     v0::traits::{EventConsumer, PersistenceOptions, SequencerPersistence},
+    v0_3::CombinedStakeTable,
     Leaf, Leaf2, NetworkConfig, Payload, SeqTypes,
 };
 use hotshot::InitializerEpochInfo;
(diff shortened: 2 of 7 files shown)


sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
```diff
@@ -2,4 +2,5 @@ CREATE TABLE epoch_drb_and_root (
     epoch BIGINT PRIMARY KEY,
     drb_result BLOB,
     block_header BLOB
-);
\ No newline at end of file
+    stake BLOB
```
sveitser, line 5: Also needs a migration for postgres I think.

tbro: Ahh yes. Also I'm unsure how we run migrations. Since this is going in later than the other fields in `epoch_drb_and_root` migration, should we assume that migration has already run and add this in a seperate file?

tbro: 210f285aeb86c5079264b5da127f56e2a870056c

sequencer/src/persistence/sql.rs
```diff
@@ -1018,6 +1019,43 @@ impl SequencerPersistence for Persistence {
...
+        &self,
+        epoch: EpochNumber,
+        stake: CombinedStakeTable,
+    ) -> anyhow::Result<()> {
+        let stake_table_bytes = bincode::serialize(&stake).context("serializing stake table")?;
+
+        let mut tx = self.db.write().await?;
+        tx.upsert(
+            "epoch_drb_and_root",
+            ["epoch", "drb_result"],
```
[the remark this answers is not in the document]
tbro: 1fc0cf54dce4848acded6734bdb14353a39b263d

sequencer/api/migrations/sqlite/V302__epoch_drb_and_root.sql
tbro: For posterity, and to answer my own question, best practice is to add a new migration to append the new field to the table. But since this has not been deployed in any long running environment, it doesn't matter. So resolving.

sequencer/src/persistence/sql.rs
```diff
@@ -1018,6 +1019,39 @@ impl SequencerPersistence for Persistence {
         Ok(())
     }
 
+    async fn load_stake(&self, epoch: EpochNumber) -> anyhow::Result<Option<StakeTables>> {
+        let result = self
+            .db
+            .read()
+            .await?
+            .fetch_optional(
+                query("SELECT stake FROM epoch_drb_and_root WHERE epoch = ?1")
```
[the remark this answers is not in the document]
tbro: cf159ec851b2d59a274d4d85a7db09b04aeadbaf

sequencer/src/persistence/fs.rs
```diff
@@ -884,6 +889,43 @@ impl SequencerPersistence for Persistence {
...
     }
 
+    async fn load_stake(&self, epoch: EpochNumber) -> anyhow::Result<Option<StakeTables>> {
+        let inner = self.inner.read().await;
+        let path = &inner.stake_table_dir_path();
+        if !path.is_file() {
+            return Ok(None);
+        }
+
+        let file_path = path.join(epoch.to_string()).with_extension("txt");
```
imabdulbasit, line 899: think the above ` if !path.is_file()` check should be for the file path

PR #2762, opened 2025-03-11
[light-client] add Schnorr keys to the HotShot realm

### LINKED ISSUE #2685: [Light Client] bundle epoch QC with Schnorr signatures
Related to #2581 

See discussions here: https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/prover.20service.20stake.20table.20update

We need to update the epoch transition logic as well:

Schnorr signatures will be included in the epoch QC as well to utilize the persistent storage. In this case, we can easily recover from failed prover-service/relay server during the epoch transition.

mrain:
Pre requisite for  #2685 

### This PR:

- Add schnorr keys to the hotshot realm.
- Add light client state update vote, certificate and vote_accumulator

### This PR does not:

- No vote logic has been added yet.


sequencer/src/persistence/sql.rs
```diff
@@ -1805,6 +1806,44 @@ impl SequencerPersistence for Persistence {
...
+        .await?;
+        tx.commit().await
+    }
+
+    async fn load_state_cert(&self) -> anyhow::Result<LightClientStateUpdateCertificate<SeqTypes>> {
+        match self
+            .db
+            .read()
+            .await?
+            .fetch_one("SELECT state_cert, MAX(epoch) from state_cert GROUP BY epoch")
```
imabdulbasit, line 1832: I don't think this would work because state_cert can not be grouped. 
I think what you want is last state_cert
so can do smth like
`SELECT state_cert from state_cert ORDER BY epoch DESC LIMIT 1`

sequencer/src/persistence/sql.rs
```diff
@@ -1805,6 +1806,44 @@ impl SequencerPersistence for Persistence {
...
+        {
+            Ok(row) => {
+                if let Some(data) = row.get::<Option<Vec<u8>>, _>("state_cert") {
+                    bincode::deserialize(&data)
+                        .context("deserializing light client state update certificate")
+                } else {
+                    Ok(LightClientStateUpdateCertificate::genesis())
+                }
+            },
+            Err(_) => Ok(LightClientStateUpdateCertificate::genesis()),
```
imabdulbasit, line 1843: why we return default? The error means that query is failing so should return an err

sequencer/src/persistence/fs.rs
```diff
@@ -1277,6 +1302,37 @@ impl SequencerPersistence for Persistence {
...
+                "reading light client state update certificate {}",
+                path.display()
+            ))?;
+            let cert = bincode::deserialize::<LightClientStateUpdateCertificate<SeqTypes>>(&bytes)
+                .context(format!(
+                    "parsing light client state update certificate {}",
+                    path.display()
+                ))?;
+
+            if cert.epoch > result.epoch {
```
imabdulbasit, line 1329: why deserialize the cert and then have this check? There is already an epoch from the sub directory. Maybe this check should be moved up and have another check that sub dir epoch == cert.epoch

sequencer/src/persistence/sql.rs
```diff
@@ -1805,6 +1806,44 @@ impl SequencerPersistence for Persistence {
...
+    async fn load_state_cert(&self) -> anyhow::Result<LightClientStateUpdateCertificate<SeqTypes>> {
+        match self
+            .db
+            .read()
+            .await?
+            .fetch_one("SELECT state_cert, MAX(epoch) from state_cert GROUP BY epoch")
+            .await
+        {
+            Ok(row) => {
+                if let Some(data) = row.get::<Option<Vec<u8>>, _>("state_cert") {
```
imabdulbasit, line 1836: so I don't understand why we are deserializing it as Option<Vec<u8>>? When we store we are serialzing the LightClientStateUpdateCertificate into bytes

sequencer/src/persistence/sql.rs
imabdulbasit, line 1832: think you want to use fetch_optional here

imabdulbasit: The fetch_one will return an error if there is no row

mrain: Changed in [60ddcad](https://github.com/EspressoSystems/espresso-network/pull/2762/commits/60ddcada8e0f37d832f2a9322726839109c12964)

sequencer/src/persistence/fs.rs
mrain: Yes this is redundant, [60ddcad](https://github.com/EspressoSystems/espresso-network/pull/2762/commits/60ddcada8e0f37d832f2a9322726839109c12964)

sequencer/src/persistence/sql.rs
mrain: Use `fetch_optional` to return an Option: [60ddcad](https://github.com/EspressoSystems/espresso-network/pull/2762/commits/60ddcada8e0f37d832f2a9322726839109c12964)

sequencer/src/persistence/sql.rs
```diff
@@ -1824,23 +1824,23 @@ impl SequencerPersistence for Persistence {
...
-                    Ok(LightClientStateUpdateCertificate::genesis())
-                }
-            },
-            Err(_) => Ok(LightClientStateUpdateCertificate::genesis()),
+            .fetch_optional("SELECT state_cert from state_cert ORDER BY epoch DESC LIMIT 1")
+            .await?
+        else {
+            return Ok(None);
+        };
+        if let Some(data) = row.get::<Option<Vec<u8>>, _>("state_cert") {
```
imabdulbasit, line 1839: why is this deserialized as Option?

mrain: check [5ccfe57](https://github.com/EspressoSystems/espresso-network/pull/2762/commits/5ccfe570049d0bd9533944dbb5de361864762787)

Also add some new test cases

imabdulbasit (approved):

PR #2766, opened 2025-03-12
increase restart test timeout: 5s -> 10s

sveitser:
This allows me to locally run with half a CPU. Maybe this is good enough for the CI?


sveitser:
Does not seem to help.

PR #2770, opened 2025-03-13
Nuke update_undecided_state from Storage

bfish713:
Closes #<ISSUE_NUMBER>


sequencer/api/migrations/postgres/V501__epoch_tables.sql
```diff
@@ -51,4 +42,4 @@ CREATE TABLE epoch_migration (
     completed bool DEFAULT FALSE
 );
 
-INSERT INTO epoch_migration (table_name) VALUES ('anchor_leaf'), ('da_proposal'), ('vid_share'), ('undecided_state'), ('quorum_proposals'), ('quorum_certificate');
\ No newline at end of file
+INSERT INTO epoch_migration (table_name) VALUES ('anchor_leaf'), ('da_proposal'), ('vid_share'), ('quorum_proposals'), ('quorum_certificate');
```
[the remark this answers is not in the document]
imabdulbasit: There must be a new migration to drop the table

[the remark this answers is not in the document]
bfish713: Yeah this is a problem, I wasn't sure what to do.  I don't think just deleting this is correct

sequencer/api/migrations/postgres/V15__undecided_state.sql
jbearer, line 1: We can't delete migrations as they are already deployed. Instead create a new migration with a higher `V` number to do a `DROP TABLE undecided_state`

sequencer/api/migrations/postgres/V501__epoch_tables.sql
```diff
@@ -18,15 +18,6 @@ CREATE TABLE vid_share2 (
 );
 
 
-CREATE TABLE undecided_state2 (
```
jbearer, line 21: Same here, this should be a new migration doing `DROP TABLE undecied_state2`

sequencer/api/migrations/postgres/V501__epoch_tables.sql
jbearer, line 45: I don't know what this is exactly but I guess we want something like `DELETE FROM epoch_migration WHERE table_name = 'undecided_state'`?

sequencer/api/migrations/sqlite/V105__undecided_state.sql
jbearer, line 1: Same here

sequencer/api/migrations/sqlite/V301__epoch_tables.sql
jbearer, line 1: And here

jbearer (changes requested):
Looks good except for migrations

bfish713:
Fix the sql I think, not sure how I can test it though

sequencer/api/migrations/postgres/V501__epoch_tables.sql
```diff
@@ -17,7 +17,6 @@ CREATE TABLE vid_share2 (
     data BYTEA
 );
 
-
```
jbearer, line 20: Unfortunately even this blank line diff might cause an issue. I believe the migration runner does a simple text comparison with the deployed migrations before applying new ones to make sure it's not doing anything inconsistent

sequencer/api/migrations/sqlite/V105__undecided_state.sql
```diff
@@ -4,4 +4,4 @@ CREATE TABLE undecided_state (
     id INT PRIMARY KEY,
     leaves BLOB NOT NULL,
     state  BLOB NOT NULL
-);
+);
```
jbearer, line 7: whitespace change

sequencer/api/migrations/sqlite/V301__epoch_tables.sql
```diff
@@ -17,7 +17,6 @@ CREATE TABLE vid_share2 (
     data BLOB
 );
 
-
```
jbearer, line 20: whitespace change

jbearer (approved):

PR #2772, opened 2025-03-13
 Reward Calculation

imabdulbasit:
This PR:
- handles the new stake table contract events
- adds reward merkle and SQL storage for it
- catchup for reward merkle tree
- reward calculation and distributions

It does not include:
-test for catchup, reward distribution


types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,217 @@ use hotshot_types::{
...
+    // TODO: return RESULT
+    // TODO: Handle consensus keys update event
+    let mut st_map = BTreeMap::new();
+    let mut bls_keys = HashSet::new();
+    let mut schnor_keys = HashSet::new();
+    for (reg, log) in registered {
+        let bls = bls_alloy_to_jf2(reg.blsVk.clone());
+        let state_ver_key = edward_bn254point_to_state_ver(reg.schnorrVk.clone());
+        if bls_keys.contains(&bls) {
+            tracing::error!("bls key {bls:?} already used");
```
sveitser, line 76:
```suggestion
            tracing::warn!("bls key {bls:?} already used");
```

types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,217 @@ use hotshot_types::{
...
+            continue;
+        }
+
+        if schnor_keys.contains(&state_ver_key) {
+            tracing::error!("schnor verifying key {bls:?} already used");
```
sveitser, line 81:
```suggestion
            tracing::warn!("schnor verifying key {bls:?} already used");
```

types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,217 @@ use hotshot_types::{
...
+            continue;
         }
+        st_map.insert(
+            (
+                log.block_number.unwrap(),
+                log.transaction_index.unwrap(),
```
sveitser, line 87: the log index should be the index in the block, so I think we don't need the tx index?

https://docs.rs/alloy/0.12.5/alloy/rpc/types/struct.Log.html#structfield.log_index

types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,217 @@ use hotshot_types::{
...
+    delegated: Vec<(Delegated, Log)>,
+    undelegated: Vec<(Undelegated, Log)>,
+    keys_update: Vec<(ConsensusKeysUpdated, Log)>,
+) -> IndexMap<Address, Validator<BLSPubKey>> {
+    // TODO: return RESULT
+    // TODO: Handle consensus keys update event
+    let mut st_map = BTreeMap::new();
+    let mut bls_keys = HashSet::new();
+    let mut schnor_keys = HashSet::new();
+    for (reg, log) in registered {
```
sveitser, line 72: Add a TODO comment for bls signature verification. We want to move this from the contract to the GCL.

types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,217 @@ use hotshot_types::{
...
+                log.block_number.unwrap(),
+                log.transaction_index.unwrap(),
+                log.log_index.unwrap(),
+            ),
+            DelegationChange::Remove(undelg),
+        );
+    }
+
+    let mut validators = IndexMap::new();
+    for staker in st_map.values() {
```
sveitser, line 132: @imabdulbasit do I understand correctly that we first process all the validator registrations an de-registrations and then the delegations? That seems incorrect to me. I think it's easier to reason about correctness if we process all the events in order, for example by making a an enum that can hold all the events, then sorting that by (block, log_index) and iterating over all the events and process them.

imabdulbasit: oh that there could be multiple events for (block, log_index)?

imabdulbasit: So, my thinking was we only need to process delegation events for validators that exist?

imabdulbasit: So, the (block, log_index) would make each event unique, right? I can make the enum change

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/83f18136918149a6f57275631f212ef3b93e06c1

types/src/v0/impls/stake_table.rs
imabdulbasit: added

types/src/v0/impls/stake_table.rs
```diff
@@ -26,84 +37,207 @@ use hotshot_types::{
...
+                account,
+                blsVk,
+                schnorrVk,
+                commission,
+            }) => {
+                // TODO(abdul): BLS and Schnorr signature keys verification
+                let stake_table_key = bls_alloy_to_jf2(blsVk.clone());
+                let state_ver_key = edward_bn254point_to_state_ver(schnorrVk.clone());
+                if bls_keys.contains(&stake_table_key) {
+                    tracing::warn!("bls key {stake_table_key:?} already used");
```
sveitser, line 90: @imabdulbasit Can re return an error from this function for anything that is a bug in the smart contract? This would mean the confirmation layer encounters a bug in the contract so it's IMO best to stop running so that we can investigate. If we return an Error here and then panic in the outer layer at least we will be alerted that something is wrong. I think this is less dangerous than continuing operation after we find a bug in the contract. I think it will also make it more likely we discover bugs in testing and it makes this function easier to test on bad inputs.

imabdulbasit: why is this a bug? Contract supports multiple BLS keys? I thought we would just ignore those events

types/src/v0/impls/reward.rs
```diff
@@ -0,0 +1,421 @@
...
+                    balance
+                },
+            }
+        })?;
+        Ok::<(), anyhow::Error>(())
+    };
+
+    // Distribute leader reward
+    let leader_reward = RewardAmount::from(
+        (U256::from(validator.commission) * block_reward().0) / U256::from(10_000),
```
bfish713, line 347: Where are the magic number `10_000` and `100` below coming from?

types/src/v0/impls/stake_table.rs
sveitser: Currently the contract doesn't allow a BLS key to be used twice. We can of course allow it but it doesn't cost much gas to check and mark as used, it's true that we currently don't do it for schnorr keys but I think we should also do it for schnorr keys. I just catches some potential configuration errors and I don't think we really gain anything from allowing it in the contract.

types/src/v0/impls/reward.rs
```diff
@@ -0,0 +1,421 @@
...
+                    tracing::warn!("overflowed reward balance for account {}", account);
+                    balance
+                },
+            }
+        })?;
+        Ok::<(), anyhow::Error>(())
+    };
+
+    // Distribute leader reward
+    let leader_reward = RewardAmount::from(
```
bfish713, line 346: If the leader has no delegators does this calculation give the leader all the rewards?  I feel like this function is missing a piece where the leader should get rewards based on how much it is personally staking, not just the commissions.  Maybe this is accounted for because the leader is in the delegators?

bfish713 (commented):
It's probably tooo late now, but in the future can you split PRs like this into smaller chunks.  In this PR there are at least 2 distinct peices.  The reward state part and the integration with the new stake table contract.

types/src/v0/impls/reward.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/ddf6c0f5dab9c80dbe54bb8bd9123240c7ef28da

types/src/v0/impls/reward.rs
imabdulbasit: Yes, the leader will be among the delegators. I believe the way the contract is designed, the leader will be a delegator if it wants to stake

imabdulbasit: I guess there should a check here that leader is among the delegators?

sequencer/src/api.rs
```diff
@@ -386,6 +393,79 @@ impl<
...
+        &self,
+        instance: &NodeState,
+        height: u64,
+        view: ViewNumber,
+        accounts: &[RewardAccount],
+    ) -> anyhow::Result<RewardMerkleTree> {
+        // Check if we have the desired state in memory.
+        match self
+            .as_ref()
+            .get_reward_accounts(instance, height, view, accounts)
```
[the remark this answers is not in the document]
imabdulbasit: we do fetch it from storage if we don't have them in memory

types/src/v0/impls/reward.rs
```diff
@@ -0,0 +1,425 @@
...
+    mut reward_state: RewardMerkleTree,
+    validator: Validator<BLSPubKey>,
+) -> anyhow::Result<RewardMerkleTree> {
+    let mut update_balance = |account: &RewardAccount, amount: RewardAmount| {
+        reward_state = reward_state.persistent_update_with(account, |balance| {
+            let balance = balance.copied();
+            match balance.unwrap_or_default().0.checked_add(amount.0) {
+                Some(updated) => Some(updated.into()),
+                None => {
+                    tracing::warn!("overflowed reward balance for account {}", account);
```
sveitser, line 337: This seems like a really unexpected error. Should we really continue? It means we cannot actually calculate the correct rewards so it's a breakdown of the protocol. Shouldn't we return an error, panic or whatever.

types/src/v0/impls/reward.rs
```diff
@@ -0,0 +1,425 @@
...
+    update_balance(&RewardAccount(validator.account.to_ethers()), leader_reward)?;
+
+    let delegators_reward = block_reward()
+        .checked_sub(&leader_reward)
+        .context("overflow")?;
+    // Distribute delegator rewards
+    let total_stake = validator.stake.to_ethers();
+    for (delegator_address, delegator_stake) in &validator.delegators {
+        let delegator_reward = RewardAmount::from(
+            delegator_stake.to_ethers() * U256::from(delegators_reward) / total_stake,
```
sveitser, line 358: Can we rip the reward computation out into a pure function so we can write some unittests for it? It seems correct to me but I think it would be nice is if we had some tests that we don't lose anything due to rounding errors.

Also we don't have overflow checks for the multiplication and in release mode it will overflow, I think it shouldn't happen here either, but if it's going to happen, here it's more likely here than when updating balances.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/69f42ceca394a959a51930b8d5ad821ba6ff694b

types/src/v0/impls/reward.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/0db1340254018a76e0ec25ad91351e9576a4969c

sequencer/api/catchup.toml
```diff
@@ -80,4 +80,27 @@ This Endpoint fetches a leaf chain which decides a specified block height.  This
 For catching up the stake table, where `:height` is the block height of the epoch root you want to catchup to
 
 Returns a list of leaves which includes `:height` as the last leaf and should prove the block with `:height` was decided.  
-"""
\ No newline at end of file
+"""
+
+[route.reward_account]
+PATH = ["/:height/:view/reward-account/:address"]
```
sveitser, line 86: Why do we require a height and a view number to query rewards? In the implementation it looks like the height isn't used except for logging. It would seem more natural to me if we could look up reward balances by height and not view number.

sequencer/src/api.rs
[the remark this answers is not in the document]
sveitser: I mean that at each layer we should keep the accounts that we found and then only get the ones we didn't find from the next layer.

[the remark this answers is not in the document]
sveitser: I think we can do it as follow up work though, since this shouldn't affect basic correctness of this first implementation.

sequencer/api/catchup.toml
sveitser: ah sorry missed that this is the catchup API

sequencer/src/api.rs
```diff
@@ -386,6 +393,79 @@ impl<
...
+                // at all, we can create a new view based on the recovered leaf and add it to
+                // our state map. In this case, we must also add the leaf to the saved leaves
+                // map to ensure consistency.
+                let mut state = ValidatedState::from_header(leaf.block_header());
+                state.reward_merkle_tree = tree.clone();
+                (Arc::new(state), None)
+            },
+        };
+        if let Err(err) = consensus.update_leaf(leaf, Arc::clone(&state), delta) {
+            tracing::warn!(?view, "cannot update fetched account state: {err:#}");
```
sveitser, line 463: Do you know why this could happen? And if it happens, is it safe to continue?

sequencer/api/migrations/sqlite/V304__reward_merkle_tree.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE reward_merkle_tree (
+  path JSONB NOT NULL,
```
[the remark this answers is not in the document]
imabdulbasit: Thanks for pointing this out. this should be JSONB as we had a migration for postgres to change array to jsonb

sequencer/src/api.rs
imabdulbasit: can only happen if the Consensus::update_validated_state_map() fails, which can only fail in two conditions: 
new update has `None` for delta and existing view has a delta, this means new view has less info than existing stored data. The other case is if we are not updating the Leaf i.e ViewInner is not the leaf which won't hit in this case.


I think its okay to return accounts even if we can't update the state but I am not too sure

sequencer/api/migrations/sqlite/V304__reward_merkle_tree.sql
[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/6b7422b53ea5ad5fbe02cb9f77db66eadd9c32ef

types/src/v0/impls/reward.rs
```diff
@@ -0,0 +1,496 @@
...
+    let validator = membership.get_validator_config(&epoch, leader).unwrap();
+    let mut reward_accounts = vec![validator.account.to_ethers().into()];
+    let delegators = validator
+        .delegators
+        .keys()
+        .cloned()
+        .map(|a| a.to_ethers().into())
+        .collect::<Vec<RewardAccount>>();
+
+    reward_accounts.extend(delegators.clone());
```
sveitser, line 421: @imabdulbasit if the validator is also a delegator we will have it twice in the `reward_accounts`. It's fine I guess but we could use a set to avoid it.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/ebe341b390c4222fe0d135a8065b2ee85f0da61a

types/src/v0/impls/stake_table.rs
```diff
@@ -26,89 +30,254 @@ use hotshot_types::{
...
+                validator_entry.stake = validator_entry
+                    .stake
+                    .checked_sub(amount)
+                    .with_context(|| "stake is less than undelegated amount")?;
+
+                let delegator_stake = validator_entry
+                    .delegators
+                    .get_mut(&delegator)
+                    .with_context(|| format!("delegator {delegator:#x} not found"))?;
+                *delegator_stake = delegator_stake.checked_sub(amount).unwrap();
```
sveitser, line 142: This should not be unwrap but return an error like above `delegated amount less than undelegated amount`

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/fc8d14dc3780c6d5295723e136e92aba8d1ec2bb

types/src/v0/impls/stake_table.rs
```diff
@@ -26,89 +30,254 @@ use hotshot_types::{
...
+fn validate_validaors(validators: IndexMap<Address, Validator<BLSPubKey>>) -> anyhow::Result<()> {
+    for (address, validator) in validators {
+        // Ensure validator has at least one delegator
+        if validator.delegators.is_empty() {
+            bail!("Validator {address:?} must have at least one delegator",);
+        }
+
+        // Ensure total stake is not zero
+        if validator.stake.is_zero() {
+            bail!("Validator {address:?} must have a non-zero total stake",);
```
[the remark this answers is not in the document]
imabdulbasit: ah okay, it looks a bit weird to have validator with 0 stake but yeah we need to remove and not return an err

types/src/v0/impls/stake_table.rs
```diff
@@ -167,61 +323,92 @@ struct Committee {
...
-                )
-            })
-            .collect();
+        self.state.insert(
+            epoch,
+            EpochCommittee {
+                eligible_leaders: self.non_epoch_committee.eligible_leaders.clone(),
+                stake_table,
+                validators,
+                address_mapping: HashMap::new(),
```
sveitser, line 374: This should be the computed `address_mapping` I think.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2772/commits/53df0ee1e26943fd5a2dcc996edf7687aecac31a

sveitser (approved):
I do think we will discover a few bugs once we add more tests and follow up work but I think this is a good start and already huge PR and we should merge it.

PR #2773, opened 2025-03-13
Persist stake table from `add_epoch_root`

tbro:
Persist stake table from `add_epoch_root`

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -30,7 +30,7 @@ use itertools::Itertools;
 use thiserror::Error;
 
 use super::{
-    traits::{MembershipPersistence, SequencerPersistence, StateCatchup},
+    traits::{MembershipPersistence, StateCatchup},
     v0_3::{DAMembers, StakeTable, StakeTables},
     Header, L1Client, Leaf2, NodeState, PubKey, SeqTypes,
 };
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -306,7 +306,7 @@ impl EpochCommittees {
         }
     }
 
-    /// Get the stake tables. Try to load from persistence and fall back to fetching from l1.
+    /// Get the stake table by epoch. Try to load from DB and fall back to fetching from l1.
     async fn get_stake_table(
         &self,
         epoch: Epoch,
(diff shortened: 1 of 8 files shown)


bfish713 (commented):
Can we add some tests for the both FS and SQL storing and loading the stake table

types/src/v0/impls/stake_table.rs
```diff
@@ -299,6 +305,37 @@ impl EpochCommittees {
...
+            .await
+            .map_err(GetStakeTablesError::PersistenceLoadError)?
+        {
+            Ok(stake_tables)
+        } else {
+            self.l1_client
+                .get_stake_table(contract_address.to_alloy(), l1_block)
+                .await
+                .map_err(GetStakeTablesError::L1ClientFetchError)
+        }
```
bfish713, line 328: We need to actually store the stake table here right?

bfish713 (changes requested):
We never call `store_stake`

types/src/v0/impls/stake_table.rs
tbro: Yea I changed it b/c I think we should try from persistence before we attempt fetching from the l1. But I guess we should do both: try `load_stake`, if that fails, fetch_from_l1, if that succeeds, persist.

tbro: ea89daf2756d620fa1330bbc2f0ffb20be99f9b9

tbro:
> Can we add some tests for the both FS and SQL storing and loading the stake table

added tests to both fs and sql.

bfish713 (approved):

types/src/v0/impls/stake_table.rs
```diff
@@ -299,6 +310,37 @@ impl EpochCommittees {
...
     }
+
+    /// Get the stake table by epoch. Try to load from DB and fall back to fetching from l1.
+    async fn get_stake_table(
+        &self,
+        epoch: Epoch,
+        contract_address: Address,
+        l1_block: u64,
+    ) -> Result<StakeTables, GetStakeTablesError> {
+        if let Some(stake_tables) = self
```
imabdulbasit, line 321: I think this will always return None when calling add_epoch_root()

tbro: Possibly in the current code, yes. But I think in the future we will be storing stake from other parts of the code which may run before `add_epoch_root`.

PR #2774, opened 2025-03-13
Fix for using wrong path for block headers in load_start_epoch_info f…

pls148:
…s implementation.

Mistakenly was loading the epoch root block header from the drb result file.


bfish713 (approved):
Can you also write a quick test for this as a follow up

PR #2784, opened 2025-03-17
Remove redundant code

lukaszrzasik:
Closes #<ISSUE_NUMBER>


hotshot-task-impls/src/helpers.rs
```diff
@@ -928,15 +928,7 @@ pub async fn wait_for_next_epoch_qc<TYPES: NodeType>(
...
     .await
-    else {
-        // Check again, there is a chance we missed it
-        if let Some(next_epoch_qc) = consensus.read().await.next_epoch_high_qc() {
-            if next_epoch_qc.data.leaf_commit == high_qc.data.leaf_commit {
-                return Some(next_epoch_qc.clone());
-            }
-        };
-        return None;
-    };
```
ss-es, line 939: so this is redundant because the `next_epoch_high_qc` in consensus should have come from a `NextEpochQc2Formed` event anyway, and this extra check is basically only accounting for the very unlikely case where we update consensus but the receiver is full or slow (which is kind of silly to account for anyway)?

ss-es (approved):
I think this looks good to me but can you confirm my understanding of the change is correct?

PR #2792, opened 2025-03-18
Fix get_leaf_chain in sequencer sql

pls148:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-query-service/migrations/postgres/V500__types_migration.sql
+++ b/hotshot-query-service/migrations/postgres/V500__types_migration.sql
@@ -1,11 +1,13 @@
 CREATE TABLE leaf2
 (
-    height     BIGINT  PRIMARY KEY REFERENCES header (height) ON DELETE CASCADE,
+    height     BIGINT NOT NULL REFERENCES header (height) ON DELETE CASCADE,
+    view       BIGINT NOT NULL,
     hash       VARCHAR NOT NULL UNIQUE,
     block_hash VARCHAR NOT NULL REFERENCES header (hash) ON DELETE CASCADE,
     leaf JSONB NOT NULL,
     qc   JSONB NOT NULL
 );
+ALTER TABLE leaf2 ADD CONSTRAINT leaf2_pk PRIMARY KEY (height, view);
 
 CREATE TABLE types_migration ( 
     id SERIAL PRIMARY KEY,
--- a/hotshot-query-service/src/data_source/storage/sql/transaction.rs
+++ b/hotshot-query-service/src/data_source/storage/sql/transaction.rs
@@ -35,7 +35,7 @@ use hotshot_types::{
     traits::{
         block_contents::BlockHeader,
         metrics::{Counter, Gauge, Histogram, Metrics},
-        node_implementation::NodeType,
+        node_implementation::{ConsensusTime, NodeType},
         EncodeBytes,
     },
 };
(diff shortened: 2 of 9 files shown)


hotshot-query-service/migrations/postgres/V500__types_migration.sql
```diff
@@ -1,11 +1,13 @@
 CREATE TABLE leaf2
 (
-    height     BIGINT  PRIMARY KEY REFERENCES header (height) ON DELETE CASCADE,
+    height     BIGINT  REFERENCES header (height) ON DELETE CASCADE,
+    view       BIGINT,
```
imabdulbasit, line 4: this should be `NOT NULL`, and also should have a index for it

imabdulbasit: or maybe the PK index would take of it, I am not sure

hotshot-query-service/src/data_source/storage/sql/transaction.rs
```diff
@@ -467,6 +467,7 @@ where
 {
     async fn insert_leaf(&mut self, leaf: LeafQueryData<Types>) -> anyhow::Result<()> {
```
imabdulbasit, line 468: can you also edit the types migration to include this field

sequencer/src/api/sql.rs
```diff
@@ -166,19 +167,51 @@ impl CatchupStorage for SqlStorage {
...
-            .context(format!("leaf chain {height} not available"))?;
-        let mut chain = vec![];
+            .context(format!("leaf {height} not available"))?;
+        let mut last_leaf: Leaf2 = leaf.leaf().clone();
+        let mut chain = vec![last_leaf.clone()];
+        let mut h = height + 1;
+
+        let mut found = false;
+        while !found {
+            for leaf in tx.get_leaves(h).await? {
```
bfish713, line 180: Are these for sure going to be in view sorted order?

sequencer/src/api/sql.rs
```diff
@@ -166,19 +167,51 @@ impl CatchupStorage for SqlStorage {
...
+            .get_leaf((height as usize).into())
             .await
-            .context(format!("leaf chain {height} not available"))?;
-        let mut chain = vec![];
+            .context(format!("leaf {height} not available"))?;
+        let mut last_leaf: Leaf2 = leaf.leaf().clone();
+        let mut chain = vec![last_leaf.clone()];
+        let mut h = height + 1;
+
+        let mut found = false;
```
bfish713, line 178: Nit: we don't really need this variable, we always break right after setting it true

bfish713 (changes requested):
Can you add a test showing that this works when adding multiple leaves with the same height and different views?

hotshot-query-service/migrations/postgres/V500__types_migration.sql
bfish713: I don't think we lookup by view, so I think that means we don't need an index

imabdulbasit: Yeah, I am just not sure if partial index performance < separate index, but okay lets leave it for now

imabdulbasit: > I don't think we lookup by view, so I think that means we don't need an index

There is an `ORDER BY view` so that requires it.

imabdulbasit:
I don't see the api endpoint for `get_leaves`

imabdulbasit (approved):

imabdulbasit (approved):

bfish713 (approved):

PR #2797, opened 2025-03-19
Add staking-cli docker image

sveitser:
- Run test demo on each PR.
- Save and load images via github artifacts.
- Deduplicate some of the CI docker building logic.
- Use ubuntu:latest for all docker images.
- Some cleanup in scripts to build docker images locally.


.github/actions/docker-image/action.yml
```diff
@@ -0,0 +1,41 @@
+name: Espresso Docker Image
```
sveitser, line 1: i just realized since the service names are standardized now we could matrix with one value per service then we don't need this action

sveitser:
Manually triggered run for docker build, to make sure push still works: https://github.com/EspressoSystems/espresso-network/actions/runs/13950270206

sveitser:
Not sure if this is going to work but it's definitely very close.

The latest run of the build job is here: https://github.com/EspressoSystems/espresso-network/actions/runs/13955857754

tbro (approved):

PR #2799, opened 2025-03-19
Remove undecided_state2 from Migrations

bfish713:
### This PR:

The migration to add this table never hit anything in production so we can just remove all references to it

### This PR does not:

Should not change anything functional.  

### Key places to review:

everything

### DIFF
--- a/sequencer/api/migrations/postgres/V403__drop_undecided_state.sql
+++ b/sequencer/api/migrations/postgres/V403__drop_undecided_state.sql
@@ -1,0 +1,1 @@
+DROP TABLE undecided_state;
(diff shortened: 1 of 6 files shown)


sequencer/api/migrations/postgres/V403__drop_undecided_state.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE undecided_state;
```
imabdulbasit, line 1: why does this migration has V403?

imabdulbasit (changes requested):

sequencer/api/migrations/postgres/V403__drop_undecided_state.sql
imabdulbasit: sorry but why this migration has changed from V503 to V403?

bfish713: because it's happening before the V5XX migrations hit

imabdulbasit (approved):

PR #2802, opened 2025-03-19
Add failover/retry to fee contract proxy validation in genesis.

jbearer:
Without this, a sequencer node cannot restart if the first client in its list is faulty, even though our L1 client implementation supports failover.

Request from @Ancient123 

### This PR:
* using the configured L1 client for fee contract validation, rather than constructing a new one with only one provider URL
* adding retries in `validate_fee_contract` so that it succeeds as long as the L1 client eventually fails over to a healthy provider

### DIFF
--- a/sequencer/src/lib.rs
+++ b/sequencer/src/lib.rs
@@ -438,6 +438,7 @@ pub async fn init_node<P: SequencerPersistence + MembershipPersistence, V: Versi
         .with_metrics(metrics)
         .connect(l1_params.urls)
         .with_context(|| "failed to create L1 client")?;
+    genesis.validate_fee_contract(&l1_client).await?;
 
     l1_client.spawn_tasks().await;
     let l1_genesis = match genesis.l1_finalized {
--- a/types/src/v0/v0_1/l1.rs
+++ b/types/src/v0/v0_1/l1.rs
@@ -215,14 +215,14 @@ pub struct SwitchingTransport {
 /// This is cloneable and returns a reference to the same underlying data.
 #[derive(Debug, Clone)]
 pub(crate) struct SingleTransport {
+    pub(crate) generation: usize,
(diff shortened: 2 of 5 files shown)


sequencer/src/lib.rs
```diff
@@ -438,6 +438,7 @@ pub async fn init_node<P: SequencerPersistence + MembershipPersistence, V: Versi
         .with_metrics(metrics)
         .connect(l1_params.urls)
         .with_context(|| "failed to create L1 client")?;
+    genesis.validate_fee_contract(&l1_client).await?;
```
rob-maron, line 441: If we don't call this at all on startup, what could happen? We call an invalid contract?

types/src/v0/v0_1/l1.rs
```diff
@@ -215,14 +215,14 @@ pub struct SwitchingTransport {
...
+    pub(crate) generation: usize,
     pub(crate) client: Http<Client>,
     pub(crate) status: Arc<RwLock<SingleTransportStatus>>,
 }
 
 /// The status of a single transport
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub(crate) struct SingleTransportStatus {
-    pub(crate) url_index: usize,
```
rob-maron, line 225: I like this change

rob-maron (approved):

sequencer/src/lib.rs
jbearer: Yeah I think it was just added as a sanity check because there were a couple cases where we incorrectly called the implementation contract instead of the proxy, and it wasn't obvious from the failure mode what was happening

rob-maron: Looks good to me then. Just wanted to make sure we couldn't move forward consensus wise without it (and refusing to start was necessary)

PR #2803, opened 2025-03-19
[VID] proof of incorrect encoding

### LINKED ISSUE #2801: [VID] add proof of incorrect encoding
after #2800

mrain:
Closes #2801


vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+            return Err(VidError::Argument(
+                "Insufficient shares to generate the proof of incorrect encoding.".to_string(),
+            ));
+        }
+        if self.raw_shares.len() > param.total_weights {
+            return Err(VidError::Argument("To many shares".to_string()));
+        }
+        let (mt, raw_shares) = AvidMScheme::raw_encode(param, &self.witness)?;
+        if mt.commitment() == commit.commit {
+            return Ok(Err(()));
```
alxiong, line 122: this is slightly confusing. (even tho i understand the intent) 

I'm not sure about the boundary between "invalid input" thus directly `Err()` and "valid input but failed verification, thus `Ok(Err())`. why isn't here also `Err(VidError::InvalidParam)`?

vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// Proof content, see [`MalEncodingProof`].
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct MalEncodingProofRawShare {
+    /// The index of this raw share
+    pub index: usize,
+    /// A Merkle tree proof against the provided Merkle tree commitment.
+    #[serde(with = "canonical")]
+    pub mt_proof: MerkleProof,
+}
```
alxiong, line 40: what about: 
- rename `witness` as `interpolated_poly` or `recovered_poly` or `recovered_data`? 
- remove the `MalEncodingProofRawShare` struct, as seems to be unnecessary abstraction, instead, simply `raw_shares: Vec<(usize, MerkleProof)>`, with comment explaining the tuple.
- avoid language like "proof content" this sounds like `raw_shares` is the actual proof, which it isn't. The whole bundle serve as a proof.

vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+    },
+    VerificationResult, VidError, VidResult, VidScheme,
+};
+
+/// A proof of incorrect encoding.
+/// It consists of a witness vector, which is the claimed low degree polynomial,
+/// and a list of [`MalEncodingProofRawShare`]s. Each raw share claims that, in the Merkle tree committed by the provided commitment,
+/// some code word of the witness vector appears at the `index` position.
+/// If we have enough raw shares, and the Merkle commitment of the encoding of witness vector doesn't match the provided commitment,
+/// we assert that the provided commitment cannot open to be a correct encoding.
```
alxiong, line 22: i think we can do better job at explaining how this work. (i think cryptographers should understand this, but other colleague might find harder time to digest this)

concretely, let's not lay out so many conditional logic as if explaining the code's various `if` checks. Instead, walk through the scenario and explain the protocol. Feel free to edit on top of the following:

```
## Proof of Incorrect Encoding
When the disperser is malicious, he can disperse an incorrectly encoded block, resulting in a merkle root of a Merkle tree containing invalid share (i.e. inconsistent with shares from correctly encoded block).
Disperser would disperse them to all replicas with valid Merkle proof against this incorrect root, or else the replicas won't even vote if the merkle proof is wrong.
By the time of reconstruction, replicas can come together with `> threshold` shares to interpolate back the original block (in polynomial form), and by recomputing the corresponding encoded block on this recovered polynomial, we can derive another merkle root of encoded shares.
If the merkle root matches the one dispersed earlier, then replicas know the encoding was correct. 
If not, then this mismatch can serves as a proof of incorrect encoding.

In short, the proof contains the recovered poly (from the received shares) and the merkle proofs (against the wrong root) being distributed by the malicious disperser. 
```

vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl AvidMScheme {
+    /// Generate a proof of incorrect encoding
+    /// See [`MalEncodingProof`] for details.
+    pub fn proof_of_incorrect_encoding(
+        param: &AvidMParam,
+        commit: &AvidMCommit,
+        shares: &[AvidMShare],
+    ) -> VidResult<MalEncodingProof> {
```
alxiong, line 49: I also wonder if we should further mention that caller of this code needs to check authenticity of `shares` and `commit`, i.e. whether they are signed by the disperser (or contains in the same VID proposal), since we don't check this internally.

alxiong (commented):
The logic looks correct to me, left some mostly doc or ergonomic comments.

vid/src/avid_m/proofs.rs
mrain: We did check that the shares should correspond to the commit.

vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut visited_indices = HashSet::new();
+        for share in shares {
+            for (index, mt_proof) in share
+                .content
+                .range
+                .clone()
+                .zip(share.content.mt_proofs.iter())
+            {
+                if index > param.total_weights {
+                    return Err(VidError::Argument("Invalid share".to_string()));
```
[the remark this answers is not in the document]
mrain: Done in [4ccb13a](https://github.com/EspressoSystems/espresso-network/pull/2803/commits/4ccb13ae68f4f8ca7fbe573420f571a3309e0ffc)

vid/src/avid_m/proofs.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Verify a proof of incorrect encoding
+    pub fn verify(
+        &self,
+        param: &AvidMParam,
+        commit: &AvidMCommit,
+    ) -> VidResult<VerificationResult> {
+        // First check that all shares are valid.
+        if self.raw_shares.len() < param.recovery_threshold {
+            return Err(VidError::Argument(
+                "Insufficient shares to generate the proof of incorrect encoding.".to_string(),
```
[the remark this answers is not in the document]
mrain: Done in [4ccb13a](https://github.com/EspressoSystems/espresso-network/pull/2803/commits/4ccb13ae68f4f8ca7fbe573420f571a3309e0ffc)

vid/src/avid_m/proofs.rs
mrain: Agree. Done in [4ccb13a](https://github.com/EspressoSystems/espresso-network/pull/2803/commits/4ccb13ae68f4f8ca7fbe573420f571a3309e0ffc)

vid/src/avid_m/proofs.rs
mrain: Used your paragraph in [4ccb13a](https://github.com/EspressoSystems/espresso-network/pull/2803/commits/4ccb13ae68f4f8ca7fbe573420f571a3309e0ffc)

vid/src/avid_m/proofs.rs
mrain: Done also in [8e6a8d6](https://github.com/EspressoSystems/espresso-network/pull/2803/commits/8e6a8d63a2972ff020999dc8f9cc84381ac03cf8)

alxiong (approved):
LGTM, thanks!

PR #2807, opened 2025-03-20
CI: workaround docker multiplatform limitations

sveitser:
Saving and loading multiplatform images with docker is not easy and the suggestion is to use `containerd` instead. This PR works around it by following different paths for PR and non-PR build. 

- PR: build single platform image, export as artifact and load, the docker tag will be of the form `pr-123`
- non-PR: build multiplatform image, push to registry, pull from registry, the docker tag will the branch name for a manually triggered run, and I expect it will be the `tag` if the build is triggered via `tag`.

In addition to that it uses a `DOCKER_TAG` env var and extracts the tag from the metadata build step. That way we can set the correct tag via GITHUB_ENV and pull the correct images and run the demo with those.

Manually triggered run: https://github.com/EspressoSystems/espresso-network/actions/runs/13977225355


.github/workflows/build.yml
```diff
@@ -189,25 +194,39 @@ jobs:
         with:
           images: ghcr.io/espressosystems/espresso-sequencer/${{ matrix.service }}
 
-      - name: Build and push ${{ matrix.service }} docker image
+      # There is no straightforward way to import a multiplatform image from a tar file with with
```
sveitser, line 197:
```suggestion
      # There is no straightforward way to import a multiplatform image from a tar file with
```

rob-maron (approved):

PR #2809, opened 2025-03-20
Epoch Version Fixes

tbro:
Pulled from #2526


tbro:
Current status:
When running upgrade test like so:
```bash
cargo nextest run -p sequencer  test_pos_upgrade_view_based --retries 0 --verbose --nocapture
```

  I see a lot of these:
```bash
2025-03-20T19:51:14.630290Z ERROR Transaction task{id=3 view=131 epoch=1}: espresso_types::v0::impls::stake_table: We are missing the randomized committee for epoch 1
```

bfish713:
> Current status: When running upgrade test like so:
> 
> ```shell
> cargo nextest run -p sequencer  test_pos_upgrade_view_based --retries 0 --verbose --nocapture
> ```
> 
> I see a lot of these:
> 
> ```shell
> 2025-03-20T19:51:14.630290Z ERROR Transaction task{id=3 view=131 epoch=1}: espresso_types::v0::impls::stake_table: We are missing the randomized committee for epoch 1
> ```

I believe this is because we do not set the INITIAL_DRB_RESULT for the first few epochs, and thus do not have a randomized committee

PR #2815, opened 2025-03-21
Deploy to stake table contract for local demo

sveitser:
- Add `stake` oneshot service.
- Add configuration to specify initial token grant recipient when deploying the token contract.
- Use dev mnemonic accounts 10 to 14 for validators 1 to 5
- Fund the validators eth accounts with ESP tokens.
- Register the validators.
- Make the validator delegate to themselves.
- Add a CLI test for the stake-for-demo command.


imabdulbasit (approved):

staking-cli/src/demo.rs
```diff
@@ -78,8 +106,12 @@ pub async fn stake_for_demo(config: &Config) -> Result<()> {
         .await?;
         assert!(receipt.status());
 
-        let receipt = delegate(stake_table, validator_address, amount).await?;
+        tracing::info!(
+            "delegate {delegate_amount_esp} ESP for validator {val_index} from {validator_address}"
+        );
+        let receipt = delegate(stake_table, validator_address, delegate_amount).await?;
```
imabdulbasit, line 112: so is this delegating to itself? can we also have multiple delegators like some random ethereum wallets delegating their stake to each validator.

PR #2819, opened 2025-03-21
Replace sequencer version in espresso-dev-node

Ayiga:
### This PR:

Modifies the Espresso dev node to use the `FeeVersion` of the sequencer instead of the `MarketplaceVersion`.  This change is to try and more closely match our current environments rather than one that is TBD.


sveitser (approved):

Ayiga:
The `espersso-dev-node` has already been updated with the POS version making this change unnecessary now. Thanks for the updated `espresso-dev-node` images!

PR #2824, opened 2025-03-23
Refactor DRB to happen when Epoch Root is Decided

bfish713:
### This PR:

Move the logic of DRB calculation to immediately after we add the epoch root.  This makes the code much simpler and easier to reason about. Now the DRB result is set only from the DRB task when it completes.  Spawning this task after the root is decided and added to the membership also has the benefit of guaranteeing that the stake table exists in the membership when the DRB calculation completes.

This also makes it so that members that are not part of the future epoch still calculate the DRB, but I don't see much downside to this.  If their node stays online they need the DRB result to follow consensus, and if they just shut off then the calculation will cancel.  

### This PR does not:

This does not alter the DRB protocol, nor change any functionality except to make it more robust and the ordering of things more clear.  

### Key places to review:

The new DRB task spawning and that it's spawned appropriately after deciding a root

Running the hotshot tests, I also was seeing notable improvements to the pos demo with this change.


hotshot-task-impls/src/helpers.rs
```diff
@@ -180,13 +181,57 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+    let storage = storage.clone();
+    let consensus = consensus.clone();
+    tokio::spawn(async move {
+        let drb_result = tokio::task::spawn_blocking(move || {
+            hotshot_types::drb::compute_drb_result::<TYPES>(seed)
+        })
+        .await
+        .unwrap();
+        let mut consensus_writer = consensus.write().await;
+        consensus_writer.drb_results.store_result(epoch, drb_result);
```
ss-es, line 222: does it make sense to maybe move the consensus `store_result` call into `handle_drb_result`?

hotshot-task-impls/src/helpers.rs
```diff
@@ -180,13 +181,57 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+    epoch: TYPES::Epoch,
+    membership: &Arc<RwLock<TYPES::Membership>>,
+    storage: &Arc<RwLock<I::Storage>>,
+    consensus: &OuterConsensus<TYPES>,
+) {
+    let membership = membership.clone();
+    let storage = storage.clone();
+    let consensus = consensus.clone();
+    tokio::spawn(async move {
+        let drb_result = tokio::task::spawn_blocking(move || {
```
ss-es, line 216: I'm a bit confused -- this is a `tokio::spawn` with a `spawn_blocking` inside but all the handles get dropped? how can we possibly cancel/garbage collect this calculation

hotshot-task-impls/src/helpers.rs
```diff
@@ -219,6 +264,31 @@ async fn decide_epoch_root<TYPES: NodeType, I: NodeImplementation<TYPES>>(
             let mut membership_writer = membership.write().await;
             write_callback(&mut *membership_writer);
         }
+        // Cancel old DRB computation tasks.
+        let mut consensus_writer = consensus.write().await;
+        consensus_writer
+            .drb_results
+            .garbage_collect(next_epoch_number);
+        drop(consensus_writer);
```
ss-es, line 272: doesn't this just garbage collect the results not the tasks?

hotshot-task-impls/src/helpers.rs
```diff
@@ -219,6 +264,31 @@ async fn decide_epoch_root<TYPES: NodeType, I: NodeImplementation<TYPES>>(
...
+
+        // Store the DRB seed input for the epoch after the next one.
```
ss-es, line 274: I think this is no longer storing the seed, it's just spawning the task

ss-es (commented):
I think in general this makes sense, with some minor comments

the major concern I have before approving is I don't see how old DRB calculation tasks get cancelled (e.g. when we receive a decide) because the computation tasks don't have their `JoinHandle`s stored anywhere. maybe I missed something

hotshot-task-impls/src/quorum_vote/handlers.rs
```diff
@@ -532,9 +339,9 @@ pub(crate) async fn handle_quorum_proposal_validated<
         tracing::debug!("Successfully sent decide event");
 
         if version >= V::Epochs::VERSION {
-            // `leaf_views.last()` is never none if we've reached a new decide, so this is safe to
-            // unwrap.
-            store_drb_seed_and_result(task_state, &leaf_views.last().unwrap().leaf).await?;
+            for leaf_view in leaf_views {
```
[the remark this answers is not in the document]
bfish713: Yes they are all decided

hotshot-task-impls/src/helpers.rs
bfish713: I don't really think cancelation is what we want here.  During the calculation, we can't cancel anyway, and at the point the calculation is finished we might as well store the result.  Cancellation only works at await points so I think no matter what the spawn blocking part will finish.

hotshot-task-impls/src/helpers.rs
bfish713: yeah need to update the comment

hotshot-task-impls/src/helpers.rs
```diff
@@ -180,13 +181,57 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+    drb_result: DrbResult,
+) {
+    tracing::debug!("Calling add_drb_result for epoch {:?}", epoch);
+    if let Err(e) = storage
+        .write()
+        .await
+        .add_drb_result(epoch, drb_result)
+        .await
+    {
+        tracing::error!("Failed to store drb result for epoch {:?}: {}", epoch, e);
```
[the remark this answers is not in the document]
bfish713: Yeah this only logs if storage failes

hotshot-task-impls/src/helpers.rs
bfish713: tokio docs [here](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html.) mention the spawn blocking piece can't be cancelled.  Idk exactly how this relates to cancelling a task which calls spawn_blocking but I think it won't cancel until the blocking part completes anyway

ss-es: can we not add a `yield_now` or something in the `compute_drb_result` loop? e.g. every 10000 iterations we yield to allow the executor the opportunity to cancel the calculation

my worry is that the drb calculation when deployed will take (potentially) hours, and we are now spawning very long-running tasks and leaking them. I don't know what specifically will go wrong but it doesn't feel right

I think we will eventually also want to add checkpoints to the calculation so we can resume it from storage, which will probably accomplish the same thing

ss-es: I guess we can merge this PR without this, and just add that in the future?

bfish713: We can't yield within the spawn blocking, we can only do this if we run it directly on the executor thread, which we don't want to do.  Checkpointing the calculation would make sense.  Without this PR we are also leaking this calculation we can't cancel calculation (no await points) and are only dropping a join handle which doesn't just leaks the task anyway.

ss-es (approved):
looks good to me

I've created an issue on Asana for the task cancellation

PR #2829, opened 2025-03-24
Fix QC Inconsistencies between epochs

bfish713:
### This PR:

The goal of this PR is to treat next_epoch_qc and high_qc as one, as much as possible.  To do this we add block number to the vote and QC to make it easy to tell when a double QC is required.  We then try to only update high QC when it's for the last block when it's accompanied by the next epoch qc.

### Key places to review:

Most changes are meaningful but the ones in in `quorum_proposal_recv` and `quorum_proposal` are notable


hotshot-task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -151,46 +147,62 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
+                }
+            }
+        }
+
         // TODO configure timeout
         while self.view_start_time.elapsed() < wait_duration {
-            let Some(time_spent) = Instant::now().checked_duration_since(self.view_start_time)
+            let time_spent = Instant::now()
+                .checked_duration_since(self.view_start_time)
+                .ok_or(error!("Shouldn't be possible, now must be after the start"))?;
```
ss-es, line 188:
```suggestion
                .ok_or(error!("Time elapsed since the start of the task is negative. This should never happen."))?;
```

hotshot-task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -151,46 +147,62 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
-            };
-            let Ok(maybe_qc) = tokio::time::timeout(
-                time_left,
-                self.wait_for_qc_event(&mut self.receiver.clone()),
-            )
-            .await
-            else {
-                // we timeout out, don't wait any longer
-                return;
+                tracing::info!("we timeout out, don't wait any longer");
```
ss-es, line 195:
```suggestion
                tracing::info!("Some nodes did not respond with their HighQc in time. Continuing with the highest QC that we received: {highest_qc}");
```

hotshot-task-impls/src/quorum_proposal/mod.rs
```diff
@@ -382,6 +385,101 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
...
+
+                if !self.upgrade_lock.epochs_enabled(view_number).await
+                    || !is_last_block_in_epoch(block_number, self.epoch_height)
+                {
+                    return self.update_high_qc(qc).await;
+                }
+                let Some(next_epoch_qc) =
+                    self.formed_extended_quorum_certificates.get(&view_number)
+                else {
+                    return Ok(());
```
ss-es, line 452: why do we return `Ok(())` if the eqc for the view number doesn't exist? maybe I misunderstood the description of the function

ss-es: I guess maybe my question is: is this function really intended to "check" anything or is it just wrapping a bunch of storage logic?

ss-es (commented):
some comments/questions

hotshot-types/src/message.rs
```diff
@@ -239,7 +239,10 @@ pub enum GeneralConsensusMessage<TYPES: NodeType> {
     ProposalResponse2(Proposal<TYPES, QuorumProposal2<TYPES>>),
 
     /// Message for the next leader containing our highest QC
-    HighQc(QuorumCertificate2<TYPES>),
+    HighQc(
+        QuorumCertificate2<TYPES>,
+        Option<NextEpochQuorumCertificate2<TYPES>>,
+    ),
```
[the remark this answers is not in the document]
bfish713: Yes it's part of hotstuff 2 protocol so only after the upgrade

hotshot-task-impls/src/quorum_proposal/mod.rs
bfish713: The intent of the function is this, given either next_epoch_qc or qc store the next epoch QC and QC if we have both of them for the same view and commit.  Also if given a QC store it if we don't/shouldn't have an eqc

We return Ok(()) here because we don't want to store anything (we have QC but still need next epoch qc) and it's not really an error (this function should only error if storage fails)

hotshot-task-impls/src/consensus/handlers.rs
```diff
@@ -205,9 +205,38 @@ pub async fn send_high_qc<TYPES: NodeType, V: Versions, I: NodeImplementation<TY
...
+                    .await
+                    .next_epoch_high_qc()
+                    .ok_or(warn!(
+                        "High QC is for the last block but we don't have next epoch QC."
+                    ))?
+                    .clone(),
+            )
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
bfish713: This isn't a proposal, it's a QC, the nodes are signing the `block_number`

PR #2836, opened 2025-03-24
Fix view number for staged upgrade certificate

ss-es:
### This PR:
Changes the view number on the staged upgrade certificate to a future view, beyond the upgrade block. This is required because, otherwise, the upgrade certificate might apply retroactively and result in an inconsistent state for older views


bfish713 (commented):
Holding off on review/approval because the upgrade tests are failing

PR #2839, opened 2025-03-24
Add a tool for compiling stats about a namespace

jbearer:
Sometimes for marketing purposes we might want to get aggregated statistics about a rollup/namespace. Eventually we should have indexes in the DB that makes this easy through an endpoint or the block explorer. Until then, here's a tool to manually go through a range of blocks and inspect a certain namespace.

### Example usage

Counting statistics for Rari deployment:

```
$ utils ns-aggregator -f 999258 -n 1380012617 -j 8 https://query.main.net.espresso.network/v0
Scanned range [999258, 1733648) for namespace 1380012617
103087 transactions
566669054 bytes
```

### DIFF
--- a/sequencer/src/bin/utils/ns_aggregator.rs
+++ b/sequencer/src/bin/utils/ns_aggregator.rs
@@ -1,26 +1,38 @@
 use std::cmp::min;
 
-use anyhow::ensure;
+use anyhow::{ensure, Context};
 use clap::Parser;
 use espresso_types::{Header, NamespaceId, NamespaceProofQueryData};
 use futures::{future::try_join_all, stream::StreamExt};
 use sequencer::SequencerApiVersion;
 use surf_disco::Url;
 
+/// Count transactions and bytes confirmed in a given namespace.
 #[derive(Debug, Parser)]
 pub struct Options {
+    /// Block to start counting from.
     #[clap(short, long, default_value = "0")]
     from_block: usize,
 
+    /// Last block to count (inclusive).
+    ///
+    /// If not specified, will count until the end of the chain at the moment the program is
+    /// invoked.
     #[clap(short, long)]
     to_block: Option<usize>,
 
+    /// Namespace to aggregate.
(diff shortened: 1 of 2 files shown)


sequencer/src/bin/utils/ns_aggregator.rs
```diff
@@ -0,0 +1,100 @@
...
+    let to = match opt.to_block {
+        Some(to) => to,
+        None => client.get("status/block-height").send().await?,
+    };
+    ensure!(to >= from, "to-block < from-block");
+
+    let tasks = (0..opt.jobs).map(|i| {
+        let chunk_size = (to - from) / opt.jobs;
+        let start = from + i * chunk_size;
+        let end = min(start + chunk_size, to);
```
fkrell, line 43: `to` here in inclusive. Edge case: to = 1, from = 0, jobs=2 (for whatever reason). start = end = 0, block 1 never accounted for. no?

jbearer: Ah `to` was meant to be exclusive here, but I guess that could be confusing. I'll change it to be inclusive and fix this

jbearer:
This requires `v1` APIs now since it's built against the latest VID types. That means it won't work on mainnet until the next release (before the PoS upgrade, though). Tested on Decaf with

```
$ utils ns-aggregator -f 3000000 -t 3100000 -n 1918988905 -j 8 https://query.decaf.testnet.espresso.network/v1
Scanned range [3000000, 3100001) for namespace 1918988905
3039 transactions
14911848 bytes
```

sequencer/src/bin/utils/ns_aggregator.rs
jbearer: [a419fc2](https://github.com/EspressoSystems/espresso-network/pull/2839/commits/a419fc21c5eb4468b7d0d04028f234d6d3d84b2f)

sveitser (approved):

PR #2843, opened 2025-03-25
2840 stake table upgradeability tests

### LINKED ISSUE #2840: stake table: upgradeability tests

alysiahuggins:
Closes #2840

### This PR:
* Adds a new upgradeable contract `StakeTableV2Test` with a new versioning scheme (`getVersion()`)
* Implements `StakeTableUpgradeTest`, which validates:
  - Successful upgrade via `upgradeToAndCall`
  - Version bump from 1 to 2
  - Admin-only upgrade restrictions
  - Reinitialization guards
  - Storage layout compatibility
* Adds a Node.js script (`compare-storage-layout.js`) that:
  - Extracts and normalizes storage layouts for old and new contracts
  - Checks layout compatibility using `forge inspect`

### This PR does not:
* Modify logic within the original `StakeTable` contract
* Handle future versions beyond `v2`- future upgrades will need additional test cases and layout validation

### Key places to review:
* `StakeTableUpgradeTest.sol` – contains all upgradeability test logic including:
  - Upgrade permission test
  - Version transition test
  - Reinitializer protection
  - Integration of `vm.ffi` + `compare-storage-layout.js`
* `contracts/test/script/compare-storage-layout.js` – normalize and compare storage layouts

 ### How to test this PR:  
Run `forge test --match-contract StakeTableUpgradeTest`

### Things tested:
* End-to-end upgrade process from `StakeTable` to `StakeTableV2Test`
* `initializeV2()` runs once with `reinitializer(2)`
* Custom errors for reinitialization and unauthorized upgrades are properly caught
* Manual test: `node contracts/test/script/compare-storage-layout.js StakeTable StakeTableV2Test` outputs `true` or fails with a clear diff


contracts/test/script/compare-storage-layout.js
```diff
@@ -0,0 +1,68 @@
+const { execSync } = require("child_process");
```
sveitser, line 1: It's a bit difficult to follow this script or be certain that it would always fail if there we made a mistake.

Are there now existing solution for checking?

I found for example https://github.com/Rubilmax/foundry-storage-check but that doesn't seem to be geared towards using a library and isn't published to npm.

contracts/test/StakeTable.t.sol
```diff
@@ -545,3 +547,137 @@ contract StakeTable_register_Test is Test {
...
+    }
+
+    function test_upgrade_reverts_when_not_admin() public {
+        address notAdmin = makeAddr("not_admin");
+
+        (uint8 majorVersion,,) = StakeTableV2Test(stakeTableRegisterTest.proxy()).getVersion();
+        assertEq(majorVersion, 1);
+
+        vm.startPrank(notAdmin);
+        try S(stakeTableRegisterTest.proxy()).upgradeToAndCall(address(new StakeTableV2Test()), "")
```
sveitser, line 599: What's the difference of this flow vs using `expectRevert`?

alysiahuggins: I think foundry's expectRevert() expects a direct revert from the call but since I'm interacting with the proxy, it's a delegate call and the the revert is isn't at the top level, so the expectRevert() wasn't picking it up.  I spent a lo t of time trying to figure this out. But i just realized that there's a way to circumvent this by adding 
`/// forge-config: default.allow_internal_expect_revert = true` at the top of the test function. So i'll update the tests with that as it's much simpler and easier to follow https://book.getfoundry.sh/cheatcodes/expect-revert

alysiahuggins: ok that doesn't work so will have to leave the current solution but will remove it where possible (e.g. the initialization reverts)

contracts/test/script/compare-storage-layout.js
alysiahuggins: true, i'll look for command line tools to help to see if that simplifies a lot more

contracts/test/StakeTable.t.sol
sveitser: I was really confused because this all works fine in the fee contract tests but not here so something seemed off. I don't know why it doesn't work but my guess is that if written as in the PR now the expect revert only applies to the deployment of the impl contract and that doesn't revert so we don't catch the expected revert.

alysiahuggins: thank you so so much for this! I had to remove multiple evm reads and writes from one program instruction. updated here https://github.com/EspressoSystems/espresso-network/commit/fc49aafca7104ae15dfeacbfc5c79d90767880a9

contracts/test/script/compare-storage-layout.js
alysiahuggins: to no avail with making it work with command line but i simplified the way the type field was cleaned and also compared to a contract that definitely does not have the same storage layout to prove that the script works as expected https://github.com/EspressoSystems/espresso-network/commit/6f1802f6a750d79000685ecaff9edfe8a3c88ea2

and also compared with "invalid" staketable contracts that either have a missing field or the fields reordered https://github.com/EspressoSystems/espresso-network/commit/8356f22661c6f4b7281dcd7e3e9c97b70ab1e824

sveitser (approved):

PR #2846, opened 2025-03-25
2513 add timelocks to the upgradable staketable contract

### LINKED ISSUE #2513: Add timelocks to the upgradable stakeTable contract
questions for product:
1. do you want separate timelocks for various contracts? (this allows for different delays, executors/proposers, overall isolate risk) 
2. which addresses for admin, proposer and executor of timelock?
3. what delay period?
4. do you want to renounce the admin role eventually as the admin of a timelock has the ability to modify the timelock contract without delays (e.g update proposers & executors list, Changing governance config)

alysiahuggins:
Closes #2513 

### This PR:
- Adds support for upgrades to StakeTable using TimelockController
- Updates the deployment script (DeployStakeTableScript) to inject timelockAddress during initialization
- Implements Foundry tests for scheduling and executing UUPS proxy upgrades via a Timelock
- Adds Timelock contract

### This PR does not:
- Include integration with off-chain proposal tooling or frontends

### Key places to review:
- test/StakeTableTimelockTest.sol
- contracts/test/script/StakeTable.s.sol

### How to test this PR: 
`forge test --match-contract StakeTableTimelockTest`

### Things tested 
- Upgrade fails if attempted before delay
- Upgrade fails if scheduled/executed by unauthorized proposer or executor
- Upgrade fails if attempted directly (bypassing Timelock)
- Upgrade fails if there is a salt mismatch
- Verifies that the timelock contract address is correctly stored on StakeTable

---
  - https://app.asana.com/0/0/1210125958075813


contracts/src/StakeTable.sol
```diff
@@ -17,7 +15,7 @@ using EdOnBN254 for EdOnBN254.EdOnBN254Point;
 /// @title Ethereum L1 component of the Espresso Global Confirmation Layer (GCL) stake table.
 ///
 /// @dev All functions are marked as virtual so that future upgrades can override them.
-contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpgradeable {
+contract StakeTable is Initializable, InitializedAt, UUPSUpgradeable {
```
sveitser, line 18: I wonder why not keep this contract as is and set the timelock as owner. When I read the doc string on the contract it sounds to me like this is an intended use case.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0a77e54c307d9becf0473b73541c4f9d3b2743a3/contracts/governance/TimelockController.sol#L12

alysiahuggins: ah i meant it as the owner and swapped onlyOwner() for onlyTimelock(), but you're right, i should be using the existing Ownable libraries,  i'll switch to that, thank you!

alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/2846/commits/b3c1b2f79fbb549bd1f75e6b13884e53c1609375

contracts/test/StakeTable.t.sol
```diff
@@ -755,3 +763,340 @@ contract StakeTableUpgradeTest is Test {
...
+
+        vm.assertTrue(timelock.isOperationReady(txId));
+
+        vm.startPrank(executors[0]);
+        timelock.execute(address(proxy), 0, data, bytes32(0), bytes32(0));
+        vm.stopPrank();
+        vm.assertTrue(timelock.isOperationDone(txId));
+    }
+
+    function test_expect_revert_when_timelock_is_not_owner() public {
```
alxiong, line 850: i thought the convention for case is `test_RevertWhen_TimelockIsNotOwner()`, as described in [pointer 6 here](https://book.getfoundry.sh/guides/best-practices#general-test-guidance)

alxiong (commented):
i'm not sure how many of these tests are redundantly testing OZ's own library rather than our own logic.

It seems to me that all we did here is deploy an OZ timelock, and insert its address as our owner. I'm slightly inclined to think that these tests are not doing much. (an analgy is we don't need to test `Ownable`'s behavior, because we use OZ's lib as a taken)

contracts/src/Timelock.sol
```diff
@@ -0,0 +1,15 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/governance/TimelockController.sol";
+
+/// @title TimelockController
+/// @notice A timelock controller for the StakeTable contract
+contract Timelock is TimelockController {
```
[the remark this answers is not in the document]
alysiahuggins: yea if not additional logic is needed, then no need to create a contract. We can just deploy the timelockcontroller. But if we do plan to extend in the future, a separate contract is needed.

Any preferences? @sveitser @0xkato

contracts/test/StakeTable.t.sol
alysiahuggins: Thanks for pointing that out, i wasn't aware of the convention.

alxiong: tbh, i'm defintely not a fan of these `mixedNotSnake_NotCamel_WhimsicalLooking` suggestions, but i still tend to follow them... 🤦

contracts/src/Timelock.sol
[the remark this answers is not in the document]
0xkato: If we are not planning on having additional logic we should just keep it simple and inherit it imo.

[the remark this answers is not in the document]
alysiahuggins: so doing a forge bind with multiple contract directory locations isn't available so i left it for a simpler forge bind set up, added logic to deploy it via rust as well https://github.com/EspressoSystems/espresso-network/commit/79fb17e595e6634f8b6ef49d2819ad0e546170e5

contracts/test/StakeTable.t.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/a6d8d29351be37c97c34957fab18ff9e0adf0437

contracts/src/Timelock.sol
```diff
@@ -0,0 +1,15 @@
...
+    constructor(
+        uint256 minDelay,
+        address[] memory proposers,
+        address[] memory executors,
+        address admin
+    ) TimelockController(minDelay, proposers, executors, admin) { }
```
alxiong, line 14: if it's the same, we don't need to re-declare it, right?

alxiong (commented):
contract and deployer LGTM.

contracts/src/StakeTable.sol
```diff
@@ -146,6 +146,9 @@ contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpg
     /// A zero amount would lead to a no-op.
     error ZeroAmount();
 
+    /// The caller is not the timelock
+    error NotTimelock();
```
sveitser, line 150: This looks like it's unused now?

contracts/test/StakeTable.t.sol
```diff
@@ -838,3 +842,359 @@ contract StakeTableUpgradeTest is Test {
...
+            )
+        );
+        timelockController.execute(address(stakeTable), 0, data, wrongSalt, bytes32(0));
+        vm.stopPrank();
+    }
+
+    function test_UnauthorizedCannotUpgrade() public {
+        vm.startPrank(makeAddr("notAdmin"));
+
+        try stakeTable.upgradeToAndCall(address(new S()), "") {
```
sveitser, line 1075: Similar to a previous PR, do we need `try / catch` only because we have a deployment inside the function argument statement?

contracts/src/StakeTable.sol
alysiahuggins: ah yes thanks https://github.com/EspressoSystems/espresso-network/commit/ca73a8188320ee73979ab5d349ebf1438097f230

contracts/test/StakeTable.t.sol
```diff
@@ -838,3 +842,359 @@ contract StakeTableUpgradeTest is Test {
...
+                selector := mload(add(lowLevelData, 32))
+            }
+            assertEq(selector, OwnableUpgradeable.OwnableUnauthorizedAccount.selector);
+        }
+    }
+
+    function test_AdminCanGrantRolesWithoutDelay() public {
+        vm.startPrank(timelockAdmin);
+        timelockController.grantRole(timelockController.PROPOSER_ROLE(), timelockAdmin);
+        timelockController.grantRole(timelockController.EXECUTOR_ROLE(), timelockAdmin);
```
sveitser, line 1090: Wondering if there are no asserts on purpose or if they were forgotten.

utils/src/deployer.rs
```diff
@@ -993,4 +1062,34 @@ mod tests {
...
+            admin,
+        )
+        .await?;
+
+        // Verify deployment
+        let timelock = Timelock::new(timelock_addr, &provider);
+        assert_eq!(timelock.getMinDelay().call().await?._0, min_delay);
+
+        // there are assert statements in the deploy_timelock function that test the roles (proposers, executors, admin)
+        // so we don't need to test them again here
```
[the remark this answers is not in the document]
alysiahuggins: good idea https://github.com/EspressoSystems/espresso-network/commit/57a97b693b95ef0e5b947435e2dd071c560ff5a3

contracts/test/StakeTable.t.sol
```diff
@@ -838,3 +842,359 @@ contract StakeTableUpgradeTest is Test {
...
+        bytes32 txId =
+            timelockController.hashOperation(address(stakeTable), 0, data, bytes32(0), bytes32(0));
+
+        timelockController.schedule(address(stakeTable), 0, data, bytes32(0), bytes32(0), DELAY);
+
+        vm.stopPrank();
+
+        vm.assertFalse(timelockController.isOperationReady(txId));
+
+        vm.warp(block.timestamp + DELAY + 1);
```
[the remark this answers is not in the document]
alysiahuggins: ah that's fair https://github.com/EspressoSystems/espresso-network/commit/8c2ba6ff3b206489da33b39a0bfd68d40b44b62b

contracts/test/StakeTable.t.sol
alysiahuggins: ah must have been forgetten https://github.com/EspressoSystems/espresso-network/commit/52411e315d12816a7ee2e8b3ae31964194b28dcd

contracts/src/Timelock.sol
alxiong: i know we need this contract to inherit, but we don't need to declare this constructor? if it's the same as the parent contract. 

i guess it doesn't hurt. so nvm my comment.

sveitser: I don't quite understand this. If I change

```diff
diff --git a/justfile b/justfile
index feecdd883..298ebbe92 100644
--- a/justfile
+++ b/justfile
@@ -143,7 +143,7 @@ build-docker-images:
     scripts/build-docker-images-native

 # generate rust bindings for contracts
-REGEXP := "^LightClient(V\\d+)?$|^LightClientArbitrum(V\\d+)?$|^FeeContract$|PlonkVerifier(V\\d+)?$|^ERC1967Proxy$|^LightClient(V\\d+)?Mock$|^StakeTable$|^EspToken$|^Timelock$"
+REGEXP := "^LightClient(V\\d+)?$|^LightClientArbitrum(V\\d+)?$|^FeeContract$|PlonkVerifier(V\\d+)?$|^ERC1967Proxy$|^LightClient(V\\d+)?Mock$|^StakeTable$|^EspToken$|^TimelockController$"
 gen-bindings:
     # Update the git submodules
     git submodule update --init --recursive
```

I do get bindings for the TimelockController.

sveitser: We also do this for `ERC1967Proxy`.

utils/src/deployer.rs
```diff
@@ -993,4 +1062,58 @@ mod tests {
         assert_eq!(stake_table.lightClient().call().await?._0, lc_addr);
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_deploy_timelock() -> Result<()> {
```
sveitser, line 1067: This function currently isn't hooked up in `deploy.rs`. So it will never be called. Is this intentional?

utils/src/deployer.rs
```diff
@@ -652,6 +660,67 @@ pub async fn is_contract(provider: impl Provider, address: Address) -> Result<bo
...
 }
 
+/// Deploy and initialize a Timelock contract
+///
+/// Parameters:
+/// - `min_delay`: The minimum delay for operations
+/// - `proposers`: The list of addresses that can propose
+/// - `executors`: The list of addresses that can execute
+/// - `admin`: The address that can perform admin actions
+pub async fn deploy_timelock(
```
sveitser, line 670: This function currently isn't hooked up in deploy.rs. So it will never be called. Is this intentional?

alysiahuggins: yes that was intention since it requires some product input, so this comment that briefly mentions it https://github.com/EspressoSystems/espresso-network/pull/2846#discussion_r2072760335

sveitser (approved):

Issue #2536, opened 2025-02-05
Resolve issue with alloy types serialization in header


sveitser:
See referenced comments below.

PR #2848, opened 2025-03-26
Migrate to alloy and Contract upgrade script in rust

### LINKED ISSUE #2142: Migrate from ethers-rs to alloy-rs
ASAP, we should remove `ethers-rs`. See doc: https://alloy.rs/migrating-from-ethers/reference.html

cc @sveitser @philippecamacho @alysiahuggins @Sneh1999 

Reminder for subtasks
- update gas oracle (from blocknative in `ethers-rs` to alloy-recommended); also add support for Arbitrum mainnet beying just ETH L1, since we are deploying a clone there too for more frequent update. 

- #2456 

### Acceptance criteria:
- [ ] ethers-rs is no longer used as a dependency

alxiong:
Closes #2142, #2535, #2536, #2555

### This PR:

Apparently, this PR is so monstrous that it's impossible for a reviewer to check everything. Honestly, I don't know how to do it any other way, or how to make reviewers' life easier. But here's my notes while I was working on it, which contains some key changes made: 

CHANGELOG

- deleted Anvil and AnvilOption in favor of using alloy's builtin
- `deployer.rs::deploy()` is bloated and multiplexed, I modularize them into separate functions with tests of their success
  - removed `enum ContractGroup`, avoid multiplexing, use separate option flags instead
- `init_signer()` contains trivial logic that I decide to push to site of usage/edge, to save another layer of indirection
- primitives::U256 has to be replaced with alloy::primitives::U256 for consistency
- NodeInfo can be directly translated from StakeTableEntry, avoid unnecessary and multiple intermediate structs and conversion between them
- bindings are now stored as module in adapter, which enables deriving traits for them for easily. (instead of living in a foreign crate)
- L1Client now holds a provider and the inner transport used to instantiate the provider, so that we placate the Provider<N: Network> and use SwitchingTransport details
- new PrimitiveSignature for builder signature has new serialization logic: https://docs.rs/alloy-primitives/0.8.22/src/alloy_primitives/signature/primitive_sig.rs.html#415
- U256, B256, Address all have different serialized bytes, and I introduce `alloy-compat` crate to bridge over, and use `serde(with)` for backward compat serialization on fields of outer struct
  ```
     "builder_signature": {
       "r": "0xa66e49fb371b83d36a6f326b95eb6146c397f3b037fd04b5f3c973d0e5624105",
       "s": "0x43c6d12317c58dd8f831d34e2fa64bba04f51374ef7fce826121ba5dc5df64cf",
  <    "v": 27
  >    "v": "0x0",
  >    "yParity": "0x0"
     },
  ```
- update read_contract and submit_state_and_proof to V2 apis
- update hotshot config fetching logic, instead of reqwest, use surf_disco and rely on auto-redirect to the correct version. removed local declaration of PublicNetworkConfig, import from types/
- removed sequencer/genesis.rs unit tests about fee contract not being a proxy, since thoese are already tested in `deployer.rs` (see comments there)
- sequencer/bin/deploy.rs: options about ContractGroup is changed into --deploy-[fee|stake-table|light-client-v1] etc. (updated docker compose files as well)

### This PR doesn't:

based off `main`, thus need to resolve more conflicts once we get enough review on the diff.


Cargo.toml
```diff
@@ -8,14 +8,14 @@ edition = "2021"
 resolver = "2"
 
 members = [
+    "alloy-compat",
+    "alloy-compat"
+,
```
sveitser, line 13:
```suggestion
    "alloy-compat",
```

alxiong: done in [b8b6804](https://github.com/EspressoSystems/espresso-network/pull/2848/commits/b8b6804089b90200a932e0012c0390c4b5f3ed50)

utils/src/deployer.rs
```diff
@@ -113,565 +119,603 @@ impl Contracts {
...
-                    l1.clone(),
-                    (fee_contract.address().to_ethers(), data.to_ethers()),
-                )?,
-            )
+    // instantiate a proxy instance, cast as LightClient's ABI interface
+    let lc_proxy = LightClient::new(lc_proxy_addr, &provider);
+
+    // set permissioned prover
+    if let Some(prover) = prover {
+        tracing::info!(%lc_proxy_addr, %prover, "Set permissioned prover ");
```
sveitser, line 260: @alxiong in the native demo this fails with `    /**Custom error with signature `OwnableUnauthorizedAccount(address)` and selector `0x118cdaa7`.`

So I think if we set another admin we need to do it after we set the permissioned prover.

alxiong: > if we set another admin we need to do it after we set the permissioned prover.

But that's what I do atm, see: https://github.com/EspressoSystems/espresso-network/blob/79c95ba8382ced91553484167edb86ad92c72fd9/sequencer/src/bin/deploy.rs#L223-L237

I set the permissioned prover with deployer (inside `deployer::deploy_light_client_proxy()`), then I did the `transfer_ownership(multisig)`. 
^^ this should be correct, right?

alxiong: when I look into the log, we panic at the fee contract ownership transfer (as I PM'ed you on Zulip), not light client issue.

alxiong:
As discussed on Zulip, I will first merge this to #2706, sync with main and resolve all conflicts, before fixing the remaining `demo-native` failure. (before finally merging into main)

PR #2851, opened 2025-03-26
[query-service] version gating the VidCommon endpoint

mrain:
Closes #<ISSUE_NUMBER>

### This PR:
- The old availability shall return the old VID common struct `ADVZCommon`
- Starting from api version 1.0.0, it returns the new enum type `VidCommon`

### DIFF
--- a/hotshot-query-service/src/availability.rs
+++ b/hotshot-query-service/src/availability.rs
@@ -1055,6 +1055,328 @@ mod test {
         network.shut_down().await;
     }
 
+    async fn validate_old(client: &Client<Error, MockBase>, height: u64) {
+        // Check the consistency of every block/leaf pair.
+        for i in 0..height {
+            // Limit the number of blocks we validate in order to
+            // speeed up the tests.
+            if ![0, 1, height / 2, height - 1].contains(&i) {
+                continue;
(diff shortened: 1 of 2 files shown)


hotshot-query-service/src/availability.rs
```diff
@@ -340,6 +384,58 @@ where
...
+                        }
+                        .boxed()
+                    })
+                    .await
+            }
+            .try_flatten_stream()
+            .boxed()
+        })?;
+    } else {
+        api.at("get_vid_common", move |req, state| {
```
tbro, line 422: Can we add a test for the new endpoint. Currently there is a `test_api` test which tests all the old version (in conjunction w/ other endpoints).  But I don't think the new endpoints are tested in any way.

mrain: I thought the `test_api` is testing the new version "1.0.0" but not the old ones.

mrain: tests for old apis is done in [3068a50](https://github.com/EspressoSystems/espresso-network/pull/2851/commits/3068a50139b1c58409d79a3baa29d28b5d34294f)

PR #2856, opened 2025-03-27
Small trait QOL changes

rob-maron:
Breaking this out as part of my request/response PR. This lets us `Box/Dyn` the storage. Does not touch any of the logic


tbro (approved):
I think if CI passes it should be good.

pls148 (approved):

PR #2857, opened 2025-03-27
Export a proper library from `staking-cli`

tbro:
`staking-cli` has a lot of functions we can use in `TestNetwork`. In order to use them we need it to act as a library. This PR also adds the lib to `sequencer` crate for use in follow up PRs.


sveitser:
If we want to use code in other crates can we move the code that is needed elsewhere to the contract adapter crate instead, or wherever else it makes sense? I think it would be unexpected for the sequencer crate to depend on the staking CLI crate.

staking-cli/src/lib.rs
```diff
@@ -1,100 +1,26 @@
...
-mod registration;
+pub mod parse;
+pub mod registration;
 
-#[cfg(any(test, feature = "testing"))]
 pub mod deploy;
 
 pub const DEV_MNEMONIC: &str = "test test test test test test test test test test test junk";
 
-#[derive(Parser)]
```
sveitser, line 39: What do we gain from moving this code out of lib.rs?

tbro: This is a library now. Libraries don't have `main` functions.

tbro:
> If we want to use code in other crates can we move the code that is needed elsewhere to the contract adapter crate instead, or wherever else it makes sense? I think it would be unexpected for the sequencer crate to depend on the staking CLI crate. Changes in staking-cli can then break the sequencer crate, or at least sequencer tests.

It's only unexpected b/c the crate has `cli` in the name. It is very common for libraries to ship clis.

staking-cli/src/lib.rs
sveitser: So we can call the function something else instead of moving the code?

sveitser:
As I said, we have a contract adapter crate I think we should put code to interact with contracts that is used in multiple crates there. Or @alxiong let me know if you have other plans after the alloy refactor.

sveitser (approved):
Some concerns. But OK.

PR #2867, opened 2025-03-28
make fetch timeout configurable

imabdulbasit:


sequencer/src/api/options.rs
```diff
@@ -571,6 +572,8 @@ pub struct Query {
     /// Peers for fetching missing data for the query service.
     #[clap(long, env = "ESPRESSO_SEQUENCER_API_PEERS", value_delimiter = ',')]
     pub peers: Vec<Url>,
+    #[clap(long, env = "ESPRESSO_SEQUENCER_FETCH_TIMEOUT", value_parser = parse_duration)]
+    pub fetch_timeout: Option<Duration>,
```
ss-es, line 576: is there a reason for this to be an `Option` and not just a hard limit?

imabdulbasit: so it this env is not set, we use the default availability options which has 500ms fetch timeout

PR #2868, opened 2025-03-28
Req/resp state catchup

rob-maron:


sequencer/src/api.rs
```diff
@@ -2171,11 +2155,215 @@ mod test {
...
+            .filter(|event| future::ready(matches!(event.event, EventType::Decide { .. })))
+            .take(3)
+            .collect::<Vec<_>>()
+            .await;
+
+        tracing::error!("restarting node");
+        let node = network
+            .cfg
+            .init_node(
+                1,
```
rob-maron, line 2362: This diff also got messed up. The deal here is that I duplicated the catchup tests and am running them without `state_peers` (to test the new catchup)

sequencer/src/lib.rs
```diff
@@ -554,6 +573,8 @@ pub async fn init_node<P: SequencerPersistence + MembershipPersistence, V: Versi
         validator_config,
         coordinator,
         instance_state,
+        storage,
+        state_catchup_providers,
```
rob-maron, line 577: We need these because later we add the `RequestResponseProtocol` (after we have constructed it)

sequencer/src/request_response/catchup/state.rs
```diff
@@ -0,0 +1,351 @@
...
+                        .with_context(|| format!("response was missing account {account}"))?;
+                    proof
+                        .verify(&fee_merkle_tree_root)
+                        .with_context(|| format!("invalid proof for account {account}"))?;
+                    proofs.push(proof);
+                }
+
+                Ok(proofs)
+            }
+        };
```
rob-maron, line 117: Would like @imabdulbasit to review these validation functions, but I basically copied and pasted them from what we already had (with some extra clones). Maybe the merkle tree one would be good to review because it has the mutable variable (`mt`)

sequencer/src/request_response/data_source.rs
```diff
@@ -2,21 +2,224 @@
...
-            Request::Example => Ok(Response::Example),
+            Request::Accounts(height, view, accounts) => {
+                // Try to get accounts from memory first, then fall back to storage
+                if let Some(state) = self.consensus.state(ViewNumber::new(*view)).await {
+                    if let Ok(accounts) =
+                        retain_accounts(&state.fee_merkle_tree, accounts.iter().copied())
+                    {
+                        return Ok(Response::Accounts(accounts));
+                    }
+                }
```
rob-maron, line 61: I also copied these mostly from `ExtensibleDataSource`

sequencer/src/request_response/data_source.rs
```diff
@@ -2,21 +2,224 @@
...
+                snapshot.non_membership_remember(account, proof).unwrap()
+            },
+            LookupResult::NotInMemory => {
+                bail!("missing account {account}");
+            },
+        }
+    }
+
+    Ok(snapshot)
+}
```
rob-maron, line 225: cc @imabdulbasit I copied this from `retain_accounts` fn. I think moving it out is cleaner

sequencer/src/request_response/recipient_source.rs
```diff
@@ -1,44 +1,70 @@
...
-                // NOTE: This is just an example request, therefore this is hardcoded to get members of epoch 0.
-                // When doing a real request, please choose the appropriate epoch to get members from.
-                memberships
-                    .stake_table(Some(EpochNumber::new(0)))
-                    .iter()
-                    .map(|entry| entry.stake_table_entry.stake_key)
-                    .collect()
+        // Attempt to get the membership for the current epoch
+        let membership = match self.memberships.wait_for_catchup(epoch_number).await {
+            Ok(membership) => membership,
```
rob-maron, line 48: @bfish713 this works - but is there a better function? If I call this on Epoch 0 or 1 it consistently gives me the warning and falls back to `membership_for_epoch`

types/src/v0/traits.rs
```diff
@@ -232,140 +231,6 @@ pub trait StateCatchup: Send + Sync {
...
 
-#[async_trait]
-impl<T: StateCatchup + ?Sized> StateCatchup for Box<T> {
-    async fn try_fetch_leaves(&self, retry: usize, height: u64) -> anyhow::Result<Vec<Leaf2>> {
-        (**self).try_fetch_leaves(retry, height).await
-    }
-
-    async fn fetch_leaf(
-        &self,
-        height: u64,
```
rob-maron, line 243: Don't need this anymore since we are not doing `Box<StateCatchup>` or `Vec<StateCatchup>`

sequencer/src/catchup.rs
```diff
@@ -723,6 +721,258 @@ impl StateCatchup for NullStateCatchup {
...
+        let providers = self.providers.lock().clone();
+
+        if providers.is_empty() {
+            return Err(anyhow::anyhow!("no providers were initialized"));
+        }
+
+        // Spawn futures for each provider
+        let mut futures = FuturesUnordered::new();
+        for provider in providers {
+            futures.push(AbortOnDropHandle::new(tokio::spawn(closure(provider))));
```
imabdulbasit, line 761: this would make parallel calls to storage and peers at the same time. I believe it's better to try storage first, and then make calls to peers

sequencer/src/request_response/data_source.rs
```diff
@@ -2,21 +2,223 @@
...
+                let accounts = self
+                    .storage
+                    .as_ref()
+                    .with_context(|| "storage was not initialized")?
+                    .get_accounts(&self.node_state, *height, ViewNumber::new(*view), accounts)
+                    .await
+                    .with_context(|| "failed to get accounts from sql storage")?
+                    .0;
+
+                Ok(Response::Accounts(accounts))
```
imabdulbasit, line 72: I think some code for adding the fetched accounts back to memory is missing. Is that intentional?

imabdulbasit (commented):
Overall it looks good to me

request-response/src/lib.rs
```diff
@@ -300,10 +334,20 @@ impl<
...
 
+                    // Modify the response validation function to return an `Arc<dyn Any>`
+                    let response_validation_fn =
+                        Box::new(move |request: &Req, response: Req::Response| {
+                            let fut = response_validation_fn(request, response);
+                            Box::pin(
+                                async move { fut.await.map(|ok| Arc::new(ok) as ThreadSafeAny) },
+                            ) as ResponseValidationFuture
+                        });
+
```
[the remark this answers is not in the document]
rob-maron: Basically it allows us to only make one concurrent request for the same data. So if two spots in our code ask for the Leaf for view 22, it will only ask once and will return the same result to both at the same time.

sequencer/src/request_response/recipient_source.rs
rob-maron: Yeah I can do that. I asked Brendon what to use and he said either

rob-maron: I did this, but I think it makes more sense not to use `wait_for_catchup` at all (from what I understand, `membership_for_epoch` is the fallible version). Lmk what you think

sequencer/src/request_response/data_source.rs
rob-maron: No it wasn't, let me see if I can find what you're referencing

rob-maron: Can you point me to where this is?

request-response/src/lib.rs
```diff
@@ -256,7 +277,14 @@ impl<
                 })?;
 
             // Request the data, handling the errors appropriately
-            match self.request(request_message, estimated_request_ttl).await {
+            match self
```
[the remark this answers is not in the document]
rob-maron: For the first q, you mean limiting the amount of parallel requests we can have going right? If so I think the queue limits are enough, but curious to hear your thoughts if you think otherwise

For the second q, I assumed by the time it got to the end of the FIFO list it would be done or is unfulfillable. Maybe we don't need a limit at all though? Do we expect our own nodes to be calling `request()` an egregious amount of times? 

Either way an old request will need to be re-signed and re-requested from the network when the TTL expires (this is what `request_indefinitely` does)

sequencer/src/context.rs
```diff
@@ -191,16 +183,29 @@ impl<N: ConnectedNetwork<PubKey>, P: SequencerPersistence, V: Versions> Sequence
         };
 
         // Create the request-response protocol
-        let request_response_protocol = RequestResponse::new(
+        let request_response_protocol = RequestResponseProtocol::new(
             request_response_config,
             RequestResponseSender::new(outbound_message_sender),
```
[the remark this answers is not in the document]
rob-maron: Yes this is completely correct. Sending an outbound message to `outbound_message_sender` goes to other nodes and sending to `request_response_sender` goes to the request response protocol handler

sequencer/src/request_response/recipient_source.rs
imabdulbasit: yeah, it totally depends if you want to return an error and not wait

sequencer/src/request_response/data_source.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/blob/c13a0216cf25cd972dca1d4e69feafebe2317e14/sequencer/src/api.rs#L336

request-response/src/lib.rs
```diff
@@ -272,11 +300,17 @@ impl<
     /// - If the request times out
     /// - If the channel is closed (this is an internal error)
     /// - If the request we sign is invalid
-    pub async fn request(
+    pub async fn request<F, Fut, O>(
```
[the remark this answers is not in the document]
imabdulbasit: Addressed on zulip

sequencer/src/request_response/recipient_source.rs
rob-maron: I think returning an error is fine here since we will just print out the error and re-request. Plus it will fall back to genesis anyway

request-response/src/lib.rs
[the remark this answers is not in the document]
imabdulbasit: discussed on zulip:
`response_validation_fn` gets executed once

sequencer/src/request_response/data_source.rs
imabdulbasit: I guess this code is still there, but when we remove the storage+http requests idk how we want to get this done

imabdulbasit: I also don't think adding it to in-memory state is very important but maybe I am missing smth @jbearer

sequencer/src/catchup.rs
rob-maron: Sure, I made the remote providers wait 1s before doing anything

sequencer/src/context.rs
```diff
@@ -191,16 +183,29 @@ impl<N: ConnectedNetwork<PubKey>, P: SequencerPersistence, V: Versions> Sequence
         };
 
         // Create the request-response protocol
-        let request_response_protocol = RequestResponse::new(
+        let request_response_protocol = RequestResponseProtocol::new(
             request_response_config,
```
imabdulbasit, line 187: regarding this config:
should these limits be higher? 
```
  max_outgoing_responses: 10,
  max_incoming_responses: 5,
```

rob-maron: Yes definitely, let me update them

rob-maron: Adjusted, lmk what you think

sequencer/src/request_response/data_source.rs
rob-maron: I think we do want this. The original issue was we would ask a node for some account data and get it. When we had the majority of nodes pointed to only catch up from us, this caused an issue where a quorum couldn't get the data necessary since we didn't save it to our own node after fetching from a peer

I do agree it is less important though, since most operators now have a larger list and this request/response protocol has been tested to get data even if only one node on the network has it

imabdulbasit: I do see consensus inside the datasource, so I think it should be added since its just copy and paste of this code.

imabdulbasit (approved):
LGTM!

sequencer/src/request_response/data_source.rs
rob-maron: So I do see that the storage stuff happens through `CatchupDataSource`. Is this on a different layer than what we're doing in `StateCatchup`? e.g. I don't see it happening in `StatePeers` at all

imabdulbasit: `CatchupDataSource` is for  API layer

rob-maron: Ah I see

PR #2874, opened 2025-03-28
Bf/threshold

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-task-impls/src/quorum_proposal_recv/mod.rs
+++ b/hotshot-task-impls/src/quorum_proposal_recv/mod.rs
@@ -146,7 +146,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
                 if self.consensus.read().await.cur_view() > proposal.data.view_number()
                     || self.cur_view > proposal.data.view_number()
                 {
-                    tracing::error!("Throwing away old proposal");
+                    tracing::error!(
+                        "Throwing away old proposal for view {:?}",
+                        proposal.data.view_number()
+                    );
                     return;
                 }
                 let proposal_epoch = option_epoch_from_block_number::<TYPES>(
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -715,8 +715,12 @@ impl Membership<SeqTypes> for EpochCommittees {
 
     /// Get the voting success threshold for the committee
     fn success_threshold(&self, epoch: Option<Epoch>) -> primitive_types::U256 {
-        let quorum_len = self.stake_table(epoch).len();
-        primitive_types::U256::from(((quorum_len as u64 * 2) / 3) + 1)
+        let total_stake = self.total_stake(epoch);
+        if total_stake < primitive_types::U256::max_value() / 2 {
+            ((total_stake * 2) / 3) + 1
+        } else {
+            ((total_stake / 3) * 2) + 1
+        }
     }
 
     /// Get the voting success threshold for the committee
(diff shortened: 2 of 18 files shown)


hotshot-task-impls/src/quorum_proposal_recv/mod.rs
```diff
@@ -146,7 +146,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
                 if self.consensus.read().await.cur_view() > proposal.data.view_number()
                     || self.cur_view > proposal.data.view_number()
                 {
-                    tracing::error!("Throwing away old proposal");
+                    tracing::error!(
```
ss-es, line 149: minor, feel like this should be

```suggestion
                    tracing::warn!(
```

types/src/v0/impls/stake_table.rs
```diff
@@ -715,33 +714,40 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+            ((total_stake / 3) * 2) + 1
+        }
     }
 
     /// Get the voting success threshold for the committee
-    fn da_success_threshold(&self, epoch: Option<Epoch>) -> NonZeroU64 {
+    fn da_success_threshold(&self, epoch: Option<Epoch>) -> primitive_types::U256 {
         let da_len = self.da_stake_table(epoch).len();
-        NonZeroU64::new(((da_len as u64 * 2) / 3) + 1).unwrap()
+        primitive_types::U256::from(((da_len as u64 * 2) / 3) + 1)
```
ss-es, line 729: I think we must use total da stake here, unless we make a much bigger chance to how DA votes are accumulated

I believe `VoteAccumulator::accumulate()` will add up the total voting stake when forming the DAC, comparing it to the threshold

types/src/v0/impls/stake_table.rs
```diff
@@ -715,33 +714,40 @@ impl Membership<SeqTypes> for EpochCommittees {
...
     /// Get the voting success threshold for the committee
-    fn success_threshold(&self, epoch: Option<Epoch>) -> NonZeroU64 {
-        let quorum_len = self.stake_table(epoch).len();
-        NonZeroU64::new(((quorum_len as u64 * 2) / 3) + 1).unwrap()
+    fn success_threshold(&self, epoch: Option<Epoch>) -> primitive_types::U256 {
+        let total_stake = self.total_stake(epoch);
+        if total_stake < primitive_types::U256::max_value() / 2 {
+            ((total_stake * 2) / 3) + 1
+        } else {
+            ((total_stake / 3) * 2) + 1
```
ss-es, line 722: minor, but this isn't exactly right

e.g. if total stake is 302, the threshold should be floor(302*2/3) + 1 = 202, but this returns floor(302/3)*2 + 1 = 201. I think you need another case depending on whether `total_stake % 3` is 1 (then add 1) or 2 (then add 2)

types/src/v0/impls/stake_table.rs
```diff
@@ -715,33 +714,40 @@ impl Membership<SeqTypes> for EpochCommittees {
...
-        ))
-        .unwrap()
+    fn upgrade_threshold(&self, epoch: Option<Epoch>) -> primitive_types::U256 {
+        let total_stake = self.total_stake(epoch);
+
+        let normal_threshold = self.success_threshold(epoch);
+        let higher_threshold = if total_stake < primitive_types::U256::max_value() / 9 {
+            (total_stake * 9) / 10
+        } else {
+            (total_stake / 10) * 9
```
ss-es, line 747: but in contrast I think here this is completely fine, it doesn't really matter for consensus safety -- the threshold was arbitrary anyway

ss-es (approved):
approving, but I think the success_threshold needs to be fixed before merging

types/src/v0/impls/stake_table.rs
bfish713: I did it this way because if the stake is really that large I figured being off by 1 doesn't matter... I think we just do + 2 thus the threshold is always > 2/3 + 1.

PR #2886, opened 2025-03-31
Fix Catchup Leaf Fetching

bfish713:
I### This PR:
Fixes stake table catchup.

During catchup we were requesting the wrong DRB block height and our verification was explicitly checking for the epoch root block.  This would always faile when we asked for DRB because it's a different block number

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -805,13 +805,13 @@ impl Membership<SeqTypes> for EpochCommittees {
                 block_height,
                 stake_table.clone(),
                 success_threshold,
-                epoch_height,
+                block_height,
             )
             .await?;
         //DRB height is decided in the next epoch's last block
-        let drb_height = block_height + epoch_height + 3;
+        let drb_height = block_height + epoch_height + 5;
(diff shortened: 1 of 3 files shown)


types/src/v0/impls/stake_table.rs
```diff
@@ -805,13 +805,13 @@ impl Membership<SeqTypes> for EpochCommittees {
                 block_height,
                 stake_table.clone(),
                 success_threshold,
-                epoch_height,
+                block_height,
             )
             .await?;
         //DRB height is decided in the next epoch's last block
-        let drb_height = block_height + epoch_height + 3;
+        let drb_height = block_height + epoch_height + 5;
```
ss-es, line 812: I think it would be nice if we could put a function in `hotshot-types/src/utils.rs` that we call here, rather than the manual `+ 5`

ss-es (approved):
looks good to me

types/src/v0/impls/stake_table.rs
bfish713: I agree, this one is a bit of a weird one off though, maybe I can use existing functions instead

PR #2890, opened 2025-04-01
Bf/catchup

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-types/src/epoch_membership.rs
+++ b/hotshot-types/src/epoch_membership.rs
@@ -194,7 +194,7 @@ where
             ))
             .await
         else {
-            anytrace::bail!("get epoch drb failed for in epoch {:?}", root_epoch + 1);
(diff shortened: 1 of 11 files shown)


hotshot-types/src/epoch_membership.rs
```diff
@@ -179,6 +180,23 @@ where
...
+
+        // get the DRB from the last block of the epoch right before the one we're catching up to
+        let Ok(drb) = drb_membership
+            .get_epoch_drb(transition_block_for_epoch(
+                *(root_epoch + 1),
+                self.epoch_height,
+            ))
+            .await
+        else {
+            anytrace::bail!("get epoch drb failed for in epoch {:?}", root_epoch + 1);
```
ss-es, line 197:
```suggestion
            return anytrace::Err(anytrace::warn!("get epoch drb failed for in epoch {:?}", root_epoch + 1));
```

just to be explicit about the level

ss-es (approved):
think this looks fine to me

PR #2903, opened 2025-04-02
Update state-signer and relay-server for epoch

### LINKED ISSUE #1022: [light-client] Relay server should maintain a stake table

alxiong:
Closes #1022

### This PR:

On top of #2706, trying to update state signer logic for the new epoch-aware contract and state prover.


alxiong:
Two issues to fix: 

- in CI, an outdated version docker image for `deploy` binary is used (thus `--deploy-fee` not found err) 
- locally, "Error: error sending request for url (http://localhost:8545/)" having trouble connecting to L1

i'm investigating both

alxiong:
Progressing forward in debugging. contracts all deployed, relay server started and fetching stake tables from seq0 successfully. 

only problem now is 

```
[prover-service	] 2025-04-03T07:21:46.966562Z ERROR hotshot_state_prover::service: Cannot sync the light client state, will retry: Error when communicating with the state relay server: Error 404: The light client state signatures are not ready.
```

alxiong:
@mrain @sveitser native demo passing now! 🎉

hotshot-state-prover/src/bin/state-prover.rs
```diff
@@ -47,7 +47,7 @@ struct Args {
     l1_provider: Url,
 
     /// Address of LightClient contract on layer 1.
-    #[clap(long, env = "ESPRESSO_SEQUENCER_LIGHTCLIENT_ADDRESS")]
+    #[clap(long, env = "ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS")]
```
sveitser, line 50: @alxiong while I do like the new name because it's more consistent than the current one, this is a breaking change for deployments. It would be better to support the old variable and issue a depreciation warning for a while.

cc @rob-maron @Ancient123 @babdor for visibility about the change of name of the env var.

alxiong: is this a breaking change tho? 
previously we set `ESPRESSO_SEQUENCER_LIGHTCLIENT_ADDRESS=$ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS` in the `.env`

it was just a convenient alias as I assume. we are not actually ever gonna use the lightclient alone (without the proxy) iiuc.

PR #2904, opened 2025-04-02
Fix breaking changes with `20250228-patch3`

ss-es:


types/src/v0/impls/state.rs
```diff
@@ -764,15 +764,21 @@ fn validate_builder_fee(
...
         } else {
-            fee_info
+            if !fee_info.account().validate_fee_signature(
+                &signature,
+                fee_info.amount().as_u64().unwrap(),
+                proposed_header.metadata(),
+            ) || !fee_info
                 .account()
-                .validate_fee_signature(
+                .validate_fee_signature_with_vid_commitment(
```
tbro, line 773: I think this used to use `validate_sequencing_fee_signature_marketplace`.

types/src/v0/impls/state.rs
```diff
@@ -764,15 +764,21 @@ fn validate_builder_fee(
                 .then_some(())
                 .ok_or(BuilderValidationError::InvalidBuilderSignature)?;
         } else {
-            fee_info
+            if !fee_info.account().validate_fee_signature(
```
tbro, line 767: Is it possible for code style to match between if statement branches. Above we use ensure, here we invert the bool.

types/src/v0/impls/state.rs
tbro: Marketplace signatures are different

types/src/v0/impls/header.rs
```diff
@@ -489,7 +489,13 @@ impl Header {
         {
             if version < MarketplaceVersion::version() {
```
tbro, line 490: I think we will also need to do this is `validate_and_apply_header`.

PR #2913, opened 2025-04-03
staking-cli: add approve command

sveitser:
This must be run before `delegate` so that the stake table contract has the permission to move funds to itself during the delegation transactions.


staking-cli/src/delegation.rs
```diff
@@ -5,7 +5,22 @@ use alloy::{
...
-use contract_bindings_alloy::staketable::StakeTable::StakeTableInstance;
+use contract_bindings_alloy::{
+    esptoken::EspToken::EspTokenInstance, staketable::StakeTable::StakeTableInstance,
+};
+
+pub async fn approve<P: Provider<T>, T: Transport + Clone>(
+    token: EspTokenInstance<T, P>,
+    stake_table_address: Address,
+    amount: U256,
+) -> Result<TransactionReceipt> {
```
alxiong, line 16: i'll rectify this API as I have suggested on Zulip. 
accept `provider, address` instead of accepting an instance.

PR #2914, opened 2025-04-03
fix lint error due to toolchain upgrade

mrain:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/hotshot-types/src/light_client.rs
+++ b/hotshot-types/src/light_client.rs
@@ -278,7 +278,7 @@ impl<F: PrimeField> GenericPublicInput<F> {
 
 pub fn hash_bytes_to_field<F: RescueParameter>(bytes: &[u8]) -> Result<F, RescueError> {
     // make sure that `mod_order` won't happen.
-    let bytes_len = ((<F as PrimeField>::MODULUS_BIT_SIZE + 7) / 8 - 1) as usize;
+    let bytes_len = <F as PrimeField>::MODULUS_BIT_SIZE.div_ceil(8) as usize;
     let elem = bytes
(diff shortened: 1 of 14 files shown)


sveitser (approved):

hotshot-types/src/light_client.rs
```diff
@@ -278,7 +278,7 @@ impl<F: PrimeField> GenericPublicInput<F> {
 
 pub fn hash_bytes_to_field<F: RescueParameter>(bytes: &[u8]) -> Result<F, RescueError> {
     // make sure that `mod_order` won't happen.
-    let bytes_len = ((<F as PrimeField>::MODULUS_BIT_SIZE + 7) / 8 - 1) as usize;
+    let bytes_len = <F as PrimeField>::MODULUS_BIT_SIZE.div_ceil(8) as usize;
```
alxiong, line 281: this is wrong. we forget the `-1`. the `+7)/8` is the `div_ceil()`

to see why: this bytes_len is suppose to avoid `mod p` to happen. if `MODULUS_BIT_SIZE = 9`, `9.div_ceil()=2`, and if I give a 10 bits value which is still within 2 bytes, it will wrap around `modulus`. 

I'm fixing the `-1` in my PR.

sveitser: Thanks @alxiong but I suggest to make separate PRs for fixes.

alxiong: true. but i need to sync with main and resolve conflicts. ... already fixed locally. if changes are significant, i promise to leave for future PRs

PR #2917, opened 2025-04-03
[light-client] endpoint to fetch light client state update certificate

### LINKED ISSUE #2793: [Light Client] add a query service endpoint to fetch the light client state update cert
Blocked by #2762

mrain:
Closes #2793

### This PR:

Add an endpoint in hotshot query service to fetch the light client state update certificate by epoch.


sequencer/src/persistence/sql.rs
```diff
@@ -2067,6 +2067,42 @@ impl Provider<SeqTypes, LeafRequest<SeqTypes>> for Persistence {
...
+    async fn fetch(&self, req: StateCertRequest) -> Option<StateCertQueryData<SeqTypes>> {
+        let mut tx = match self.db.read().await {
+            Ok(tx) => tx,
+            Err(err) => {
+                tracing::warn!("could not open transaction: {err:#}");
+                return None;
+            },
+        };
+
+        match query_as::<(Vec<u8>,)>("SELECT state_cert FROM state_cert WHERE epoch = $1 LIMIT 1")
```
jbearer, line 2082: This is fetching from the same table as `get_state_cert` in the `SqlStorage`? If so, I think it's pointless. The other impls in this file are attempting to help catchup missing data from the main query service storage by looking at temporary consensus storage (different tables). This is just replicating the query that the query service already makes so it doesn't add anything.

hotshot-query-service/src/data_source/storage.rs
```diff
@@ -214,6 +216,10 @@ where
         common: VidCommonQueryData<Types>,
         share: Option<VidShare>,
     ) -> impl Send + Future<Output = anyhow::Result<()>>;
+    fn insert_state_cert(
```
jbearer, line 219: Where does this get called?

hotshot-query-service/src/availability/data_source.rs
```diff
@@ -233,6 +234,8 @@ where
         hash: TransactionHash<Types>,
     ) -> Fetch<TransactionQueryData<Types>>;
 
+    async fn get_state_cert(&self, epoch: u64) -> Fetch<StateCertQueryData<Types>>;
```
jbearer, line 237: This doesn't seem to be called anywhere? To hook this up we would need to add a new GET endpoint handler in `hotshot-query-service/src/availability.rs` to handle the `get_state_cert` request and call this function

jbearer (changes requested):
Dunno about this change. It's not clear to me where the state certs are getting added to the database. And repurposing the existing fetching infrastructure without any changes to deal with data that is indexed by epoch (as opposed to block height) seems very error prone. This problem could maybe be fixed by indexing the state certs by block height, then we could still support queries by epoch by converting the epoch number to a block number just before lookup.

However, I thought the plan here was to add the schnorr signatures to the QC struct for epoch transitioning QCs. In that case, the signatures should already be stored in the db because we store QCs, so we shouldn't have to add all this low-level infrastructure to store and retrieve the certificates; we could just add a high-level endpoint handler which fetches the QC and pulls out the schnorr signatures

mrain:
> However, I thought the plan here was to add the schnorr signatures to the QC struct for epoch transitioning QCs. In that case, the signatures should already be stored in the db because we store QCs, so we shouldn't have to add all this low-level infrastructure to store and retrieve the certificates; we could just add a high-level endpoint handler which fetches the QC and pulls out the schnorr signatures

I made such an attempt before, and found out that there're other troubles on the hotshot side. For example, previously the hotshot storage APIs simply convert the new types (including QC) to the old ones and store them.
Now I slightly regret my decision.

hotshot-query-service/src/availability/query_data.rs
```diff
@@ -800,8 +801,18 @@ where
...
 pub struct Limits {
     pub small_object_range_limit: usize,
     pub large_object_range_limit: usize,
 }
+
+impl<Types: NodeType> HeightIndexed for StateCertQueryData<Types> {
+    fn height(&self) -> u64 {
+        self.0.epoch.u64()
+    }
+}
```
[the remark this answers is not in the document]
mrain: Now I change it to be the block number [9c85bbb](https://github.com/EspressoSystems/espresso-network/pull/2917/commits/9c85bbb25d6e0a9af2266e39b2f5644933337adb)

hotshot-query-service/src/data_source/storage.rs
mrain: called here: https://github.com/EspressoSystems/espresso-network/pull/2917/files#diff-d6d79e95846b7f888aec7a5acbaa9e53645539fbf0334e30bc674c8317b7f6f2R118

hotshot-query-service/src/availability/data_source.rs
mrain: I forgot to add the handler, done in [9c85bbb](https://github.com/EspressoSystems/espresso-network/pull/2917/commits/9c85bbb25d6e0a9af2266e39b2f5644933337adb)

sequencer/src/persistence/sql.rs
mrain: Got it. The other place that might contain this information is the next quorum proposal.

hotshot-query-service/src/data_source/fetching/state_cert.rs
```diff
@@ -0,0 +1,120 @@
...
+            Ok(state_cert) => {
+                fetcher.store_and_notify(state_cert).await;
+            },
+            Err(QueryError::NotFound | QueryError::Missing) => {
+                tracing::debug!("state cert not found; trying fetch");
+            },
+            Err(QueryError::Error { message }) => {
+                bail!("failed to get state cert for epoch {}: {message}", req.0);
+            },
+        }
```
jbearer, line 79: This logic does not make sense. It is saying, if we can already get the state cert from the database (via `tx`) then store it in the database and notify the caller. This is redundant, if this function is being called we should assume the cert is _not_ in the database.

Then, if it wasn't in the database (`NotFound` or `Missing`) we do nothing. So it will never be fetched from a peer. If we want to enable catchup/fetching of state certs from peers, this needs to do something like `fetcher.state_cert_fetcher.spawn_fetch(req)`

mrain: Got it. If we don't want to implement peer fetching at this point, should we just return nothing for `FetchingDataSource::get_state_cert()`

jbearer: The top level `get_state_cert` has to return something if you want the query to work. But if you're fine with it failing whenever we don't have the state cert locally, rather than trying to get it from peers, you can have `active_fetch` just return `Ok(())` (see e.g. `Fetchable<Types> for TransactionQueryData<Types>` which also doesn't implement fetching)

jbearer (approved):
This looks great! Thank you. Just note I reviewed the query service and sequencer persistence changes, but not the hotshot changes, as I'm not too familiar with that code

bfish713 (approved):

PR #2919, opened 2025-04-03
Use `l1_finalized` when adding epoch root

ss-es:


tbro (approved):

types/src/v0/impls/stake_table.rs
```diff
@@ -770,7 +770,7 @@ impl Membership<SeqTypes> for EpochCommittees {
         };
 
         let stake_tables = self
-            .get_stake_table_by_epoch(epoch, address, block_header.height())
+            .get_stake_table_by_epoch(epoch, address, block_header.l1_head())
```
tbro, line 773: Please use `l1_finalized`

bfish713 (approved):

PR #2922, opened 2025-04-03
cannot undelegate whilst there is a pending withdrawal/undelegation

### LINKED ISSUE #2920: prevent subsequent undelegation whilst another delegation is pending
see comment here

https://github.com/EspressoSystems/espresso-network/blob/rewards_audit/contracts/src/StakeTable.sol#L422

alysiahuggins:
Closes #2920

### This PR:
- Fixes a bug that allowed multiple undelegations from the same delegator to the same validator without waiting for the first to be claimed
- Introduces a new UndelegationAlreadyExists error to guard against undelegation overrides
- Makes the delegations mapping public for better testability and transparency
- Adds a new test case test_PoC_Undelegation_Overrides to validate the behavior and cover edge cases around undelegation logic. (to be renamed)

### This PR does not:
- allow for multiple withdrawals at once
- Modify any behavior related to validator registration or exit beyond the undelegation guard

### Key places to review:
- `StakeTable.sol`: look at the added revert check using `UndelegationAlreadyExists` before processing a new undelegation.
- `StakeTable.t.sol`: the new test `test_PoC_Undelegation_Overrides `walks through undelegation scenarios and ensures the new guard behaves correctly

### How to test this PR:  
- `forge test --match-test test_PoC_Undelegation_Overrides`


contracts/test/StakeTable.t.sol
```diff
@@ -550,6 +550,61 @@ contract StakeTable_register_Test is Test {
...
+        stakeTable.undelegate(validator, 1 ether);
+        stakeTable.claimWithdrawal(validator);
+        assertEq(token.balanceOf(delegator), INITIAL_BALANCE - 3 ether + 2 ether);
+
+        assertEq(stakeTable.delegations(validator, delegator), 1 ether);
+
+        // now the delegator can undelegate again
+        stakeTable.undelegate(validator, 1 ether);
+
+        assertEq(stakeTable.delegations(validator, delegator), 0);
```
sveitser, line 600: Suggest to check the emitted event instead of storage. In general, I think whenever it's not part of the API intended to be used by consumers we should try to avoid checking storage because otherwise we have to rewrite tests a lot when we make changes.

contracts/test/StakeTable.t.sol
```diff
@@ -550,6 +550,61 @@ contract StakeTable_register_Test is Test {
...
+        assertEq(token.balanceOf(address(stakeTable)), 1 ether);
+
+        vm.expectRevert(S.PrematureWithdrawal.selector);
+        stakeTable.claimWithdrawal(validator);
```
sveitser, line 604: At the end we could check the token balance, since that's part of the API of the token.

contracts/src/StakeTable.sol
```diff
@@ -394,6 +397,13 @@ contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpg
             revert ValidatorAlreadyExited();
         }
 
+        if (
+            undelegations[validator][delegator].amount > 0
```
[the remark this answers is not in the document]
alysiahuggins: so that they can only undelegate if they withdraw the previously undelegated amount - unless we want to maintain another data structure for withdrawals.
Because if they undelegate, and then the exitEpoch passes, then they undelegate again, so there is a new unlocksAt value, then they have to wait for that epoch to pass to retrieve the previously undelegated amount which was already ready to be claimed

contracts/test/StakeTable.t.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/2922/commits/2e0d60e6fc2756377432e048e33c127d392f2db8

contracts/src/StakeTable.sol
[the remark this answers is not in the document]
sveitser: we updated this in https://github.com/EspressoSystems/espresso-network/pull/2922/commits/be85b80abde17b3592c0d82be9c48d92b4d9042a

sveitser (approved):

0xkato (approved):
LGTM!

PR #2923, opened 2025-04-03
Get stake table contract address from `ChainConfig`

tbro:
When we upgrade from fee version, stake table contract address will *not* have been passed in to `EpochCommitee::new_stake` on instantiation. So when `add_epoch_root` tries to query stake table, address will always be `None`. 

FYI this has been sitting around in some other branches for a while. This PR just splits this out in order to get it on main.

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -10,10 +10,12 @@ use alloy::{
 };
 use anyhow::{bail, Context};
 use async_lock::RwLock;
+use committable::Committable;
 use contract_bindings_alloy::staketable::StakeTable::{
     ConsensusKeysUpdated, Delegated, Undelegated, ValidatorExit, ValidatorRegistered,
 };
-use ethers_conv::ToEthers;
+
+use ethers_conv::{ToAlloy, ToEthers};
 use hotshot::types::{BLSPubKey, SignatureKey as _};
 use hotshot_contract_adapter::stake_table::{bls_alloy_to_jf2, edward_bn254point_to_state_ver};
 use hotshot_types::{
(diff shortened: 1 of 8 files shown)


types/src/v0/impls/stake_table.rs
```diff
@@ -896,6 +901,31 @@ impl Membership<SeqTypes> for EpochCommittees {
         self.add_drb_result(epoch + 1, initial_drb_result);
     }
 }
+/// Retrieve and verify `ChainConfig`
+// TODO move to appropriate object (Header?)
+pub(crate) async fn get_chain_config(
```
sveitser, line 906: Is this code path exercised by any of the tests?

types/src/v0/impls/stake_table.rs
```diff
@@ -764,19 +766,22 @@ impl Membership<SeqTypes> for EpochCommittees {
...
         block_header: Header,
     ) -> Option<Box<dyn FnOnce(&mut Self) + Send>> {
-        let Some(address) = self.contract_address else {
-            tracing::debug!("`add_epoch_root` called with `self.contract_address` value of `None`");
+        let chain_config = get_chain_config(self.chain_config, &self.peers, &block_header)
+            .await
+            .ok()?;
+
+        let Some(address) = chain_config.stake_table_contract else {
+            tracing::error!("No stake table contract address found in Chain config");
```
sveitser, line 774: What happens if we end up here? It seems to me that this is a misconfiguration and we might as well panic, but not sure.

types/src/v0/impls/stake_table.rs
tbro: Yes it would mean that we couldn't resolve the `ChainConfig` on the `Header` which would be severe. I doubt this is even possible since we have already produced at least `epoch_height -5` blocks .

types/src/v0/impls/stake_table.rs
tbro: Yes it would mean that we couldn't resolve the `ChainConfig` on the `Header` which would be severe. I doubt this is even possible since we have already produced at least `epoch_height -5` blocks .

sveitser (approved):
LGTM but some questions

PR #2925, opened 2025-04-04
Enable Demo Native for PoS version

bfish713:
Closes #<ISSUE_NUMBER>

### This PR:

Moves changes from ab/fix-config and release-fb into main. This should be the changes we actually need without brining in too much of the debugging things.

The changes are:
- Change ValidatorConfig stake_amount to be U256
- Calculate rewards for the current leader, not previous
- Remove trying to load stake table when we are adding the epoch root
- Only apply rewards when the the current height is in the third epoch, not the parent height
- Adds demo-native-pos-base which tests running starting from epoch version
- Adds "pos" feature for using the epoch base version

### Key places to review:

Changes to:
`stake_table.rs`
`rewards.rs`

### How to test this PR:

run `just demo-native-pos-base` to make sure the new thing works
run `just demo-native` to confirm it's not broken

### Things tested 

The 2 things from how to test.


builder/src/bin/permissionless-builder.rs
```diff
@@ -120,10 +120,15 @@ async fn main() -> anyhow::Result<()> {
     let upgrade = genesis.upgrade_version;
 
     match (base, upgrade) {
+        (FeeVersion::VERSION, FeeVersion::VERSION) => {
```
tbro, line 123: This may work, not sure, but it is confusing. If we want to disable the upgrade can we use the same syntax everywhere (i.e. `(FeeVersion::version, _)`.

tbro:
```suggestion
        (FeeVersion::VERSION, _) => {
```

builder/src/bin/permissionless-builder.rs
```diff
@@ -120,10 +120,15 @@ async fn main() -> anyhow::Result<()> {
...
+            run::<SequencerVersions<FeeVersion, V0_0>>(genesis, opt).await
+        },
         (FeeVersion::VERSION, MarketplaceVersion::VERSION) => {
```
tbro, line 126:
```suggestion
        (FeeVersion::VERSION, EpochVersion::VERSION) => {
```

hotshot-task-impls/src/consensus/mod.rs
```diff
@@ -232,7 +232,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
                 if high_qc_updated || next_high_qc_updated {
                     // Send ViewChange indicating new view and new epoch.
                     let next_epoch = high_qc.data.epoch().map(|x| x + 1);
-                    tracing::info!("Entering new epoch: {:?}", next_epoch);
+                    tracing::error!("Entering new epoch: {:?}", next_epoch);
```
tbro, line 235: Should this be a `warn!`?

hotshot-task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -627,7 +628,7 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             signature,
             _pd: PhantomData,
         };
-        tracing::debug!(
+        tracing::error!(
```
tbro, line 631: If I understand this is logged on every proposal, so `debug!` or `info!`.

types/src/v0/impls/state.rs
```diff
@@ -886,10 +887,10 @@ impl ValidatedState {
         // so that marketplace version also supports this,
         // and the marketplace integration test passes
         if version == EpochVersion::version()
```
tbro, line 889:
```suggestion
        if version >= EpochVersion::version()
```
And remove TODO.

types/src/v0/impls/reward.rs
```diff
@@ -409,7 +409,7 @@ pub async fn first_two_epochs(height: u64, instance_state: &NodeState) -> anyhow
         .first_epoch()
         .context("The first epoch was not set.")?;
 
-    Ok(epoch == first_epoch || epoch == first_epoch + 1)
+    Ok(epoch == first_epoch || epoch == first_epoch + 1 || epoch == EpochNumber::new(0))
```
tbro, line 412: Update doc comment b/c we are checking for first epoch, second epoch and epoch 0.

sequencer/src/run.rs
```diff
@@ -33,33 +33,55 @@ pub async fn main() -> anyhow::Result<()> {
...
+                modules,
+                opt,
+                // Specifying V0_0 disables upgrades
+                SequencerVersions::<espresso_types::EpochVersion, espresso_types::V0_0>::new(),
+            )
+            .await
+        },
         #[cfg(all(feature = "fee", feature = "marketplace"))]
-        (FeeVersion::VERSION, MarketplaceVersion::VERSION) => {
+        (espresso_types::FeeVersion::VERSION, espresso_types::MarketplaceVersion::VERSION) => {
```
tbro, line 59: NIT: we should just import `FeeVersion` and avoid these last few changes in the diff.

hotshot-task-impls/src/quorum_proposal/handlers.rs
bfish713: yeah, should be info! I think.  I change it for debugging and always for get to reduce it

tbro (approved):

PR #2926, opened 2025-04-04
CI: slow test fixes

sveitser:
- don't use release profile (use default: "test")
- remove stuff that shouldn't be necessary


alxiong (approved):

alxiong:
Well, it seems that the build is still 60 min... and running

tbro:
> Well, it seems that the build is still 60 min... and running

Looks like the test artifact build, specifically, is faster.

alxiong:
yah true. Thanks Mathis for this. cut down almost in half. great saving.

PR #2931, opened 2025-04-04
fix reconstruct_state() for rewards

imabdulbasit:
fixes state construction for rewards


sequencer/src/api/sql.rs
```diff
@@ -512,12 +512,11 @@ async fn reconstruct_state<Mode: TransactionMode>(
     }
 
     if let Some(accounts) = reward_accounts {
-        // TODO: don't think this is needed?
-        // What if we are reconstructing a state from the leaf which is on pre-epoch version?
-        // that is okie?
-        // if parent.block_header().reward_merkle_tree_root().is_none() {
-        //     bail!("reward merkle tree root not set");
-        // }
+        tracing::error!(
```
bfish713, line 515: lower this log level

types/src/v0/impls/reward.rs
```diff
@@ -413,19 +413,22 @@ pub async fn first_two_epochs(height: u64, instance_state: &NodeState) -> anyhow
...
-    // This block is for the next block we need the leader for that block not the parent block
-    let epoch = EpochNumber::new(epoch_from_block_number(new_height, epoch_height));
+    let parent_view = parent_leaf.view_number();
+
+    let epoch_height = instance_state.epoch_height.context("epoch height")?;
+    if epoch_height == 0 {
+        bail!("epoch height is 0. can not catchup reward accounts");
+    }
+    let epoch = EpochNumber::new(epoch_from_block_number(height, epoch_height));
+
```
bfish713, line 431: I'd rather not include these changes, to me it just makes it less clear to read with the height being passed in.  It's not clear from the fn signature which `height`

types/src/v0/impls/state.rs
```diff
@@ -889,15 +889,21 @@ impl ValidatedState {
...
+            .await?;
 
             // apply rewards
 
             validated_state
-                .distribute_rewards(&mut delta, validator)
-                .context("failed to distribute rewards")?
+                .distribute_rewards(&mut delta, validator.clone())
+                .context("failed to distribute rewards")?;
+            tracing::info!("Rewards distributed for validator={:?}", validator.account);
```
bfish713, line 906: Don't think we need these changes

process-compose.yaml
bfish713, line 1: What do these changes do

imabdulbasit: api peers is for query service peers.  Some of the nodes are pointing to peers that don't even run the query service

types/src/v0/impls/state.rs
imabdulbasit: changed to debug level

types/src/v0/impls/reward.rs
imabdulbasit: reverted

bfish713 (approved):

PR #2937, opened 2025-04-05
Remove `todo_by`

rob-maron:


sequencer/src/network/cdn.rs
```diff
@@ -79,12 +78,7 @@ impl<T: SignatureKey> SignatureScheme for WrappedSignatureKey<T> {
             Err(_) => return false,
         };
 
-        todo_by!(
-            "2025-4-4",
-            "Only accept the namespaced message once everyone has upgraded"
-        );
-        public_key.0.validate(&signature, message)
-            || public_key.0.validate(&signature, &namespaced_message)
+        public_key.0.validate(&signature, &namespaced_message)
```
rob-maron, line 81: Everyone has the new version, even on mainnet

bfish713 (approved):

PR #2939, opened 2025-04-06
backport types migration changes

imabdulbasit:
cherry picks all the fixes from release branch


tbro:
I don't see anything wrong w/ it but its hard to tell unless we're already very familiar w/ this part of the code. If we added test cases for the new code it would be easy to tell if it works as expected.

imabdulbasit:
> I don't see anything wrong w/ it but its hard to tell unless we're already very familiar w/ this part of the code. If we added test cases for the new code it would be easy to tell if it works as expected.

I'll add a test later for this one. This has already been tested on devnet and this PR is backporting those changes

PR #2953, opened 2025-04-07
Cherry-pick fixes 

imabdulbasit:


hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -949,7 +954,15 @@ impl<Types: NodeType> MigrateTypes<Types> for SqlStorage {
 
             query.execute(tx.as_mut()).await;
 
-            tx.commit().await?;
+            // update migrated_rows column with the offset
+            tx.upsert(
+                "types_migration",
+                ["id", "completed", "migrated_rows"],
+                ["id"],
+                [(0_i64, false, offset)],
```
sveitser, line 951: @imabdulbasit Shouldn't this be `offset + limit` because when fetching we get the rows from `offset` until `offset + limit`?

sequencer/src/persistence/sql.rs
```diff
@@ -1532,6 +1553,13 @@ impl SequencerPersistence for Persistence {
 
             let mut tx = self.db.write().await?;
             query.execute(tx.as_mut()).await?;
+            tx.upsert(
+                "epoch_migration",
+                ["table_name", "completed", "migrated_rows"],
+                ["table_name"],
+                [("vid_share".to_string(), false, offset)],
```
sveitser, line 1560: Same question here. Since we're only adding batch_size to offset afterwards.

sveitser (approved):
LGTM. But I'm not sure the tracking of migrated rows is correct.

hotshot-query-service/src/data_source/storage/sql.rs
imabdulbasit: uh yeah, need to move the offset+ batch_size above this query

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2953/commits/dfc6559d0e07a5410743e203924c1d154cfa99d0

PR #2957, opened 2025-04-07
staking-cli: ESP amount, utility commands

sveitser:
- Use ESP amounts (10^18 becomes 1). Fractional values are also supported (e. g. "0.123").
- Add info, transfer, balance, allowance commands

The info command shows what stakers are currently active by fetchign the
events from the L1 stake table.

Add env var overrides to pick up contract addresses from the env vars we
use for the espresso network services in case these env vars aren't
already set.

This makes running `staking-cli info` work without other parameters in
environments where the env vars are already present.

To test, for example start demo native and run `just demo-native` and
when the stake-for-demo job is done:

```
   env STAKE_TABLE_ADDRESS=0xf7cd8fa9b94db2aa972023b379c7f72c65e4de9d \
    ESP_TOKEN_ADDRESS=0xe1da8919f262ee86f9be05059c9280142cf23f48
    cargo run --bin staking-cli info
 ```

Optionally a block number, hash or tag can be passed to info: 123,
latest, finalized, etc.

### CODE REVIEW

File: staking-cli/src/bin/staking-cli.rs [hunk 1]
```diff
@@ -157,6 +166,18 @@ pub async fn main() -> Result<()> {
         _ => {}, // Other commands handled after shared setup.
     }
 
+    // When the staking CLI is used for our testnet, the env var names are different.
+    let config = config.apply_env_var_overrides()?;
```
imabdulbasit, line 170: why are the env var different?

staking-cli/src/bin/staking-cli.rs
```diff
@@ -157,6 +166,18 @@ pub async fn main() -> Result<()> {
...
+
+    // Clap serde will put default value if they aren't set. We check some
+    // common configuration mistakes.
+    if config.stake_table_address == Address::ZERO {
+        exit("Stake table address is not set")
```
imabdulbasit, line 175: maybe mention the env variable that is not set?

staking-cli/src/bin/staking-cli.rs
```diff
@@ -157,6 +166,18 @@ pub async fn main() -> Result<()> {
...
+    };
+    if config.token_address == Address::ZERO {
+        exit("ESP token address is not set")
```
imabdulbasit, line 178: same here

staking-cli/src/bin/staking-cli.rs
sveitser: Because they are really long for the sequencer, so I thought it's weird to use those for the staking-cli.

staking-cli/src/info.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+pub fn display_stake_table(stake_table: Vec<Validator<BLSPubKey>>) -> Result<()> {
+    let mut stake_table = stake_table.clone();
+    stake_table.sort_by(|a, b| a.stake.cmp(&b.stake));
+
+    for validator in stake_table.iter() {
+        let comm: Commission = validator.commission.try_into()?;
+        let bls_key = validator.stake_table_key.to_string();
+        let end = bls_key.chars().map(|c| c.len_utf8()).take(40).sum();
```
imabdulbasit, line 31: what is this for?

staking-cli/src/info.rs
```diff
@@ -0,0 +1,41 @@
...
+        tracing::info!(
+            "Validator {}: {}... comm={} stake={} ESP",
```
imabdulbasit, line 33: can we also display delegator addresses?

staking-cli/src/bin/staking-cli.rs
imabdulbasit: can we not use different long name but same env?

staking-cli/src/info.rs
sveitser: to truncate the bls key string so that it fits on the terminal

staking-cli/src/bin/staking-cli.rs
sveitser: I mean the env vars are long and IMO don't really make sense for the stake table CLI. But yeah we could use the same. I will do a follow up PR to address your suggestions.

PR #2959, opened 2025-04-07
`epoch_start_block` from genesis file

tbro:
`epoch_start_block` needs to be configurable, so we take it from genesis file. 

Also adds a test for epoch version genesis file.

Plucked from https://github.com/EspressoSystems/espresso-network/commit/b0cd03ad7883dd256319a6e9af271a515e7546cd

### DIFF
--- a/sequencer/src/genesis.rs
+++ b/sequencer/src/genesis.rs
@@ -52,6 +52,7 @@ pub struct Genesis {
     #[serde(with = "version_ser")]
     pub upgrade_version: Version,
     pub epoch_height: Option<u64>,
+    pub epoch_start_block: Option<u64>,
     pub chain_config: ChainConfig,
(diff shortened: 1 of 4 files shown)


sequencer/src/genesis.rs
```diff
@@ -1254,7 +1255,7 @@ mod test {
     }
 
     #[test]
-    fn test_marketplace_and_fee_upgrade_toml() {
+    fn test_fee_and_epoch_upgrade_toml() {
         let toml = toml! {
             base_version = "0.1"
             upgrade_version = "0.2"
```
imabdulbasit, line 1261: would be nice if you can include the epoch_height and epoch_start_block here as well

imabdulbasit (approved):

PR #2970, opened 2025-04-08
Add decaf config file and make the default

sveitser:
Don't abort if token address isn't set since it's not required for stake table interactions.

### DIFF
--- a/staking-cli/config.decaf.toml
+++ b/staking-cli/config.decaf.toml
@@ -1,0 +1,7 @@
+# TODO, use ledger for singing?
+# Fill in your mnemonic
+mnemonic = "test test test test test test test test test test test junk"
+account_index = 0
+rpc_url = "https://ethereum-sepolia-rpc.publicnode.com"
+token_address = "0xb3e655a030e2e34a18b72757b40be086a8f43f3b"
+stake_table_address = "0x40304fbe94d5e7d1492dd90c53a2d63e8506a037"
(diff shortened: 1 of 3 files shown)


staking-cli/config.decaf.toml
```diff
@@ -0,0 +1,7 @@
+# TODO, use ledger for singing?
+# Fill in your mnemonic
+mnemonic = "test test test test test test test test test test test junk"
+account_index = 0
+rpc_url = "https://ethereum-sepolia-rpc.publicnode.com"
+token_address = "0xb3e655a030e2e34a18b72757b40be086a8f43f3b"
+stake_table_address = "0x40304fbe94d5e7d1492dd90c53a2d63e8506a037"
```
rob-maron, line 7: Are we using `toml` now instead of the `.env` file? As opposed to what we have in the doc

rob-maron (approved):

staking-cli/config.decaf.toml
rob-maron: Does this fill in and env does the override?

sveitser: This file is only used if one first runs the `init` command. I figured for simplicity we don't use it for node operators just now cuz it will be a bit messy if they use the docker container.

sveitser: The idea is to have all the static config in a file so it doesn't need to be specified each time. Was thinking more about end users (delegators) when I made this.

PR #2985, opened 2025-04-09
optimize stake table events fetching

imabdulbasit:
- divides stake table events fetching into chunks to avoid large block range rpc calls
- uses contract init block to avoid starting from 0

### DIFF
--- a/types/src/v0/impls/l1.rs
+++ b/types/src/v0/impls/l1.rs
@@ -874,44 +874,57 @@ impl L1Client {
     pub async fn get_stake_table(
         &self,
         contract: Address,
-        block: u64,
+        to_block: u64,
     ) -> anyhow::Result<IndexMap<Address, Validator<BLSPubKey>>> {
         // TODO stake_table_address needs to be passed in to L1Client
         // before update loop starts.
+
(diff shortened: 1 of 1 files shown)


types/src/v0/impls/l1.rs
```diff
@@ -870,51 +870,177 @@ impl L1Client {
...
 
-        let registered = stake_table_contract
-            .ValidatorRegistered_filter()
-            .from_block(0)
-            .to_block(block)
-            .query()
-            .await?;
+        // Retrieve the L1 block number when this contract was initialized.
+        // This helps avoid starting the event fetching from block number 0.
+        let from = if let Ok(init_block) = stake_table_contract.initializedAtBlock().call().await {
```
sveitser, line 887: @imabdulbasit we should retry this too

sveitser: Yes it gets set when the contract is deployed.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/2985/commits/ca16b3df4164d7d849af656d2fce5422ccdacc77

sveitser (approved):

PR #3000, opened 2025-04-09
Add callback to store add_drb_result from within catchup

pls148:
This one is a little tricky. I started by trying to add either `I: NodeImplementation<TYPES>` or `S: Storage<TYPES>` as a parameter to `EpochMembershipCoordinator`. This eventually bubbles up to the point where either some types that CAN'T implement `NodeImplementation` need to, or traits that need vtables are unable to because they contain generic functions. Ultimately, not everything that wants `EpochMembershipCoordinator` even has valid storage.

There's only one function from storage needed by `EpochMembershipCoordinator`, and that's add_drb_result. So, we create an async callback type with some gnarly looking bounds:
```rust
pub type StorageAddDrbResultFn<TYPES> = Arc<
    Box<
        dyn Fn(<TYPES as NodeType>::Epoch, DrbResult) -> BoxFuture<'static, Result<()>>
            + Send
            + Sync
            + 'static,
    >,
>;
```

This type is added to EpochMembershipCoordinator, wrapped in an Option to disable this behavior, as some things which need an EpochMembershipCoordinator don't have storage at all.

I had to change the point where storage is wrapped in Arc<RwLock<>> to make this work. The function `storage_add_drb_result` deserves extra attention to make sure my pinning is right, but it type checks and hopefully looks straightforward enough.

---
  - https://app.asana.com/0/0/1210111843237781

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ env:
 
 jobs:
   build:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: ubuntu-24.04-8core
     steps:
       - uses: rui314/setup-mold@v1
 
@@ -60,7 +60,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           prefix-key: v1-rust
-          cache-provider: buildjet
+          cache-provider: github
 
       - name: Build
         # Build in release without `testing` feature, this should work without `hotshot_example` config.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ env:
 jobs:
   build-test-artifacts-postgres:
     name: Build test artifacts (postgres)
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: ubuntu-24.04-8core
     steps:
       - uses: rui314/setup-mold@v1
 
@@ -40,7 +40,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           cache-all-crates: true
-          cache-provider: buildjet
+          cache-provider: github
 
       - name: Show memory
         run: free -h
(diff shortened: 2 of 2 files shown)


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -36,6 +37,9 @@ pub struct EpochMembershipCoordinator<TYPES: NodeType> {
     /// wait for the actual catchup and allert future callers when it's done
     catchup_map: Arc<Mutex<EpochMap<TYPES>>>,
 
+    /// Callback function to store a drb result when one is calculated during catchup
+    storage_add_drb_result_fn: Option<StorageAddDrbResultFn<TYPES>>,
+
```
bfish713, line 42: Is there a reason to not just take the storage, I feel like having a callback here, that pretty much just calls storage to add DRB doesn't gain us much and just makes the code a bit confusing.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -206,9 +216,19 @@ where
...
+                    .unwrap();
+
+            if let Some(cb) = &self.storage_add_drb_result_fn {
+                tracing::info!("Writing drb result from catchup to storage for epoch {epoch}");
+                if let Err(e) = cb(epoch, drb).await {
+                    tracing::warn!("Failed to add drb result to storage: {e}");
+                }
+            }
+
+            drb
```
bfish713, line 231: This code should move out of this else clause right?  We want to store the in the case we fetched it and the case we calculated it

bfish713 (commented):
Don't see any issues besides the one comment

ss-es (approved):
looks good overall, just some minor comments

crates/hotshot/types/src/traits/storage.rs
```diff
@@ -148,3 +151,30 @@ pub trait Storage<TYPES: NodeType>: Send + Sync + Clone {
         block_header: TYPES::BlockHeader,
     ) -> Result<()>;
 }
+
+pub type StorageAddDrbResultFn<TYPES> = Arc<
```
ss-es, line 155: also, does this really need to be in an `Arc`?

PR #3004, opened 2025-04-10
Fix the admin address in dev node

ImJeremyHe:


alxiong (approved):

sequencer/src/lib.rs
```diff
@@ -846,11 +846,6 @@ pub mod testing {
             self
         }
 
-        pub fn epoch_start_block(mut self, epoch_start_block: u64) -> Self {
-            self.config.epoch_start_block = epoch_start_block;
-            self
-        }
```
alxiong: if we remove this, shouldn't we also remove `epoch_height()` above, we might as well specify all of these via env var.

alxiong: ah nvm, this is unrelated to this PR, you can ignore it.

sveitser (approved):

PR #3006, opened 2025-04-10
Refactor state prover logic

alxiong:
Closes #<ISSUE_NUMBER>

### This PR:

In debugging a error reported by @ImJeremyHe during our testnet, I refactor (and thus attempt to fix) our state prover to the following logic: 

ProverServiceLogic(1)

### DIFF
--- a/hotshot-state-prover/src/circuit.rs
+++ b/hotshot-state-prover/src/circuit.rs
@@ -307,7 +307,7 @@ where
     Ok((
         circuit,
         GenericPublicInput::new(
-            lightclient_state.clone(),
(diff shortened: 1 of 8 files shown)


hotshot-state-prover/src/circuit.rs
```diff
@@ -307,7 +307,7 @@ where
     Ok((
         circuit,
         GenericPublicInput::new(
-            lightclient_state.clone(),
```
tbro, line 310: Should `Clone` be removed from derived impls on the type?

mrain (approved):

PR #3007, opened 2025-04-10
L1Client: prepend latest block to block stream

sveitser:
When watching blocks we only subscribe to new blocks so if the L1 does not produce any blocks we never update the state of the L1 client. For testing we may want to use anvil with on demand mining, this ensures that our L1 client quickly has blocks to work with. I think as part of a larger refactor we should make the L1 finalized non-optional and fetch the latest L1 finalized block on startup.

The test added fails on `main` but passes with the commit afterwards.


sveitser:
Hmm the new test fails on the CI

sveitser:
We only put finalized blocks into the cache when we get the next finalized block so this currently doesn't really work because the cache will still be uninitialized.

tbro (approved):

PR #3008, opened 2025-04-10
staking-cli: improve info command

sveitser:
- Include all validators, not just the ones with stake.
- Show delegators.
- Test info command against fix output.
- Make TestSystem keys deterministic.
- Allow compact and full BLS key output.


staking-cli/src/info.rs
```diff
@@ -12,30 +12,49 @@ pub async fn stake_table_info(
...
-            comm,
             format_ether(validator.stake),
         );
+
+        if validator.delegators.is_empty() {
+            tracing::info!(" - No delegators");
+            continue;
+        }
+
+        // sort delegators by address for easier reading
```
imabdulbasit, line 49: shouldn't sort be by stake?

sveitser: Thought about it but I just think it's easier to find stuff like this.

imabdulbasit (approved):

PR #3010, opened 2025-04-10
fix prover-service logic

mrain:
Closes #<ISSUE_NUMBER>


alxiong (approved):
This is correct. agree with Chengyu's fix

PR #3029, opened 2025-04-11
Fix tests

tbro:
Fix broken test. Test passes locally. 

First guess is limited resources on CI. So trying spawning less tasks. Another possibility might be more retries, w/ the idea that when other tests finish there will be more bandwidth and retry might succeed.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -17,7 +17,7 @@ retries  = 10
 
 [[profile.default.overrides]]
 filter = 'test(test_hotshot_event_streaming_epoch_progression)'
-slow-timeout = { period = "4m", terminate-after = 4 }
+slow-timeout = { period = "8m", terminate-after = 4 }
 
 # The restart tests run an entire sequencing network, and so are quite resource intensive.
 [[profile.default.overrides]]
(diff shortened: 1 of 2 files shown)


sveitser:
@tbro have you reached any conclusions here to what the issue is?

imabdulbasit:
> Locally I'm hitting: #3033

merged main with this fix. CI should pass on this branch

.config/nextest.toml
```diff
@@ -15,10 +15,20 @@ test(test_process_client_handling_stream_subscribe_voters)
...
 # The restart tests run an entire sequencing network, and so are quite resource intensive.
 [[profile.default.overrides]]
 filter = 'test(slow_test_restart)'
+slow-timeout = { period = "1m", terminate-after = 3 }
+threads-required = 'num-cpus'
+
+[[profile.default.overrides]]
+filter = 'package(request-response)'
 threads-required = 'num-cpus'
+retries  = 20
```
sveitser, line 31: Is this needed? Isn't setting `threads-required = 'num-cpus'` should be enough?

tbro: let's find out

sveitser (approved):

PR #3031, opened 2025-04-11
fix: run request-response tests serially

sveitser:
Unable to reproduce failure locally, not sure this will help.


sveitser:
We merged https://github.com/EspressoSystems/espresso-network/pull/3029 so there's not much need for this PR anymore.

PR #3042, opened 2025-04-11
Check if we have stake table before querying L1

ss-es:


types/src/v0/impls/stake_table.rs
```diff
@@ -760,6 +760,14 @@ impl Membership<SeqTypes> for EpochCommittees {
         epoch: Epoch,
         block_header: Header,
     ) -> Option<Box<dyn FnOnce(&mut Self) + Send>> {
+        if self.state.contains_key(&epoch) {
```
tbro, line 763: Doesn't hotshot block on `None`? I feel we should instead make the l1 fetch conditional on this result.

tbro (changes requested):

types/src/v0/impls/stake_table.rs
ss-es: we shouldn't -- I also checked the actual call sites, and it agrees with my understanding

if `add_epoch_root` returns `None`, that just means that no update of the membership struct is required, i.e. the stake table is already present for whatever reason

the requirement we have is actually that `add_epoch_root` can never fail, i.e. after we call `add_epoch_root` we must *always* get an updater for `Membership` or `None` if no update is necessary. we are not allowed to e.g. timeout and just return `None`  if we can't get the stake table on a valid post-epoch block header

ss-es: but this is a good point, I'm not sure we really do that

returning `None` should not be used for errors, but we are definitely using it to propagate errors in the following code. need to think about that

tbro: Ok. So `chain_config.stake_table_contract` and `l1_finalized()` shoujld always be `Some`, so if we change those to panic, it might clear up the semantics a little. The other calls I think are already in line w/ what you are saying, they return `None` if no update is needed.

tbro (approved):

PR #3047, opened 2025-04-11
Move contract deployment into test network

tbro:
POS specific setup is moved inside the `TestNetwork` family and version gated. Now in order to run tests that require POS contracts, you need to call `pos_hook<Version>` before calling `build()` on the `TestNetworkConfigBuilder`. This PR also updates `test_hotshot_event_streaming_epoch_progression` to use this setup (as opposed to "inline" setup as before) to demonstrate a working model. 

The expectation is that going forward tests that require working epochs will use this instead of duplicating the same setup everywhere. 

As a followup we should have `dev-node` use this logic.

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -961,13 +961,6 @@ pub mod test_helpers {
             let mut cfg = cfg;
             let mut marketplace_builder_url = "http://example.com".parse().unwrap();
             let mut builder_tasks = Vec::new();
-            // Self::version_hooks(
-            //     &mut cfg,
-            //     &mut marketplace_builder_url,
-            //     &mut builder_tasks,
-            //     bind_version,
-            // )
-            // .await;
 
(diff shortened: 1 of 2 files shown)


sequencer/src/api.rs
```diff
@@ -831,6 +837,101 @@ pub mod test_helpers {
...
+                signer,
+                stake_table_address,
+                contracts
+                    .address(Contract::EspTokenProxy)
+                    .expect("ESP token deployed"),
+                staking_priv_keys,
+            )
+            .await?;
+
+            let chain_config = ChainConfig {
```
imabdulbasit, line 920: can you check node state also gets the same chain config

tbro: Yes `NodeState` is inialized from these values later in the pipeline: 
https://github.com/EspressoSystems/espresso-network/blob/ea9bd39f200256325307dd002dda671578d08024/sequencer/src/lib.rs#L1097-L1127

PR #3048, opened 2025-04-11
skip reward_account dependency if version < 0.3

imabdulbasit:


sequencer/src/api/sql.rs
```diff
@@ -513,15 +517,20 @@ async fn reconstruct_state<Mode: TransactionMode>(
...
-    );
+
+    if !reward_accounts.is_empty() {
+        state.reward_merkle_tree = load_reward_accounts(tx, from_height, &reward_accounts)
+            .await
+            .context(
+                "unable to reconstruct state because reward accounts are not available at origin",
+            )?
+            .0;
+        ensure!(
```
imabdulbasit, line 528: not very sure about this check

sequencer/src/api/sql.rs
```diff
@@ -659,6 +669,12 @@ async fn reward_header_dependencies<Mode: TransactionMode>(
             catchup.add_chain_config(cf);
         };
 
+        let version = header.version();
```
imabdulbasit, line 672: I think the above chain config catchup is not needed anymore because we already do it for fee header deps?

sequencer/src/api/sql.rs
```diff
@@ -629,12 +638,14 @@ async fn reward_header_dependencies<Mode: TransactionMode>(
     let epoch_height = instance.epoch_height;
 
     let Some(epoch_height) = epoch_height else {
-        bail!("epoch height not set");
+        tracing::warn!("epoch height is not set. returning empty reward_header_dependencies");
+        return Ok(HashSet::new());
```
jbearer, line 640: Is this acceptable? It won't lead to false positives? Under what scenario is the epoch height not set?

imabdulbasit: so If we are running on version < EpochVersion, this function will be called anyway but it shouldn't error out I think

imabdulbasit: epoch height is an Optoon in NodeState. I guess it doesn't matter because of the way it is set lol

jbearer: Oh I see. This is basically checking whether the 0.3 upgrade is configured at all (statically), and the later version check is further checking whether the upgrade has been enabled yet for each header?

sequencer/src/api/sql.rs
```diff
@@ -619,45 +624,31 @@ async fn fee_header_dependencies<Mode: TransactionMode>(
...
+async fn reward_header_dependencies(
     instance: &NodeState,
     leaves: impl IntoIterator<Item = &Leaf2>,
 ) -> anyhow::Result<HashSet<RewardAccount>> {
     let mut reward_accounts = HashSet::default();
     let epoch_height = instance.epoch_height;
 
     let Some(epoch_height) = epoch_height else {
-        bail!("epoch height not set");
+        tracing::warn!("epoch height is not set. returning empty reward_header_dependencies");
```
jbearer, line 635: Nit: idk if this should be a warn log because it can happen in totally normal usage prior to rewards. Maybe info?

jbearer (approved):
LGTM

sequencer/src/api/sql.rs
imabdulbasit: yes

imabdulbasit: tho NodeState always passes Some(epoch_height) which is not okay, but will fix it later

PR #3052, opened 2025-04-12
fix: installing just with snap takes ages

sveitser:
Weirdly, installing just took almost 9 minutes in this run

https://github.com/EspressoSystems/espresso-network/actions/runs/14417749807/job/40436465426

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,11 +54,6 @@ jobs:
     steps:
       - uses: rui314/setup-mold@v1
 
-      - name: Install just command runner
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -54,11 +54,6 @@ jobs:
     steps:
       - uses: rui314/setup-mold@v1
 
-      - name: Install just command runner
```
sveitser, line 57: I think we're not using just for this step.

imabdulbasit (approved):

PR #3060, opened 2025-04-14
add audit_fix job

tbro:
Attempt to automate cargo-audit fixes

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -16,6 +16,7 @@ on:
   # Run the audit job once a day on main.
   schedule:
     - cron: "0 0 * * *"
+
 jobs:
   security_audit:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/audit.yml
```diff
@@ -25,3 +26,25 @@ jobs:
...
+
+  audit_fix:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Run cargo audit fix
+        uses: simonhyll/cargo-audit@v1
```
sveitser, line 38: I think we could run `cargo audit fix` instead?

tbro: The theoretical  benefit of this is that it creates a PR with the fix for us.

sveitser: I mean instead of using this action here. The PR is created in the next step.

tbro: Oh yea. I read that wrong.

tbro: So I think in order to use `cargo audit fix` we need to `cargo install cargo-audio --features fix` ([reference](https://github.com/rustsec/rustsec/blob/main/cargo-audit/README.md#cargo-audit-fix-subcommand)). Which I'm pretty sure `rustsec/audit-check` does not do ([reference](https://github.com/petalmd/rustsec-audit-check/blob/main/src/main.ts#L17C1-L17C46)).  So I guess the choice is to either use the method currently employed (dockerfile) or build `cargo-audit`.

sveitser (approved):

PR #3069, opened 2025-04-15
Ab/release combined water conflicts resolved

imabdulbasit:


sequencer/src/state_signature/relay_server.rs
```diff
@@ -308,14 +308,6 @@ impl StateRelayServerDataSource for StateRelayServerState {
             },
         };
         let epoch = epoch_from_block_number(block_height, blocks_per_epoch);
-
-        // if epoch is activated and block_height is greater than epoch root, ignore and drop
-        if block_height > self.epoch_start_block.expect("init_genesis wrong")
```
mrain, line 313: can we keep these changes in the file?

mrain (approved):
Took a pass on the prover-service/vid related changes, LGTM with a minor comment.

data/genesis/demo-pos.toml
```diff
@@ -1,6 +1,7 @@
 base_version = "0.2"
 upgrade_version = "0.3"
-epoch_height = 20
+epoch_height = 200
```
tbro, line 3: These values look high for local testing. Should we create another configuration for water, and lower these?

data/genesis/demo-pos.toml
```diff
@@ -21,15 +21,14 @@ number = 0
 
 [[upgrade]]
 version = "0.3"
-start_proposing_view = 10
-stop_proposing_view = 50
+start_proposing_view = 200
```
tbro, line 24: Same as above.

sequencer/src/state_signature/relay_server.rs
imabdulbasit: do you want only this change from main or the whole file>

data/genesis/demo-pos.toml
imabdulbasit: upgrade doesn't work with epoch height < 200, so these changes are fine

types/src/v0/impls/state.rs
```diff
@@ -745,7 +745,8 @@ fn validate_builder_fee(
         // TODO Marketplace signatures are placeholders for now. In
         // finished Marketplace signatures will cover the full
         // transaction.
-        if version.minor >= MarketplaceVersion::MINOR {
+        if version >= MarketplaceVersion::VERSION {
+            tracing::debug!("Validating sequencing fee signature.");
```
tbro, line 749:
```suggestion
            tracing::debug!("Validating (Marketplace) sequencing fee signature.");
```

Also add a log to the other branch of the if statement.

types/src/v0/impls/reward.rs
```diff
@@ -409,7 +409,7 @@ pub async fn first_two_epochs(height: u64, instance_state: &NodeState) -> anyhow
         .first_epoch()
         .context("The first epoch was not set.")?;
 
-    Ok(epoch == first_epoch || epoch == first_epoch + 1 || epoch == EpochNumber::new(0))
+    Ok(epoch <= first_epoch + 1)
```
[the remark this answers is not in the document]
imabdulbasit: epoch <= first_epoch + 1 would cover that? 
lets say epoch == 0 and first epoch = 6
0 <= 6 +1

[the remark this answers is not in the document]
tbro: my mistake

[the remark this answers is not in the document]
imabdulbasit: this was added because hotshot had an error where it was setting first epoch as 0. I believe that was fixed

types/src/v0/impls/state.rs
```diff
@@ -755,16 +756,20 @@ fn validate_builder_fee(
...
-            fee_info
-                .account()
-                .validate_fee_signature(
-                    &signature,
-                    fee_info.amount().as_u64().unwrap(),
-                    proposed_header.metadata(),
-                )
-                .then_some(())
-                .ok_or(BuilderValidationError::InvalidBuilderSignature)?;
+        } else if !fee_info.account().validate_fee_signature(
```
tbro, line 759: So maybe we want to keep this as it is since we know that it is working. But I don't think it is great to have different coding styles across  branches of an `if` statement. I think the `else` block could be better written to use `or_else` combinator. Then we could log both branches. But maybe it's better to keep the tested code for now.

imabdulbasit: yeah I also don't like this, but I wanna keep these changes in this PR. We can make a followup PR to change this

data/genesis/demo-pos.toml
ss-es: I think we could get it down to `50` with some refactoring, but we'd need to change some upgrade constants (currently set via `UPGRADE_CONSTANTS`, which is an associated constant on `NodeType`)

But this is not entirely straightforward and I think requires some changes to the hotshot testing infrastructure

hotshot-task-impls/src/helpers.rs
```diff
@@ -616,6 +622,7 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
     }
 
     let consensus_reader = consensus.read().await;
+    //let parent_view_number = consensus_reader.high_qc().view_number();
```
ss-es, line 625: delete this comment?

vid/Cargo.toml
```diff
@@ -14,7 +14,7 @@ ark-serialize = { workspace = true }
 ark-std = { workspace = true }
 displaydoc = "0.2"
 jf-crhf = { git = "https://github.com/EspressoSystems/jellyfish", tag = "jf-crhf-v0.1.1" }
-jf-merkle-tree = { git = "https://github.com/EspressoSystems/jellyfish", tag = "jf-merkle-tree-v0.2.1" }
+jf-merkle-tree = { git = "https://github.com/EspressoSystems/jellyfish", rev = "bd6098da2c4a53bf7f4042c96630e2d3cb1d844b" }
 jf-poseidon2 = { git = "https://github.com/EspressoSystems/jellyfish", tag = "jf-poseidon2-v0.1.0", features = ["bn254"] }
 jf-utils = { git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.5" }
```
ss-es, line 19: cc @mrain 

I think we shouldn't change this in this PR but should we rework this to `workspace = true` for all dependencies?

ss-es (approved):
looks fine to me

hotshot-task-impls/src/helpers.rs
imabdulbasit: done

vid/Cargo.toml
imabdulbasit: breaks lot of things https://github.com/EspressoSystems/espresso-network/pull/3073

PR #3075, opened 2025-04-15
staking-cli: avoid binary re-compiliation on CI

sveitser:
With current target set we get a targed subdirectory inside the target directory. This directory isn't there when we run `cargo build`.


sveitser:
This did not help.

imabdulbasit (approved):

PR #3080, opened 2025-04-16
Reorganize hotshot crates

ss-es:
### This PR:
Reorganizes HotShot crates (specifically consensus-related crates) under `crates/hotshot/*`. This significantly speeds up my own personal workflow (especially with debugging) by clearly separating code into a subdirectory rather than cluttering at the top-level, and should also help make relationships between crates clearer.

### This PR does not:
No files or crate names are changed, only the directory structure. This should not affect import statements and `git` should recognize all files as being moved (not deleted/created). No `cargo` commands should be affected because crate names are unchanged, e.g. `cargo test -p hotshot` will work exactly as it did before


Cargo.toml
```diff
@@ -14,25 +14,25 @@ members = [
     "contracts/rust/adapter",
     "contracts/rust/diff-test",
     "contracts/rust/gen-vk-contract",
-    "hotshot",
-    "hotshot-builder-api",
+    "crates/hotshot/builder-api",
```
tbro, line 17: My expectation of a `crates` folder would be that all the crates would be found there. However we are only moving some of the crates into the `crates` folder.

ss-es: yeah, we discussed doing this incrementally in the standup today

after this, I'll PR separately for `sequencer` and `builder` over the next week. I didn't want to rush to try and categorize all crates in one go

tbro:
Please add an explanation to the PR description as to why we are making this change. What problem does it solve? What improvement should we expect? Also please add to the description any change in how we will interact with the repo. I'm guessing `cargo test -p sequencer` will continue to work. But `cargo test -p hotshot` will not. 

In addition it would be nice if this information made it into the commit message on merge.

PR #3082, opened 2025-04-16
staking-cli: decode contract reverts

sveitser:
This shows descriptive error messages in the CLI instead of raw Ethereum revert data.

Added some tests but to see in the command line, run for example

```
cargo run --bin staking-cli -- init --mnemonic "test test test test test test test test test test test junk" --account-index 0
cargo run --bin staking-cli -- delegate --validator-address 0x0774A80189030B3673983ebd4D81DCB5AF21c78A --amount 1 
```


alxiong (approved):
this is great! 

only one question: should we centralize these impl in a file instead of floating around? what about put them into `contract-adapter` crate?

sveitser:
Yeah we can move it to the contract adapter.

PR #3083, opened 2025-04-16
POS upgrade test

tbro:
Adds a unit test for POS upgrades, principally  addition of `set_upgrades` method  to `TestNetworkBuilder`.

Also:

  * improvements for upgrade testing "framework"
  * removal of upgrade tests for version we won't be upgrading to (again, i.e. `FeeVersion`).
  * deployment code shared w/ `pos_hook` moved to a common place for de-duplication
  * some shared functions needed contract deployment moved to an accessible location
  * created a new type for upgrade map, in order to add some convenience functions
  * moved upgrade configurations into static methods on `Upgrade`.

---
  - https://app.asana.com/0/0/1210024927640840

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -2650,66 +2650,6 @@ mod test {
     }
 
     #[tokio::test(flavor = "multi_thread")]
-    async fn test_fee_upgrade_view_based() {
-        setup_test();
-
-        let mut upgrades = std::collections::BTreeMap::new();
-        type MySequencerVersions = SequencerVersions<V0_1, FeeVersion>;
-
-        let mode = UpgradeMode::View(ViewBasedUpgrade {
-            start_voting_view: None,
-            stop_voting_view: None,
-            start_proposing_view: 1,
-            stop_proposing_view: 10,
-        });
-
-        let upgrade_type = UpgradeType::Fee {
-            chain_config: ChainConfig {
-                max_block_size: 300.into(),
-                base_fee: 1.into(),
-                ..Default::default()
-            },
-        };
-
-        upgrades.insert(
-            <MySequencerVersions as Versions>::Upgrade::VERSION,
-            Upgrade { mode, upgrade_type },
-        );
-        test_upgrade_helper::<MySequencerVersions>(upgrades, MySequencerVersions::new()).await;
-    }
-
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_fee_upgrade_time_based() {
-        setup_test();
-
-        let now = OffsetDateTime::now_utc().unix_timestamp() as u64;
-
-        let mut upgrades = std::collections::BTreeMap::new();
-        type MySequencerVersions = SequencerVersions<V0_1, FeeVersion>;
-
-        let mode = UpgradeMode::Time(TimeBasedUpgrade {
-            start_proposing_time: Timestamp::from_integer(now).unwrap(),
-            stop_proposing_time: Timestamp::from_integer(now + 500).unwrap(),
-            start_voting_time: None,
-            stop_voting_time: None,
-        });
-
-        let upgrade_type = UpgradeType::Fee {
-            chain_config: ChainConfig {
-                max_block_size: 300.into(),
-                base_fee: 1.into(),
-                ..Default::default()
-            },
-        };
-
-        upgrades.insert(
-            <MySequencerVersions as Versions>::Upgrade::VERSION,
-            Upgrade { mode, upgrade_type },
-        );
-        test_upgrade_helper::<MySequencerVersions>(upgrades, MySequencerVersions::new()).await;
-    }
-
-    #[tokio::test(flavor = "multi_thread")]
     async fn test_pos_upgrade_view_based() {
         setup_test();
 
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -2730,13 +2730,14 @@ mod test {
             }
         };
 
+        let wanted_view = upgrade.new_version_first_view + wait_extra_views;
         // Loop until we get the `new_version_first_view`, then test the upgrade.
         loop {
             let event = events.next().await.unwrap();
             let view_number = event.view_number;
 
             tracing::debug!(?view_number, ?upgrade.new_version_first_view, "upgrade_new_view");
-            if view_number > upgrade.new_version_first_view + wait_extra_views {
+            if view_number > wanted_view {
                 let states: Vec<_> = network
                     .peers
                     .iter()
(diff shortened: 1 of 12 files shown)


staking-cli/src/demo.rs
```diff
@@ -187,3 +204,143 @@ pub async fn stake_for_demo(config: &Config, num_validators: u16) -> Result<()>
     tracing::info!("completed staking for demo");
     Ok(())
 }
+
+pub async fn pos_deploy_routine(
```
sveitser, line 208: Could the main function in `bin/deploy.rs` handle this deployment instead? I feel like we should move that function out of the binary so it can be used elsewhere to avoid code duplication. @alxiong do you know if this is possible now or do we still have that issues with two way dependency between prover service and deploy code because the deploy code needs

https://github.com/EspressoSystems/espresso-sequencer/blob/6bfe9aca6b7c70a29b517eee31bcbd39d1ce8ac9/hotshot-state-prover/src/service.rs#L214

and the prover service deploys contracts for tests. What about splitting the prover into a library and service crate?

tbro: > Could the main function in `bin/deploy.rs` handle this deployment instead?

Not entirely as is, but yes placing this here was just to unblock the test. We can find a better organization.

tbro: https://app.asana.com/1/1208976916964769/project/1209380107411004/task/1210024497670042

staking-cli/src/demo.rs
```diff
@@ -313,3 +322,154 @@ pub async fn stake_for_demo(config: &Config, num_validators: u16) -> Result<()>
     tracing::info!("completed staking for demo");
     Ok(())
 }
+
+/// Commonly used contract deployment routine.
+// TODO move to proper place for shared code. See:
```
sveitser, line 327: I think when we add something that we expect to be removed soon it's preferable to add it right where it's used, instead of in another crate. This would also not require duplicating the staking_priv_keys function and we could put gate everything with `cfg(test)` or the testing feature.

sequencer/src/lib.rs
```diff
@@ -851,11 +868,56 @@ pub mod testing {
...
+                    let blocks_per_epoch = self.config.epoch_height;
+                    let epoch_start_block = self.config.epoch_start_block;
+
+                    let initial_stake_table =
+                        stake_table(self.config.known_nodes_with_stake.clone());
+
+                    let staking_private_keys =
+                        staking_priv_keys(&self.priv_keys, &self.state_key_pairs, NUM_NODES);
+
+                    let address = pos_deploy_routine(
```
sveitser, line 886: If we remove `stake_in_contract_for_test` from `pos_deploy_routine` and instead call it here then we don't need to pass `staking_priv_keys` to `pos_deploy_routine` and don't need to copy that code into the staking-cli crate.

types/src/v0/impls/instance_state.rs
```diff
@@ -264,6 +267,25 @@ impl NodeState {
...
+        self.0
+            .get(&version)
+            .unwrap()
+            .upgrade_type
+            .chain_config()
+            .unwrap()
+    }
+}
+
+impl From<BTreeMap<Version, Upgrade>> for UpgradeMap {
```
sveitser, line 283: use `derive_more::From`?

types/src/v0/impls/instance_state.rs
```diff
@@ -264,6 +267,25 @@ impl NodeState {
     }
 }
 
+/// NewType to hold upgrades and some convenience behavior.
+pub struct UpgradeMap(pub BTreeMap<Version, Upgrade>);
```
sveitser, line 271: IIUC this is intended only as test code due to the `unwrap`? Can we feature gate it as such so we don't at some point end up using it in normal code?

sequencer/src/api.rs
```diff
@@ -2647,145 +2589,50 @@ mod test {
...
 
-        let chain_config_upgrade = upgrades
-            .get(&<MockSeqVersions as Versions>::Upgrade::VERSION)
-            .unwrap()
-            .upgrade_type
-            .chain_config()
-            .unwrap();
+        let test_config = TestConfigBuilder::default()
+            .epoch_height(200)
+            .epoch_start_block(321)
```
sveitser, line 2626: Why is this 321?

This test took 6 minutes when I ran it. Do you think there's a way to perform it with fewer views?

staking-cli/src/demo.rs
```diff
@@ -34,6 +40,9 @@ use crate::{
     Config,
 };
 
+pub type StakeTableVecBased = StakeTable<BLSPubKey, StateVerKey, CircuitField>;
+
+pub const STAKE_TABLE_CAPACITY_FOR_TEST: u64 = 3;
```
sveitser, line 45: These seem to be only used in tests, so suggest moving them there.

types/src/v0/impls/instance_state.rs
```diff
@@ -325,6 +347,43 @@ impl Upgrade {
             },
         }
     }
+    pub fn pos_view_based(address: Address) -> Upgrade {
+        let chain_config = ChainConfig {
+            base_fee: 0.into(),
+            stake_table_contract: Some(address),
```
sveitser, line 353: More descriptive name? For example:
```suggestion
    pub fn pos_view_based(stake_table_contract: Address) -> Upgrade {
        let chain_config = ChainConfig {
            base_fee: 0.into(),
            stake_table_contract: Some(stake_table_contract),
```

staking-cli/src/demo.rs
tbro: We need it in a place where it can be reached by all the tests, not just the ones in this PR. We have many more tests to write. Some of them will live in the `tests` crate. Importing `seqeuncer` crate everywhere is not good IMO.

tbro:
Merging for the following reasons:

  * this PR has been awaiting review since friday
  * if I change things now who knows when it will get merged due slow iterations, updating from main, etc
  * we have many more tests to write  this PR isn't the finished product but just some small steps.
  * I consider having tests such as this one merged and running on PRs/main far greater priority than any of the comments here.

sequencer/src/api.rs
tbro: These are the values deemed correct from testing:
https://github.com/EspressoSystems/espresso-network/blob/dee7f5e80d2ef8e237ddbe0773ff6dc8f449d4db/data/genesis/demo-pos.toml#L3-L4

There was [some discussion](https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/confirmation.20layer.20pos.20tests/near/512388886). I did try lower values with no success. Could be that the values I tried were too low. But there is a trade-off between how much trial and error is worth while when we currently have something working.

PR #3086, opened 2025-04-17
rfc: move decode_revert to contract adapter crate

sveitser:
Fix `just fmt` after crates re-org.

I attempted to also rework `contract_send` but it got messy when I added another generic to it. We use that function almost never so we can think about maybe doing things a bit different instead.


contracts/rust/adapter/Cargo.toml
```diff
@@ -5,6 +5,10 @@ version = { workspace = true }
 authors = { workspace = true }
 edition = { workspace = true }
 
+# Contract bindings docs fail to compile
+[lib]
+doctest = false
```
alxiong, line 10: Can't we add module level instruction? Guarded at bindings module only?

alxiong (approved):
Do you know the reason for CI failure?

contracts/rust/adapter/Cargo.toml
sveitser: I tried 

```
#[cfg(not(doctest))]
#[allow(dead_code)]
pub(crate) mod bindings;
``` 
but that just fails compile with doctests. Since we don't have any doctests in this crate, what's wrong with disabling it?

alxiong: nothing wrong. i thought it would work. it's fine.

PR #3089, opened 2025-04-17
Source-local message deduplication

rob-maron:
Previously, the cache just deduplicated messages as long as we had seen them recently. This came with the side effect of potentially ignoring transactions we had recently seen.

## This PR:

This PR Updates the message deduplication logic to only deduplicate between the two sources (primary and secondary, CDN and Libp2p). 

Basically, now we will only ignore a message if we had recently seen it on one network and we are now seeing it on the other. If we get the same message from the CDN or Libp2p twice, it will go through. But if we see it once on the CDN, and once on Libp2p, the second will be ignored.


rob-maron:
> I'm missing context as to why we dropped the number of cache entries from 200k to 5k, but otherwise this seems pretty straightforward

Thanks Phil!

The change there was because instead of being a general message cache, it is now supposed to only deduplicate between the two networks. So I would expect if we were to see the same message on both networks that they are not more than 5,000 apart

PR #3090, opened 2025-04-17
simple builder: keep block header input in state

sveitser:
This fixes the flaky query_service: availability::test::test_api

Not sure if it will break other things. Let's see what CI says.


ss-es (approved):

imabdulbasit (approved):
Thanks for debugging this

PR #3092, opened 2025-04-18
post deploy verification checks and contract verification

### LINKED ISSUE #3064: Verify inputs to deployment scripts
Verify that the contract inputs to deployments are valid.

- LightClient inputs (PlonkVerifier)
- LightClientV2 inputs (PlonkVerifier, PlonkVerifierV2)

alysiahuggins:
Closes #3064 

### This PR:
- Implements post-deployment verification checks for various contracts in `deployer.rs`
- Ensures that contracts that require contracts as inputs, have that contract address verified e.g. light client needs the PlonkVerifier address

### This PR does not:
- verify any specific details about the PlonkVerifier contract as the constants are internal and the verification functions require actual proofs

### Key places to review:
- post deployment verification checks
- especially check the staketable deployment verifications - it expects that the token will be deployed before the staketable is deployed

### How to test this PR:  
`cargo test -p sequencer-utils -- deployer`


alxiong (commented):
Changes in this PR is a little different from what I expect. These asserts are sort of already covered by existing unit tests in `deployer.rs` (those constant setting are already checked) 

Put it another way, i don't there's any possibility for anyone to run our `deployer.rs`'s function and wrongly upgrade LightClientV2 to a wrong PlonkVerifierV2 address. 
This is because deploying PV2 and immediately use its address to upgrade light client is bundled together inside the `deployer.rs::upgrade_light_client_v2()` whose correctness is tested in unit tests. 



I thought what we miss is, which causes our limbo in testnet attempt, is we try to use forge script to upgrade. 
So i was expecting to have a function (in whatever language, rust or forge script or bash via `cast`) that I can run in a command line to test whether a certain address is the proper `PlonkVeriferV2.sol`, and we will inject this checking step inside the flow of upgrading light client manual.

utils/src/deployer.rs
```diff
@@ -316,6 +318,22 @@ pub async fn deploy_light_client_proxy(
             .await?;
     }
 
+    // post deploy verification checks
+    assert_eq!(lc_proxy.getVersion().call().await?.majorVersion, 1);
+    assert_eq!(lc_proxy.owner().call().await?._0, admin);
+    assert_eq!(
+        lc_proxy.permissionedProver().call().await?._0,
+        prover.unwrap()
+    );
```
alxiong, line 327: should only check inside `if let Some(prover)` block

utils/src/deployer.rs
```diff
@@ -596,6 +672,19 @@ pub async fn is_proxy_contract(provider: impl Provider, addr: Address) -> Result
     Ok(impl_address != Address::default())
 }
 
+pub async fn verify_contract_address(provider: impl Provider, address: Address) -> Result<bool> {
```
alxiong, line 675: rename to `is_contract()`? 
(we probably will use `verify_contract_address()` for CREATE2 consistency check.)

utils/src/deployer.rs
alysiahuggins: good call, updated here https://github.com/EspressoSystems/espresso-network/commit/6fc2d192bd2dd2c8fa15f2fc926bf9c484988ea0

utils/src/deployer.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/6fc2d192bd2dd2c8fa15f2fc926bf9c484988ea0

alxiong (approved):

alxiong:
```
deploy-sequencer-contracts-1   | 2025-04-23T16:19:25.435532Z  INFO sequencer_utils::deployer: deployed ESPRESSO_SEQUENCER_STAKE_TABLE_ADDRESS at 0xed1db453c3156ff3155a97ad217b3087d5dc5f6e
deploy-sequencer-contracts-1   | 
deploy-sequencer-contracts-1   | thread 'main' panicked at /home/runner/actions-runner/_work/espresso-network/espresso-network/utils/src/deployer.rs:568:5:
deploy-sequencer-contracts-1   | assertion failed: is_contract(&provider, light_client_addr).await?
```

@alysiahuggins this is the key failure log

it seems that when we try to `deploy_stake_table_proxy(,..,light_client_addr)`, the light client address is wrong.

utils/src/deployer.rs
```diff
@@ -503,6 +564,10 @@ pub async fn deploy_stake_table_proxy(
         .await?;
     let stake_table = StakeTable::new(stake_table_addr, &provider);
 
+    // verify light client and token addresses are contracts
+    assert!(is_contract(&provider, light_client_addr).await?);
```
alxiong, line 568: chatted with @sveitser, the design of orchestrator and the question of "should we load init stake table from a static config file" is still open for discussion. 

for now, stop-gap fix is just remove this check, add a TODO that follows https://github.com/EspressoSystems/espresso-network/issues/3163 
@alysiahuggins

PR #3095, opened 2025-04-19
Fix upgrade

ss-es:


hotshot-task-impls/src/upgrade.rs
```diff
@@ -304,8 +309,10 @@ impl<TYPES: NodeType, V: Versions> UpgradeTaskState<TYPES, V> {
                         epoch_from_block_number(self.epoch_start_block, self.epoch_height);
                     let last_proposal_epoch =
                         epoch_from_block_number(last_proposal_block, self.epoch_height);
-                    let upgrade_finish_epoch =
-                        epoch_from_block_number(new_version_first_view + 10, self.epoch_height);
+                    let upgrade_finish_epoch = epoch_from_block_number(
+                        last_proposal_block + (new_version_first_view - last_proposal_view) + 10,
```
bfish713, line 313: isn't this basically `last_proposal_block + TYPES::UPGRADE_CONSTANTS.finish_offset + 10`

I'm trying to describe what this code is checking and this should be, the block the upgrade finishes on should be in the same epoch it started.  so we basically need current block + how many views it'll take from now to apply + buffer

ss-es: conceptually it's almost the same thing as what you just said, except that I think `last_proposal_block + finish_offset + 10` could be earlier than `new_version_first_view`

ss-es: the way I think of this is just (block height of last proposal) + (the maximum number of blocks we could possibly produce by `new_version_first_view`)

hotshot-task-impls/src/upgrade.rs
```diff
@@ -160,6 +160,46 @@ impl<TYPES: NodeType, V: Versions> UpgradeTaskState<TYPES, V> {
                     proposal.data.view_number()
                 );
 
+                let epoch_upgrade_checks = if V::Upgrade::VERSION == V::Epochs::VERSION {
```
bfish713, line 163: Just to confirm this is and should be a copy of the same checks on the proposal side.  Looks the same to me but just wanted a double check

PR #3104, opened 2025-04-22
Periodically update stake table events persistence

imabdulbasit:
This PR:
- spawns a background update loop for updating the stake table events persistence
- takes env `ESPRESSO_SEQUENCER_L1_STAKE_TABLE_UPDATE_INTERVAL` for update interval. Default is 1 hour
- at every interval, it updates the stake table events in persistence


sveitser:
> Default is 2minutes

I think this means 30 * 24 * 5 = 3600 calls per day, which seems a bit wasteful. For local networks we probably want this to be smaller but for public testnets maybe 1 hour is okay?

types/src/v0/impls/stake_table.rs
```diff
@@ -335,32 +337,145 @@ pub struct EpochCommittees {
...
+                state_lock.last_finalized
+            };
+
+            let Some(finalized_block) = new_finalized else {
+                sleep(l1_retry).await;
+                continue;
+            };
+
+             // Only fetch updates if the finalized block has advanced enough.
+            if finalized_block - last_synced_block >= max_range {
```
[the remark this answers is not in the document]
imabdulbasit: The idea was to avoid wasting rpc calls by making multiple requests for a range that can be fetched with a single rpc call

imabdulbasit:
> > Default is 2minutes
> 
> I think this means 30 * 24 * 5 = 3600 calls per day, which seems a bit wasteful. For local networks we probably want this to be smaller but for public testnets maybe 1 hour is okay?

We don't make rpc call every 2m because of this condition `if finalized_block - last_synced_block >= max_range `

types/src/v0/impls/stake_table.rs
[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3104/commits/758d1c306ec8bd0a95410b7743196a6ec74c9c2f

types/src/v0/impls/stake_table.rs
```diff
@@ -372,23 +460,22 @@ impl StakeTableFetcher {
...
+                .collect(),
+            _ => contract_events,
         };
 
         // There are no duplicates because the RPC returns all events,
         // which are stored directly in persistence as is.
         // However, this step is taken as a precaution.
         // The vector is already sorted above, so this should be fast.
-        events.dedup_by_key(|(k, _)| (k.0, k.1));
+        events.dedup();
```
sveitser, line 478: I think we said we want to emit a warning log here if we had duplicates?

types/src/v0/v0_1/l1.rs
```diff
@@ -151,6 +151,17 @@ pub struct L1ClientOptions {
     /// Typically this would be a WebSockets endpoint while the main provider uses HTTP.
     #[clap(long, env = "ESPRESSO_SEQUENCER_L1_WS_PROVIDER", value_delimiter = ',')]
     pub l1_ws_provider: Option<Vec<Url>>,
+    
+    /// Interval at which the background update loop polls the L1 stake table contract for new events
+    /// and updates local persistence.
+    ///
+    #[clap(
+        long,
+        env = "ESPRESSO_SEQUENCER_L1_STAKE_TABLE_UPDATE_INTERVAL",
```
sveitser, line 160: As discussed, let's set `1m` in `.env`.

types/src/v0/impls/stake_table.rs
```diff
@@ -475,7 +475,12 @@ impl StakeTableFetcher {
         // which are stored directly in persistence as is.
         // However, this step is taken as a precaution.
         // The vector is already sorted above, so this should be fast.
+        let len_before_dedup = events.len();
         events.dedup();
+        let len_after_dedup = events.len();
+        if len_before_dedup != len_after_dedup {
+            tracing::warn!("Duplicate events found and removed. This should not normally happen.")
```
sveitser, line 482: Argh it's annoying we don't get the duplicate events returned. It would be so hard to debug this just with the log message.

PR #3105, opened 2025-04-22
Add support for legacy (pre-epoch) events in hotshot-event-service

pls148:
Closes #<ISSUE_NUMBER>

### Key places to review:

---
  - https://app.asana.com/0/0/1210003190971094

### DIFF
--- a/crates/hotshot/types/src/event.rs
+++ b/crates/hotshot/types/src/event.rs
@@ -8,12 +8,13 @@
 
 use std::sync::Arc;
 
+use hotshot_utils::anytrace::*;
 use serde::{Deserialize, Serialize};
 
 use crate::{
     data::{
-        DaProposal, DaProposal2, Leaf, Leaf2, QuorumProposal, QuorumProposalWrapper,
-        UpgradeProposal, VidDisperseShare,
+        vid_disperse::ADVZDisperseShare, DaProposal, DaProposal2, Leaf, Leaf2, QuorumProposal,
+        QuorumProposalWrapper, UpgradeProposal, VidDisperseShare,
     },
     error::HotShotError,
     message::{convert_proposal, Proposal},
@@ -37,11 +38,11 @@ pub struct Event<TYPES: NodeType> {
 }
 
 impl<TYPES: NodeType> Event<TYPES> {
-    pub fn to_legacy(self) -> LegacyEvent<TYPES> {
-        LegacyEvent {
+    pub fn to_legacy(self) -> anyhow::Result<LegacyEvent<TYPES>> {
+        Ok(LegacyEvent {
             view_number: self.view_number,
-            event: self.event.to_legacy(),
-        }
+            event: self.event.to_legacy()?,
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


builder/src/non_permissioned.rs
```diff
@@ -310,7 +310,7 @@ mod test {
         events_service_client.connect(None).await;
 
         let subscribed_events = events_service_client
-            .socket("hotshot-events/events")
+            .socket("hotshot-events/v1/events")
```
imabdulbasit, line 313: v1 is the version which you get when the api is definied. If you don't hardcode this version, then the client would automatically redirect to the latest version. In this case, I think it would redirect to `v1/hotshot-events/events`

imabdulbasit (changes requested):

crates/hotshot/types/src/event.rs
```diff
@@ -64,11 +88,59 @@ impl<TYPES: NodeType> LeafInfo<TYPES> {
...
+    /// Decided leaf.
+    pub leaf: Leaf<TYPES>,
+    /// Validated state.
+    pub state: Arc<<TYPES as NodeType>::ValidatedState>,
+    /// Optional application-specific state delta.
+    pub delta: Option<Arc<<<TYPES as NodeType>::ValidatedState as ValidatedState<TYPES>>::Delta>>,
+    /// Optional VID share data.
+    pub vid_share: Option<VidDisperseShare<TYPES>>,
+    /// Optional light client state update certificate.
+    pub state_cert: Option<LightClientStateUpdateCertificate<TYPES>>,
```
imabdulbasit, line 116: can you double check this? AFAIR state_cert wasn't there in legacy type

crates/hotshot/types/src/event.rs
```diff
@@ -64,11 +88,59 @@ impl<TYPES: NodeType> LeafInfo<TYPES> {
...
+#[serde(bound(deserialize = "TYPES: NodeType"))]
+pub struct LegacyLeafInfo<TYPES: NodeType> {
+    /// Decided leaf.
+    pub leaf: Leaf<TYPES>,
+    /// Validated state.
+    pub state: Arc<<TYPES as NodeType>::ValidatedState>,
+    /// Optional application-specific state delta.
+    pub delta: Option<Arc<<<TYPES as NodeType>::ValidatedState as ValidatedState<TYPES>>::Delta>>,
+    /// Optional VID share data.
+    pub vid_share: Option<VidDisperseShare<TYPES>>,
```
imabdulbasit, line 114: VidDisperseShare is now an enum.  This should be `ADVZDisperseShare`

marketplace-builder/src/builder.rs
```diff
@@ -0,0 +1,799 @@
+use std::{arch::global_asm, collections::HashSet, num::NonZeroUsize, sync::Arc, time::Duration};
```
imabdulbasit, line 1: this crate has also been removed

imabdulbasit (approved):

PR #3106, opened 2025-04-23
test for duplicate events from contract

imabdulbasit:
adds a test to check that contract does not return any duplicate events

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -3264,6 +3264,9 @@ mod test {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_stake_table_duplicate_events_from_contract() -> anyhow::Result<()> {
+        // TODO(abdul): This test currently uses TestNetwork only for contract deployment and for L1 block number.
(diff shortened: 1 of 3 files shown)


sequencer/src/api.rs
```diff
@@ -3259,4 +3261,122 @@ mod test {
...
+            if header.height() == 0 {
+                continue;
+            }
+            let l1_block = header.l1_finalized().expect("l1 block not found");
+
+            let events = StakeTableFetcher::fetch_events_from_contract(
+                l1_client.clone(),
+                stake_table,
+                None,
+                l1_block.number(),
```
sveitser, line 3349: Does this test only use the l1 block number from the the test network? Is it worth spinning up a test network then?

imabdulbasit: I can copy some code from `pos_deploy_routine()` to remove this test network

sveitser: Let's avoid copying that code again.

sveitser (approved):

PR #3109, opened 2025-04-23
skip reward header dependencies for first two epochs in state reconstruction

imabdulbasit:
- Only get reward header dependencies for third epoch onwards
- wait for catchup request to finish


jbearer (approved):

sequencer/src/api/sql.rs
```diff
@@ -648,15 +667,31 @@ async fn reward_header_dependencies(
             continue;
         }
 
-        let epoch = epoch_from_block_number(height, epoch_height);
+        let first_epoch = first_epoch.context("first epoch not found")?;
+
+        let epoch = EpochNumber::new(epoch_from_block_number(height, epoch_height));
```
sveitser, line 672: nit: `proposal_epoch`

Just because we have two `epoch` values we are comparing, both should have different names than just "epoch".

sveitser (approved):

PR #3121, opened 2025-04-24
add reference tests for NsProof

sveitser:
Draft: data doesn't look quite right to me yet.


types/src/reference_tests.rs
```diff
@@ -77,6 +84,68 @@ async fn reference_payload() -> Payload {
...
+    let payload = reference_payload().await;
+    let ns_index = payload
+        .ns_table()
+        .find_ns_id(&(REFERENCE_NAMESPACE_ID.into()))
+        .unwrap();
+    let enc = payload.encode();
+    let mut scheme = advz_scheme(10);
+    let disperse = VidScheme::disperse(&mut scheme, &enc).unwrap();
+
+    let ns_proof = NsProof::new(&payload, &ns_index, &VidCommon::V0(disperse.common)).unwrap();
```
mrain, line 97: we may directly use `ADVZNsProof::new()` here, so you don't have to unwrap down below.

mrain (approved):
LGTM with one minor comment

types/src/reference_tests.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/3121/commits/748073b1e1bf8bb31d1103be21621086319d77f7

PR #3122, opened 2025-04-24
Pin ubuntu to 24.04 for docker builds

rob-maron:


sveitser:
I thought `latest` is always the LTS version, maybe not?

rob-maron:
> I thought `latest` is always the LTS version, maybe not?

It should be, good point

The reason for this was that we have an operator building their own images and the build process between our PoS patches (patch3 -> patch5) was much different. Maybe we should just change versions on `main` whenever there is a new LTS?

fkrell (approved):

sveitser:
I guess if someone doesn't pull since April 24 then latest might be a 22.04? So yes, let's hardcode 24.04. I think I changed it to latest because we were on 22.04 and CI is using on ubuntu-latest and that broke at some point.

PR #3128, opened 2025-04-24
fix: namespaced vid range check and doc

mrain:
Closes #<ISSUE_NUMBER>


vid/src/avid_m/namespaced.rs
```diff
@@ -169,6 +171,12 @@ impl NsAvidMScheme {
         if shares.is_empty() {
             return Err(VidError::InsufficientShares);
         }
+        if shares
+            .iter()
+            .any(|share| ns_id >= share.ns_lens.len() || ns_id >= share.content.len())
```
alxiong, line 176: `ns_id >= share.content.len()`, why this?

mrain: avoid panic in later `share.inner_ns_share(ns_id)` and `shares[0].ns_commits[ns_id]`

alxiong: got it, can you add a comment about how the content is expected to be as long as the number of namespace, even if the namespace has empty payload, they should have a non-empty `RawAvidShare` with some metadata?

alxiong (approved):

PR #3129, opened 2025-04-24
Return error instead of empty vec for stake table endpoints

imabdulbasit:
Currently, if the stake table is not present for an epoch, it returns an empty vec. This PR changes it to return the error


sequencer/src/api/endpoints.rs
```diff
@@ -324,15 +324,23 @@ where
...
+                }))
         }
         .boxed()
     })?
     .at("stake_table_current", |_, state| {
         async move {
             Ok(state
                 .read(|state| state.get_stake_table_current().boxed())
-                .await)
+                .await
```
tbro, line 339: I think it is fine, but I'm wondering is there really a case when we don't have a current stake table?

tbro (approved):

sequencer/src/api/endpoints.rs
imabdulbasit: Yeah, I think we can hit this case. The current epoch comes from the hotshot and we basically pass this to the stake-table/epoch endpoint to get the stake table. This can fail if membership does not have it

PR #3130, opened 2025-04-24
Remove failed stake table catchup request to allow new catchup request

imabdulbasit:
This PR:
Removes failed catchup request from the map, to allow new catchup request to be spawned

### DIFF
--- a/crates/hotshot/types/src/epoch_membership.rs
+++ b/crates/hotshot/types/src/epoch_membership.rs
@@ -247,8 +247,14 @@ fn spawn_catchup<T: NodeType>(coordinator: EpochMembershipCoordinator<T>, epoch:
             tx
         };
         // do catchup
-        let ret = coordinator.catchup(epoch).await;
-        let _ = tx.broadcast_direct(ret).await;
+
+        let result = coordinator.clone().catchup(epoch).await;
+        let _ = tx.broadcast_direct(result.clone()).await;
+
+        if let Err(err) = result {
+            tracing::warn!("failed to catchup for epoch={epoch:?}. err={err:#}");
+            coordinator.catchup_map.lock().await.remove(&epoch);
+        }
     });
 }
 /// Wrapper around a membership that guarantees that the epoch


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -228,7 +228,7 @@ where
         else {
             return self.clone().catchup(epoch).await;
         };
-        let Ok(Ok(mem)) = rx.recv_direct().await else {
+        let Ok(mem) = rx.recv_direct().await else {
```
bfish713, line 231: this looks wrong, we get mem is a Result here, the outer ok is just if the recv suceeds

imabdulbasit: so I think it should return the result received so that wait_for_catchup() returns the error if catchup failed

imabdulbasit: or I guess we do want to block. I'll revert this change

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -247,8 +247,14 @@ fn spawn_catchup<T: NodeType>(coordinator: EpochMembershipCoordinator<T>, epoch:
             tx
         };
         // do catchup
-        let ret = coordinator.catchup(epoch).await;
-        let _ = tx.broadcast_direct(ret).await;
+
+        let result = coordinator.clone().catchup(epoch).await;
+        let _ = tx.broadcast_direct(result.clone()).await;
```
bfish713, line 252: Is the error handled properly on the receiving end?

imabdulbasit: No it spawns catchup again
`
 let Ok(Ok(mem)) = rx.recv_direct().await else {
            return self.clone().catchup(epoch).await;
        };
        
        `
        
        I thought this was intentional, but maybe not?

imabdulbasit: I thought this was intentional, but maybe not?

bfish713: Yeah I think it's fine, I added detail to the other comment

bfish713 (approved):

bfish713:
can you backport this to release as well

crates/hotshot/types/src/epoch_membership.rs
bfish713: I in the case of `Ok(Err(_))` we want to just return the `Err` this way the error will bubble up.

Maybe it's ok as is, we'll just loop trying to catchup to the new

imabdulbasit:
> can you backport this to release as well

which release branch?

bfish713:
> > can you backport this to release as well
> 
> which release branch?

I just added the label, forgot anyone could

PR #3134, opened 2025-04-24
Add epoch stake table retrieval support to node-metrics

Ayiga:
### This PR:

Adds support for updating the local stake table when the epoch changes.

### DIFF
--- a/node-metrics/src/service/data_state/mod.rs
+++ b/node-metrics/src/service/data_state/mod.rs
@@ -22,13 +22,16 @@ use hotshot_types::{
         stake_table::{SnapshotVersion, StakeTableScheme},
         BlockPayload, EncodeBytes,
     },
+    utils::epoch_from_block_number,
 };
 pub use location_details::LocationDetails;
 pub use node_identity::NodeIdentity;
 use time::OffsetDateTime;
 use tokio::{spawn, task::JoinHandle};
 
-use crate::api::node_validator::v0::LeafAndBlock;
+use crate::api::node_validator::v0::{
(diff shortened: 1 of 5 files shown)


node-metrics/src/service/data_state/mod.rs
```diff
@@ -210,6 +286,81 @@ pub fn create_block_detail_from_block(block: &BlockQueryData<SeqTypes>) -> Block
...
+        let previous_epoch = epoch_number_helper(
+            block_height.saturating_sub(1),
+            epoch_starting_block,
+            num_blocks_per_epoch,
+        );
+        let upcoming_epoch =
+            epoch_number_helper(block_height, epoch_starting_block, num_blocks_per_epoch);
+
+        if upcoming_epoch != previous_epoch {
+            tracing::debug!(
```
jbearer, line 330: If we straight up replace the stake table as soon as we enter a new epoch, there is going to be a period where we have blocks in our "last 50" window from both the old and new epoch, but we are interpreting the QC bitmaps using only the new stake table. So the participation numbers might go a little wonky the first 50 blocks of each epoch.

This is honestly not that big of a deal I think especially because we don't expect the stake table to change that much from one epoch to the next usually. But we should fix it eventually

jbearer (approved):
This makes sense to me, just one comment about potential future work

node-metrics/src/service/data_state/mod.rs
Ayiga: For clarity, the votes are transferred to users independently of the Block data transferred.  The Votes are transferred in a `BitVec` whose ordering is the same as the collection of our stored Node Identities.

PR #3136, opened 2025-04-25
CI: dependabot: we don't allow merge, use squash

sveitser:


alxiong (approved):

sveitser:
I'll bypass since this only affects a CI job that is currently not working.

PR #3148, opened 2025-04-25
CI: fix cargo audit fix

sveitser:
The idea is to create a PR to fix cargo audit that targets the base branch when that is feasible (so on PRs, and branch builds, but no for tag builds).

~~Target branch is `[ma/testing-audit-fail](https://github.com/EspressoSystems/espresso-network/tree/ma/testing-audit-fail)` to simulate cargo audit failure.~~

- [x] change target to `main`.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -28,13 +28,26 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
   audit_fix:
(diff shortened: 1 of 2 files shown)


sveitser:
So this sort of works now: see https://github.com/peter-evans/create-pull-request/issues/3742 for a PR

imabdulbasit (approved):

PR #3155, opened 2025-04-25
added code to create an upgrade proposal via rust

### LINKED ISSUE #3077: add upgrading to LCV2 and LCAV2 to rust deployer scripts
currently there are foundry scripts but we want it included in the rust deployer

alysiahuggins:
Closes #3077

### This PR:
- Updates the `upgradeProxy.ts` script to handle command-line arguments more for when it's called using rust whilst ensuring it's backward compatible until we phase out deployments and upgrades via foundry scripts
- In `deployer.rs` added
1. `upgrade_light_client_v2_multisig_owner`
- Creates the upgrade proposal to the LightClientProxy using a multisig wallet.
- Deploys necessary contracts (PlonkVerifierV2, LightClientV2).
- Prepares initialization data and calls a TypeScript script to propose the upgrade.
2. `call_upgrade_proxy_script`
- Executes a TypeScript script to handle the upgrade proposal.

### This PR does not:
- Implement the execution of the upgrade proposal on the blockchain; it only sends the proposal.
- Address any issues related to the execution of transactions on networks not supported by the Safe SDK.

### Key places to review:
- `upgrade_light_client_to_v2_multisig_owner ` in `deployer.rs`

### How to test this PR: 
- `yarn jest` (for typescript tests)
- `cargo test -p sequencer-utils test_upgrade_light_client_to_v2_multisig_owner -- --nocapture`

### Things tested
- Adds tests for the processRustCommandLineArguments function to ensure correct parsing of command-line arguments.
- Verifies that the upgrade proposal is sent successfully, but not its execution.


contracts/rust/deployer/src/lib.rs
```diff
@@ -704,6 +826,44 @@ pub async fn is_contract(provider: impl Provider, address: Address) -> Result<bo
...
 
+/// this depends on upgradeProxy.ts which has to be ran on a real network supported by the safeSDK
+pub async fn call_upgrade_proxy_script(
+    proxy_addr: String,
+    new_impl_addr: String,
+    init_data: String,
+    rpc_url: String,
+    safe_address: String,
+) -> Result<(String, bool), anyhow::Error> {
+    let script_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
sveitser, line 837: The `env!` directive is evaluated at compile time so this is likely not what we want here.

contracts/rust/deployer/src/lib.rs
```diff
@@ -1046,6 +1206,86 @@ mod tests {
...
+    // This test is used to test the upgrade of the LightClientProxy via the multisig wallet
+    // It only tests the upgrade proposal via the typescript script and thus requires the upgrade proposal to be sent to a real network
+    // However, the contracts are deployed on anvil, so the test will pass even if the upgrade proposal is not executed
+    // The test assumes that there is a file .env.deployer.rs.test in the root directory with the following variables:
+    // RPC_URL=
+    // SAFE_MULTISIG_ADDRESS=0x0000000000000000000000000000000000000000
+    // SAFE_ORCHESTRATOR_PRIVATE_KEY=0x0000000000000000000000000000000000000000000000000000000000000000
+    // Ensure that the private key has proposal rights on the Safe Multisig Wallet and the SDK supports the network
+    async fn test_upgrade_light_client_to_v2_multisig_owner_helper(is_mock: bool) -> Result<()> {
+        dotenvy::from_filename_override(".env.deployer.rs.test").ok();
```
sveitser, line 1218: We shouldn't manipulate the environment from tests. `cargo test` doesn't use a single process per test (nextest does though) so it can lead to weird things happening.

I think we can use https://docs.rs/dotenvy/latest/dotenvy/fn.from_filename_iter.html convert that into a HashMap and read the values from there instead of loading it into the environment and reading them back.

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
```diff
@@ -32,60 +41,120 @@ async function main() {
...
     console.log(
-      `The other owners of the Safe Multisig wallet need to sign the transaction via the Safe UI https://app.safe.global/transactions/queue?safe=sep:${safeAddress}`,
+      `The other owners of the Safe Multisig wallet need to sign the transaction via the Safe UI https://app.safe.global/transactions/queue?safe=sep:${upgradeData.safeAddress}`,
     );
   } catch (error) {
     throw new Error("An error occurred: " + error);
   }
 }
 
+export function processRustCommandLineArguments(args: string[]): UpgradeData {
```
sveitser, line 58: Can we use a command line parser lib so we can outsource parsing of the command  line arguments. I don't know the node ecosystem well. I know about https://github.com/yargs/yargs, maybe @Ayiga has a preferred one?

docker/upgrade.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ubuntu:24.04
+
+ARG TARGETARCH
+
+RUN apt-get update \
+    &&  apt-get install -y curl libcurl4 wait-for-it tini \
+    nodejs  \
+    npm \
+    && npm install -g ts-node typescript \
+    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y \
```
sveitser, line 10: I think we don't need a rust toolchain in this container, but maybe I'm missing something.

alysiahuggins: that's right, i was trying some rust tests and forgot to remove it https://github.com/EspressoSystems/espresso-network/commit/f0a148fe64314e44205e5a21be00aa9fbbcfa06a

contracts/rust/deployer/src/lib.rs
alysiahuggins: thank you for that https://github.com/EspressoSystems/espresso-network/commit/f0a148fe64314e44205e5a21be00aa9fbbcfa06a

contracts/rust/deployer/src/lib.rs
```diff
@@ -491,6 +500,119 @@ pub async fn upgrade_light_client_v2(
...
+            };
+
+            // prepare init calldata
+            let lcv2 = LightClientV2::new(lcv2_addr, &provider);
+            let init_data = lcv2
+                .initializeV2(blocks_per_epoch, epoch_start_block)
+                .calldata()
+                .to_owned();
+
+            println!("Init Data to be signed.\n Function: initializeV2\n Arguments:\n blocks_per_epoch: {:?}\n epoch_start_block: {:?}", blocks_per_epoch, epoch_start_block);
```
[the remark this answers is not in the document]
alysiahuggins: ah, updated here https://github.com/EspressoSystems/espresso-network/commit/f0a148fe64314e44205e5a21be00aa9fbbcfa06a

contracts/rust/deployer/src/lib.rs
```diff
@@ -704,6 +826,46 @@ pub async fn is_contract(provider: impl Provider, address: Address) -> Result<bo
     Ok(true)
 }
 
+/// this depends on upgradeProxy.ts which has to be ran on a real network supported by the safeSDK
+pub async fn call_upgrade_proxy_script(
+    proxy_addr: String,
```
sveitser, line 831: Why not pass `Address` types? Can call `to_string()` on them to turn them into a string.

sequencer/src/bin/deploy.rs
```diff
@@ -147,17 +150,34 @@ struct Options {
...
+    /// The epoch start block
+    #[clap(long, env = "ESPRESSO_SEQUENCER_EPOCH_START_BLOCK")]
+    epoch_start_block: Option<u64>,
+
     #[clap(flatten)]
     logging: logging::Config,
 }
 
+impl Options {
+    fn validate_upgrade_choice(&self) -> anyhow::Result<()> {
```
sveitser, line 166: This can be done with clap.

https://github.com/EspressoSystems/espresso-network/blob/3426a6b0c2a7fe4a3bb8775e06e893693f0d5291/sequencer/src/persistence/sql.rs#L188

alysiahuggins: ooop thank you!

https://github.com/EspressoSystems/espresso-network/commit/050be5fbc9a578ff375e8b7e65593dd4a9d546ea

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
alysiahuggins: no longer required as per our convo

.github/workflows/test.yml
```diff
@@ -218,7 +219,7 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
 
       - uses: actions/checkout@v4
-
+
```
sveitser: unwanted change?

README.md
```diff
@@ -150,12 +150,68 @@ To additionally verify the contract on etherscan set the `ETHERSCAN_API_KEY` env
...
+**Configuration**
+
+You can configure the deployer using CLI flags or environment variables. Most options can be set via environment
+variables (see the code for the full list `sequencer/src/bin/deploy.rs`). Common environment variables:
+
+- `ESPRESSO_SEQUENCER_L1_PROVIDER` — L1 JSON-RPC endpoint
+- `ESPRESSO_SEQUENCER_ETH_MNEMONIC` — Mnemonic for the deployer wallet
+- `ESPRESSO_SEQUENCER_ETH_MULTISIG_ADDRESS` — Multisig admin address
+- `ESPRESSO_DEPLOYER_ACCOUNT_INDEX` — Account index in the wallet
+- `ESPRESSO_SEQUENCER_URL` — Sequencer node URL for HotShot config You can use a `.env` file and load it with:
```
sveitser:
```suggestion
- `ESPRESSO_SEQUENCER_URL` — Sequencer node URL for HotShot config 

You can use a `.env` file and load it with:
```

contracts/rust/deployer/Cargo.toml
```diff
@@ -20,5 +21,8 @@ tokio = { workspace = true }
 tracing = "0.1.37"
 vbs = { workspace = true }
 
+[features]
+safesdk = []
```
sveitser: What is the rationale behind adding this feature? Is it just to avoid running a test by default?

scripts/build-docker-images-native
```diff
@@ -130,6 +136,7 @@ images=(
  state-relay-server
  staking-cli
  submit-transactions
+ upgrade
```
sveitser: I think we don't have this image anymore?

contracts/rust/deployer/src/lib.rs
```diff
@@ -762,6 +939,71 @@ pub async fn is_contract(provider: impl Provider, address: Address) -> Result<bo
...
+        .arg(dry_run.to_string())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .output();
+
+    let output = output.unwrap();
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    // if stderr is not empty, return the stderr
+    if !stderr.is_empty() {
```
sveitser: It's more customary to check for a zero exit code: https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.success

It's not normally a requirement for processes to only print to stderr if they fail.

docker-compose.yaml
```diff
@@ -55,6 +55,34 @@ services:
       deploy-sequencer-contracts:
         condition: service_completed_successfully
 
+  upgrade-prover-contracts:
+    image: ghcr.io/espressosystems/espresso-sequencer/deploy:${DOCKER_TAG:-main}
+    command: deploy --use-mock --upgrade-light-client-v2 --use-multisig --dry-run
```
sveitser, line 60: Is this service added to see if we can run `deploy` with those argument in the docker container successfully?

contracts/rust/deployer/Cargo.toml
alysiahuggins: ah to run just tests related to the safe sdk

docker-compose.yaml
alysiahuggins: yes

contracts/rust/deployer/Cargo.toml
alysiahuggins: well more so to run specifically the safesdk related tests

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
```diff
@@ -10,82 +10,164 @@ export interface UpgradeData {
...
+    const [upgradeData, dryRun] = processCommandLineArguments();
+    console.log("Upgrade Data : ", JSON.stringify(upgradeData));
+    if (!upgradeData.rpcUrl) {
+      upgradeData.rpcUrl = getEnvVar("RPC_URL");
+    }
+    if (!upgradeData.safeAddress) {
+      upgradeData.safeAddress = getEnvVar("SAFE_MULTISIG_ADDRESS");
+    }
+
+    if (!dryRun) {
```
sveitser, line 31: It would be preferable to have dry-run print out the payload to stdout so we can assert it's a certain value for some fixed inputs in a test. Currently I think we only test that argument parsing doesn't fail for some inputs. I would expect dry-run to do exactly the same thing than a normal run except for not applying a change but showing what what change it would apply instead. Currently we actually change the arguments we pass to the TS script if `--dry-run` is passed to the rust deploy binary

https://github.com/EspressoSystems/espresso-network/blob/a74bd45e44614785666536906a0ec4638d3cd814/contracts/rust/deployer/src/lib.rs#L541-L549

contracts/rust/deployer/Cargo.toml
sveitser: Then I think we can put `safesdk` in the name of the test or put those into a separate file or test module. Adding a feature seems overkill for this.

.github/workflows/test.yml
alysiahuggins: yes, https://github.com/EspressoSystems/espresso-network/commit/e7ec2b6fd5cc7f58068ab867633dbcf9b5e504a6

scripts/build-docker-images-native
alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/e7ec2b6fd5cc7f58068ab867633dbcf9b5e504a6

contracts/rust/deployer/src/lib.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/e7ec2b6fd5cc7f58068ab867633dbcf9b5e504a6

docker-compose.yaml
sveitser: I'm thinking maybe we can just put an assert into the docker build? I think this currently mainly tests that we can run the typescript code and the argument parsing is correct. It's not  exactly what I had in mind with the dry-run flag but yeah it's not a bad idea to have this either.

If we're not checking the output we get from the typescript script I would probably add a subcommand to the deploy binary like `verify-installation` and call it at the end of the docker build (inside the dockerfile) to ensure that the rust script in the container can execute the typescript script.

docker/deploy.Dockerfile
```diff
@@ -2,7 +2,39 @@ FROM ghcr.io/espressosystems/ubuntu-base:main
...
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# Copy project metadata and install JS deps
+COPY package.json yarn.lock ./
+
+RUN yarn global add typescript ts-node && \
+    yarn && \
+    rm -rf /usr/local/share/.cache ~/.cache node_modules/.cache
```
sveitser: Unfortunately this sort of 10x-es the build time for the docker image that was recently optimized https://github.com/EspressoSystems/espresso-network/actions/runs/15393016536

The problem is that the emulated build is rather slow. Moving to seperate native x86 and ARM docker builds and digest merge is some work though. Hmm.

sveitser: I'll let it run to completion and see what takes how long. If it's mainly apt then it's reasonably easy to make another base image where all the system dependencies are installed. If the culprit is yarn itself things are a bit more tricky.

sveitser: It's even worse it actually fails.

contracts/rust/deployer/Cargo.toml
alysiahuggins: ok I've removed it

contracts/script/multisigTransactionProposals/safeSDK/upgradeProxy.ts
alysiahuggins: ok i've changed it so that it uses the vars passed in

alysiahuggins: added asserts for the results of a dry run https://github.com/EspressoSystems/espresso-network/commit/d7c6e72c667a50d9618eb01e2300b8edff1c5367

utils/Cargo.toml
```diff
@@ -15,6 +15,8 @@ ark-serialize = { workspace = true, features = ["derive"] }
 async-trait = { workspace = true }
 clap = { workspace = true }
 committable = "0.2"
+derive_more = { workspace = true }
+dotenvy = "0.15.7"
```
sveitser: Is this needed? It seems there are no other changes to this crate.

sequencer/src/bin/deploy.rs
```diff
@@ -122,6 +126,18 @@ struct Options {
...
 
+    /// Option to deploy contracts owned by multisig
+    #[clap(long, default_value = "false")]
+    pub use_multisig: bool,
+
+    /// Option to test upgrade stake table v2 multisig owner dry run
+    #[clap(long, default_value = "false")]
+    pub dry_run: bool,
+
+    /// Option to test locally but with a real eth network
```
sveitser, line 137: What does "real" mean here?

README.md
```diff
@@ -150,12 +150,70 @@ To additionally verify the contract on etherscan set the `ETHERSCAN_API_KEY` env
...
+source .env
+set +a
+```
+
+#### Deployment via Docker
+
+You can run the deployer in a container:
+
+```bash
+docker run --env-file .env --rm ghcr.io/espressosystems/espresso-sequencer/deploy:main /bin/deploy --deploy-light-client-v1 ...
```
sveitser: Does  this actually work?

```
docker run --env-file .env --rm ghcr.io/espressosystems/espresso-sequencer/deploy:main /bin/deploy --deploy-light-client-v1
error: invalid value 'http://demo-l1-network:${ESPRESSO_SEQUENCER_L1_PORT}' for '--rpc-url <RPC_URL>': invalid port number

For more information, try '--help'.
```

README.md
```diff
@@ -150,12 +150,70 @@ To additionally verify the contract on etherscan set the `ETHERSCAN_API_KEY` env
...
+
+```bash
+RUST_LOG=info cargo run --bin deploy -- [FLAGS]
+RUST_LOG=debug cargo run --bin deploy -- [FLAGS]
+```
+
+For Docker:
+
+```bash
+docker run --env-file .env.docker -e RUST_LOG=debug ...
```
sveitser, line 224: The `.env.docker` file doesn't seem to be checked in.

sequencer/src/bin/deploy.rs
alysiahuggins: "real" means a live ethereum network, as opposed to a local development chain or a fork/simulated env like anvil/hardhat

contracts/rust/deployer/src/lib.rs
```diff
@@ -762,6 +922,71 @@ pub async fn is_contract(provider: impl Provider, address: Address) -> Result<bo
...
+    init_data: String,
+    rpc_url: String,
+    safe_addr: Address,
+    dry_run: Option<bool>,
+) -> Result<(String, bool), anyhow::Error> {
+    let dry_run = dry_run.unwrap_or(false);
+    tracing::info!("Dry run: {}", dry_run);
+    tracing::info!("Sending the upgrade proposal to multisig: {}", safe_addr);
+
+    let script_path = if let Ok(env_path) = std::env::var("MULTISIG_UPGRADE_ENTRYPOINT_PATH") {
```
sveitser: I'm not sure we need to this env var here. I thought we assured that the entrypoint script is in the path.

contracts/rust/deployer/src/lib.rs
```diff
@@ -1114,6 +1342,170 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_upgrade_light_client_to_v2_multisig_owner_dry_run() -> Result<()> {
+        test_upgrade_light_client_to_v2_multisig_owner_helper(false, true, false).await
+    }
+
+    // We expect no init data for the second upgrade because the proxy was already initialized
+    #[tokio::test]
+    async fn test_upgrade_light_client_to_v2_twice_multisig_owner_dry_run() -> Result<()> {
+        test_upgrade_light_client_to_v2_multisig_owner_helper(false, true, true).await
```
sveitser: The downside of using booleans as arguments is that when reading this code here it's hard to tell what the different values mean. This can be avoided by defining enums instead. Then we also get compile errors if the order is messed up.

contracts/rust/deployer/src/lib.rs
```diff
@@ -500,6 +509,157 @@ pub async fn upgrade_light_client_v2(
...
+                lcv2_addr,
+                init_data.to_string(),
+                params.rpc_url,
+                owner_addr,
+                Some(dry_run),
+            )
+            .await;
+
+            if let Err(ref err) = result {
+                tracing::error!("LightClientProxy upgrade failed: {:?}", err);
```
sveitser: You can usually do `anyhow::bail!(...)` if you want to return with an error. And if you want to propagate the error to the caller you can use `?` if the error you have can be converted to the error the (outer) function returns. This reduces the amount of error handling code we need to write.

contracts/rust/deployer/src/lib.rs
sveitser: @alysiahuggins so I don't quite understand why we inject the path via an env var here we should have the entrypoint script in the PATH both inside the dev env and inside docker so I think the env var isn't necessary.

sequencer/src/bin/deploy.rs
sveitser: I think what this flag does is mainly mocking out calls to the espresso network. So I think it would be better to call it something like `--mock-espresso`.

sveitser:
The ARM build failed with `make not found`. This is rather confusing. Another run: https://github.com/EspressoSystems/espresso-network/actions/runs/15437539302

contracts/rust/deployer/src/lib.rs
alysiahuggins: thanks for this, it's not
https://github.com/EspressoSystems/espresso-network/commit/a5083dc31cde57a300c3c2d2296404a8953c75b5

contracts/rust/deployer/src/lib.rs
alysiahuggins: thank you! https://github.com/EspressoSystems/espresso-network/commit/a5083dc31cde57a300c3c2d2296404a8953c75b5

contracts/rust/deployer/src/lib.rs
alysiahuggins: thank you 
https://github.com/EspressoSystems/espresso-network/commit/e34ebf240cb255e91e01194a1b87ab1c81e25637 & https://github.com/EspressoSystems/espresso-network/commit/95c4aea35c00a49fcee0ca6d402e099f503a8284

sequencer/src/bin/deploy.rs
alysiahuggins: called it mock-espresso-live-network as it does use a live eth network so didn't want it to mask that fact

https://github.com/EspressoSystems/espresso-network/pull/3155/commits/335847c2edc0fb71476877ba0986ed543d477e3f

README.md
alysiahuggins: thanks updated it to the docker compose command
https://github.com/EspressoSystems/espresso-network/commit/c23f8ff53c7f715402dcb62eb31868680c009bd0

README.md
alysiahuggins: added `.env.docker.example` https://github.com/EspressoSystems/espresso-network/commit/c23f8ff53c7f715402dcb62eb31868680c009bd0

utils/Cargo.toml
alysiahuggins: you're right, it's unused
https://github.com/EspressoSystems/espresso-network/commit/ec2cfaf48d106a8a1f25fa6d18e38c60f9cde5aa

sveitser (approved):

PR #3167, opened 2025-04-28
skip sepolia fork test unless RPC url provided

sveitser:
The tests needs an archive node and I couldn't find a free RPC url. We should set it via a github secret. Until we get around to to that the test will be skipped if the URL isn't provided via env var.


sveitser:
The fork test is skipped on CI. so I will bypass.

PR #3168, opened 2025-04-28
Proposal pipeline cleanup

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* renames `is_high_qc_for_last_block` to `is_high_qc_for_epoch_transition` to better reflect what it actually does
* removes the unnecessary call to `wait_for_next_epoch_qc` in `send_high_qc`
* removes the unnecessary calls to `validate_qc_and_next_epoch_qc` in `send_high_qc` and substitutes them with simple leaf commit comparison
* removes the receiver parameter from `send_high_qc` because it is no longer needed; also propagates that change up the call tree.
* substitutes calls to `membership_for_epoch` with calls to `stake_table_for_epoch` where knowledge about the random leader is not required
* `wait_for_highest_qc` and `wait_for_qc_event` now return an optional next epoch QC; previously we obtained it but never returned.
* removes the unnecessary call to `wait_for_next_epoch_qc` in `publish_proposal`; it's not required because we either got the next epoch QC as part of the proposal dependency or we received it as part of a transition QC or a highest QC; or we simply had it in the consensus state
* removes the definition of `wait_for_next_epoch_qc` as it is no longer used
* correctly sets `epoch_transition_indicator` in the quorum proposal task


crates/hotshot/types/src/consensus.rs
```diff
@@ -1209,8 +1209,8 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         is_epoch_transition(block_height, self.epoch_height)
     }
 
-    /// Returns true if our high QC is for the last block in the epoch
-    pub fn is_high_qc_for_last_block(&self) -> bool {
+    /// Returns true if our high QC is for one of the epoch transition blocks
+    pub fn is_high_qc_for_epoch_transition(&self) -> bool {
```
bfish713, line 1213: Lets get rid of the leaf part, we can just get the height from the high_qc directly

crates/hotshot/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -446,7 +446,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
         event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
     ) -> Result<()> {
         let epoch_number = self.cur_epoch;
-        let epoch_transition_indicator = if self.consensus.read().await.is_high_qc_for_last_block()
+        let epoch_transition_indicator = if self
+            .consensus
+            .read()
+            .await
+            .is_high_qc_for_epoch_transition()
```
bfish713, line 453: This returns true if the QC is for the last block.  In this case we are not in transition when it comes to proposing.  For example with epoch_height == 300 if our high QC is for block 600 then we are proposing block 601 and we are definitively must be in epoch 3 (High QC is the EQC).  

I see the Restart tests are failing/flaky and I suspect it related to this fact + the changess below to not pass in the hard coded `NotInTransition`

bfish713 (changes requested):
Requesting changes to investigate why the restart tests seem to regress with this change

PR #3175, opened 2025-04-28
Add fallback deserialization to avoid duplicate http requests for fetching legacy data

imabdulbasit:
Closes https://app.asana.com/1/1208976916964769/project/1209413670216270/task/1210014495940934

### This PR:

This PR removes duplicate fetch request in the query service from the same provider.

For the proof-of-stake upgrade, we have new types. The new binary migrates all the query service data to the new types. When the new binary is deployed on the testnet, it may take some time to migrate all the existing data, and so the node has to catch up on all the missing data, which means it needs to fetch from the peers running old binary. The new binary, therefore, registers two availability module versions: `v0` and `v1`. Version v1 returns the new types, while v0 returns the downgraded (legacy) types. The old peers only have `v0` availability module

A quick solution was implemented, which makes a request to the provider expecting the new type. If that fails, another request is made for the legacy type from the same provider. This legacy type is then converted to the new type and returned as the result.

This means two http requests to fulfil one missing pre-epoch req.

This PR adds a fallback deserialization to solve this. It returns the bytes from the provider and first tries to deserialize them into the new type. If that fails, another attempt is made to parse the bytes as the old legacy type. This eliminates the need for two catch-up requests.

https://github.com/EspressoSystems/surf-disco/pull/68 added the bytes() to return response bytes.

### Key places to review:
The fetch() for all the requests.

### DIFF
--- a/hotshot-query-service/src/fetching/provider/query_service.rs
+++ b/hotshot-query-service/src/fetching/provider/query_service.rs
@@ -2454,17 +2454,19 @@ mod test {
             }
 
             // Append the latest known leaf to the local store
-
+            // This would trigger fetches for the corresponding missing data
+            // such as header, vid and payload
+            // This would also trigger fetches for the parent data
             ds.append(leaves.last().cloned().unwrap().into())
                 .await
                 .unwrap();
 
-            // This should now fetch from the provider
-            let leaf = ds.get_leaf(test_leaf.height() as usize).await;
-            let payload = ds.get_payload(test_payload.height() as usize).await;
-            let common = ds.get_vid_common(test_common.height() as usize).await;
-
+            // check that the data has been fetches and matches the network data source
             {
+                let leaf = ds.get_leaf(test_leaf.height() as usize).await;
+                let payload = ds.get_payload(test_payload.height() as usize).await;
+                let common = ds.get_vid_common(test_common.height() as usize).await;
+
                 let truth = network.data_source();
                 assert_eq!(
                     leaf.await,
--- a/hotshot-query-service/src/fetching/provider/query_service.rs
+++ b/hotshot-query-service/src/fetching/provider/query_service.rs
@@ -23,7 +23,7 @@ use hotshot_types::{
 };
 use jf_vid::VidScheme;
 use surf_disco::{Client, Url};
-use vbs::version::StaticVersionType;
+use vbs::{version::StaticVersionType, BinarySerializer};
 
 use super::Provider;
 use crate::{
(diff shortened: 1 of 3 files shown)


hotshot-query-service/src/fetching/provider/query_service.rs
```diff
@@ -56,68 +55,56 @@ impl<Ver: StaticVersionType> QueryServiceProvider<Ver> {
...
-                    Err(err) => {
-                        tracing::error!(%err, "unable to compute VID commitment");
-                        return None;
-                    },
-                };
-                if commit != advz_commit {
-                    tracing::error!(?req, ?commit, "received inconsistent payload");
-                    return None;
-                }
+        let payload =
```
sveitser, line 66: Rename this function to `decode_legacy_payload` or similar?

hotshot-query-service/src/fetching/provider/query_service.rs
```diff
@@ -56,68 +55,56 @@ impl<Ver: StaticVersionType> QueryServiceProvider<Ver> {
...
+            tracing::error!(?req, ?commit, "received inconsistent legacy payload");
+            return None;
         }
+
+        Some(payload.data)
     }
 
     async fn fetch_legacy_vid_common<Types: NodeType>(
         &self,
+        bytes: Vec<u8>,
```
sveitser, line 100: Similarly for this function

hotshot-query-service/src/fetching/provider/query_service.rs
```diff
@@ -2327,4 +2376,185 @@ mod test {
...
+            ));
+
+            let ds = data_source(&db, &provider).await;
+            network.start().await;
+
+            let leaves = network.data_source().subscribe_leaves(1).await;
+            let leaves = leaves.take(5).collect::<Vec<_>>().await;
+            let test_leaf = &leaves[0];
+            let test_payload = &leaves[2];
+            let test_common = &leaves[3];
```
sveitser, line 2435: Do you know why we use indices 0, 2 and 3 here?

sveitser: Ah I found a comment in another test

hotshot-query-service/src/fetching/provider/query_service.rs
```diff
@@ -2327,4 +2376,185 @@ mod test {
...
+                fetch.try_resolve().unwrap_err();
+            }
+
+            // Append the latest known leaf to the local store
+
+            ds.append(leaves.last().cloned().unwrap().into())
+                .await
+                .unwrap();
+
+            // This should now fetch from the provider
```
sveitser, line 2461: Why exactly does this fetch now?

imabdulbasit: we appended latest leaf before these calls, now these get calls would fetch from the provider and verify the whole leaf chain.

hotshot-query-service/src/fetching/provider/query_service.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3175/commits/cc91b39dcb0248118394e8973fb26f5fa25429d0

hotshot-query-service/src/fetching/provider/query_service.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3175/commits/cc91b39dcb0248118394e8973fb26f5fa25429d0

sveitser (approved):
LGTM but let's separate the test into 4 tests.

PR #3180, opened 2025-04-29
Refactor deploy code to reduce duplication

### LINKED ISSUE #3164: Make deploy rust code available to all workspace crates

alxiong:
Closes #3164 

### This PR:

- introduce new crate `espresso-contract-deployer` that encompass the old `utils/deployer.rs`, and further extract out the core logic from `sequencer/bin/deploy.rs` using a builder pattern, so that it can be reused at different granular level
- remove `utils/block-native.rs`
- clean up `staking-cli/demo.rs`, particularly removed the previously duplicated `fn pos_deploy_routine()`, instead use the new crate's `DeployerArgsBuilder::build().deploy_all()` instead. 

### This PR does not:

- we didn't really cut any dependency, only centralized deploy logic and duplicate those steps.


alxiong:
@sveitser @tbro ready for review, all tests passed now.

.env
```diff
@@ -62,9 +62,10 @@ ESP_TOKEN_INITIAL_GRANT_RECIPIENT_ADDRESS=0xdD2FD4581271e230360230F9337D5c0430Bf
 ESPRESSO_SEQUENCER_PERMISSIONED_PROVER=0x14dc79964da2c08b23698b3d3cc7ca32193d9955
 
 # Contracts
-ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0x82dc47734901ee7d4f4232f398752cb9dd5daccc
-ESPRESSO_SEQUENCER_STAKE_TABLE_PROXY_ADDRESS=0x0c8e79f3534b00d9a3d4a856b665bf4ebc22f2ba
-ESPRESSO_SEQUENCER_ESP_TOKEN_PROXY_ADDRESS=0x8ce361602b935680e8dec218b820ff5056beb7af
+ESPRESSO_SEQUENCER_LIGHT_CLIENT_PROXY_ADDRESS=0x196dbcbb54b8ec4958c959d8949ebfe87ac2aaaf
```
sveitser, line 65: Is it intentional that these addresses all change? It's annoying for the integrations team if the LC address also changes in the dev node because they are hardcoded in nitro tests where that hardcoding is difficult to remove.

cc @ImJeremyHe @Sneh1999

contracts/rust/deployer/src/lib.rs
```diff
@@ -3,16 +3,63 @@ use std::{collections::HashMap, io::Write};
...
+/// use `<HttpProviderWithWallet as WalletProvider>::default_signer_address(&provider)` to get wallet address
+pub type HttpProviderWithWallet = FillProvider<
+    JoinFill<JoinedRecommendedFillers, WalletFiller<EthereumWallet>>,
+    RootProvider,
+    Ethereum,
+>;
+
+/// a handy thin wrapper around wallet builder and provider builder that directly
+/// returns an instantiated `Provider` with default fillers with wallet, ready to send tx
+#[inline]
```
sveitser, line 36: just curious, why `#[inline]`?

.env
alxiong: I didn't investigate why these changes, the native demo failed and I checked the reason being the contract address mismatch, thus I update them as printed out in the logs of déployer. I'm not sure what changes cause them to be different tho. I didn't add new tx or change order

contracts/rust/deployer/src/lib.rs
alxiong: Compiler is probably smart enough to auto inline, I think this function is too simple to cause JUMP overhead

sveitser: I think we don't need this since we also call these functions very rarely and not a billion times in some tight loop. I don't know what exactly the overhead would be in terms of time but I'm guessing it's on the order of nanoseconds?

.env
sveitser: This is even weirder because the PR doesn't change `bin/deploy.rs` which is what's used to deploy for the demo. :thinking:

sveitser: Sorry I got tricked by the github UI. The code was updated.

contracts/rust/deployer/src/lib.rs
sveitser: AFAIK the only time this is useful if you we something to be inlined across crate boundaries and have concrete evidence from profiling or otherwise that it actually makes a difference.

contracts/rust/deployer/src/network_config.rs
```diff
@@ -0,0 +1,119 @@
...
+            match surf_disco::Client::<tide_disco::error::ServerError, StaticVersion<0, 1>>::new(
+                sequencer_url.clone(),
+            )
+            .get::<Vec<PeerConfig<SeqTypes>>>(&format!("node/stake-table/{}", epoch.u64()))
+            .send()
+            .await
+            {
+                Ok(resp) => break Ok(resp),
+                Err(e) => {
+                    tracing::error!("Failed to fetch the network config: {e}");
```
sveitser, line 51:
```suggestion
                    tracing::error!("Failed to fetch the stake table: {e}");
```

This isn't from this PR but since the code was moved, might as well fix it now.

.env
sveitser: no that code isn't called like this, sorry

alxiong: i see, so do you suggest that i change the ordering in `deploy_all()`? that would be bending backward.

I'm happy to update the genesis.toml file though

sveitser: Sorry not trying to nitpick, I just think we may be changing something we don't want to change by accident.

contracts/rust/deployer/src/lib.rs
alxiong: removed in [58a21a3](https://github.com/EspressoSystems/espresso-network/pull/3180/commits/58a21a3e236eeb0305869b9326df1012764bd36d)

.env
alxiong: i updated the toml file in [58a21a3](https://github.com/EspressoSystems/espresso-network/pull/3180/commits/58a21a3e236eeb0305869b9326df1012764bd36d)

alxiong: 💡 this is why the contract addresses changed:

 after cross-compare the logs, we didn't transfer ownership of the fee proxy on main. Since fee contract is the first we deploy, thus this extra transfer ownership tx cause the address of the rest of the deployments to be different.

alxiong:
May I get a re-review @sveitser @tbro

sveitser (approved):

PR #3181, opened 2025-04-29
Support predeployments in espresso-dev-node

QuentinI:
Closes #0000

### This PR:
Adds `--l1-deployment` option to espresso devnode with three possible values:
- "deploy" doesn't change the node's behaviour
- "dump" dumps a json description of L1 state after deployments complete and exits immediately. This is supported only when devnode starts a local Anvil node.
- "skip" skips all deployments for L1

Motivation for this comes from us at OP integration using devnode in tests and waiting 2+ minutes per integration test for devnode to deploy everything is very annoying. Dumping the chain state and then pre-deploying it on our L1 instance before starting the devnode significantly reduces test runtime.

### This PR does not:
Extend this behaviour to altchains.

### Key places to review:
Everything


sveitser:
@QuentinI I'm a bit confused as to why contract deployment takes a long time. Since by default anvil is configured to mine immediately on demand the mining it should be quick, shouldn't it? 

Also I'm curious if you're using this together with persisting the hotshot state or if you use a clean hotshot state each time. Last @imabdulbasit and I looked at that we figured that that feature might require some work.

Lastly it would be nice to have some tests for this. Maybe @imabdulbasit can help with this?

Overall LGTM.

QuentinI:
> @QuentinI I'm a bit confused as to why contract deployment takes a long time. Since by default anvil is configured to mine immediately on demand the mining it should be quick, shouldn't it?

It takes a long time for us because in OP integration we use optimism's e2e integration framework, which doesn't allow us to produce blocks on L1 on-demand.

> Also I'm curious if you're using this together with persisting the hotshot state or if you use a clean hotshot state each time.

No, only initial L1 state is dumped. I figured as we don't launch hotshot at all until after the deployment steps, this should be fine, and I haven't yet run into any issues.

> Lastly it would be nice to have some tests for this.

Sure! I didn't really have much time to work on testing for this right now, but we can leave this unmerged until I have bandwidth to add some tests and we'll just use this branch in OP Integration.

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -210,15 +250,36 @@ async fn main() -> anyhow::Result<()> {
...
         max_block_size,
         epoch_height,
+        contracts,
+        l1_deployment,
     } = cli_params;
 
     logging.init();
 
-    let (l1_url, _anvil) = if let Some(url) = rpc_url {
+    let mut anvil_statefile = NamedTempFile::new()?;
```
sveitser, line 259: Should we make this configurable via CLI argument so users can choose it? How are you currently determining what file to load?

QuentinI: This PR doesn't add loading into dev-node itself, it wouldn't make much difference since anvil node mines blocks on-demand anyways. We can only dump anvil state, which then can be loaded by the user when running devnode instances on their L1 instance that for whatever reason doesn't mine on-demand.

sveitser: Thanks I understand better now.

sveitser (approved):
LGTM. I think this is a great feature for integrations. We can add a test and documentation in a follow up issue.

PR #3182, opened 2025-04-29
deprecate hotshot-stake-table

### LINKED ISSUE #2971: [Light Client][Tech Debt] deprecate hotshot-stake-table crate
We should migrate to the stake table implementation in hotshot .

mrain:
Closes #2971


hotshot-state-prover/src/service.rs
```diff
@@ -101,7 +100,8 @@ impl ProverServiceState {
         let stake_table = fetch_stake_table_from_sequencer(&config.sequencer_url, None)
             .await
             .with_context(|| "Failed to initialize stake table")?;
-        let st_state = compute_stake_table_commitment(&stake_table, config.stake_table_capacity);
+        let st_state = compute_stake_table_commitment(&stake_table, config.stake_table_capacity)
```
alxiong, line 103: will it be clearer to introduce a `StakeTable` wrapper type around `Vec<PeerConfig>`? and implement `.commitment()` as a method. 

we can auto into or deref on the wrapper type. 

if that's too troublesome in your liking, then adding an alias would also help. because it's not straightforward for first comer to realize this type is what we call stake_table.

alxiong (commented):
Thanks for the cleanup, I have reviewed `hotshot-state-prover` and other places that I'm familiar with.

the only comment i have is below. A StakeTable handle/type is particularly needed now, especially when we have `StakeTableEntry` but not `StakeTable`, could be confusing.

Ayiga:
`node-metrics` changes look fine, and everything runs well in my local testing / evaluation.

sveitser:
staking-cli changes look fine

hotshot-state-prover/src/service.rs
mrain: Do you think that we should change (maybe) all occurrences of `Vec<PeerConfig>`?

alxiong: hmm.... good question. @imabdulbasit @Ayiga @bfish713 wdyt? 
I think only when the intention of the caller is clearly expecting a full stake table. there are cases where caller is just getting some slices or a few members of it, in those cases, maybe leave it as `Vec<PeerConfig>`?

mrain: I also agree that using a wrapper/type alias would be much clearer. But perhaps we need some more discussions on it and do it in another PR.

alxiong: Sure I don't mind another linked issue

mrain: https://github.com/EspressoSystems/espresso-network/issues/3188

bfish713: I would be in favor of the wrapper type, but it does seem like it can be done seperately

alxiong (approved):

PR #3192, opened 2025-04-30
Revert back to primary L1 provider some time after a failover

jbearer:
https://app.asana.com/1/1208976916964769/project/1209503267467519/task/1210113656728028?focus=true

### This PR:
Causes any L1 requests made at least a specified duration after the last failover from the primary provider, to switch back to the primary provider.

### DIFF
--- a/types/src/v0/impls/l1.rs
+++ b/types/src/v0/impls/l1.rs
@@ -310,11 +310,11 @@ impl Service<RequestPacket> for SwitchingTransport {
             let mut current_transport = self_clone.current_transport.read().clone();
 
             // Revert back to the primary transport if it's time.
-            if current_transport
+            let should_revert = current_transport
                 .status
                 .write()
-                .should_revert(current_transport.revert_at)
-            {
+                .should_revert(current_transport.revert_at);
+            if should_revert {
                 // Switch to the next generation which maps to index 0.
                 let n = self_clone.urls.len();
                 // Rounding down to a multiple of n gives us the last generation of the primary transport.
--- a/types/src/v0/impls/l1.rs
+++ b/types/src/v0/impls/l1.rs
@@ -180,7 +180,7 @@ impl SwitchingTransport {
         let metrics = L1ClientMetrics::new(&**opt.metrics, urls.len());
 
         // Create a new `SingleTransport` for the first URL
-        let first_transport = Arc::new(RwLock::new(SingleTransport::new(&first_url, 0)));
+        let first_transport = Arc::new(RwLock::new(SingleTransport::new(&first_url, 0, None)));
 
         Ok(Self {
             urls: Arc::new(urls),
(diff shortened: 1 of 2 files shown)


types/src/v0/v0_1/l1.rs
```diff
@@ -136,6 +136,15 @@ pub struct L1ClientOptions {
     )]
     pub l1_consecutive_failure_tolerance: usize,
 
+    /// Revert back to the first provider this duration after failing over.
+    #[clap(
+        long,
+        env = "ESPRESSO_SEQUENCER_L1_FAILOVER_REVERT",
+        default_value = "30m",
```
Ancient123, line 143: Seems like a pretty good default to me!

types/src/v0/impls/l1.rs
```diff
@@ -250,15 +250,33 @@ impl SingleTransportStatus {
...
+        let Some(revert_at) = revert_at else {
+            return false;
+        };
+        if Instant::now() >= revert_at {
+            self.shutting_down = true;
+            return true;
+        }
+
+        false
+    }
```
rob-maron, line 269: Can we reset `revert_at` here instead of later? I think it makes the state a tiny bit easier to understand

types/src/v0/impls/l1.rs
```diff
@@ -291,6 +309,21 @@ impl Service<RequestPacket> for SwitchingTransport {
             // Clone the current transport
             let mut current_transport = self_clone.current_transport.read().clone();
 
+            // Revert back to the primary transport if it's time.
+            if current_transport
+                .status
+                .write()
+                .should_revert(current_transport.revert_at)
+            {
```
rob-maron, line 317: I know the switching code is synchronous, but what do you think about something like
```rust
let should_revert = current_transport
    .status
    .write()
    .should_revert(current_transport.revert_at);

if should_revert {
```

types/src/v0/impls/l1.rs
```diff
@@ -291,6 +309,21 @@ impl Service<RequestPacket> for SwitchingTransport {
...
+                // Switch to the next generation which maps to index 0.
+                let n = self_clone.urls.len();
+                // Rounding down to a multiple of n gives us the last generation of the primary transport.
+                let prev_primary_gen = (current_transport.generation / n) * n;
+                // Adding n jumps to the next generation.
+                let next_gen = prev_primary_gen + n;
+                current_transport = self_clone.switch_to(next_gen, current_transport);
+            }
```
rob-maron, line 325: This works but curious on the generation logic (not introduced in this PR). What is the downside of storing the index and switching based on that and the number of URLs? Then the logic could be `switch_to(0, ..)` right?

jbearer: I added it because it made [this](https://github.com/EspressoSystems/espresso-network/blob/40ada470d2b9d10c0579db3d31d0874821931004/types/src/v0/impls/l1.rs#L983) a lot simpler if it's an always-increasing counter rather than a cycling index. 

Actually now I think this change sort of breaks that, in that `retry_on_all_providers` might now return without trying every single provider, if it reverts back to the primary in the middle. I don't think that's a big deal though. It's a best-effort semantics anyways

types/src/v0/impls/l1.rs
jbearer: Oh wait we can't (at least not easily) because `revert_at` is on the `SingleTransport` itself, not the `SingleTransportStatus` (which is the receiver of this method). And I think it makes sense to keep it there because the way it works out, `revert_at` doesn't actually need to be in the second layer of locking

Ancient123 (approved):

PR #3197, opened 2025-05-01
Register v1 version for all API modules

imabdulbasit:
Adds v1 versioning to all APIs to ensure consistency and avoid confusion, since the availability API already has both v0 and v1 versions.
All APIs—except for availability v0 and v1—register the same endpoints with identical behavior.

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -2120,6 +2120,7 @@ mod test {
 
     async fn run_catchup_test(url_suffix: &str) {
         setup_test();
+
         // Start a sequencer network, using the query service for catchup.
         let port = pick_unused_port().expect("No ports free");
         let anvil = Anvil::new().spawn();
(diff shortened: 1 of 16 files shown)


sequencer/src/api.rs
```diff
@@ -2190,35 +2193,50 @@ mod test {
...
-                    "waiting for blocks from {proposers:?}, block {height} is from {leaf_builder}",
-                );
-                proposers[leaf_builder] = true;
-            }
-
-            if proposers.iter().all(|has_proposed| *has_proposed) {
-                break;
+                if proposers.iter().all(|has_proposed| *has_proposed) {
+                    break;
+                }
```
jbearer, line 2220: What is this change? It looks semantically the same just with more indentation?

jbearer (approved):

sequencer/src/api.rs
imabdulbasit: I asked chatgpt to split into three tests and change the url, and it changes some of these things as well lol. I have reverted them.

imabdulbasit: Looks like I should have explicitly said not to change any of the test code lmao

PR #3202, opened 2025-05-02
Request/response fixes

rob-maron:
## This PR:
1. Makes all response validation happen in the fallible versions of the `StateCatchup` functions. Before this, we had been doing some validation in the top-level, infallible functions and some in the lower-level ones. This change makes it so all are validated, so a response from `try_xyz` is always already valid
2. Various usability improvements
3. `try_xyz` implementations for `ParallelStateCatchup` and `RequestResponseProtocol` with sane defaults so we don't hit anything `unreachable` if we want to manually use the lower-level functions

### DIFF
--- a/crates/hotshot/types/src/utils.rs
+++ b/crates/hotshot/types/src/utils.rs
@@ -104,12 +104,17 @@ pub type StateAndDelta<TYPES> = (
 );
 
 pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
-    leaf_chain: Vec<Leaf2<T>>,
-    stake_table: Vec<PeerConfig<T>>,
+    mut leaf_chain: Vec<Leaf2<T>>,
+    stake_table: &[PeerConfig<T>],
     success_threshold: U256,
     expected_height: u64,
     upgrade_lock: &crate::message::UpgradeLock<T, V>,
 ) -> anyhow::Result<Leaf2<T>> {
+    // Sort the leaf chain by view number
+    leaf_chain.sort_by_key(|l| l.view_number());
+    // Reverse it
+    leaf_chain.reverse();
+
     // Check we actually have a chain long enough for deciding
     if leaf_chain.len() < 3 {
         return Err(anyhow!("Leaf chain is not long enough for a decide"));
@@ -140,7 +145,7 @@ pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
     newest_leaf
         .justify_qc()
         .is_valid_cert(
-            StakeTableEntries::<T>::from(stake_table.clone()).0,
+            StakeTableEntries::<T>::from(stake_table.to_vec()).0,
             success_threshold,
             upgrade_lock,
         )
@@ -148,7 +153,7 @@ pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
     parent
         .justify_qc()
         .is_valid_cert(
-            StakeTableEntries::<T>::from(stake_table.clone()).0,
+            StakeTableEntries::<T>::from(stake_table.to_vec()).0,
             success_threshold,
             upgrade_lock,
         )
@@ -156,7 +161,7 @@ pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
     grand_parent
         .justify_qc()
         .is_valid_cert(
-            StakeTableEntries::<T>::from(stake_table.clone()).0,
+            StakeTableEntries::<T>::from(stake_table.to_vec()).0,
             success_threshold,
             upgrade_lock,
         )
@@ -169,7 +174,7 @@ pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
         ensure!(last_leaf.justify_qc().data().leaf_commit == leaf.commit());
         leaf.justify_qc()
             .is_valid_cert(
-                StakeTableEntries::<T>::from(stake_table.clone()).0,
+                StakeTableEntries::<T>::from(stake_table.to_vec()).0,
                 success_threshold,
                 upgrade_lock,
             )
--- a/sequencer/src/catchup.rs
+++ b/sequencer/src/catchup.rs
@@ -17,8 +17,8 @@ use espresso_types::{
     v0::traits::StateCatchup,
     v0_1::{RewardAccount, RewardAccountProof, RewardMerkleCommitment, RewardMerkleTree},
     v0_99::ChainConfig,
-    BackoffParams, BlockMerkleTree, FeeAccount, FeeAccountProof, FeeMerkleCommitment,
-    FeeMerkleTree, Leaf2, NodeState, PubKey, SeqTypes, ValidatedState,
+    BackoffParams, BlockMerkleTree, EpochVersion, FeeAccount, FeeAccountProof, FeeMerkleCommitment,
+    FeeMerkleTree, Leaf2, NodeState, PubKey, SeqTypes, SequencerVersions, ValidatedState,
 };
 use futures::{
     future::{Future, FutureExt, TryFuture, TryFutureExt},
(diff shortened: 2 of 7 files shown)


sequencer/src/catchup.rs
```diff
@@ -542,8 +569,28 @@ impl<T> StateCatchup for SqlStateCatchup<T>
...
+        // Verify the leaf chain
+        let leaf = verify_leaf_chain(
+            leaf_chain,
+            &stake_table,
+            success_threshold,
+            height,
+            &UpgradeLock::<SeqTypes, SequencerVersions<EpochVersion, EpochVersion>>::new(),
+        )
+        .await
+        .with_context(|| "failed to verify leaf chain")?;
```
rob-maron, line 591: Not sure if we need this, but better to be safer IMO

sequencer/src/catchup.rs
```diff
@@ -772,30 +846,15 @@ impl ParallelStateCatchup {
...
         // Make sure we have at least one provider
         let providers = self.providers.lock().clone();
-
         if providers.is_empty() {
             return Err(anyhow::anyhow!("no providers were initialized"));
         }
 
-        // Get the remote request delay so we can throw it into the future
-        let remote_request_delay = self.remote_request_delay;
-
```
rob-maron, line 782: Removed this stuff for now, the logic for doing local and then remote was complex because they would need to use completely different functions (`try_xyz` vs `xyz`). I think it's better to have something that works for now

sequencer/src/request_response/catchup/state.rs
```diff
@@ -32,52 +33,111 @@ impl<
...
-        unreachable!()
+    async fn try_fetch_leaf(
+        &self,
+        _retry: usize,
+        height: u64,
+        stake_table: Vec<PeerConfig<SeqTypes>>,
+        success_threshold: U256,
+    ) -> anyhow::Result<Leaf2> {
+        // Timeout after a few batches
+        let timeout_duration = self.config.request_batch_interval * 3;
```
rob-maron, line 44: Request/response is probabilistic, and it doesn't know how many nodes are running on the network in total. So it seemed reasonable that a `try` would try for a few batches (6s in our configured defaults) and then give up

types/src/v0/traits.rs
```diff
@@ -38,52 +38,55 @@ use serde::{de::DeserializeOwned, Serialize};
...
+            .retry(self, |provider, retry| {
+                let stake_table_clone = stake_table.clone();
+                async move {
+                    provider
+                        .try_fetch_leaf(retry, height, stake_table_clone, success_threshold)
+                        .await
+                }
+                .boxed()
+            })
+            .await
```
rob-maron, line 77: Git nuked this, but I just removed the leaf verification logic from here in favor of it happening in the lower functions

sequencer/src/catchup.rs
```diff
@@ -330,15 +333,36 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
-    async fn try_fetch_leaves(&self, retry: usize, height: u64) -> anyhow::Result<Vec<Leaf2>> {
-        self.fetch(retry, |client| async move {
-            let leaf = client
-                .get::<Vec<Leaf2>>(&format!("catchup/{}/leafchain", height))
-                .send()
-                .await?;
-            anyhow::Ok(leaf)
-        })
+
+    async fn try_fetch_leaf(
```
imabdulbasit, line 337: can you please also rename this method to try_fetch_leaf_chain

sequencer/src/catchup.rs
```diff
@@ -330,15 +333,36 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
+        let leaf_chain = self
+            .fetch(retry, |client| async move {
+                let leaf = client
+                    .get::<Vec<Leaf2>>(&format!("catchup/{}/leafchain", height))
+                    .send()
+                    .await?;
+                anyhow::Ok(leaf)
+            })
+            .await
+            .with_context(|| format!("failed to fetch leaf at height {height}"))?;
```
imabdulbasit, line 354:
```suggestion
            .with_context(|| format!("failed to fetch leaf chain at height {height}"))?;
```

sequencer/src/catchup.rs
```diff
@@ -554,14 +601,28 @@ where
...
-        Ok(self
+    ) -> anyhow::Result<Vec<FeeAccountProof>> {
+        // Get the accounts
+        let (fee_merkle_tree_from_db, _) = self
             .db
             .get_accounts(instance, block_height, view, accounts)
-            .await?
-            .0)
+            .await
+            .with_context(|| "failed to get reward accounts from DB")?;
```
imabdulbasit, line 612:
```suggestion
            .with_context(|| "failed to get fee accounts from DB")?;
```

sequencer/src/catchup.rs
imabdulbasit: The fetch() function used to perform the verification. I see that the verification was removed from there and added to try_fetch(), so it's needed here

types/src/v0/traits.rs
```diff
@@ -212,15 +209,7 @@ pub trait StateCatchup: Send + Sync {
                             err.context(format!(
                                 "fetching reward accounts {accounts:?}, height {height}, view {view:?}"
                             ))
-                        })?;
```
imabdulbasit, line 215: wondering why all the verification was moved into try_fetch functions?

imabdulbasit: is it just because we use try_fetch_leaf_chain() at one place?

sequencer/src/catchup.rs
rob-maron: Ah right, I meant in terms of the data already coming from our trusted DB. But it makes sense to check either way

sequencer/src/catchup.rs
imabdulbasit: the first provider is always the database, so a hacky way could be to check that, and then spawn parallel requests

imabdulbasit: the previous implementation, checked every provider serially so that is what was happening anyway

sequencer/src/catchup.rs
imabdulbasit: yeah, we were checking that before as well so we have to keep it here

sequencer/src/catchup.rs
rob-maron: I have the changes locally, this can come after that. It was a big PR already and it was introducing a lot more code

sequencer/src/catchup.rs
rob-maron: What about the `fetch_leaf` function? I changed this because it is no longer returning the entire leaf chain, but the validated leaf

types/src/v0/traits.rs
imabdulbasit: yeah that sounds okay to me

sequencer/src/catchup.rs
imabdulbasit: oh so this one already has the validation, then we can't rename it to `try_fetch_leaf_chain`.

imabdulbasit (approved):

crates/hotshot/types/src/utils.rs
```diff
@@ -140,23 +145,23 @@ pub async fn verify_leaf_chain<T: NodeType, V: Versions>(
     newest_leaf
         .justify_qc()
         .is_valid_cert(
-            StakeTableEntries::<T>::from(stake_table.clone()).0,
+            StakeTableEntries::<T>::from(stake_table.to_vec()).0,
```
jbearer, line 148: Um why are we cloning the entire stake table every time we validate a cert? Does `is_valid_cert` really need to take ownership of the stake table?

rob-maron: I have no idea, just followed what we already had. But I can look into this before merging

rob-maron: Fixed in https://github.com/EspressoSystems/espresso-network/pull/3202/commits/080c7abab0efdee4af11252798a48287328bc0cc

PR #3205, opened 2025-05-04
fix: pass correct stake table contract address

leopardracer:


leopardracer:
@sveitser, @jbearer @tbro @imabdulbasit @ss-es @pls148 @bfish713 @rob-maron @lukaszrzasik@

PR #3211, opened 2025-05-05
Use `AnvilProvider` in TestNetwork to  enable interval mining after PoS contracts deployment

imabdulbasit:
Closes https://app.asana.com/1/1208976916964769/project/1209413670216270/task/1210134443040120

This PR uses `AnvilProvider`, constructed using the L1Client as the provider and with the Anvil instance, and adds it to `TestNetworkConfig`. It is used to enable interval mining with a 1-second block time after all Proof-of-Stake contracts have been deployed in pos_hook().

Previously, each transaction in the PoS deployment routine waited for finalization, requiring at least one block per transaction and increasing `TestNetwork` startup time. With this change, auto-mining is used during the deployment phase to speed up contract deployment, and interval mining is enabled only afterward. This avoids adding extra startup time.

This PR also moves the Anvil instance into the TestNetwork::default() implementation.


staking-cli/src/demo.rs
```diff
@@ -38,7 +38,7 @@ use crate::{
 };
 
 pub const STAKE_TABLE_CAPACITY_FOR_TEST: usize = 3;
-type Prov = FillProvider<
+pub type AlloyFillProvider = FillProvider<
```
sveitser, line 41: Looks like this doesn't need to be `pub`?

sveitser (approved):

PR #3212, opened 2025-05-05
3187 thoroughly test delegations and undelegations for all scenarios

### LINKED ISSUE #3187: thoroughly test delegations and undelegations for all scenarios
e.g.
https://github.com/EspressoSystems/espresso-network/blob/e8c696b3dab3d4b39b228b8910bfde417e88fd8b/contracts/src/StakeTable.sol#L463-L467basically after a node is exited you cant undelegate so you have to get the stake back via claimValidatorExit but we dont touch the validators[validator].delegatedAmount in the case so a validators delegatedAmount could still be large

alysiahuggins:
Closes #3187


contracts/test/StakeTable.t.sol
```diff
@@ -633,6 +652,554 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+
+        assertEq(token.balanceOf(delegator), INITIAL_BALANCE);
+        assertEq(token.balanceOf(address(stakeTable)), 0);
+        (validatorAmountDelegated,) = stakeTable.validators(validator);
+        assertEq(validatorAmountDelegated, 0);
+        uint256 delegatedAmount = stakeTable.delegations(validator, delegator);
+        assertEq(delegatedAmount, 0);
+    }
+
+    function test_ValidatorExitDuringUndelegationPeriod() public {
```
alysiahuggins, line 921: hey @sveitser, for this particular test, it reveals an undelegation scenario that I'm not a fan of 

undelegation scenario example:
- user delegates 4 tokens
- user undelegates 1 tokens
- validator exits
- user tries to call `claimWithdrawalExit()`
- They're only able withdraw 3 tokens
- they then have to call `claimWithdrawal()` to withdraw the previously undelegated amouont( 1 token in this example)

I wish they could have withdrawn all of their balance in one function as it's all available to be withdrawn - perhaps we should try to consolidate the functions to deal with this scenario, let me know what you think.

contracts/test/StakeTable.t.sol
```diff
@@ -633,6 +652,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+            EdOnBN254.EdOnBN254Point memory postExitSchnorrVK,
+            BN254.G1Point memory postExitSig
+        ) = genClientWallet(validator, "145");
+
+        vm.expectRevert(S.ValidatorAlreadyExited.selector);
+        stakeTable.updateConsensusKeys(postExitBlsVK, postExitSchnorrVK, postExitSig);
+        vm.stopPrank();
+    }
+
+    function test_ValidatorSelfDelegation() public {
```
alysiahuggins, line 1206: i saw this acceptance criteria in the [book](https://github.com/EspressoSystems/book/blob/main/pages/stake-table-contract-requirements.tex#L142), but i wonder why we want this behaviour @sveitser

sveitser: The previous design separated out solo stakers. This is to explicitly make it clear that the use-case of delegating to one's own node is covered by the new design.

contracts/test/StakeTable.t.sol
sveitser: It's just like this for simplicity now and product team has OK'ed it.

Concretely what would you like to add? We can't just call the two functions because it will revert unless both types of withdrawals are available. I think we could combine the functions, add up the balance to withdraw. Only revert if there is neither of the two withdrawals available.

So something like (untested):

```solidity
    function claimWithdrawal(address validator) external virtual {
        address delegator = msg.sender;
        // If entries are missing at any of the levels of the mapping this will return zero
        uint256 undelegationAmount = undelegations[validator][delegator].amount;
        uint256 withdrawalAmount;

        if (block.timestamp >= undelegations[validator][delegator].unlocksAt) {
            withdrawalAmount += undelegationAmount;
            // Mark funds as spent
            delete undelegations[validator][delegator];
        }

        if (block.timestamp >= validatorExits[validator]) {
            withdrawalAmount += delegations[validator][delegator];
            delete delegations[validator][delegator];
        }

        if (withdrawalAmount == 0) {
            revert NothingToWithdraw();
        }

        SafeTransferLib.safeTransfer(token, delegator, withdrawalAmount);

        emit Withdrawal(delegator, withdrawalAmount);
    }
```

But if that function is ran too early it needs to be run again. Alternatively we could add a new function that only works if both types of withdrawals are available. This then however makes it more work for UIs because there are 3 functions to call.

alysiahuggins: Hey Mathis, thanks for your response with a code example. 

```quote
But if that function is ran too early it needs to be run again. 
```

Too early meaning their undelegation was ready for withdrawal but the validator already exited and the escrow period for that validator exit hasn't passed yet? If yes, then they would have had to call it twice in the current version.

However if they happened to call it after the validator Exit escrow period then they would only have one call.

I mean we can leave as is but i'm thinking about a user and it's a definitely confusing.
Also if not done in the contract then it's left to the user + docs somewhere and/or a UI doing the check for existing undelegations. Not the best experience both ways but what do you think?

sveitser: Sorry didn't see this reply.

> Too early meaning their undelegation was ready for withdrawal but the validator already exited and the escrow period for that validator exit hasn't passed yet? If yes, then they would have had to call it twice in the current version.

Yes. I actually think it's not necessarily bad to have separate functions. It makes the intent clear and the fact if it worked or not more informative. For the combined functions there is a case where the function did something successfully but there are funds still have to be withdrawn and the function needs to be called again. I'm not a big fan of that.

I don't think there's a way around having the UI / user check for existing undelegations unless we rewrite the contract such that anyone can execute the claim operations on behalf of other users. It's also something we could do but I think we should leave this for later and focus on making sure what we currently have is safe and sound. For the CLI this is of course up to the user but we can have an "info" like command that shows the users what's actions they should perform or even propose the transaction to claim. For a GUI I expect it to subscribe to events and display the user when certain things are "ready" to be executed.

I also think that validator exits should be really rare but it would be nice if users didn't have to do anything to get funds back into their wallets.

wdyt?

contracts/src/StakeTable.sol
```diff
@@ -275,6 +275,9 @@ contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpg
     }
 
     function ensureValidatorActive(address validator) internal view {
+        if (validators[validator].status == ValidatorStatus.Exited) {
```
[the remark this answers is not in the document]
alysiahuggins: Yea, that makes sense. The reason I separated out the Exited case is so that checks like ensureValidatorActive can give more nuanced errors.  it’s helpful to distinguish between a validator that never registered (Status: Unknown) and one that registered but has since exited (Status: Exit).

Inactive in our case means either not registered (Unknown) or Exited (Exit). Technically, the right error for the Unknown case would be something like ValidatorNotRegistered, but since we can't rename errors in upgradeable contracts, we’d have to either add a new error or keep using ValidatorInactive. I’m fine with the latter unless there’s a strong need to make that distinction more explicit.

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Delegated(delegator, validator1, INITIAL_BALANCE / 3);
+        stakeTable.delegate(validator1, INITIAL_BALANCE / 3);
+
+        // delegate to validator 2
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Delegated(delegator, validator2, INITIAL_BALANCE / 3);
+        stakeTable.delegate(validator2, INITIAL_BALANCE / 3);
+
+        // delegate from validator 1
```
sveitser, line 724:
```suggestion
        // delegate to validator 1
```

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Withdrawal(delegator, INITIAL_BALANCE / 3);
+        stakeTable.claimWithdrawal(validator2);
+        assertEq(token.balanceOf(delegator), delegatorBalanceBefore + INITIAL_BALANCE / 3);
+        assertEq(token.balanceOf(address(stakeTable)), 0);
+
+        vm.stopPrank();
+    }
+
+    function test_RevertWhen_WithdrawFromNonExistentUndelegation() public {
```
sveitser, line 866: Was a bit confused by the test name

```suggestion
    function test_RevertWhen_UndelegateWithoutDelegation() public {
```

sveitser: Or should this test call `claimWithdrawal`?

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        vm.prank(tokenGrantRecipient);
+        token.transfer(delegator, INITIAL_BALANCE);
+
+        registerValidatorOnStakeTable(validator, seed1, COMMISSION, stakeTable);
+
+        vm.startPrank(delegator);
+        token.approve(address(stakeTable), INITIAL_BALANCE);
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Delegated(delegator, validator, INITIAL_BALANCE);
+        stakeTable.delegate(validator, INITIAL_BALANCE);
```
sveitser, line 936: I think it would  become a bit easier to read if we made a helper function for this like with the validator registration, since it's used quite a lot. Can be done as follow up work though.

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        assertEq(delegatorAmountUndelegated, INITIAL_BALANCE / 2); // undelegated amount is
+            // INITIAL_BALANCE/2 because the user undelegated INITIAL_BALANCE/2
+        assertEq(unlocksAt, block.timestamp + ESCROW_PERIOD); // unlocks at is the current timestamp
+            // + ESCROW_PERIOD because the user undelegated INITIAL_BALANCE/2
+
+        vm.startPrank(delegator);
+        vm.expectRevert(abi.encodeWithSelector(S.ValidatorAlreadyExited.selector));
+        stakeTable.undelegate(validator, INITIAL_BALANCE);
+
+        //validator already exited so call claim withdaral exit
```
sveitser, line 961:
```suggestion
        //validator already exited so call claim withdrawal exit
```

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        stakeTable.deregisterValidator();
+        vm.stopPrank();
+        (uint256 validatorAmountDelegated,) = stakeTable.validators(validator);
+        assertEq(validatorAmountDelegated, 0);
+        (uint256 delegatorAmountUndelegated, uint256 unlocksAt) =
+            stakeTable.undelegations(validator, delegator);
+        assertEq(delegatorAmountUndelegated, INITIAL_BALANCE / 2); // undelegated amount is
+            // INITIAL_BALANCE/2 because the user undelegated INITIAL_BALANCE/2
+        assertEq(unlocksAt, block.timestamp + ESCROW_PERIOD); // unlocks at is the current timestamp
+            // + ESCROW_PERIOD because the user undelegated INITIAL_BALANCE/2
```
sveitser, line 955: Are these assert _after_ the validator exit on purpose? Just asking because the changes are made by the undelegation.

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        assertEq(validatorAmountDelegated, 0);
+        uint256 delegatedAmount = stakeTable.delegations(validator, delegator);
+        assertEq(delegatedAmount, 0);
+
+        // now claim the previous undelegation
+        vm.startPrank(delegator);
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Withdrawal(delegator, INITIAL_BALANCE / 2);
+        stakeTable.claimWithdrawal(validator);
+        vm.stopPrank();
```
sveitser, line 981: Add a balance check to show everything is withdrawn?

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Withdrawal(delegator2, INITIAL_BALANCE / 3);
+        stakeTable.claimValidatorExit(validator);
+        vm.stopPrank();
+
+        vm.startPrank(delegator3);
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Withdrawal(delegator3, INITIAL_BALANCE / 3);
+        stakeTable.claimValidatorExit(validator);
+        vm.stopPrank();
```
sveitser, line 1038: Another candidate for a helper function

contracts/test/StakeTable.t.sol
```diff
@@ -637,6 +656,596 @@ contract StakeTable_register_Test is LightClientCommonTest {
...
+    }
+
+    function test_ValidatorSelfDelegation() public {
+        uint256 validatorBalance = token.balanceOf(validator);
+        assertEq(validatorBalance, INITIAL_BALANCE);
+        (
+            BN254.G2Point memory blsVK,
+            EdOnBN254.EdOnBN254Point memory schnorrVK,
+            BN254.G1Point memory sig
+        ) = genClientWallet(validator, seed1);
```
sveitser, line 1217: use the helper function?

sveitser (approved):

sveitser:
@alysiahuggins thanks. LGTM. Just some suggestions for cleanup. I'm okay with merging first and doing follow up work.

contracts/test/StakeTable.t.sol
alysiahuggins: it should have tested claimWithdrawal https://github.com/EspressoSystems/espresso-network/commit/2d2e8e95c9c95c166f165d0e1330888a9abc1515

contracts/test/StakeTable.t.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/45a8d436679775f3252ecb1704b055f92d5f7723

contracts/test/StakeTable.t.sol
alysiahuggins: it does for each user at the end but added a check that the stake table balance is zero

PR #3216, opened 2025-05-06
add_drb_result to store_drb_result

pls148:
Closes #<ISSUE_NUMBER>


bfish713 (commented):
I don't have any problem with this, but there are so many conflicts, I can look again as soon as those are resolved

bfish713 (approved):
Looks fine to me, @ss-es can you also take a look, since DRB checkpointing is doing something similar

ss-es (approved):
I think this looks good to me

PR #3217, opened 2025-05-06
split hotshot tests_6 into multiple binaries

sveitser:
Splits up tests_6 to avoid high memory usage and increase parallelization. 

Adds just recipes to build the github test matrix dynamically.

What to review:

- Check that all tests are still present. Most files are just moved, but one test file has been split up.

This is very hard to do manually, so
```
git checkout ma/split-hotshot-tests_6
# remove lines ending with : and select last 3 fields x::y::z 
cargo nextest list -p hotshot-testing --profile hotshot | grep -v ':$' | awk -F'::' '{print $(NF-2) "::" $(NF-1) "::" $NF}' | sort | tee HEAD

git checkout main
cargo nextest list -p hotshot-testing --profile hotshot | grep -v ':$' | awk -F'::' '{print $(NF-2) "::" $(NF-1) "::" $NF}' | sort | tee MAIN
```
Then looking at `diff MAIN HEAD`, it's empty with 210 tests in each.


sveitser:
@lukaszrzasik thanks for the review. I haven't merged this because it would mess with the required CI jobs. So I plan to merge this after #3257 is merged and we have updated our github branch protection rule to require these aggregate jobs instead.

sveitser:
This is finally ready to merge. Checking with the diff method described in the PR description I again get no diff but now have 212 tests here and in current `main`. So I think we still have everything covered. There were no conflicts on merging in main either.

PR #3225, opened 2025-05-07
Remove marketplace builder, builder-core, solver crates

sveitser:
First PR for marketplace removal. Mainly removes builder, builder-core, solver crates which aren't very intertwined with the rest of the code, binaries which some small changes in the sequencer code base.


sequencer/src/context.rs
```diff
@@ -146,7 +145,12 @@ impl<N: ConnectedNetwork<PubKey>, P: SequencerPersistence, V: Versions> Sequence
             initializer,
             ConsensusMetricsValue::new(metrics),
             Arc::clone(&persistence),
-            marketplace_config,
+            // TODO: MA: will be removed when more marketplace code is removed,
+            // at the moment we need to pass in a config to hotshot.
+            MarketplaceConfig {
```
sveitser, line 150: Wondering if this will break anything. I'm thinking if the marketplace feature isn't turned on it shouldn't but not sure.

imabdulbasit (approved):
I haven't reviewed every line, but most of the files are related to the marketplace. If it compiles and all the tests pass, it is okay to merge it

PR #3226, opened 2025-05-07
Remove marketplace, auction, solver, v99 code

sveitser:
This PR Removes 
- new_marketplace function from header, renames new_legacy( to new(),
- espresso types v99 module
- marketplace tests
- hotshot builder v99 module and api

Second PR for marketplace removal. Follow up from #3225 

---
  - https://app.asana.com/0/0/1210097601362705


crates/hotshot/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -501,39 +501,20 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
-                *self.view_number,
-                commitment_and_metadata.auction_result,
-                version,
-            )
-            .await
-            .wrap()
-            .context(warn!("Failed to construct marketplace block header"))?
-        };
-
+        let block_header = TYPES::BlockHeader::new_legacy(
```
imabdulbasit, line 504: should new_legacy() be renamed to new() ?

sveitser: yeah sure, a lot is missing from this PR still :D

sveitser:
@imabdulbasit I rebased this PR on `main`. I think currently hotshot crates compile, excl. `hotshot-testing`. `espresso-types` and anything downstream isn't finished.

types/src/v0/v0_3/chain_config.rs
```diff
@@ -23,6 +23,7 @@ pub struct ChainConfig {
     /// contract when they are off. In a future release, after fees are switched on and thoroughly
     /// tested, this may be made mandatory.
     #[serde(with = "ethers_serde::option_address")]
+    #[serde(default)]
     pub fee_contract: Option<Address>,
```
rob-maron, line 27: Just curious - shouldn't these already be Option-compatible?

types/src/v0/v0_3/chain_config.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{v0_1, v0_99, BlockSize, ChainId, FeeAccount, FeeAmount};
+use crate::{v0_1, BlockSize, ChainId, FeeAccount, FeeAmount};
 use alloy::primitives::{Address, U256};
 use alloy_compat::ethers_serde;
 use committable::{Commitment, Committable};
```
rob-maron, line 4: Previously were we using the v99 chain config everywhere (e.g. in storage)? If so, removing this field by changing to v3 (while not necessarily making it fail deserialization) may change the chain config commitment

imabdulbasit: The bid_recipient was None everywhere because the marketplace hasn't been deployed yet. The commitment was calculated using an if statement that includes the bid_recipient only if it is Some(_), so I don't think the commitment would change.

types/src/v0/v0_3/chain_config.rs
imabdulbasit: some of the tests started complaining about fields not present even though it should be deserialized as None. removing `#[serde(with = "ethers_serde::option_address")]` fixed those errors but we need this derive.. I don't know the exact reason why  the `#[serde(with = "ethers_serde::option_address")] ` would prevent it, but adding `#[serde(default)]` fixed it

types/src/v0/v0_3/chain_config.rs
rob-maron: Ah okay great then, I wasn't able to immediately find the commitment function

Approved

rob-maron (approved):

sequencer/src/api.rs
```diff
@@ -2637,109 +2636,95 @@ mod test {
...
-                    break; // if assertion did not panic, the test was successful, so we exit the loop
-                }
-            }
-            sleep(Duration::from_millis(200)).await;
-        }
-
-        network.server.shut_down().await;
-    }
+    // async fn test_upgrade_helper<V: Versions>(version: V) {
+    //     setup_test();
```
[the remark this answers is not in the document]
imabdulbasit: This function was only used for marketplace. Clippy started complaining when I removed marketplace tests. The epoch upgrade test do not use this function I think, so I just commented it out. I can remove this completely

types/src/v0/v0_3/chain_config.rs
imabdulbasit: here is the commitment. I am not sure if this link works.
https://github.com/EspressoSystems/espresso-network/pull/3226/files#diff-ed2d27878980f1d1c3a7cae443563bb866d4486d5898f4445d86e82366b86310L81

ss
image

PR #3229, opened 2025-05-07
Store progress at DRB checkpoints

ss-es:
The bulk of the PR is just propagating storage to the membership coordinator. The main logic to pay attention to is in `drb.rs`. Please look over the loops in the DRB calculation especially carefully; it's easy for this to be off by 1 somewhere. 

The methods will be implemented in a separate PR.


crates/hotshot/types/src/drb.rs
```diff
@@ -8,7 +8,19 @@ use std::collections::BTreeMap;
...
+    storage::StoreDrbProgressFn,
+};
+
+pub struct DrbInput {
+    /// The epoch we are calculating the result for
+    pub epoch: u64,
+    /// The iteration this seed is from. For fresh calculations, this should be `0`.
+    pub iteration: u64,
+    /// Initial seed for the DRB calculation
+    pub initial: [u8; 32],
```
mrain, line 22: The name seems misleading. Should it be the seed after iterations?

mrain: Also a general question, do we have to verify the validity of this seed?

crates/hotshot/types/src/drb.rs
```diff
@@ -55,9 +70,50 @@ pub fn difficulty_level() -> u64 {
...
+        });
+    }
+
+    // perform the remaining iterations
+    for _ in iteration..DIFFICULTY_LEVEL {
+        hash = Sha256::digest(hash).to_vec();
+        iteration += 1;
+    }
+
+    for _ in 0..DRB_CHECKPOINT_INTERVAL {
```
mrain, line 116: Not sure about the correctness here. What if the remaining iteration is not a multiple of interval?
Should we also create a reference test for it?

crates/hotshot/types/src/drb.rs
ss-es: there's no way to verify correctness or anything like that (I'm not sure what that would even mean in this case)

I agree on the naming, I'll change to `seed` I think

mrain: For correctness we could spawn multiple threads verifying the previous computation, we know the genesis seed.

crates/hotshot/types/src/drb.rs
ss-es: oops, that last loop was completely unintentional.

and yes, I'll add a test. I wasn't sure how to write one initially but I think I have a better idea now

crates/hotshot/types/src/drb.rs
ss-es: I'm not sure that adds anything -- the point of storing the partial results is that we want to avoid having to redo the work we've already done in a long-running calculation, in case e.g. the node restarts. also note that these are not being sent over the network, we're just trying to make sure we can restart the calculation from a previous state when necessary

crates/hotshot/types/src/drb.rs
ss-es: test added here: https://github.com/EspressoSystems/espresso-network/pull/3229/commits/ebeaf0794bbc42847a208ba09505c6f2cbf68d9a

I did confirm that the test failed on the initial PR (with the extra loop that you pointed out), but passes now

sequencer/src/network/cdn.rs
```diff
@@ -78,7 +78,8 @@ impl<T: SignatureKey> SignatureScheme for WrappedSignatureKey<T> {
             Err(_) => return false,
         };
 
-        public_key.0.validate(&signature, &namespaced_message)
+        public_key.0.validate(&signature, message)
```
mrain, line 81: am slightly worried about this change. The documentation said we shouldn't do this because of the signature replay attacks.

crates/hotshot/testing/src/helpers.rs
```diff
@@ -128,11 +127,8 @@ pub async fn build_system_handle_from_launcher<
         hotshot_config.known_da_nodes.clone(),
     )));
 
-    let coordinator = EpochMembershipCoordinator::new(
-        memberships,
-        Some(storage_add_drb_result(storage.clone())),
-        hotshot_config.epoch_height,
-    );
+    let coordinator =
+        EpochMembershipCoordinator::new(memberships, hotshot_config.epoch_height, &storage.clone());
```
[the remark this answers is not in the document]
ss-es: oops, leftover from when I required `&'static` I think. fixed!

sequencer/src/network/cdn.rs
ss-es: this is odd, I don't know why this change is here. I didn't make this change and it's not related to this PR

fixed

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -51,28 +56,27 @@ impl<TYPES: NodeType> Clone for EpochMembershipCoordinator<TYPES> {
...
-        storage_add_drb_result_fn: Option<StorageAddDrbResultFn<TYPES>>,
         epoch_height: u64,
+        storage: &S,
     ) -> Self {
         Self {
             membership,
             catchup_map: Arc::default(),
-            storage_add_drb_result_fn,
             epoch_height,
+            store_drb_progress_fn: store_drb_progress_fn(storage.clone()),
```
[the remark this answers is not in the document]
ss-es: I don't think so -- I think `storage_store` is a bit redundant, and it also takes `storage` as an argument anyway

I do think this is part of why we should rename `add_drb_result` -> `store_drb_result` on storage though, to avoid conflation with the membership method

builder/src/non_permissioned.rs
```diff
@@ -60,14 +61,15 @@ pub fn build_instance_state<V: Versions>(
         l1_client.clone(),
         chain_config,
     );
+    let storage = TestStorage::default();
```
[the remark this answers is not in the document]
ss-es: updated to use `NoStorage` from `sequencer`!

PR #3241, opened 2025-05-08
Fix premature rand membership fetching

lukaszrzasik:
Closes [Extra Catchup Requests Triggered causing Failed to fetch leaves logs](https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1209912122084028?focus=true)

### This PR:

* Fixes DA voting logic so that it doesn't prematurely trigger fetching of the next epoch membership
* Fixes proposal logic so that it doesn't prematurely trigger fetching of the next epoch membership
* Adds a helper function `broadcast_view_change` that checks whether epoch field should be set because we reached the epoch version view. It then broadcasts the correct `ViewChange` event. That way we avoid sending `ViewChange` event twice for the same view which most of our tasks treat as a reason for noop.
* Removes a workaround in the transactions task that sets epoch to 1 if the received epoch in `ViewChange` event is `None` and the we're already in the epoch version.
* Fixes all "missing randomized stake table" logs in the demo-native-pos

### This PR does not:

* Doesn't fix an issue with the leader selection in the first view of the epoch version.

### DIFF
--- a/.github/workflows/hotshot.yml
+++ b/.github/workflows/hotshot.yml
@@ -96,3 +96,23 @@ jobs:
         run: |
           just hotshot::example all-push-cdn -- --config_file ./crates/hotshot/orchestrator/run-config.toml
         timeout-minutes: 20
+
+  # This job enables having a single required status check in github for all hotshot tests.
+  aggregate-hotshot-tests:
+    needs: [test, test-examples]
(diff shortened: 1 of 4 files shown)


bfish713 (approved):
Looks straight forward to me, and it explains why we sometimes saw logs for missing randomized committee for epoch 1

PR #3243, opened 2025-05-08
fix: fix staking to contract in local demo

sveitser:
The "exit_on_failure" will terminate process compose if the process exits with non-zero exit code. Until we have better tests this should at least increase the visibility.


bfish713 (approved):

sveitser:
So this breaks the demo? I'm so confused.

PR #3245, opened 2025-05-08
run demo native PoS tests on the CI

sveitser:
Adds a test for running the native demo with 2 configs:

- PoS version as the base
- Upgrade from Fee to PoS version.

The tests are really slow now because I think we should run the demo for long enough to make sure we can get to the point where we are using the new stake table from the contract. I did add some extra timeout to hopefully fail much earlier if things are otherwise broken and we see no blocks or no transactions finalized.

Discovered issues:

- #3264


sveitser:
On this run: https://github.com/EspressoSystems/espresso-network/actions/runs/14975602546

One test go a connection refused from the L1 after about 800 hotshot blocks when calling getLogs. This seems rather strange.

sveitser:
One run indicated that we were running out of disk space on the runner. The test was previously running with RUST_LOG=debug, removed this now.

data/genesis/demo-pos-base.toml
```diff
@@ -1,6 +1,7 @@
 base_version = "0.3"
 upgrade_version = "0.3"
-epoch_height = 300
+epoch_height = 100
+epoch_start_block = 1
```
imabdulbasit, line 4: so I think we also need to change default value for `epoch_start_block`to 1 if its not provided in genesis.toml?

data/genesis/demo-pos.toml
```diff
@@ -21,7 +23,7 @@ number = 0
 
 [[upgrade]]
 version = "0.3"
-start_proposing_view = 200
+start_proposing_view = 150
 stop_proposing_view = 1000
```
imabdulbasit, line 27: lets decrease the `stop_proposing_view` window to something like 300 maybe?

sveitser: The test actually only waits for 200 blocks for the upgrade to happen and will fail otherwise. Not sure what setting it lower in this file accomplishes though.

data/genesis/demo-pos-base.toml
sveitser: Do you mean in the sequencer, or prover or both?

Currently it will be set to `0` at

https://github.com/EspressoSystems/espresso-network/blob/02bb5190ad9c772821d4777eb9471844c1d49540/sequencer/src/lib.rs#L372

this seems kind of wrong doesn't it?

Like if we used `main` code for `mainnet` right now it would be set to `0` there.

The hotshot network config does require a value to be set.

imabdulbasit: yeah, we need to set it to 1 instead of 0 IMO

imabdulbasit (approved):

data/genesis/demo-pos.toml
imabdulbasit: doesn't accomplish anything, its just weird to have such a large window for proposing

.github/workflows/test.yml
```diff
@@ -208,14 +208,10 @@ jobs:
...
-        version: [02]
         include:
-          - version: 02
-            test-name: test_native_demo_basic
-
-          # - version: 99
-          #   test-name: test_native_demo_upgrade
+          - test-name: test_native_demo_base
+          - test-name: test_native_demo_pos_base
+          - test-name: test_native_demo_pos_upgrade
```
alxiong, line 214: can we also add these options to justfile for easy local testing?

data/genesis/demo-pos-base.toml
sveitser: I don't understand why 1 is okay. If we run the current binary with the current mainnet genesis we would be setting it to 1 but we're not configured to run with epochs or upgrade to it.

.github/workflows/test.yml
sveitser: There are no extract config / env vars required now so `cargo test test_native_demo_base` already works. Is that sufficient?

mrain (approved):
LGTM

data/genesis/demo-pos-base.toml
sveitser: So if we set this to 1 it breaks the prover for the demo native fee version.

sveitser:
So there is more to discuss about what to do with default values and w.r.t. to the prover. I will merge this PR first since that seems out of scope of this PR for me.

PR #3249, opened 2025-05-09
Add a rewards test

imabdulbasit:
Checks
- no reward distribution for first two epochs
- validator stake is sum of total delegators stake
- reward calculation matches to what is in the reward merkle tree


sequencer/src/api.rs
```diff
@@ -3389,6 +3393,213 @@ mod test {
...
+                    .await
+                    .ok()
+                    .flatten();
+                assert!(amount.is_none(), "amount is not none for block {block}")
+            }
+        }
+
+        // Collect leaves for epoch 3 to 5 to verify reward calculations.
+        let leaves = client
+            .socket("availability/stream/leaves/41")
```
sveitser, line 3509: is this epoch height * 2 + 1 or 41 for some other reason?

sequencer/src/api.rs
```diff
@@ -3389,6 +3393,213 @@ mod test {
...
+
+            let rewards = compute_rewards(leader_validator.clone()).expect("reward computation");
+
+            // Verify that the leader commission amount is within the tolerated range.
+            // Due to potential rounding errors in decimal calculations for delegator rewards,
+            // the actual distributed commission
+            // amount may differ very slightly from the calculated value.
+            // this asserts that it is within 10wei tolerance level.
+            // 10 wei is 10* 10E-18
+            let (_, leader_commission_amount) = rewards.last().unwrap();
```
sveitser, line 3560: I think we shouldn't depend on the order in which rewards are returned. I don't think it should be a requirements of the the compute_rewards function to put the leader reward last. IIUC currently if the leader is also a delegator we will have 2 entries for rewards that go to the leader.

sveitser: How about adding a type for rewards instead of using a `Vec`? With public methods like `leader_reward` that are safe to use. I think the vec is easy to misuse. Like it's tempting to assume that it only contains each address once but it won't if the leader is also a delegator.

imabdulbasit: This vec actually contains two entries for leaders. one is for commission reward and other the delegation reward. I just wanted the commission one so that seemed the easiest in this test. I don't know how to differentiate between these two entries. Maybe I should let compute_rewards return commission separately as well

sveitser: I mean if we have our own type we can add a function to it to get the leader reward.

sveitser: I just think we should encapsulate this logic rather than having an implicit dependency on what order things are put in a vec and I think this is most easily solved by creating a new type that we return from compute rewards instead of using a vec.

sveitser (approved):
LGTM. I do really think we should add a types for rewards and stop using the Vec.

sequencer/src/api.rs
imabdulbasit: ok I'll add it

types/src/v0/impls/reward.rs
```diff
@@ -314,6 +314,30 @@ impl From<(RewardAccountProof, U256)> for RewardAccountQueryData {
...
+
+impl ComputedRewards {
+    pub fn new(delegators: Vec<(Address, RewardAmount)>, commission: RewardAmount) -> Self {
+        Self {
+            delegators,
+            commission,
+        }
+    }
+
+    pub fn leader_commission(&self) -> &RewardAmount {
```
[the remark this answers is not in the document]
imabdulbasit: leader reward would be sum of commission and how much its staking? the staking reward is included in the delegators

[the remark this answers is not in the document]
imabdulbasit: this function only returns the leader reward that is coming from commission

sequencer/src/api.rs
```diff
@@ -3389,6 +3393,221 @@ mod test {
...
+
+            // Verify that the leader commission amount is within the tolerated range.
+            // Due to potential rounding errors in decimal calculations for delegator rewards,
+            // the actual distributed commission
+            // amount may differ very slightly from the calculated value.
+            // this asserts that it is within 10wei tolerance level.
+            // 10 wei is 10* 10E-18
+            let total_reward = block_reward().0;
+            let leader_commission_basis_points = U256::from(leader_validator.commission);
+            let calculated_leader_reward_amount = leader_commission_basis_points
```
sveitser, line 3563: @imabdulbasit I guess I adapted the terminology from earlier reviews. Here we use `leader_reward` to mean the reward from commission only.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3249/commits/315c405c2747a6771cc0101ebd4e7ca3286f81e4

PR #3252, opened 2025-05-09
feat: test the outcomes of random leader election

mrain:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/types/src/drb.rs
+++ b/crates/hotshot/types/src/drb.rs
@@ -230,3 +230,58 @@ pub mod election {
         drb: [u8; 32],
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        signature_key::BLSPubKey,
+        stake_table::StakeTableEntry,
+        traits::signature_key::{BuilderSignatureKey, StakeTableEntryType},
+    };
+    use alloy::primitives::U256;
+    use rand::RngCore;
+    use sha2::{Digest, Sha256};
+    use std::collections::HashMap;
+
+    use super::election::{generate_stake_cdf, select_randomized_leader};
+
+    #[test]
+    fn test_randomized_leader() {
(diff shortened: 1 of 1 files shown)


crates/hotshot/types/src/drb.rs
```diff
@@ -230,3 +230,59 @@ pub mod election {
...
+        let mut tvd = 0.;
+        let total_stakes = stake_table_entries
+            .iter()
+            .map(|e| e.stake())
+            .sum::<U256>()
+            .to::<u64>() as f64;
+        for entry in stake_table_entries {
+            let expected = entry.stake().to::<u64>() as f64 / total_stakes;
+            let actual = *selected.get(&entry).unwrap_or(&0) as f64 / num_views as f64;
+            tvd += (expected - actual).abs();
```
ss-es, line 282: I think for total variation distance this should be `tvd = max(tvd, (expected - actual).abs())`. the metric here is even more strict

crates/hotshot/types/src/drb.rs
```diff
@@ -230,3 +230,59 @@ pub mod election {
...
+        }
+
+        // Allow a small margin of error
+        assert!(tvd < 0.03);
```
ss-es, line 286: it's not strictly necessary but can we also `assert!(tvd > 0)` as a sanity check?

ss-es (approved):
minor comments but looks good to me

crates/hotshot/types/src/drb.rs
mrain: added

crates/hotshot/types/src/drb.rs
mrain: No TVD is defined to be the maximum difference of events, so it's the accumulated difference of pdf.

PR #3253, opened 2025-05-09
Remove double fetch for builder client claim_block_header_input

pls148:
Removes a redundant second call to builder `claim_block_header_input`. Previous behavior was to call twice, deserialize the first call as `AvailableBlockHeaderInputV2`, and the second as `AvailableBlockHeaderInputV2Legacy`. This change makes a single call happen, attempting to deserialize both types.

The old behavior would first try to validate_signature on `AvailableBlockHeaderInputV2`. As this struct is a subset of `AvailableBlockHeaderInputV2Legacy`, if a legacy response was encountered we'd always end up being able to pull out a `AvailableBlockHeaderInputV2`, and thus calling the non-legacy version's `validate_signature` first, even if the extra fields had been returned. This behavior is kept while removing the double call, by trying to deserialize the response as `AvailableBlockHeaderInputV2Legacy` but trying the non-legacy `validate_signature`.

I walked through the code at https://github.com/EspressoSystems/espresso-network/blob/3d031ec74c52d8ccb52eefd11ee6a396b86fcf96/hotshot-builder-api/src/v0_1/block_info.rs but couldn't prove that there wouldn't be a scenario where we built a signature without the vid_commitment, but still shipped the vid_commitment in the builder response, so this seemed like the safe way to handle this.

If we know that the above scenario cannot happen, then it's a trivial change to only try the legacy validation when we get the legacy fields.

---
  - https://app.asana.com/0/0/1210128488239109

### DIFF
--- a/crates/hotshot/builder-api/src/v0_1/block_info.rs
+++ b/crates/hotshot/builder-api/src/v0_1/block_info.rs
@@ -96,11 +96,10 @@ pub struct AvailableBlockHeaderInputV2Legacy<TYPES: NodeType> {
 /// either version of the AvailableBlockHeaderInputV2 type. Note that we try to deserialize legacy first,
 /// as that has extra fields that are not present in the current version. When presented with a legacy
 /// input, we'll first try to validate its signature as the current version.
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, Hash)]
-#[serde(untagged, bound = "")]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum AvailableBlockHeaderInputV2Either<TYPES: NodeType> {
-    Legacy(AvailableBlockHeaderInputV2Legacy<TYPES>),
     Current(AvailableBlockHeaderInputV2<TYPES>),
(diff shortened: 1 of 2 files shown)


crates/hotshot/builder-api/src/v0_1/block_info.rs
```diff
@@ -93,6 +93,16 @@ pub struct AvailableBlockHeaderInputV2Legacy<TYPES: NodeType> {
     pub sender: <TYPES as NodeType>::BuilderSignatureKey,
 }
 
+/// either version of the AvailableBlockHeaderInputV2 type. Note that we try to deserialize legacy first,
+/// as that has extra fields that are not present in the current version. When presented with a legacy
+/// input, we'll first try to validate its signature as the current version.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, Hash)]
```
ss-es, line 99: is `Serialize`/`Deserialize` actually required here? from my reading we never transmit this type

if so, can you remove the bounds? it's very confusing because it makes it seem like the PR is introducing a new serialized type (which would be a breaking/much larger change)

ss-es (approved):
I think this looks good to me

PR #3258, opened 2025-05-12
fixup after merging stale PR

sveitser:


imabdulbasit (approved):

sveitser:
I checked it compiles locally. Since CI is toast, I will bypass, this shouldn't make things any worse.

Issue #50, opened 2022-11-29
`State`: Implement `validate_block`


LaplaceFox:
Needed to satisfy `HotShotState` trait for `State`

- [x] `validate_block` needs to check that the parent state commitment of the block matches the current state commitment of the state.
- [x] check that view number is greater than current view number
- [x] check that the first block only contains a genesis transaction.
- [x] check that no genesis transaction (except for in the first block).

PR #3259, opened 2025-05-12
Separate entry for each stake table event in persistence

imabdulbasit:
This fixes an issue where multiple tasks can request different epoch stake tables. Having separate entries means we can return earlier epoch stake tables as well. persistence will be updated every time we fetch events from the contract and the fetched L1 block is higher than what is stored. 

For example, if a HotShot calls add_epoch_root() for epoch #50, we will have all events up to epoch 50. This allows us to return events for any epoch less than 50, such as when a merklized state loop requires the stake table for an earlier epoch.


sequencer/src/persistence.rs
```diff
@@ -1145,6 +1157,41 @@ mod persistence_tests {
             .is_err());
     }
 
+    async fn assert_events<P: TestablePersistence>(
```
sveitser, line 1160: nit: maybe say something about what we're asserting? by default in rust `assert!` just means checking that something is true.

```suggestion
    async fn assert_events_eq<P: TestablePersistence>(
```

sequencer/src/persistence.rs
```diff
@@ -1316,83 +1324,138 @@ mod persistence_tests {
...
+        // sleep so that we have enough events
+        sleep(Duration::from_secs(20)).await;
 
+        fetcher.spawn_update_loop().await;
+        let mut prev_l1_block = 0;
         for _i in 0..10 {
             // Wait for more than update interval to assert that persistence was updated
-            // L1 update interval is 5s for testing
-            tokio::time::sleep(std::time::Duration::from_secs(6)).await;
+            // L1 update interval is 2s in this test
```
sveitser, line 1434: Where is that `2s` set? The stake table update interval is 7s and the L1 block time is 1s.

sequencer/src/persistence.rs
```diff
@@ -1316,83 +1324,138 @@ mod persistence_tests {
...
-                .expect("failed to load events");
+                .expect("latest l1 block");
+
+            let (l1_block, events) = persistence.load_events(block).await?;
 
-            let (l1, events) = res.unwrap();
-            assert!(!events.is_empty());
+            tracing::info!("{l1_block:?}, persistence events = {events:?}.");
+            let l1_block = l1_block.unwrap();
+            assert!(l1_block > prev_l1_block, "events not updated");
```
sveitser, line 1446: Add an assert that we are actually getting new events?

sequencer/src/persistence.rs
```diff
@@ -1316,83 +1324,138 @@ mod persistence_tests {
...
 
-            let res = persistence
-                .load_events()
+            let block = anvil_provider
+                .get_block_number()
                 .await
-                .expect("failed to load events");
+                .expect("latest l1 block");
+
+            let (l1_block, events) = persistence.load_events(block).await?;
```
sveitser, line 1442:
```suggestion
            let (l1_block, persisted_events) = persistence.load_events(block).await?;
```

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
+            return Ok(());
+        }
+
         let mut inner = self.inner.write().await;
         let dir_path = &inner.stake_table_dir_path();
         let events_dir = dir_path.join("events");
 
         fs::create_dir_all(events_dir.clone()).context("failed to create events dir")?;
+        // Read the last l1 finalized for which events has been stored
+        let last_l1_finalized_path = events_dir.join("last_l1_finalized").with_extension("txt");
```
sveitser, line 1467: just a comment: I sometimes wonder why we use the `.txt` extension but store binary data.

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
+            bytes.as_slice().read_exact(&mut buf[..8])?;
+            let stored_l1 = u64::from_le_bytes(buf);
+            if stored_l1 > l1_block {
+                tracing::debug!(?stored_l1, ?l1_block, "stored l1 is greater");
+                return Ok(());
+            }
+        }
 
-        let file_path = events_dir.with_extension("txt");
+        for (event_key, event) in events {
```
sveitser, line 1481: Here we are potentially putting an unbounded large number of files into the same directory. I think this can cause perf issues, for example when listing all files in the directory. Do you know if we do that anywhere? Pruning maybe?

We're sort of deprecating file system storage but I think it's worth leaving a comment, in case we do not mange to remove it before this goes to mainnet.

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
+        // Determine the L1 block for querying events.
+        // If the last stored L1 block is greater than the requested block, limit the query to the requested block.
+        // Otherwise, query up to the last stored block.
+        let query_l1_block = if last_l1 > l1_block {
+            l1_block
+        } else {
+            last_l1
+        };
+
+        for entry in fs::read_dir(&events_dir).context("events directory")? {
```
sveitser, line 1545: Okay for example here we list all entries.

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
 
-        let file = File::open(file_path)?;
-        let reader = BufReader::new(file);
-        let (l1_block, events) = serde_json::from_reader(reader)?;
-        Ok(Some((l1_block, events)))
+        let mut events = Vec::new();
+
+        let bytes = fs::read(&last_l1_finalized_path).context("read")?;
+        let mut buf = [0; 8];
+        bytes.as_slice().read_exact(&mut buf[..8])?;
```
sveitser, line 1532: add some context to error so it's easier to tell what's wrong? also the one above just says "read" but not which file I think it may be easier to debug if we could put the path in the error

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
             },
         )
     }
 
-    async fn load_events(&self) -> anyhow::Result<Option<(u64, Vec<(EventKey, StakeTableEvent)>)>> {
-        let inner = self.inner.write().await;
-        let dir_path = &inner.stake_table_dir_path();
+    async fn load_events(
+        &self,
+        l1_block: u64,
```
sveitser, line 1514: Maybe add some information that this is the end block like "to_l1_block"

sequencer/src/persistence/fs.rs
```diff
@@ -1454,40 +1454,140 @@ impl MembershipPersistence for Persistence {
...
-        let reader = BufReader::new(file);
-        let (l1_block, events) = serde_json::from_reader(reader)?;
-        Ok(Some((l1_block, events)))
+        let mut events = Vec::new();
+
+        let bytes = fs::read(&last_l1_finalized_path).context("read")?;
+        let mut buf = [0; 8];
+        bytes.as_slice().read_exact(&mut buf[..8])?;
+
+        let last_l1 = u64::from_le_bytes(buf);
```
sveitser, line 1534: Something like "last_processed_l1_block"?

sequencer/src/persistence/sql.rs
```diff
@@ -2096,39 +2096,118 @@ impl MembershipPersistence for Persistence {
...
+            );
+            return Ok(());
+        }
+
+        let mut query_builder: sqlx::QueryBuilder<Db> =
+            sqlx::QueryBuilder::new("INSERT INTO stake_table_events (l1_block, log_index, event) ");
+
+        let events = events
+            .into_iter()
+            .map(|((b, i), e)| {
```
sveitser, line 2134:
```suggestion
            .map(|((block, index), event)| {
```

sequencer/api/migrations/sqlite/V602__redefine_stake_table_events.sql
```diff
@@ -0,0 +1,17 @@
...
+  l1_block BIGINT NOT NULL,
+  log_index BIGINT NOT NULL,
+  event JSONB NOT NULL,
+  PRIMARY KEY (l1_block, log_index)
+);
+
+-- Tracks the last L1 block that has been finalized and whose events have been processed and stored.  
+-- This tracking is necessary to determine the starting point for fetching new contract events. 
+CREATE TABLE stake_table_events_l1_block (
+  id INTEGER PRIMARY KEY CHECK (id = 0),
```
sveitser, line 15: Is there a reason we use several different ways of doing singleton tables, e.g.

https://github.com/EspressoSystems/espresso-network/blob/ee934e1cba4dc7b84b770a74c8ebf79fced2a0fd/sequencer/api/migrations/sqlite/V203__next_epoch_qc.sql#L2

types/src/v0/impls/stake_table.rs
```diff
@@ -439,15 +439,23 @@ impl StakeTableFetcher {
...
-        let from_block = res
-            .as_ref()
-            .map(|(block, _)| block + 1)
-            .filter(|from| *from <= to_block); // Only use from_block if it's less than to_block
+        tracing::info!("loaded events from storage to_block={to_block:?}");
 
-        tracing::info!("loaded events from storage from_block={from_block:?}");
+        // No need to fetch from contract
+        // if persistence returns all the events that we need
+        if queried_l1 == Some(to_block) {
```
sveitser, line 449: Should this be `>=` ?

sveitser: I guess not because we already trim the block number in `load_events`. This confused me a bit. Especially since the variable here is called `queried_l1`. I think what we care about are 1) if we had all the results in storage and 2) if not up to what l1 block we previously fetched.

Maybe a type like

```rust
enum EventsPersistenceRead {
    Complete,
    UntilL1Block(u64)
}
```

would make this more clear?

sveitser (approved):
LGTM.

Most comments are nitpicky, if you agree with the points please address them.

I think the extra assert that we get new events in the test would be good to have. If the test breaks in some way that no new events are created then we should fail it because it's no longer testing everything we want to test.

The treatment of block numbers in `load_events` could be a bit clearer. So that when calling `load_events` we make it clear if result is complete or not and easy to determine what's missing if it's incomplete.

sequencer/src/persistence/fs.rs
imabdulbasit: yeah I don't think we can avoid this if we want separate entry  for each event

sequencer/api/migrations/sqlite/V602__redefine_stake_table_events.sql
imabdulbasit: There is no reason.  I can change it to the other way with id as bool, but I don't think it matters

sequencer/src/persistence/fs.rs
imabdulbasit: changed it to bin extension, I think that makes more sense

sequencer/api/migrations/sqlite/V602__redefine_stake_table_events.sql
sveitser: It's just a bit easier to read if we're consistent.

PR #3263, opened 2025-05-13
feat: add max_gas_price for prover service

alxiong:
Closes [Asana task](https://app.asana.com/1/1208976916964769/project/1209394409339229/task/1209939948731653?focus=true)

### This PR:

- add options to specify max gas price in gwei for prover service
- skip `sync_state()` (and subsequent submitting light client update) if the gas price is too high.

cc @sveitser @Sneh1999


.env
```diff
@@ -109,6 +109,7 @@ ESPRESSO_STATE_SIGNATURE_TOTAL_STAKE=5
 # Prover service
 ESPRESSO_PROVER_SERVICE_PORT=30050
 ESPRESSO_STATE_PROVER_UPDATE_INTERVAL=20s
+ESPRESSO_STATE_PROVER_MAX_GAS_PRICE_IN_GWEI="2.3"
```
alxiong, line 112: i set this number based on what i saw for the last month: https://etherscan.io/gastracker#chart_gasprice

hotshot-state-prover/src/service.rs
```diff
@@ -420,6 +422,24 @@ pub async fn sync_state<ApiVer: StaticVersionType>(
...
+            .await
+            .map_err(|e| ProverError::NetworkError(anyhow!("{e}")))?;
+        if cur_gas_price > max_gas_price {
+            tracing::debug!(
+                "Current gas price too high: cur={} gwei, max={} gwei",
+                format_units(cur_gas_price, "gwei")
+                    .map_err(|e| ProverError::Internal(format!("{e}")))?,
+                format_units(max_gas_price, "gwei")
+                    .map_err(|e| ProverError::Internal(format!("{e}")))?,
+            );
```
Sneh1999, line 438: I have a question here, so lets say that LightClient syncs after 12 hours. Currently the has prices are high, will that mean it wont sync for another 12 hours? Should we maybe add a retryInterval instead when there is an error?

alxiong: sure we can retry, but asking it to sleep for 10min is thread blocking? (since this prover service is a single service container). 

I tend to think the cron job ("retry logic") should live in some scripts instead of this prover service logic. 
@Ancient123 @babdor do you have opinions?

alxiong: thanks for the input, updated in [437dbb1](https://github.com/EspressoSystems/espresso-network/pull/3263/commits/437dbb1b2ecc905ede2bf52937fb05a8ce2ba56a)

now we return Err if gas price to high, so daemon mode would retry based on the original (configurable) `retry_interval`, and in the run_once mode, will log a static ERROR msg for easy filtering on datadog.

PR #3265, opened 2025-05-13
Proof of incorrect encoding

rob-maron:
## This PR
Implements an endpoint in the `availability` API which returns a proof of incorrect encoding (if applicable and with enough honest nodes). It does so by collecting the shares from other nodes for a time period (now 40s), verifying them, and then performing and returning the proof.

There are some minor changes to the request/response protocol to better allow for a request where we need `n` messages to satisfy it, as opposed to the first valid one.


mrain (approved):
LGTM!

PR #3278, opened 2025-05-14
feat: make stake_table_capacity configurable for sequencer nodes #3276

### LINKED ISSUE #3264: local demo: fix prover after epoch transition
The prover makes it through the first epoch upgrade but then signature verifications start to fail.

Once we fix this we should add asserts to the demo E2E tests to make sure the LC contract keeps making progress.

mrain:
Closes #3264

### DIFF
--- a/contracts/rust/gen-vk-contract/src/main.rs
+++ b/contracts/rust/gen-vk-contract/src/main.rs
@@ -8,7 +8,7 @@ use std::{fs::OpenOptions, io::Write, path::PathBuf, process::Command};
 use alloy::hex::ToHexExt;
 use clap::Parser;
 use hotshot_contract_adapter::sol_types::VerifyingKeySol;
-use hotshot_types::light_client::STAKE_TABLE_CAPACITY;
+use hotshot_types::light_client::DEFAULT_STAKE_TABLE_CAPACITY;
 use jf_pcs::prelude::UnivariateUniversalParams;
 
 #[derive(Parser)]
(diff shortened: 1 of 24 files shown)


crates/hotshot/types/src/light_client.rs
```diff
@@ -25,7 +25,7 @@ use crate::signature_key::BLSPubKey;
 
 /// Capacity of the stake table, used for light client
 /// TODO(Chengyu): this should be loaded from the sequencer config
```
sveitser, line 27: I guess you can remove the TODO comment?

contracts/rust/gen-vk-contract/src/main.rs
```diff
@@ -42,7 +42,7 @@ fn main() {
     let (_, vk) = if mock {
         hotshot_state_prover::preprocess(&srs, 10).expect("Circuit preprocess failed")
     } else {
-        hotshot_state_prover::preprocess(&srs, STAKE_TABLE_CAPACITY)
```
sveitser, line 45: @mrain @alxiong Does it save significant amounts of computation to set capacity 10 for the demo? Or why are we doing it?

sveitser (approved):

crates/hotshot/types/src/light_client.rs
mrain: yes!
https://github.com/EspressoSystems/espresso-network/pull/3278/commits/e005b62ca2f1668cf32d3d07c4f1ba2e683c7dfb

contracts/rust/gen-vk-contract/src/main.rs
mrain: I believe it brings down the proof generation time from ~2mins to <20s.

PR #3281, opened 2025-05-15
Add go-sdk

ImJeremyHe:
### This PR:
- Move [the go sdk](https://github.com/EspressoSystems/espresso-network-go) into this repo.

### Key places to review:
- sdks/go

---
  - https://app.asana.com/0/0/1210137380322309
  - https://app.asana.com/0/0/1210226183244011


sdks/crypto-helper/.cargo/config.toml
```diff
@@ -0,0 +1,11 @@
+[target.aarch64-apple-darwin]
```
sveitser, line 1: A comment to what this achieves or why it's necessary?

.github/workflows/build-crypto-helper.yml
```diff
@@ -0,0 +1,104 @@
...
+          if [ "$(uname -s)" != "Darwin" ]; then
+            export LD_LIBRARY_PATH="$PWD/target/lib:$LD_LIBRARY_PATH"
+          fi
+          mkdir -p target/lib && \
+          cargo build --release --locked --target ${{ matrix.target }} --manifest-path ./sdks/crypto-helper/Cargo.toml && \
+          if [ "$(uname -s)" == "Darwin" ]; then
+            cp ./target/${{ matrix.target }}/release/libespresso_crypto_helper.dylib target/lib/libespresso_crypto_helper-${{ matrix.target }}.dylib
+            shasum -a 256 target/lib/libespresso_crypto_helper-${{ matrix.target }}.dylib | awk '{print $1}' > target/lib/libespresso_crypto_helper-${{ matrix.target }}.dylib.sha256
+          fi
+          if [ "$(uname -s)" != "Darwin" ]; then
```
sveitser, line 63:
```suggestion
          else
```

sdks/go/go.mod
```diff
@@ -0,0 +1,48 @@
+module github.com/EspressoSystems/espresso-network-go
+
+go 1.22
```
sveitser, line 3: Nix shell is using 1.23

sveitser: actually that value is unused

.github/workflows/build-crypto-helper.yml
```diff
@@ -0,0 +1,104 @@
+name: Build Crypto Helper Library
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*.*.*"
```
sveitser, line 8: Wonder if we should use tags like `go-v1.2.3`?

sveitser (approved):
LGTM. I suggest someone from nitro team has a look at the go files but they are mostly copied from the previous repo.

sdks/crypto-helper/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+hotshot-types = { workspace = true }
+jf-crhf = { workspace = true }
+jf-merkle-tree = { workspace = true, features = ["std"] }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha2 = { workspace = true }
+tagged-base64 = { workspace = true }
+
+# https://tikv.github.io/doc/openssl/index.html
+openssl = { version = "0.10", features = ["vendored"] }
```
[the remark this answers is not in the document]
ImJeremyHe: I removed this and it still works. I guess we don't need it anymore

sdks/go/download/main.go
```diff
@@ -0,0 +1,228 @@
...
+		if os == "darwin" {
+			fileName = "x86_64-apple-darwin"
+		} else if os == "linux" {
+			fileName = "x86_64-unknown-linux-musl"
+		}
+	case "arm64":
+		if os == "darwin" {
+			fileName = "aarch64-apple-darwin"
+		} else if os == "linux" {
+			fileName = "aarch64-unknown-linux-musl"
```
zacshowa, line 212: Is there a case here where we don't actually assign a filename previx here if the OS is neither darwin or linux? I doubt anyone will be attempting to run this on TempleOS anytime soon, but should we also add a base case just so we don't silently fail to download the object correctly?

ImJeremyHe: Yes you are right. We should return error if file name is empty.

PR #3283, opened 2025-05-16
update rust 1.86 -> 1.87, fix clippy

sveitser:
- Update rust in flake.nix

Clippy fixes:

- remove a `must_use` that has no effect.
- replace some single case matches with `if let`
- use `Error::other` instead of `Error::new(ErrorKindOther, ...)`
- wrap `File` in `BufReader` to avoid infficient byte by byte read
- replace `distribution.iter().any(...)` with `.contains(...)`

### DIFF
--- a/crates/hotshot/orchestrator/src/lib.rs
+++ b/crates/hotshot/orchestrator/src/lib.rs
@@ -13,7 +13,7 @@ use std::{
     collections::{HashMap, HashSet},
     fs,
     fs::OpenOptions,
-    io::{self},
+    io,
     time::Duration,
 };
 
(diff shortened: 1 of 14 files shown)


crates/hotshot/orchestrator/src/lib.rs
```diff
@@ -13,7 +13,7 @@ use std::{
     collections::{HashMap, HashSet},
     fs,
     fs::OpenOptions,
-    io::{self, ErrorKind},
+    io::{self},
```
sveitser, line 16: There should be a lint for this

sequencer/src/persistence/fs.rs
```diff
@@ -509,7 +509,7 @@ impl Inner {
         if self.legacy_anchor_leaf_path().is_file() {
             // We may have an old version of storage, where there is just a single file for the
             // anchor leaf. Read it and return the contents.
-            let mut file = File::open(self.legacy_anchor_leaf_path())?;
+            let mut file = BufReader::new(File::open(self.legacy_anchor_leaf_path())?);
```
sveitser, line 512: @imabdulbasit FYI, this is https://rust-lang.github.io/rust-clippy/master/index.html#unbuffered_bytes

PR #3289, opened 2025-05-19
WIP: use StakeTableV2

sveitser:
I mainly want to test if the asana linking from github works

https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1210174253354698


contracts/rust/adapter/src/sol_types.rs
```diff
@@ -39,8 +39,9 @@ pub use crate::bindings::{
     lightclientv2mock::{self, LightClientV2Mock},
     plonkverifier::PlonkVerifier,
     plonkverifierv2::PlonkVerifierV2,
-    staketable::{
-        self, EdOnBN254::EdOnBN254Point as EdOnBN254PointSol, StakeTable,
+    staketable::StakeTable,
+    staketablev2::{
```
sveitser, line 43: @alxiong I thought it would be nice to add something in this file so that we can verify the signatures on the `ConsensusKeysUpdatedV2` and `ValidatorRegisteredV2` events. But so far I was not successful in converting the BLS sig from the contract in to the jellyfish type.

```rust
#[derive(thiserror::Error, Debug)]
pub enum StakeTableError {
    #[error("Schnorr signature invalid")]
    InvalidSchnorrSignature,
    #[error("BLS signature invalid")]
    InvalidBLSSignature,
}

impl ConsensusKeysUpdatedV2 {
    pub fn verify_signatures(&self) -> Result<(), StakeTableError> {
        let bls_vk: BLSPubKey = self.blsVK.into();
        let msg = self.account.abi_encode();
        let sig_sol: G1PointSol = self.blsSig.into(); 
        let sig_affine: Affine<EdwardsConfig> = sig_sol.into(); # Probably the wrong type
        let sig = jf_signature::bls_over_bn254::Signature { sigma: sig_affine };
        if !bls_vk.verify(&msg, &sig, CS_ID_BLS_BN254) {
           // error
        }
        // ... TODO verify schnorr
        Ok(())

    }
}

```

alxiong: one inconvenience in upstream jellyfish is that we didn't implement `VerKey::from_affine()`, we should add it in the future. right now, I'm just gonna use the more verbose deserialization.

alxiong: done in [0166bb0](https://github.com/EspressoSystems/espresso-network/pull/3289/commits/0166bb0b44bdf326c7f81e8848a8322a86c9ff0b) @sveitser

sveitser: Thanks @alxiong. I think I managed to do the Schnorr signature verification myself.

staking-cli/src/registration.rs
```diff
@@ -34,54 +43,95 @@ pub async fn register_validator(
...
+
+    let version = stake_table.getVersion().call().await?.majorVersion;
+    Ok(match version {
+        1 => {
+            // NOTE: forge bind chose the _1 suffix for this function
+            stake_table
+                .registerValidator_1(bls_vk, schnorr_vk, bls_sig.into(), commission.to_evm())
+                .send()
+                .await
+                .maybe_decode_revert::<StakeTableV2Errors>()?
```
alysiahuggins, line 60: would it be `StakeTableErrors` for v1?

staking-cli/src/deploy.rs
```diff
@@ -76,7 +74,7 @@ impl TestSystem {
         let token = EspToken::new(*token_proxy.address(), provider.clone());
 
         // `StakeTable.sol`
-        let stake_table_impl = StakeTable::deploy(provider.clone()).await?;
+        let stake_table_impl = StakeTableV2::deploy(provider.clone()).await?;
```
alysiahuggins, line 77: when this gets pushed to main, would the public-facing staking cli have these changes? if so, will we need to upgrade StakeTableV2 before merging this into main?

staking-cli/src/registration.rs
```diff
@@ -34,54 +43,95 @@ pub async fn register_validator(
...
+    let stake_table = StakeTableV2::new(stake_table_addr, &provider);
+    let (bls_vk, bls_sig) = prepare_bls_payload(&bls_key_pair, validator_address);
+    let (schnorr_vk, schnorr_sig) = prepare_schnorr_payload(&schnorr_key_pair, validator_address);
+
+    let version = stake_table.getVersion().call().await?.majorVersion;
+    Ok(match version {
+        1 => {
+            // NOTE: forge bind chose the _1 suffix for this function
+            stake_table
+                .registerValidator_1(bls_vk, schnorr_vk, bls_sig.into(), commission.to_evm())
```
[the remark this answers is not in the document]
sveitser: It seems be something sort of random. Maybe sorted by 4byte selector? I guess we could work around this by appending `v2` to the function names in the contract.

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -230,6 +259,42 @@ impl<'de> Deserialize<'de> for ValidatorRegistered {
     }
 }
 
+impl Serialize for ValidatorRegisteredV2 {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
alxiong, line 263: why are we adding on this? so `.abi_encode()` is not sufficient?

sveitser: We store the events locally in the DB for that I think they are wrapped into an enum. 

https://github.com/EspressoSystems/espresso-network/blob/7617c2204f25ae9dec89f6418ee8429133a31cbc/types/src/v0/v0_3/stake_table.rs#L96-L105

It would  probably also work with abi_encode / decode but to change now we could  have to migrate storage of any nodes that have this deployed already. We would still have to implement serde I think to work with our storage code.

Correct me if I'm wrong @imabdulbasit .

staking-cli/src/deploy.rs
sveitser: At this line here it's only affecting the staking-cli tests. My goal is to write staking-cli and GCL code so it can all work with both versions of the contract.

staking-cli/src/registration.rs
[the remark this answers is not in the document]
sveitser: I think the pragmatic solution is to just append V2 to the new function names, then forge bind doesn't have to apply this workaround for the lack of function overloading in rust.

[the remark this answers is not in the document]
sveitser: afcb93c76c2efd10f78d3f3264a33a0f2ff6a0cd

contracts/rust/deployer/src/lib.rs
```diff
@@ -1105,8 +1167,48 @@ mod tests {
...
+        )
+        .await?;
+        let stake_table = StakeTable::new(stake_table_addr, &provider);
+
+        // upgrade to v2
+        upgrade_stake_table_v2(&provider, &mut contracts).await?;
+
+        assert_eq!(stake_table.owner().call().await?._0, owner);
+        assert_eq!(stake_table.token().call().await?._0, token_addr);
+        assert_eq!(stake_table.lightClient().call().await?._0, lc_addr);
```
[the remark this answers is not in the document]
sveitser: 05e75117baea65883900a8df9b6b1d55b5f51b2e

contracts/src/StakeTableV2.sol
```diff
@@ -9,18 +9,38 @@ import { EdOnBN254 } from "./libraries/EdOnBn254.sol";
...
     /// @notice A validator is registered in the stake table
     /// @notice the blsSig and schnorrSig are validated by the Espresso Network
     event ValidatorRegisteredV2(
         address indexed account,
-        BN254.G2Point blsVk,
-        EdOnBN254.EdOnBN254Point schnorrVk,
+        BN254.G2Point blsVK,
+        EdOnBN254.EdOnBN254Point schnorrVK,
         uint16 commission,
         BN254.G1Point blsSig,
```
alysiahuggins, line 42: since we made schnorrSig bytes, any chance it makes sense to do so for blsSig too? cc @alxiong

contracts/rust/adapter/src/copy.rs
```diff
@@ -0,0 +1,12 @@
+// The bindings types are small and pure data, there is no reason they
+// shouldn't be Copy. However some of them do have a bytes field which cannot be Copy.
+impl Copy for crate::sol_types::G1PointSol {}
+impl Copy for crate::sol_types::G2PointSol {}
+impl Copy for crate::sol_types::EdOnBN254PointSol {}
+impl Copy for crate::sol_types::StakeTableV2::ValidatorRegistered {}
```
alysiahuggins, line 6: not totally sure but would this also need to be called ValidatorRegisteredV2?

contracts/rust/adapter/src/stake_table.rs
```diff
@@ -24,10 +59,120 @@ impl From<EdOnBN254PointSol> for StateVerKey {
...
+        BLSPubKey::deserialize_uncompressed(&ser_bytes[..]).unwrap()
+    };
+    let msg = account.abi_encode();
+    let sig = {
+        let sigma_affine: ark_bn254::G1Affine = bls_sig.into();
+        BLSSignature {
+            sigma: sigma_affine.into_group(),
+        }
+    };
+    if !bls_vk.validate(&sig, &msg) {
```
alysiahuggins, line 115: I might have missed it but do we have a test for bls vk validation against its sigs?

alysiahuggins (approved):
lgtm with a few clarifying qs

contracts/rust/adapter/src/stake_table.rs
sveitser: You mean unittests? I didn't add them because @alxiong wrote it, so it should be correct :joy: added some unittests here: 985ce8d18874437d80d2fae2014f3ef84dbc35bc

contracts/src/StakeTableV2.sol
sveitser: But IIRC we are missing some code so if we don't use bytes things would have gotten more complicated.

contracts/rust/adapter/src/stake_table.rs
```diff
@@ -24,10 +59,133 @@ impl From<EdOnBN254PointSol> for StateVerKey {
...
+    Ok(())
+}
+
+/// Stake-table related error from the contract
+#[derive(Debug, thiserror::Error)]
+pub enum StakeTableSolError {
+    #[error("Failed to deserialize Schnorr signature")]
+    SchnorrSigDeserializationError(#[from] ark_serialize::SerializationError),
+    #[error("v1 err: {0:?}")]
+    V1(StakeTable::StakeTableErrors),
```
sveitser, line 141: This is never used. Will remove.

contracts/src/StakeTableV2.sol
alxiong: personally I like `bytes` because it's more future proof. but whenever we want to do something with these objects in the contract, we should consider giving them a concrete types. 

In this case, we do verify the BLS signature, thus giving it a type is better IMO.

I hear Mathis's complain about no `Copy` for bytes, but my personal tolerance for `Clone` is higher and I can live with lots of `.clone()` everywhere, (clippy will inform us about unnecessary clone once alloy bindings decide to add Copy on them)

staking-cli/src/registration.rs
```diff
@@ -34,60 +43,105 @@ pub async fn register_validator(
...
+    // There is a race-condition here if the contract is upgraded while this transactions is waiting
+    // to be mined. We're very unlikely to hit this in practice, and since we only perform the upgrade
+    // on decaf this is acceptable.
+    Ok(match version {
+        StakeTableContractVersion::V1 => {
+            stake_table
+                .registerValidator(bls_vk, schnorr_vk, bls_sig.into(), commission.to_evm())
+                .send()
+                .await
+                .maybe_decode_revert::<StakeTableV2Errors>()?
```
alxiong, line 63: Should we decode as V2Error here?

staking-cli/src/registration.rs
```diff
@@ -34,60 +43,105 @@ pub async fn register_validator(
...
-    let (bls_vk_sol, sig_sol) = prepare_bls_payload(&bls_key_pair, validator_address);
-    let schnorr_vk_sol: EdOnBN254PointSol = schnorr_vk.to_affine().into();
-    Ok(stake_table
-        .updateConsensusKeys(bls_vk_sol, schnorr_vk_sol, sig_sol.into())
-        .send()
-        .await
-        .maybe_decode_revert::<StakeTableErrors>()?
-        .get_receipt()
-        .await?)
+    let stake_table = StakeTableV2::new(stake_table_addr, &provider);
```
alxiong, line 92: Can we add some comments that since V2 inherits v1, all v1 methods are available and thus the "get_version" layer still make senses, this instantiation won't fail even if the address is a v1 stake table

staking-cli/tests/cli.rs
```diff
@@ -8,12 +8,20 @@ use alloy::primitives::{
...
 use sequencer_utils::test_utils::setup_test;
-use staking_cli::{demo::DelegationConfig, deploy::Signer, *};
+use staking_cli::{demo::DelegationConfig, deploy, deploy::Signer, Config};
 
 use crate::deploy::TestSystem;
 
+#[rstest_reuse::template]
+#[rstest::rstest]
+#[case::v1(StakeTableContractVersion::V1)]
+#[case::v2(StakeTableContractVersion::V1)]
```
alxiong, line 21: Are they supposed to be the same?

sveitser: uff, no, thanks

staking-cli/src/registration.rs
sveitser: See the commend below, I think it's fine to decode as V2 errors because the V2 ABI is a superset.

PR #3292, opened 2025-05-19
feat: proof of incorrect encoding

mrain:
Closes #<ISSUE_NUMBER>

### Key places to review:

---
  - https://app.asana.com/0/0/1210239975254098

### DIFF
--- a/types/src/v0/nsproof.rs
+++ b/types/src/v0/nsproof.rs
@@ -39,7 +39,7 @@ impl NsProof {
         }
     }
 
-    pub fn v11_new_with_correct_encoding(
+    pub fn v1_1_new_with_correct_encoding(
         payload: &Payload,
         index: &NsIndex,
         common: &VidCommon,
@@ -52,7 +52,7 @@ impl NsProof {
         }
     }
 
-    pub fn v11_new_with_incorrect_encoding(
+    pub fn v1_1_new_with_incorrect_encoding(
         shares: &[AvidMShare],
         ns_table: &NsTable,
         index: &NsIndex,
@@ -66,8 +66,6 @@ impl NsProof {
             _ => None,
         }
     }
-
-    // pub fn v11_new_with_incorrect_encoding(
 
     pub fn verify(
         &self,
--- a/types/src/v0/nsproof.rs
+++ b/types/src/v0/nsproof.rs
@@ -1,5 +1,5 @@
 use hotshot_query_service::VidCommon;
-use hotshot_types::data::VidCommitment;
+use hotshot_types::{data::VidCommitment, vid::avidm::AvidMShare};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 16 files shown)


rob-maron (approved):

vid/src/avid_m/proofs.rs
```diff
@@ -82,35 +84,39 @@ impl AvidMScheme {
                     break;
                 }
             }
+            if raw_shares.len() >= param.recovery_threshold {
```
alxiong, line 87: if we want "Exactly threshold shares" then maybe we just mark `==` here, and also add a `debug_assert!()` (or a `tracing::error!()`) at the end of the loop for sanity check?

vid/src/avid_m/proofs.rs
```diff
@@ -127,6 +133,107 @@ impl MalEncodingProof {
...
+        param: &AvidMParam,
+        commit: &NsAvidMCommit,
+        shares: &[NsAvidMShare],
+    ) -> VidResult<NsAvidMBadEncodingProof> {
+        if shares.is_empty() {
+            return Err(VidError::InsufficientShares);
+        }
+        for ns_index in 0..shares[0].ns_commits().len() {
+            let result =
+                Self::proof_of_incorrect_encoding_for_namespace(param, ns_index, commit, shares);
```
alxiong, line 200: when will someone use this method? it feels strange to shares of a specific namespace, but not specifying the namespace index, and ask this method to enumerate.

vid/src/avid_m/proofs.rs
```diff
@@ -127,6 +133,107 @@ impl MalEncodingProof {
...
+        if shares.iter().any(|share| !share.contains_ns(ns_index)) {
+            return Err(VidError::IndexOutOfBound);
+        }
+        let mt = MerkleTree::from_elems(
+            None,
+            shares[0].ns_commits().iter().map(|commit| commit.commit),
+        )?;
+        if mt.commitment() != commit.commit {
+            return Err(VidError::InvalidShare);
+        }
```
alxiong, line 171: this is more of `InvalidParam`? indicating that namespace tree's commitment passed in as input argument is wrong, right?

vid/src/avid_m/proofs.rs
```diff
@@ -142,6 +249,7 @@ pub struct NsProof {
 
 impl NsAvidMScheme {
     /// Generate a proof of inclusion for a namespace payload.
+    /// WARN: for the current implementation, no proof can be generated if any namespace is malformed.
     pub fn namespace_proof(
```
alxiong, line 253: can you elaborate on this more? why is this acceptable? 
(we can discuss offline and not blocking this PR)

vid/src/avid_m/proofs.rs
mrain: there's another check just at the end of the loop

vid/src/avid_m/proofs.rs
mrain: There's a problem about current implementation. If any namespace within this block is malformed, you cannot generate a good namespace proof for any namespace. So if you failed generating the namespace proof you don't know which namespace is the cause. You are forced to use this method tho.

vid/src/avid_m/proofs.rs
```diff
@@ -127,6 +133,107 @@ impl MalEncodingProof {
...
+        &self,
+        param: &AvidMParam,
+        commit: &NsAvidMCommit,
+    ) -> VidResult<VerificationResult> {
+        if MerkleTree::verify(
+            &commit.commit,
+            self.ns_index as u64,
+            &self.ns_commit.commit,
+            &self.ns_mt_proof,
+        )?
```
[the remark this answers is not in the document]
mrain: I agree that it looks nice. However currently `NsAvidMCommit` is just a type alias for `AvidMCommit` and it'll cause confusion. Sorry for the bad type design 😢

vid/src/avid_m/proofs.rs
mrain: you are right. [9d8d35c](https://github.com/EspressoSystems/espresso-network/pull/3292/commits/9d8d35c1ddfe1507fae08b06cfe1f5319ee65661)

types/src/v0/nsproof.rs
```diff
@@ -35,6 +39,36 @@ impl NsProof {
         }
     }
 
+    pub fn v11_new_with_correct_encoding(
```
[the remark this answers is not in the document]
mrain: [9d8d35c](https://github.com/EspressoSystems/espresso-network/pull/3292/commits/9d8d35c1ddfe1507fae08b06cfe1f5319ee65661)

alxiong (approved):
vid crypto part LGTM

PR #3294, opened 2025-05-20
Remove unnecessary `dead_code` cfg

c410-f3r:
`ChallengesSol` is already public, as such, `pub fn dummy` doesn't need a `dead_code` flag.


c410-f3r:
Let me know if anything is missing

c410-f3r:
Closing due to the lack of feedback

PR #3303, opened 2025-05-22
Pre-epoch membership uses static leaders

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### Key places to review:

---
  - https://app.asana.com/0/0/1210350359469183


types/src/v0/impls/stake_table.rs
```diff
@@ -1137,11 +1137,11 @@ impl Membership<SeqTypes> for EpochCommittees {
         view_number: <SeqTypes as NodeType>::View,
         epoch: Option<Epoch>,
     ) -> Result<PubKey, Self::Error> {
-        if let Some(epoch) = epoch {
-            let Some(randomized_committee) = self.randomized_committees.get(&epoch) else {
+        if self.first_epoch.is_some() && epoch.is_some_and(|e| e >= self.first_epoch.unwrap()) {
```
ss-es, line 1140: I'm not sure if I agree with this change

the effect is now, if we e.g. have first_epoch = 1000, and we query for `membership.leader(view_number: 10, epoch: Some(1))`, we will actually get an answer. To be fair the answer will be correct, but the query is wrong -- we should never be passing `epoch: Some(1)`

I think the bigger problem is that -- I think there is only really ambiguity during the first epoch (where we use both the randomized & non-randomized stake tables), but this doesn't change the behavior for the first epoch at all!

ss-es: I'm confused though -- is `membership_for_epoch(epoch)` being called with `epoch = Some(epoch_number)` where `epoch_number < first_epoch`? in that case it's being called with the wrong thing and we need to figure out where that number is set and fix it

ss-es: does it come from the `ViewChange` event? I don't see why we couldn't just make the event broadcast `Some(first_epoch)` instead

ss-es: Yes, I think I agree with making `has_randomized_stake_table` consistent. 

So instead of the change in this PR, we would just make that take an `Option<Epoch>` instead (and return an error when that fails)?

PR #3306, opened 2025-05-23
Upgrade st2 via multisig

alysiahuggins:
### This PR:
- Adds code to upgrade to stake table v2 where the owner is a multisig

### This PR does not:
- Actually perform the upgrade

### How to test this PR: 
```bash
cargo test -p espresso-contract-deploye test_upgrade_stake_table_to_v2_multisig_owner_dry_run
```
test demo deployment in docker
```bash
just pull
./scripts/build-docker-images-native --image deploy
just demo upgrade-sequencer-contracts-v2
```

---
  - https://app.asana.com/0/0/1210174253354697


contracts/rust/deployer/src/lib.rs
```diff
@@ -861,6 +911,93 @@ async fn upgrade_stake_table_v2(
...
+                tracing::error!("StakeTableProxy upgrade failed: {:?}", err);
+            } else {
+                tracing::info!("StakeTableProxy upgrade proposal sent");
+                // IDEA: add a function to wait for the proposal to be executed
+            }
+
+            match result {
+                Ok(r) => Ok(r),
+                Err(e) => Err(anyhow!("Upgrade proposal failed: {:?}", e)),
+            }
```
```suggestion
            result.context("Upgrade proposal failed")?
```

sveitser: I think this does the same here

sveitser: Maybe the ? isn't needed either.

sequencer/src/bin/deploy.rs
```diff
@@ -85,6 +85,18 @@ struct Options {
     )]
     multisig_address: Option<Address>,
 
+    /// Address for the multisig wallet that will be the pauser
+    ///
+    /// If provided, this the multisig wallet that will be able to upgrade contracts and execute
```
sveitser: This comment should be updated.

contracts/rust/deployer/src/builder.rs
```diff
@@ -235,19 +238,40 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
-                    crate::transfer_ownership(
+                let use_multisig = self.use_multisig;
+                let dry_run = self.dry_run;
+                tracing::info!(?dry_run, ?use_multisig, "Upgrading to StakeTableV2 with ");
+                if use_multisig {
+                    crate::upgrade_stake_table_v2_multisig_owner(
+                        provider,
+                        contracts,
+                        self.rpc_url.clone(),
+                        self.multisig.unwrap(),
```
sveitser: Are we checking anywhere that these unwraps won't fail? If not, it would be nice to have a descriptive error message instead of just a panic.

sveitser (approved):
LGTM. Just some suggestions.

sequencer/src/bin/deploy.rs
alysiahuggins: ah thank you, updated https://github.com/EspressoSystems/espresso-network/commit/7fc3e3dac250a3e854a4d1796e01a280b5cdc736

contracts/rust/deployer/src/builder.rs
alysiahuggins: good point, updated https://github.com/EspressoSystems/espresso-network/commit/7fc3e3dac250a3e854a4d1796e01a280b5cdc736

PR #3308, opened 2025-05-23
Handle epoch version startup

lukaszrzasik:
Closes [Use `first_epoch` when starting in epoch version](https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1210350359469196?focus=true) and [Use `first_epoch` when starting in epoch version](https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1210350359469196?focus=true)

### This PR:

* Broadcasts initial `ViewChange` event with `first_epoch` if it's present and we're in the epoch version
* ~~Populates epochs before `first_epoch` with the genesis stake table~~
* modifies `has_randomized_stake_table` so that it returns `Result<bool>`. Returns error if `first_epoch` is not set or `epoch < first_epoch`.
* modifies `lookup_leader` so that it return error `epoch` is Some and `first_epoch` is None or `epoch < first_epoch`
* Removes ambiguity when calling `EpochNumber::genesis`. It will always return 1 now.
* Removes two unnecessary calls to `membership_for_epoch`, substituted by `stake_table_for_epoch`

### Key places to review:

---
  - https://app.asana.com/0/0/1210350359469183

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -1381,20 +1381,14 @@ impl Membership<SeqTypes> for EpochCommittees {
             .insert(epoch, randomized_committee);
     }
 
-    fn set_first_epoch(&mut self, first_epoch: Epoch, initial_drb_result: DrbResult) {
-        self.first_epoch = Some(first_epoch);
+    fn set_first_epoch(&mut self, epoch: Epoch, initial_drb_result: DrbResult) {
+        self.first_epoch = Some(epoch);
 
-        let genesis_epoch = Epoch::genesis();
-        let genesis_epoch_committee = self.state.get(&genesis_epoch).unwrap().clone();
-        // Populate epochs from genesis + 1 to first_epoch + 1 with the genesis stake table
-        for epoch_number in *genesis_epoch..=*first_epoch {
-            let epoch = <SeqTypes as NodeType>::Epoch::new(epoch_number);
-            self.state
-                .insert(epoch + 1, genesis_epoch_committee.clone());
-        }
-        // Populate first_epoch and first_epoch + 1 with the initial DRB result
-        self.add_drb_result(first_epoch, initial_drb_result);
-        self.add_drb_result(first_epoch + 1, initial_drb_result);
+        let epoch_committee = self.state.get(&Epoch::genesis()).unwrap().clone();
+        self.state.insert(epoch, epoch_committee.clone());
+        self.state.insert(epoch + 1, epoch_committee);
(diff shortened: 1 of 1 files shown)


crates/hotshot/types/src/data/vid_disperse.rs
```diff
@@ -109,7 +109,7 @@ impl<TYPES: NodeType> ADVZDisperse<TYPES> {
         data_epoch: Option<TYPES::Epoch>,
     ) -> Result<Self> {
         let num_nodes = membership
-            .membership_for_epoch(target_epoch)
+            .stake_table_for_epoch(target_epoch)
```
ss-es, line 112: good catch!

types/src/v0/impls/stake_table.rs
```diff
@@ -1131,31 +1131,60 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+        match (self.first_epoch(), epoch) {
+            (Some(first_epoch), Some(epoch)) => {
+                if epoch < first_epoch {
+                    tracing::error!(
+                        "lookup_leader called with epoch {} before first epoch {}",
+                        epoch,
+                        first_epoch,
+                    );
+                    return Err(LeaderLookupError);
+                }
```
ss-es, line 1160: can this check not be implemented via `has_randomized_stake_table()`?

types/src/v0/impls/stake_table.rs
```diff
@@ -1338,14 +1381,20 @@ impl Membership<SeqTypes> for EpochCommittees {
...
-        self.add_drb_result(epoch, initial_drb_result);
-        self.add_drb_result(epoch + 1, initial_drb_result);
+        let genesis_epoch = Epoch::genesis();
+        let genesis_epoch_committee = self.state.get(&genesis_epoch).unwrap().clone();
+        // Populate epochs from genesis + 1 to first_epoch + 1 with the genesis stake table
+        for epoch_number in *genesis_epoch..=*first_epoch {
+            let epoch = <SeqTypes as NodeType>::Epoch::new(epoch_number);
+            self.state
+                .insert(epoch + 1, genesis_epoch_committee.clone());
+        }
```
ss-es, line 1394: I don't think this is correct semantically (unless I'm missing something)

epochs 0 up to the first epoch - 1 should not have a stake table of any kind, because those are not real epochs. the `Membership` object carries the initial/starting/pre-epoch stake table separately, outside of the <epoch, stake table> map

ss-es (changes requested):
looks fine overall, but I don't know if I agree with the `set_first_epoch` changes

PR #3309, opened 2025-05-23
feat: build individual docker images locally

sveitser:
- feat: support single image build with optimized compilation
- maintain backwards compatibility: without arguments we still build
  everything.
- feat: option to clean
- fix: avoid installing rust toolchain when starting cargo build in
  container
- feat: allow compiling with profiles other than --release, here the
  *default is changed* to use the new `test-docker` profile since
  release builds are very slow and rarely needed for local development. 
  The new profile is needed because without stripping the sequencer
  image is 4.5 GB instead of 1.5 GB.
- test: add test script to test some invocations of the build script.

### How to test

These scripts are currently not run on the CI. To see if it's working locally:

1. Run `scripts/build-docker-images-native --help` and try some invocations. 
1. Run `scripts/test-build-docker-images-native`, be aware it does a `clean` the end.

---
  - https://app.asana.com/0/0/1210463623862009

### DIFF
--- a/scripts/build-docker-images-native
+++ b/scripts/build-docker-images-native
@@ -235,8 +235,8 @@ cleanup(){
 }
 
 # Copy data files to Docker context.
-mkdir -p ${workdir}/data
-cp -rv data/genesis ${workdir}/data/
+mkdir -p "${workdir}/data"
+cp -rv data/genesis "${workdir}/data/"
 
 # find all the binaries built
 mapfile -t binaries < <(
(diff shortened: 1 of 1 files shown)


scripts/build-docker-images-native
```diff
@@ -1,13 +1,85 @@
...
+- The script supports building all images (the default) or one specific image.
+- By default all images are built with the `test-docker` cargo profile.
+
+Examples:
+
+  # build everything with default profile `test-docker`
+  scripts/build-docker-images-native
+
+  # build only the deploy image
+  scripts/build-docker-images-native --image deploy
```
alysiahuggins, line 22: 👐

tbro (approved):
Looks fine to me.

PR #3310, opened 2025-05-23
Move DRB difficulty to HotShotConfig

ss-es:


crates/hotshot/types/src/lib.rs
```diff
@@ -225,6 +225,8 @@ pub struct HotShotConfig<TYPES: NodeType> {
     /// Stake table capacity for light client use
     #[serde(default = "default_stake_table_capacity")]
     pub stake_table_capacity: usize,
+    /// number of iterations in the DRB calculation
+    pub drb_difficulty: u64,
```
rob-maron, line 229: Can we add the Decaf default here?

crates/hotshot/types/src/hotshot_config_file.rs
```diff
@@ -61,6 +61,8 @@ pub struct HotShotConfigFile<TYPES: NodeType> {
     /// Stake table capacity for light client use
     #[serde(default = "default_stake_table_capacity")]
     pub stake_table_capacity: usize,
+    /// number of iterations for DRB calculation
+    pub drb_difficulty: u64,
```
rob-maron, line 65: Can we add the Decaf default here?

sequencer/src/lib.rs
```diff
@@ -368,15 +368,18 @@ where
     }
 
     let epoch_height = genesis.epoch_height.unwrap_or_default();
+    let drb_difficulty = genesis.drb_difficulty.unwrap_or_default();
```
mrain, line 371: this `unwrap_or_default()` is going to get 0 if it's not set. should we do `unwrap_or(SOME_DEFAULT_VALUE)` instead?

rob-maron (approved):

PR #3312, opened 2025-05-23
Split up add_epoch_root to allow stake table storage to happen outsid…

pls148:
…e of read lock

Closes #<ISSUE_NUMBER>

### Key places to review:

---
  - https://app.asana.com/0/0/1210003190971090

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -18,7 +18,7 @@ use hotshot_task::dependency::{Dependency, EventDependency};
 use hotshot_types::{
     consensus::OuterConsensus,
     data::{Leaf2, QuorumProposalWrapper, VidDisperseShare, ViewChangeEvidence2},
-    drb::{DrbInput, DrbResult, DrbSeedInput},
+    drb::{DrbInput, DrbResult},
     epoch_membership::EpochMembershipCoordinator,
     event::{Event, EventType, LeafInfo},
     message::{Proposal, UpgradeLock},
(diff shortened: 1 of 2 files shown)


crates/hotshot/types/src/traits/election.rs
```diff
@@ -164,17 +164,44 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
...
+    ///
+    ///            // Return a callback to be executed under the write lock
+    ///            Some(Box::new(move |committee: &mut Self| {
+    ///                // Do some synchronous work here, updating the membership
+    ///            }) as Box<dyn FnOnce(&mut Self) + Send>)
+    ///        })
+    ///    }))
+    /// ```
     fn add_epoch_root(
         &self,
```
bfish713, line 189: Any chance we can just take a `RwLock<Self>` like in `get_epoch_root` and eliminate the need to have this function return this really complex return type?

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -222,16 +222,20 @@ async fn decide_epoch_root<TYPES: NodeType, I: NodeImplementation<TYPES>>(
...
-                .add_epoch_root(next_epoch_number, decided_leaf.block_header().clone())
-                .await
+            membership_reader.add_epoch_root(next_epoch_number, decided_leaf.block_header().clone())
         };
-        if let Some(write_callback) = write_callback {
-            let mut membership_writer = membership.write().await;
-            write_callback(&mut *membership_writer);
+
+        if let Some(worker) = add_epoch_root_worker {
+            let add_epoch_root_updater = worker().await;
```
bfish713, line 233: This still isn't moving this off the critical path.  We aren't holding the membership read lock anymore but it's still going to block the consensus task until `add_epoch_root_updater` completes.  The idea I had was that removing the requirement to lock the membership would allow us to `tokio::spawn` this and thus it consensus can continue while the stake table stuff happens in the background

bfish713 (changes requested):

bfish713 (approved):

PR #3319, opened 2025-05-26
state reconstruction test

imabdulbasit:


sequencer/src/api.rs
```diff
@@ -4108,4 +4113,464 @@ mod test {
...
+                }
+                .boxed()
+            })
+            .await
+            .unwrap();
+
+        let mut events = network.peers[2].event_stream().await;
+        while let Some(event) = events.next().await {
+            if let EventType::Decide { leaf_chain, .. } = event.event {
+                let height = leaf_chain[0].leaf.height();
```
tbro, line 4221: Epoch is also on the leaf. Can't we just use that directly instead of calculating it from the height?

sequencer/src/api.rs
```diff
@@ -4108,4 +4113,464 @@ mod test {
...
+            }))
+            .pos_hook::<PosVersion>(DelegationConfig::MultipleDelegators)
+            .await
+            .unwrap()
+            .build();
+        let state = config.states()[0].clone();
+        let mut network = TestNetwork::new(config, PosVersion::new()).await;
+        network.peers.remove(0);
+
+        let node_0_storage = &storage[1];
```
tbro, line 4174: Our persistence API is confusing to me, but this looks like we are copying persistence into the new node. Is there any danger that the new node will access data of `nodes[1]`? Also, is there anything special about `1`? We get the same result if we use of `0` or `3` correct?

imabdulbasit: so I dropped the peer0 which is using storage[1] then I am reinitializing the node with that same storage

sequencer/src/api.rs
```diff
@@ -4108,4 +4113,464 @@ mod test {
...
+                .boxed()
+            })
+            .await
+            .unwrap();
+
+        let client: Client<ServerError, SequencerApiVersion> =
+            Client::new(format!("http://localhost:{node_0_port}").parse().unwrap());
+        client.connect(None).await;
+
+        // assert that the restarted node has caught up
```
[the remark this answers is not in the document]
imabdulbasit: I think the second one was not needed so I removed it
https://github.com/EspressoSystems/espresso-network/pull/3319/commits/c02efdf8a6fff93bae75b6e275701c5e4d6e54c2

sequencer/src/api.rs
```diff
@@ -4108,4 +4113,464 @@ mod test {
...
+                        )
+                        .await)
+                }
+                .boxed()
+            })
+            .await
+            .unwrap();
+
+        let mut events = network.peers[2].event_stream().await;
+        while let Some(event) = events.next().await {
```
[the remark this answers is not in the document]
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3319/commits/c02efdf8a6fff93bae75b6e275701c5e4d6e54c2

sequencer/src/api.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3319/commits/c02efdf8a6fff93bae75b6e275701c5e4d6e54c2

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+
+        // check that the node has stored atleast 6 epochs merklized state in persistence
+        loop {
+            sleep(Duration::from_secs(1)).await;
+
+            let bh = client
+                .get::<u64>("block-state/block-height")
+                .send()
+                .await
+                .expect("block height not found");
```
tbro, line 4307: We are looping to wait for expected block height, but this will panic on first iteration if we haven't yet reached that height.

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+            .unwrap();
+
+        let client: Client<ServerError, SequencerApiVersion> =
+            Client::new(format!("http://localhost:{node_0_port}").parse().unwrap());
+        client.connect(None).await;
+
+        let epoch_6_block = EPOCH_HEIGHT * 6 + 1;
+
+        // check that the node's state has reward accounts
+        loop {
```
tbro, line 4279: This could potentially loop for ever, can we put a timeout? or max retries?

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+                .expect_ok()
+                .is_ok(),
+            "inconsistent block merkle tree"
+        );
+
+        for account in &fee_accounts {
+            state.fee_merkle_tree.lookup(account).expect_ok().unwrap();
+        }
+
+        // Reconstruct state using reward accounts only
```
tbro: This comment is a little confusing. We are just reconstructing reward state, correct?

sequencer/src/api.rs
tbro: So `network.peers.remove(0)` will drop peer at index `0`. But then we are using `storage[1]`. Isn't that the storage for peer at index `1`?

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+            tracing::info!("block state: block height={bh}");
+            if bh > epoch_6_block {
+                break;
+            }
+        }
+
+        // shutdown consensus to freeze the state
+        node_0.shutdown_consensus().await;
+        // wait few seconds to be sure that the node has shutdown all the tasks
+        sleep(Duration::from_secs(3)).await;
```
tbro: Seems fragile and likely to cause confusion if this takes longer than we expect. What if we listen to `internal_event_stream` on `SequencerContext.handle` and wait for `HotshotEvent::Shutdown`.

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+        let node_0 = opt
+            .clone()
+            .serve(|metrics, consumer, storage| {
+                let cfg = network.cfg.clone();
+                let node_0_persistence = node_0_persistence.clone();
+                let state = state.clone();
+                async move {
+                    Ok(cfg
+                        .init_node(
+                            1,
```
tbro, line 4416: I think this should be `0`.

sequencer/src/api.rs
```diff
@@ -4115,4 +4120,500 @@ mod test {
...
+            .clone()
+            .into_iter()
+            .map(|(acct, _)| acct)
+            .collect::<Vec<_>>();
+
+        let client: Client<ServerError, SequencerApiVersion> =
+            Client::new(format!("http://localhost:{node_0_port}").parse().unwrap());
+        client.connect(Some(Duration::from_secs(10))).await;
+
+        sleep(Duration::from_secs(5)).await;
```
tbro: What are we waiting for?  If it is the block height, we can use the event stream to wait instead.

sequencer/src/api.rs
imabdulbasit: storage[0] is for network.server

sequencer/src/api.rs
imabdulbasit: It would return 0 if there is no block height, so it shouldn't panic

sequencer/src/api.rs
imabdulbasit: 0 is for server.

sequencer/src/api.rs
imabdulbasit: is HotshotEvent::Shutdown emitted after all the tasks are canceled?

sequencer/src/api.rs
imabdulbasit: to be sure that all the transactions have been committed in query service. The node block height is connected to what we are passing in reconstruct_state(), because we delete proposals on every decide. I'll reduce it to 3s

sequencer/src/api.rs
imabdulbasit: I think I don't need this sleep because the shutdown_consensus() waits for the background tasks to cancel

PR #3321, opened 2025-05-27
Restart test state integrity

tbro:
Adds state to the restart test `TestNetwork` in order to validate that state is consistent among nodes and is progressing (has not been reverted to a previous state). This is accomplished  by not restarting *all* the nodes. That is, if one node has not restarted it can be used as a reference that restarted nodes have not reverted. The test basically works by comparing leaf commitments. If all nodes agrees with some other node about a commitment at a height, than all nodes are valid (even if not all nodes have knowledge of all the same commitments). 

To convince yourself this is working as expected run the restart tests with `RUST_LOG=info` and look for messages "verified leaf commitment" and "number of validated nodes".

These checks have been added to every test that uses `test_restart_helper`, which I believe in their totality covers the cases we want to cover (namely that a quorum of nodes cannot revert state). It also includes tests which restart *all* nodes, which will *not* address the specific case we are targeting, but also shouldn't cause any harm in my opinion. 

---
  - https://app.asana.com/0/0/1210003190971086


imabdulbasit (commented):
I might have missed something in standup but for validating the state integrity, we should also validate that restarted nodes validatedstate matches the one which did not restart?

PR #3328, opened 2025-05-28
feat: stake table contract pausability

alysiahuggins:
### This PR:
- adds StakeTableV3, extending StakeTableV2 with AccessControl and Pausable features
- Adds a PAUSER_ROLE and corresponding pause/unpause functions
- restricts critical  functions (delegate, undelegate, claimWithdrawal, claimWithdrawalExit, deregister) using `whenNotPaused` modifier 

### This PR does not:
- modify logic of existing stake/unstake flows outside of access control changes

### Key places to review:
- StakeTableV3
- StakeTableV3Test

### How to test this PR: 
```bash
forge test --match-contract StakeTableV3
```

### Things tested
Tested the V3 upgrade flow, role assignments, pausing/unpausing, that only pausers can pause, that paused functions revert, and that unpaused functions still work. Also verified storage layout compatibility across versions.

---
  - https://app.asana.com/0/0/1210253168280789

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -258,3 +258,12 @@ opt-level = 0
 opt-level = 0
 [profile.test.package.hotshot-state-prover]
 opt-level = 3
+
+# Geared towards building our docker images for local development
+# - Compile quickly
+# - Optimized enough to run demo
+# - Avoids huge binaries
+[profile.test-docker]
+inherits = "test"
+strip = true
+debug = "none"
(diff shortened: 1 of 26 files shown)


contracts/src/StakeTableV3.sol
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: MIT
```
sveitser: @alysiahuggins have we deployed the V2 to a public testnet already? If not, do we need the V3 or should we merge this with V2?

alysiahuggins: we haven't but separated it just incase we wanted to push v2 first before v3 - can merge if that's the preference though

sveitser: I think it's a bit easier for us to merge them so we don't have to distinguish between two stake table upgrades that have yet to happen. Both updates are pretty small in terms of API changes and change in functionality.

alysiahuggins: https://github.com/EspressoSystems/espresso-network/commit/1f81a84991d554ae09016659817bb8e37c27fcad

PR #3330, opened 2025-05-29
Update restart-tests to use  EpochVersion

tbro:
Updates restart tests to use `EpochVersion` and adds logic to deploy stake table contracts and delegate stake. In addition, restart tests will now wait for epoch 3 before testing for progress after restarts. This enforces that stake tables are actually working and in use by the network.  

For local testing make sure `anvil` is 2.1 or greater. On 2.0 blocks seemed slow to finalize. Also, some logs have been added at the `info` level, so if you want more information add `RUST_LOG=info`.

---
  - https://app.asana.com/0/0/1210134443040112


imabdulbasit (commented):
I think we should be able to restart without waiting for 3 epochs cc @sveitser . What if something goes wrong in the first two epochs because we're using a fixed stake table? I'm not sure, but it would be nice to have an option to wait for 3 epochs or not. Also, do you think it would be better to have an option to enable PoS or not?

imabdulbasit (approved):

PR #3336, opened 2025-06-02
Fix incorrect function name in Update modifyStateHistoryRetentionPeri…

fatwhorefrommd:
This PR fixes a critical bug in the ABI declaration used to encode the setStateHistoryRetentionPeriod function call. The original function name was misspelled as setstateHistoryRetentionPeriod (lowercase s), which caused ethers.Interface.encodeFunctionData to throw an error due to a missing match in the contract ABI.

Without this fix, the proposal transaction fails to encode properly, preventing it from being submitted and signed via the Safe Multisig interface.

### DIFF
--- a/contracts/script/multisigTransactionProposals/safeSDK/modifyStateHistoryRetentionPeriod.ts
+++ b/contracts/script/multisigTransactionProposals/safeSDK/modifyStateHistoryRetentionPeriod.ts
@@ -63,10 +63,10 @@ export async function proposeSetStateHistoryRetentionTransaction(
   stateHistoryRetentionPeriod: number,
 ) {
   // Define the ABI of the function to be called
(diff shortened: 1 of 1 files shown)


contracts/script/multisigTransactionProposals/safeSDK/modifyStateHistoryRetentionPeriod.ts
```diff
@@ -63,10 +63,10 @@ export async function proposeSetStateHistoryRetentionTransaction(
   stateHistoryRetentionPeriod: number,
 ) {
   // Define the ABI of the function to be called
-  const abi = ["function setstateHistoryRetentionPeriod(uint32)"];
```
sveitser, line 66: @alysiahuggins since the 4byte selector seems to be case sensitive in the function name this seems like a valid fix.

PR #3339, opened 2025-06-02
take initial token amount from the contract

imabdulbasit:


types/src/v0/impls/stake_table.rs
```diff
@@ -975,6 +981,80 @@ impl StakeTableFetcher {
...
+        let chain_config = *self.chain_config.lock().await;
+
+        let Some(stake_table_contract) = chain_config.stake_table_contract else {
+            bail!("No stake table contract address found in Chain config");
+        };
+
+        let provider = self.l1_client.provider.clone();
+        let stake_table = StakeTableV2::new(stake_table_contract, provider.clone());
+        let stake_table_init_block = stake_table
+            .initializedAtBlock()
```
sveitser, line 1014: Should query w.r.t. to the finalized block too.

types/src/v0/impls/stake_table.rs
```diff
@@ -975,6 +981,80 @@ impl StakeTableFetcher {
...
+
+        let token_address = stake_table
+            .token()
+            .block(BlockId::finalized())
+            .call()
+            .await
+            .context("Failed to get token address")?
+            ._0;
+
+        // In the transfer event, the token1 represents the address of the sender
```
sveitser:
```suggestion
        // In the transfer event, the topic1 represents the address of the sender
```

types/src/v0/impls/stake_table.rs
```diff
@@ -975,6 +981,80 @@ impl StakeTableFetcher {
...
+        // here the sender is the zero address as the tokens were minted
+        let token = EspToken::new(token_address, provider.clone());
+        let mut address_bytes = [0u8; 32];
+        address_bytes[12..].copy_from_slice(Address::ZERO.as_slice());
+        let topic1 = FixedBytes::<32>::from_slice(&address_bytes);
+
+        let mut from_block = stake_table_init_block.saturating_sub(5000u64);
```
sveitser: Can we use the max block range env var here?

Also I think we could first do a query without limits since that will work for many common RPC (infura, alchemy) and only if that fails try to scan. When we scan we should also inform via logging, and set some maximum number of block we look back because otherwise this might hang for a very, very long time.

sveitser: Also there should be a comment that says it's assuming there is only one mint event before the stake table is deployed and that this is safe because further minting is disabled until rewards withdrawal is implemented. Once we have reward withdrawals and minting there is some potential for things going horribly wrong because those will probably also be from address zero.

To make this safer maybe it would be better after all to search the Initialized event of the token contract because there should definitely only be one of those. Alternatively we can keep it as is but then get the transaction hash, fetch the receipt and try decoding an Initialized event. If we don't find that we should abort.

I think starting with the Initialized event of the token contract is safer though. Sorry for the back and forth.

types/src/v0/v0_1/state.rs
```diff
@@ -89,21 +89,11 @@ pub struct RewardAccount(pub Address);
...
-const TOTAL_SUPPLY: u128 = 10_000_000_000 * 10_u128.pow(18);
-const INFLATION_RATE: u128 = 300; // 3% in basis points
-const BLOCK_TIME_SECONDS: u128 = 2;
-const SECONDS_PER_YEAR: u128 = 60 * 60 * 24 * 365;
-const BLOCKS_PER_YEAR: u128 = SECONDS_PER_YEAR / BLOCK_TIME_SECONDS;
-const REWARD_PER_BLOCK: u128 =
-    ((TOTAL_SUPPLY * INFLATION_RATE) / BLOCKS_PER_YEAR) / COMMISSION_BASIS_POINTS as u128;
+ 
+pub(crate) const INFLATION_RATE: u128 = 300; // 3% in basis points
+pub(crate) const BLOCK_TIME_SECONDS: u128 = 2;
```
sveitser: Maybe `ASSUMED_BLOCK_TIME_SECONDS`? This isn't really the block time :D

types/src/v0/impls/stake_table.rs
sveitser: So to be more precise, my suggestion is to 

1. Find the initialized event
2. Get the TX hash of that transaction from the log
3. Get the receipt of the TX
4. Decode the Transfer event for the transaction.
5. Sanity check that the `from` address is zero

sequencer/api/node.toml
```diff
@@ -11,3 +11,7 @@ DOC = "Get the stake table for the given epoch"
 PATH = ["validators/:epoch_number"]
 ":epoch_number" = "Integer"
 DOC = "Get the validators map for the given epoch."
+
+[route.get_block_reward]
```
sveitser, line 15: Should this take an epoch number in case we want to update it? We can also do that later. I'm not quite sure what we need this API for hence my question.

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
         // Process the sorted events and return the resulting stake table.
         validators_from_l1_events(sorted.into_iter().map(|(_, e)| e))
     }
+    /// This function is used to calculate the reward for a block.
+    /// It fetches the initial supply from the token contract.
+    ///
+    /// Assumptions:
+    /// - This assumes only one mint event** occurred before the stake table was deployed.
```
sveitser: Update this comment, we don't make this assumption anymore since we implement the safer approach below. We should leave a comment about how checking the mint event is not a good idea though.

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+            .context("Failed to get token address")?
+            ._0;
+
+        let token = EspToken::new(token_address, provider.clone());
+
+        // Attempt to find Initialized event
+        let init_logs = token
+            .Initialized_filter()
+            .from_block(0u64)
+            .to_block(BlockNumberOrTag::Finalized)
```
sveitser, line 1042: Won't this fail due to rate limiting on some providers. I don't think we will get `None` in that case, I think we get an error, don't we?

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+            .query()
+            .await?;
+
```
sveitser: Add a sanity check that we only get a single event? It should never happen, so fine to panic.

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+        let initial_supply = if let Some((_, init_log)) = init_logs.first().cloned() {
+            // Get the transaction that emitted the Initialized event
+            let init_tx = provider
+                .get_transaction_receipt(init_log.transaction_hash.unwrap())
+                .await?
+                .context("Failed to get transaction for Initialized event")?;
```
sveitser: Add the tx hash to the log message?

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+
+            let mint_transfer = init_tx
+                .decoded_log::<EspToken::Transfer>()
+                .context("Failed to decode Transfer log")?;
```
sveitser: I don't think this will ever fail but it would also be better to log the input that failed decoding.

sequencer/api/node.toml
imabdulbasit: when we make other parameters configuratble, then we can add epoch number and this endpoint will be useful. I also think this endpoint doesn't do anything now

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+
+            tracing::debug!("mint transfer event ={mint_transfer:?}");
+            ensure!(
+                mint_transfer.from == Address::ZERO,
+                "First transfer should be a mint from the zero address"
+            );
+
+            mint_transfer.value
+        } else {
+            const MAX_BLOCKS_SCANNED: u64 = 200_000;
```
sveitser: As mentioned above, I think this needs to happen also if our first query for the initialized event returned an error. I think the return would be an empty list only if no event was found and in that case it doesn't make sense to scan for the log because the RPC thinks no such log exists. So in the case where we don't get a log I think we should return an error.

types/src/v0/impls/stake_table.rs
imabdulbasit: ok so we want to do the scanning if this fails?

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,136 @@ impl StakeTableFetcher {
...
+            let mut total_scanned = 0;
+
+            tracing::error!("No Initialized event found for fetching block reward");
+
+            // Fallback: manually search for mint transfer (from zero address)
```
sveitser: I think we should scan for the Initialized event, not the mint transfer. Then get the mint from the txn receipt, like we do in case we got the Initialized event without scanning.

types/src/v0/v0_3/stake_table.rs
```diff
@@ -65,7 +65,7 @@ pub type IndexedStake = (
 );
 
 #[derive(Clone, derive_more::derive::Debug)]
-pub struct StakeTableFetcher {
```
sveitser, line 68: Don't need to update this but I wonder why the change. If in the code we use as `stake_table::Fetcher` then yes, the new name is better, but if not, the old name is more descriptive and not overly verbose, isn't it?

imabdulbasit: it is also fetching block reward so I changed the name

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,137 @@ impl StakeTableFetcher {
...
+            tracing::debug!(tx_hash = ?init_log.transaction_hash, "Found token `Initialized` event");
+            init_log
+        } else {
+            const MAX_BLOCKS_SCANNED: u64 = 200_000;
+            let mut total_scanned = 0;
+
+            let mut from_block = stake_table_init_block.saturating_sub(max_events_range);
+            tracing::warn!("Initial token `Initialized` event not found. Scanning backwards from block {from_block}");
+
+            loop {
```
sveitser: This looks correct now. But I think we never execute the scanning code because the first call never fails. How about making a helper function for the scanning code so we can write a unittest for it?

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,137 @@ impl StakeTableFetcher {
...
+            .token()
+            .block(BlockId::finalized())
+            .call()
+            .await
+            .context("Failed to get token address")?
+            ._0;
+
+        let token = EspToken::new(token_address, provider.clone());
+
+        // Attempt to find Initialized event
```
sveitser:
```suggestion
        // Attempt to find Initialized event. Most RPCs will allow this query over all 
        // blocks because the result set is a single event. If we do get an error we 
        // will scan with fixed block ranges instead.
```

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,137 @@ impl StakeTableFetcher {
...
+            let (_, init_log) = init_logs[0].clone();
+
+            tracing::debug!(tx_hash = ?init_log.transaction_hash, "Found token `Initialized` event");
+            init_log
+        } else {
+            const MAX_BLOCKS_SCANNED: u64 = 200_000;
+            let mut total_scanned = 0;
+
+            let mut from_block = stake_table_init_block.saturating_sub(max_events_range);
+            tracing::warn!("Initial token `Initialized` event not found. Scanning backwards from block {from_block}");
```
sveitser: I think this can be INFO. Also IMO we shouldn't say "not found". The issue is that the RPC returned an error. So something like "INFO: RPC returned error: {err}, will fallback to scanning over fixed block range" would be more descriptive.

types/src/v0/v0_3/stake_table.rs
sveitser: I see.

types/src/v0/impls/stake_table.rs
imabdulbasit: warn level was intentional because it would be only called once, so its better to be loud to debug it if we got stuck somewhere in scan

types/src/v0/impls/stake_table.rs
```diff
@@ -974,6 +981,165 @@ impl StakeTableFetcher {
...
+                );
+                return Err(anyhow::anyhow!(
+                    "No Initialized event found after scanning {} blocks",
+                    MAX_BLOCKS_SCANNED
+                ));
+            }
+            let init_logs = token
+                .Initialized_filter()
+                .from_block(from_block)
+                .to_block(stake_table_init_block)
```
sveitser: This should be something like

```suggestion
                .to_block(from_block + max_events_range)
```

shouldn't it?

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3339/commits/8ab5a1b07144b13ca7f88c62c9f835ee372190a7

PR #3342, opened 2025-06-03
feat: legacy prover for v1 light client contract

mrain:
Closes #<ISSUE_NUMBER>

### This PR:

Brings back the legacy prover for light client contract v1
Basically duplicating all the components for convenience.

### This PR does not:
There are some leftover works in #3347 . In short, the sequencer node and state-relay are providing different signatures. We need to update those parts and have new endpoints to make the legacy fully functional.

### Key places to review:

---
  - https://app.asana.com/0/0/1210329661126931


crates/hotshot/types/src/traits/signature_key.rs
```diff
@@ -451,6 +451,19 @@ pub trait StateSignatureKey:
         next_stake_table_state: &StakeTableState,
     ) -> bool;
 
+    /// Sign the state for legacy light client
+    fn legacy_sign_state(
```
alxiong, line 455: Q: if we upgrade further in the future, current `sign_state()` will become `legacy_sign_state()`? 
basically, if we only keep two versions then this naming is perfectly fine, otherwise, call it `_v1`, `_v2` etc. are better

hotshot-state-prover/src/legacy/mod.rs
```diff
@@ -0,0 +1,12 @@
+//! Legacy prover
```
alxiong, line 1: in this module doc, specify the commit from which this `legacy/` prover is based from. also note the diff if any.

mrain: Out logic is directly modified from the existing impl. I did copy the circuit implementation from previous commit, doc added here: [6d98dd2](https://github.com/EspressoSystems/espresso-network/pull/3342/commits/6d98dd24bc8478c9d9b8670eccd5f67d9296c710)

crates/hotshot/types/src/traits/signature_key.rs
mrain: I think keeping the current state is fine. When another upgrade happens in the future, we can rename/reorg them there.

alxiong (approved):
LGTM!

PR #3344, opened 2025-06-03
CI: native docker builds

sveitser:
QEMU is very slow which limits us to what we can do in docker builds within reasonable time. We do want images to build quickly in order to deploy them. This PR builds native images on AMD and ARM architectures and then merges them into a multiplatform image.

Concretely in #3077 we need to add a typescript script to the deploy image and installing this takes ages with QEMU.

Since we use public arm runners now we could consider always doing the ARM and multiplatform builds. I don't have much of a preference here, it seems a bit wasteful but the workflow file would become a bit simpler, arm and amd builds would become almost the same and we could use a matrix for that too. If anyone has opinions here please weigh in. Edit: I would like to safe this for follow up work, this PR is already large, and annoying to test.

The workflow file contains some duplication, notably the github action matrix over services. I didn't want to create another job and we don't add binaries very often so I think dynamically outputting a matrix from a job isn't really worth it. Don't mind changing it tough.

### Follow up
- https://github.com/EspressoSystems/espresso-network/issues/3348

---
  - https://app.asana.com/0/0/1210452759586916

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,20 @@
 name: Build
-
+# This workflow file is pretty complex, due to optimizations and features:
+#   - using public runners
+#   - rust caching
+#   - native docker multiplatform builds (without QEMU)
+#
+# 1. We build the binaries in 4 jobs for each of the 2 architectures. The jobs are broken up so that
+#    we get a decent amount of parallelism and a low overall runtime.
+# 2. We build the native docker image for each of 2 architectures using the binary artifacts from
+#    the first step. We push the images to the registry by digest.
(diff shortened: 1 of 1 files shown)


sveitser:
Next run: https://github.com/EspressoSystems/espresso-network/actions/runs/15414641399

sveitser:
The digest artifacts all used the same name, leading to a name collision. Next run: https://github.com/EspressoSystems/espresso-network/actions/runs/15414914866

sveitser:
Previous build passed :partying_face: 

However artifact download on buildjet is really slow, trying the public github arm runners here: https://github.com/EspressoSystems/espresso-network/actions/runs/15415259894

PR #3352, opened 2025-06-03
fix: lookup_leader and inport SeedableRng Update randomized_committee…

fatwhorefrommd:
Description:
This pull request addresses two critical compilation errors in randomized_committee_members.rs:

Missing SeedableRng import
The code calls StdRng::seed_from_u64(...) but SeedableRng was not imported, resulting in an undefined trait.

Invalid dereference of view_number
Since view_number has type <TYPES as NodeType>::View (a newtype over u64), using *view_number does not compile. We must use .u64() to obtain a raw u64.

With these changes:

We import SeedableRng so that StdRng::seed_from_u64(...) is available.

We replace every *view_number with view_number.u64() to obtain the underlying u64.


rob-maron:
Hi, the code compiles fine

PR #3354, opened 2025-06-04
Improve logging in Espresso go multiclient

zacshowa:
Closes # Appchain investigation

### This PR:
In the short term, this PR is to assist with debugging the Appchain mainnet incident, In the long term. I would like to change the return signatures for some of the functions of the `MultipleNodesClient` with partial error information for adding debugging logs in consumer programs.

Adds some improved logging to the espresso go client. It also adds an error collection array to all methods that iterate over the individual clients for the multi node client to collect error messages.
This PR also modifies the tests to account for the modified return values.

### This PR does not:

Modify the client API such that the MultipleNodesClient will capture and return errors from specific nodes so that the Caller may report individual error messages in whatever logging framework they choose.

I can't think of a particularly great/non intrusive way to propagate errors to the caller. My intuition is to modify the return signatures to return an `[]error` value. This new return value would be different not be the `error` field that indicates the function call failed. Regardless to properly propagate the errors to the caller, the return signature will need to change to avoid breaking error semantics for consumers.

### Key places to review:
All 2 files of the PR.

### How to test this PR:

from the `/espresso-network/sdks/go` repo run `go test -run ^TestFetchWithMajority$ github.com/EspressoSystems/espresso-network/sdks/go/client`

### DIFF
--- a/sdks/go/client/multiple_nodes_client.go
+++ b/sdks/go/client/multiple_nodes_client.go
@@ -36,13 +36,17 @@ func NewMultipleNodesClient(urls []string) (*MultipleNodesClient, error) {
 }
 
 func (c *MultipleNodesClient) FetchLatestBlockHeight(ctx context.Context) (uint64, error) {
+	var errs []error
 	for _, node := range c.nodes {
 		height, err := node.FetchLatestBlockHeight(ctx)
 		if err == nil {
 			return height, nil
+		} else {
+			fmt.Printf("Encountered an error while attempting to fetch latest block height from one node. Error: %v\nNode: %v\n", err, node)
+			errs = append(errs, err)
 		}
 	}
-	return 0, errors.New("fetch latest block height failed with all nodes")
+	return 0, fmt.Errorf("fetch latest block height failed with all nodes, Errors: %v\n", errs)
 }
 
 func (c *MultipleNodesClient) FetchHeaderByHeight(ctx context.Context, height uint64) (types.HeaderImpl, error) {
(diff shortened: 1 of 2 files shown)


ImJeremyHe (approved):

ImJeremyHe:
[CodeReviewComments](https://go.dev/wiki/CodeReviewComments):

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

Can you fix the error strings?

sdks/go/client/multiple_nodes_client.go
```diff
@@ -36,13 +36,17 @@ func NewMultipleNodesClient(urls []string) (*MultipleNodesClient, error) {
 }
 
 func (c *MultipleNodesClient) FetchLatestBlockHeight(ctx context.Context) (uint64, error) {
+	var errs []error
 	for _, node := range c.nodes {
 		height, err := node.FetchLatestBlockHeight(ctx)
 		if err == nil {
 			return height, nil
+		} else {
+			fmt.Printf("Encountered an error while attempting to fetch latest block height from one node. Error: %v\nNode: %v\n", err, node)
```
Sneh1999: would we be even able to see this printf, I am not sure

sdks/go/client/multiple_nodes_client.go
```diff
@@ -129,13 +133,17 @@ func (c *MultipleNodesClient) FetchVidCommonByHeight(ctx context.Context, blockH
 
 func (c *MultipleNodesClient) SubmitTransaction(ctx context.Context, tx common.Transaction) (*common.TaggedBase64, error) {
 	// Check if one node is successfully able to submit the transaction
+	var errs []error
 	for _, node := range c.nodes {
 		hash, err := node.SubmitTransaction(ctx, tx)
 		if err == nil {
 			return hash, nil
+		} else {
+			fmt.Printf("Encountered an error while attempting to submit transaction with one node.\n Error: %v \n Node: %v \n", err, node)
```
Sneh1999: same here

sdks/go/client/multiple_nodes_client.go
```diff
@@ -184,10 +193,14 @@ func FetchWithMajority[T any](ctx context.Context, nodes []*T, fetchFunc func(*T
 					}
 
 				}
+			} else {
+				fmt.Printf("Encountered an error while attempting to fetch with majority.\n Error: %v \n", res.err)
+				errs = append(errs, res.err)
```
Sneh1999, line 200: should we also add `fmt.Printf("error: failed to normalize json value: %v, error: %v", res.value, err)` to the errors list?

sdks/go/client/multiple_nodes_client.go
zacshowa: I think we should be able to see them, but I'm not 100% sure. I think it makes the most sense to use some standard library logging functionality rather than the go-ethereum package if we are doing the logging in the client itself. 

That's why my ultimate goal would be to alter the function signature to return the intermediate errors in some array that doesn't represent the function call failing so that the caller can log them however they choose to.

Sneh1999 (approved):
LGTM! Thanks @zacshowa  🫡

PR #3356, opened 2025-06-04
DRB difficulty upgrade

ss-es:
- Adds a new field `drb_upgrade_difficulty` to `HotShotConfig` set via the genesis file (e.g. `decaf.toml`)
- Adds a new version `DrbDifficultyUpgrade` to the `Versions` trait
- Adds logic to spawn the DRB calculation with either the base or upgrade difficulty level depending on the version

There are some expectations I have that would be good to double-check:
- because of the way DRB checkpoint works (storing the `DrbInput` struct which itself carries the difficulty level), I expect there should be no subtleties with storing/loading partial DRB calculations
- the DRB difficulty selector is only added to `EpochMembership` when `start_consensus` is called. I reasoned that this was the most natural place to put this, but if you can think of any issues or have a better suggestion please let me know

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -57,7 +57,7 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
     qc: &QuorumCertificate2<TYPES>,
     event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
     event_receiver: Receiver<Arc<HotShotEvent<TYPES>>>,
-    membership_coordinator: EpochMembershipCoordinator<TYPES, V>,
+    membership_coordinator: EpochMembershipCoordinator<TYPES>,
     consensus: OuterConsensus<TYPES>,
     sender_public_key: TYPES::SignatureKey,
     sender_private_key: <TYPES::SignatureKey as SignatureKey>::PrivateKey,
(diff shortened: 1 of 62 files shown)


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -168,25 +168,38 @@ pub async fn handle_drb_result<TYPES: NodeType, I: NodeImplementation<TYPES>>(
...
     let membership = membership.clone();
     let storage = storage.clone();
     let store_drb_progress_fn = store_drb_progress_fn(storage.clone());
     let load_drb_progress_fn = load_drb_progress_fn(storage.clone());
-    let consensus = consensus.clone();
-    let difficulty_level = consensus.read().await.drb_difficulty;
+    let version = upgrade_lock.version_infallible(view).await;
+
+    let consensus_reader = consensus.read().await;
+    let difficulty_level = if version >= V::DrbDifficultyUpgrade::VERSION {
```
pls148: Would be more readable to have this move to UpgradeLock like UpgradeLock::epochs_enabled

bfish713 (approved):
I think this looks good, my one concern is that some parts of the code will trigger catchup (e.g. rewards calculation in query service catchup) before consensus is fully online.  Looks like the code will handle this, but we should explicitly look out for and test it.

crates/hotshot/task-impls/src/helpers.rs
bfish713: Agree, but this upgrade is going to contain more than just DRB, I Think we should just name it `is_version_0_4` or something

PR #3360, opened 2025-06-04
Add Metric for Proposal Latency

bfish713:
Closes #<ISSUE_NUMBER>

### This PR:

Adds a new metric for which is the number of seconds between proposal creation and finalization

It also moves the `timestamp` function to be part of the `BlockHeader` trait

### This PR does not:

This is not a high enough resolution metric for detailed metrics.

### Key places to review:

Changes in helpers

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -44,6 +44,7 @@ use hotshot_types::{
     vote::{Certificate, HasViewNumber},
 };
 use hotshot_utils::anytrace::*;
+use time::OffsetDateTime;
 use tokio::time::timeout;
 use tracing::instrument;
 use vbs::version::StaticVersionType;
@@ -303,6 +304,26 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
     }
 }
 
+async fn update_metrics<TYPES: NodeType>(
+    consensus: &OuterConsensus<TYPES>,
+    leaf_views: &[LeafInfo<TYPES>],
+) {
+    for leaf_view in leaf_views {
+        let consensus_reader = consensus.read().await;
+        let now = OffsetDateTime::now_utc().unix_timestamp() as u64;
+        let proposal_timestamp = leaf_view.leaf.block_header().timestamp();
+
+        let Some(proposal_to_decide_time) = now.checked_sub(proposal_timestamp) else {
+            tracing::error!("Failed to calculate proposal to decide time: {proposal_timestamp}");
+            continue;
+        };
+        consensus_reader
+            .metrics
+            .proposal_to_decide_time
+            .add_point(proposal_to_decide_time as f64);
+    }
+}
+
 /// calculate the new decided leaf chain based on the rules of HotStuff 2
 ///
 /// # Panics
@@ -413,6 +434,7 @@ pub async fn decide_from_proposal_2<TYPES: NodeType, I: NodeImplementation<TYPES
             )
             .await;
         }
+        update_metrics(&consensus, &res.leaf_views).await;
     }
 
     res
--- a/crates/hotshot/types/src/consensus.rs
+++ b/crates/hotshot/types/src/consensus.rs
@@ -394,6 +394,12 @@ pub struct ConsensusMetricsValue {
     pub number_of_empty_blocks_proposed: Box<dyn Counter>,
     /// Number of events in the hotshot event queue
     pub internal_event_queue_len: Box<dyn Gauge>,
+    /// Time from proposal creation to decide time
+    pub proposal_to_decide_time: Box<dyn Histogram>,
+    /// Time from proposal received to proposal creation
+    pub previous_proposal_to_proposal_time: Box<dyn Histogram>,
+    /// Finalized bytes per view
+    pub finalized_bytes: Box<dyn Histogram>,
 }
 
 impl ConsensusMetricsValue {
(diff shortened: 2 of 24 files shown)


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -303,6 +304,26 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
...
+        let proposal_timestamp = leaf_view.leaf.block_header().timestamp();
+
+        let Some(proposal_to_decide_time) = now.checked_sub(proposal_timestamp) else {
+            tracing::error!("Failed to calculate proposal to decide time: {proposal_timestamp}");
+            continue;
+        };
+        consensus_reader
+            .metrics
+            .proposal_to_decide_time
+            .add_point(proposal_to_decide_time as f64);
```
ss-es, line 324: would it be good to save more information here? e.g. index by view number, leader etc

ss-es: I'm not sure how the metrics work currently, I guess it might need a new e.g. `Graph` type or something to do that

hotshot-query-service/src/data_source.rs
```diff
@@ -1244,33 +1247,41 @@ pub mod node_tests {
...
-                    // itself or the last block if the window is nonempty), so we can `unwrap` here.
-                    assert!(next.timestamp() >= prev.unwrap().timestamp());
+        let check_invariants = |res: &TimeWindowQueryData<Header<MockTypes>>,
+                                start,
+                                end,
+                                check_prev| {
+            let mut prev = res.prev.as_ref();
+            if let Some(prev) = prev {
+                if check_prev {
+                    assert!(<TestBlockHeader as BlockHeader<MockTypes>>::timestamp(prev) < start);
```
ss-es: is this all just to use the new `timestamp` method on the block header?

ss-es (commented):
not sure why tests are failing, but overall the changes look fine to me -- just some minor comments

crates/hotshot/task-impls/src/helpers.rs
bfish713: Yeah I was trying to avoid adding a completely new metric type compared to what's already there.  The histogram should let us see each point on the datadog dashboard I think.  At least that was my hope

hotshot-query-service/src/data_source.rs
bfish713: yeah compiler didn't like it unless I sued the type here, I don't know if we can get around that

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -303,6 +304,26 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
     }
 }
 
+async fn update_metrics<TYPES: NodeType>(
+    consensus: &OuterConsensus<TYPES>,
+    leaf_views: &[LeafInfo<TYPES>],
+) {
+    for leaf_view in leaf_views {
+        let consensus_reader = consensus.read().await;
```
pls148: you're acquiring a read lock once every pass of this for loop just to do some trivial math. I'd move the readlock acq out of the for loop.

pls148: also lets you avoid having an async yield point once every loop iteration since that's the only async call in this fn

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -303,6 +304,26 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
...
+        let now = OffsetDateTime::now_utc().unix_timestamp() as u64;
```
pls148: I'd also hoist now() out of the for loop, recalculating it here just adds noise to the data

hotshot-query-service/src/data_source.rs
pls148: maybe clean it up with some helper function for testing like

```
    fn block_header_timestamp(header: &Header<MockTypes>) -> u64 {
        <TestBlockHeader as BlockHeader<MockTypes>>::timestamp(header)
    }
```

that'd be a lot more readable imo

crates/hotshot/types/src/consensus.rs
```diff
@@ -394,6 +394,12 @@ pub struct ConsensusMetricsValue {
     pub number_of_empty_blocks_proposed: Box<dyn Counter>,
     /// Number of events in the hotshot event queue
     pub internal_event_queue_len: Box<dyn Gauge>,
+    /// Time from proposal creation to decide time
```
ss-es, line 397: can you elaborate a bit more in the comments here? e.g. time between this node creating a proposal and reaching a decide on that proposal

I think the next one is also a bit unclear, I imagine it's the time between receiving the proposal from the preceding view to creating a new proposal (including forming the QC on it etc)

crates/hotshot/types/src/consensus.rs
```diff
@@ -394,6 +394,12 @@ pub struct ConsensusMetricsValue {
...
+    pub proposal_to_decide_time: Box<dyn Histogram>,
+    /// Time from proposal received to proposal creation
+    pub previous_proposal_to_proposal_time: Box<dyn Histogram>,
```
ss-es, line 400: is it intentional that this (and the next one) are not actually used? is that for a future PR?

crates/hotshot/types/src/consensus.rs
bfish713: Yeah exactly what you said.  It's more for diagnosing slow leaders and for accounting for the extra time not included in the above metric

crates/hotshot/types/src/consensus.rs
bfish713: Yes I already have a pr for finalized bytes and will do one for this metric today

PR #3361, opened 2025-06-04
Distinguish between namespace indexes vs IDs

jbearer:
Does [this](https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1210253275558855?focus=true), among other things.

The earlier change to separate namespace information in transactions in the database was incorrect, because it conflated namespace indexes (i.e. the offset of a given namespace within the namespace table, which may vary from block to block for a given namespace) with global namespace IDs.

I discovered this when I tried to implement transaction filtering by namespace for the explorer, which should have been easy after the previous change. It turned out not to be possible, because what was stored in the database was only the namespace index, not the ID.

### This PR:

This change addresses the problem and hopefully reduces confusion by creating separate types for namespace index vs ID. Both are now stored in the database (the migration to populate the namespace ID is a little gross, but thankfully the information is all there in the namespace table).

We add a test for the new functionality, and implement the explorer namespace filtering feature as a sanity check that the database schema is now correct and sufficently expressive. Thus, in addition to running unit tests, you can test this change by running a local demo, opening the block explorer in a browser, and clicking on a namespace link in the transactions panel. This will take you to a page listing transactions in that namespace, which should be correctly populated. Prior to this change, that page was empty.

Incidentally, this change also fixes what I believe is a bug: when building a block, bytes added to the namespace table are counted against the total block size for the purpose of truncating, but these bytes actually live in the header, and don't count when enforcing maximum block size in the actual proposal validation. I discovered this while working on the new unit test, since that also deals with transaction and namespace sizes.

### This PR does not:

Note that the updated migration for Postgres populates all the new fields using a coputation on existing data, but the sqlite migration creates a new _empty_ table. This is basically because sqlite doesn't have a base64 decode function, which is needed to get the namespace ID out of the namespace table. There are extensions for this, but there is not a good package manager, so it's difficult to install and manage these extensions in a nice way. For now, I think it's ok for the migration to adjust the schema but not populate old data, since we don't have any sqlite nodes in production anyway.

### Key places to review:
* Both database migrations
* New unit test in `api.rs`
* New queries and filtering in `queries/explorer.rs`

### DIFF
--- a/types/src/v0/impls/block/test.rs
+++ b/types/src/v0/impls/block/test.rs
@@ -117,9 +117,7 @@ async fn enforce_max_block_size() {
     let tx_count_expected = test.all_txs().len();
 
     let chain_config = ChainConfig {
-        max_block_size: BlockSize::from(
-            (payload_byte_len_expected + ns_table_byte_len_expected) as u64,
-        ),
+        max_block_size: BlockSize::from(payload_byte_len_expected as u64),
         ..Default::default()
     };
 
@@ -142,9 +140,7 @@ async fn enforce_max_block_size() {
     // WARN log should be emitted
 
     let chain_config = ChainConfig {
-        max_block_size: BlockSize::from(
-            (payload_byte_len_expected + ns_table_byte_len_expected - 1) as u64,
-        ),
+        max_block_size: BlockSize::from((payload_byte_len_expected - 1) as u64),
         ..Default::default()
     };
     let instance_state = NodeState::default().with_chain_config(chain_config);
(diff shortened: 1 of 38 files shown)


types/src/v0/impls/block/test.rs
```diff
@@ -142,9 +140,7 @@ async fn enforce_max_block_size() {
     // WARN log should be emitted
 
     let chain_config = ChainConfig {
-        max_block_size: BlockSize::from(
-            (payload_byte_len_expected + ns_table_byte_len_expected - 1) as u64,
-        ),
+        max_block_size: BlockSize::from((payload_byte_len_expected - 1) as u64),
```
imabdulbasit, line 143: This also looks correct to me, as `vid_share.data.payload_byte_len()`, is being passed to validate_and_apply_header()

PR #3362, opened 2025-06-04
Add transaction count and payload size aggregator stats for each namespace 

imabdulbasit:
This PR adds aggregator stats for each namespace and adds endpoints for transaction counts and payload sizes per namespace. In the database, a namespace ID of -1 represents overall stats of all namespaces. The aggregator task now inserts cumulative stats at every block height for each namespace which also helps to support range endpoints. A test is also included to test all namespace stats endpoints.

Closes: https://app.asana.com/1/1208976916964769/project/1209413670216270/task/1210253275558851
https://app.asana.com/1/1208976916964769/project/1209413670216270/task/1210253275558853


hotshot-query-service/api/node.toml
```diff
@@ -34,9 +34,11 @@ Returns an integer.
 """
 
 [route.count_transactions]
-PATH = ["transactions/count", "transactions/count/:to", "transactions/count/:from/:to"]
+PATH = ["transactions/count", "transactions/count/:to", "transactions/count/:from/:to", "transactions/count/:namespace",
```
jbearer: I think `transactions/count/:to` and `transactions/count/:namespace` overlap in which requests they match. It might need to be something like `transactions/count/namespace/:namespace`

hotshot-query-service/src/data_source/storage/sql/queries/node.rs
```diff
@@ -293,19 +309,48 @@ impl<Mode: TransactionMode> AggregatesStorage for Transaction<Mode> {
     }
 
     async fn load_prev_aggregate(&mut self) -> anyhow::Result<Option<Aggregate>> {
-        let res  : Option<(i64, i64,i64)> = query_as(
-            "SELECT height, num_transactions, payload_size FROM aggregate ORDER BY height DESC limit 1",
+        // Get the maximum height for which we have stored aggregated results
+        // then query all the namespace info for that height
+        let res: Option<(i64,)> = query_as(
+            "SELECT height FROM aggregate WHERE namespace = -1 ORDER BY height DESC LIMIT 1",
```
jbearer: Nit: instead of `ORDER BY height DESC LIMIT 1` maybe `SELECT max(height)` would be a bit more clear?

hotshot-query-service/src/data_source/storage/sql/queries/node.rs
```diff
@@ -335,25 +380,48 @@ impl<Types: NodeType> UpdateAggregatesStorage<Types> for Transaction<Write> {
...
+
+                        rows.push((
+                            block.height as i64,
+                            ns_id as i64,
+                            tx_count[&key] as i64,
+                            size[&key] as i64,
+                        ));
+                    }
+
+                    Some(Ok((block.height as i64, tx_count.clone(), size.clone())))
```
jbearer, line 428: Is there a way to avoid cloning here? These could be maps with many namespaces so could potentially be a lot of memory usage

imabdulbasit: I tried to remove it but wasn't able to. There were lifetime errors

jbearer (approved):

PR #3364, opened 2025-06-05
Add metrics

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* Adds VID disperse calculation time metric
* Adds `validate_and_apply_header` duration metric
* Adds `update_leaf` duration metric

### Key places to review:

---
  - https://app.asana.com/0/0/1210402626691489
  - https://app.asana.com/0/0/1210402626691486


crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -452,10 +452,18 @@ pub(crate) async fn update_shared_state<
...
+    let update_leaf_duration = now.elapsed();
+
+    consensus_writer
+        .metrics
+        .validate_and_apply_header_duration
+        .add_point(validation_duration.as_secs_f64());
+    consensus_writer
+        .metrics
+        .update_leaf_duration
+        .add_point(update_leaf_duration.as_secs_f64());
```
bfish713, line 466: can we get higher resolution duration here, milliseconds would be ok, nanos preferable

lukaszrzasik: The problem is that `add_point` expects f64 and [as_millis_f64](https://doc.rust-lang.org/core/time/struct.Duration.html#method.as_millis_f64) is 

> a nightly-only experimental API

I'd expect that 64-bit floating point in seconds should give us quite good resolution. I don't expect many digits on the integer side.

bfish713: Yeah you are right this should be good for our purpose

bfish713 (approved):

PR #3370, opened 2025-06-05
Upgrade esp via multisig

alysiahuggins:
### This PR:
- adds code to upgrade to v2 via eoa or multisig

### This PR does not:
- change contract logic

### How to test this PR: 
Rust Tests
```bash
cargo test -p espresso-contract-deployer -- test_upgrade_esp_token_v2_multisig_owner_dry_run
```
Docker Demo
```bash
just pull
./scripts/build-docker-images-native --image build --image deploy
just demo deploy-sequencer-contracts
```

---
  - https://app.asana.com/0/0/1210237922954059


contracts/rust/deployer/src/lib.rs
```diff
@@ -1763,4 +1879,187 @@ mod tests {
         );
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_upgrade_esp_token_v2() -> Result<()> {
+        setup_test();
+        setup_test();
```
```suggestion
```

contracts/rust/deployer/src/lib.rs
```diff
@@ -1763,4 +1879,187 @@ mod tests {
...
+        }
+
+        // transfer ownership to multisig
+        let _receipt = transfer_ownership(
+            &provider,
+            Contract::EspTokenProxy,
+            esp_token_proxy_addr,
+            multisig_admin,
+        )
+        .await?;
```
sveitser, line 2033: Our transfer ownership function should probably check that the receipt status is 1. Currently I think we don't do that. Don't need to fix it in this PR though.

PR #3371, opened 2025-06-06
feat: legacy state endpoints & prover update

### LINKED ISSUE #3347: [light-client] sequencer node and relay-server need to provide legacy state signatures
Discussion:
https://espressosys.slack.com/archives/C08RH5EQ66A/p1748890474566139

mrain:
Closes #3347


sequencer/src/state_signature/relay_server.rs
```diff
@@ -111,13 +133,16 @@ impl StateRelayServerState {
         self.thresholds
             .insert(first_epoch, one_honest_threshold(genesis_total_stake));
 
-        let mut genesis_known_nodes = HashMap::<StateVerKey, U256>::new();
+        // let mut genesis_known_nodes = HashMap::<StateVerKey, U256>::new();
```
alxiong:
```suggestion
```

sequencer/src/state_signature/relay_server.rs
```diff
@@ -230,6 +245,19 @@ impl StateRelayServerState {
         }
     }
 
+    /// Centralizing all garbage-collection logic, won't panic, won't error, simply do nothing if nothing to prune.
+    /// `until_height` is inclusive, meaning that would also be pruned.
+    pub fn prune_for_legacy(&mut self, until_height: u64) {
+        while let Some(&height) = self.queue_for_legacy.first() {
+            if height > until_height {
+                return;
```
alxiong, line 252: try use [`.retain()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) api, it's cleaner

hotshot-state-prover/src/legacy/service.rs
```diff
@@ -97,6 +100,15 @@ pub async fn read_contract_state(
     Ok((state.into(), st_state.into()))
 }
 
+/// Check if the contract is a legacy contract. Returns true if the version check succeeds, false otherwise.
+pub async fn is_contract_legacy(provider: impl Provider, address: Address) -> bool {
+    let contract = LightClient::new(address, &provider);
+    match contract.getVersion().call().await {
+        Ok(version) => version.majorVersion == 1,
+        Err(_) => false,
```
alxiong: should we return `Result<bool>` instead and return err here?

hotshot-state-prover/src/bin/state-prover.rs
```diff
@@ -169,16 +169,37 @@ async fn main() {
...
         };
     } else {
         // Run light client state update once
-        if let Err(err) = run_prover_once(config, StaticVersion::<0, 1> {}).await {
-            tracing::error!("Error running prover once: {:?}", err);
+        let result = if is_legacy {
+            hotshot_state_prover::legacy::service::run_prover_once(config, StaticVersion::<0, 1> {})
+                .await
+        } else {
+            hotshot_state_prover::service::run_prover_once(config, StaticVersion::<0, 1> {}).await
```
alxiong: i know we never really pay much attention to the version specified here, but should we assign different versions to the legacy v.s. the latest prover?

sequencer/src/state_signature.rs
```diff
@@ -148,7 +149,15 @@ impl<ApiVer: StaticVersionType> StateSigner<ApiVer> {
                     }
                 }
 
-                let signature = self.sign_new_state(&state, self.voting_stake_table).await;
+                let Ok(signature) = self.sign_new_state(&state, self.voting_stake_table).await
+                else {
+                    tracing::error!("Failed to sign new state");
+                    return;
```
alxiong, line 155: i'm confused by the logic, so right now the state-signer would always sign both and try to send both to relay-server and see which one succeed? 
this feels unnecessary? can we branch the action by the `is_legacy()` status here?

sequencer/src/state_signature/relay_server.rs
```diff
@@ -230,6 +245,19 @@ impl StateRelayServerState {
         }
     }
 
+    /// Centralizing all garbage-collection logic, won't panic, won't error, simply do nothing if nothing to prune.
+    /// `until_height` is inclusive, meaning that would also be pruned.
+    pub fn prune_for_legacy(&mut self, until_height: u64) {
```
alxiong, line 249: reminder to gc after epoch is activated, and we should test this in a unit tests.

hotshot-state-prover/src/legacy/service.rs
mrain: I'm not sure what will happen if you are calling `LightClient::getVersion()` on a V2 contract

alxiong: The MajorVersion will be 2. 
The Err case here is for when you pass in the wrong address, the call might fail

hotshot-state-prover/src/bin/state-prover.rs
mrain: I don't known when is this parameter introduced. But it doesn't seem to be meanful.

mrain: I don't know when is this parameter introduced. But it doesn't seem to be very meanful

sequencer/src/state_signature/relay_server.rs
mrain: Code will be similar with `retain()` since you need to remove things from other collections. It's okay to keep it like it is now

sequencer/src/state_signature.rs
mrain: no we don't know when is contract upgraded. so we need to keep both signatures served, at least before the epoch upgrade.

sequencer/src/state_signature/relay_server.rs
mrain: it may not be necessary. the legacy signatures will be served until `epoch_start_block - 1`. And it's safer to serve it for a little longer after the epoch upgrade. Note that all signatures will be gc'ed once we start serving `epoch_start_block - 1`.

sequencer/src/state_signature.rs
```diff
@@ -206,7 +233,14 @@ impl<ApiVer: StaticVersionType> StateSigner<ApiVer> {
...
         );
-        signature
+        Ok(signature)
+    }
+
+    async fn legacy_sign_new_state(
+        &self,
+        state: &LightClientState,
+    ) -> Result<StateSignature, SignatureError> {
+        <SchnorrPubKey as StateSignatureKey>::legacy_sign_state(&self.sign_key, state)
```
alxiong, line 243: what's the point of this function? so thinly wrapped.

sequencer/src/state_signature.rs
alxiong: i don't understand why state-signer need to know "when is contract upgraded"? state-signer only acts according to the hotshot state, right?

sequencer/src/state_signature.rs
mrain: just for convenience, nothing more

PR #3372, opened 2025-06-06
Add txn bytes to block payload, add metric for finalized bytes

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -322,6 +322,12 @@ async fn update_metrics<TYPES: NodeType>(
             .metrics
             .proposal_to_decide_time
             .add_point(proposal_to_decide_time as f64);
+        if let Some(txn_bytes) = leaf_view.leaf.block_payload().map(|p| p.txn_bytes()) {
+            consensus_reader
+                .metrics
+                .finalized_bytes
+                .add_point(txn_bytes as f64);
+        }
     }
 }
 
--- a/crates/hotshot/types/src/traits/block_contents.rs
+++ b/crates/hotshot/types/src/traits/block_contents.rs
@@ -127,6 +127,9 @@ pub trait BlockPayload<TYPES: NodeType>:
         &'a self,
         metadata: &'a Self::Metadata,
     ) -> impl 'a + Iterator<Item = Self::Transaction>;
+
+    /// Get the number of bytes of transactions in the payload.
+    fn txn_bytes(&self) -> usize;
 }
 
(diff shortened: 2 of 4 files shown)


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -303,6 +304,33 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
...
+
+        let Some(proposal_to_decide_time) = now.checked_sub(proposal_timestamp) else {
+            tracing::error!("Failed to calculate proposal to decide time: {proposal_timestamp}");
+            continue;
+        };
+        consensus_reader
+            .metrics
+            .proposal_to_decide_time
+            .add_point(proposal_to_decide_time as f64);
+        if let Some(txn_bytes) = leaf_view.leaf.block_payload().map(|p| p.txn_bytes()) {
```
ss-es, line 338: does it matter if `block_payload` is missing at this point? from my understanding most nodes won't have the payload right? (only DA?)

crates/hotshot/types/src/traits/block_contents.rs
```diff
@@ -188,6 +191,9 @@ pub trait BlockHeader<TYPES: NodeType>:
     /// Get the block number.
     fn block_number(&self) -> u64;
 
+    /// Get the timestamp.
+    fn timestamp(&self) -> u64;
```
ss-es: I guess this would be a `u128` soon and we'd be casting the current timestamps to a `u128`?

ss-es (commented):
looks straightforward on the whole, just a quick question before approving

hotshot-query-service/src/data_source.rs
```diff
@@ -1232,7 +1237,9 @@ pub mod node_tests {
             let leaf = leaves.next().await.unwrap();
             let header = leaf.header().clone();
             if let Some(last_timestamp) = test_blocks.last_mut() {
-                if last_timestamp[0].timestamp() == header.timestamp() {
+                if <TestBlockHeader as BlockHeader<MockTypes>>::timestamp(&last_timestamp[0])
```
pls148: block_header_timestamp?

crates/hotshot/task-impls/src/helpers.rs
bfish713: Yeah not much we can do, if it's missing we just won't have the stat for it

hotshot-query-service/src/data_source.rs
bfish713: Oh I need to update after updating the base pr

PR #3377, opened 2025-06-10
feat: multiple l1 urls for state-prover

mrain:
Closes #<ISSUE_NUMBER>

### This PR:
Uses the [`SwitchingTransport`](https://github.com/EspressoSystems/espresso-network/blob/200e0373d4c0d78b84b959fd71e6a3d01543703d/types/src/v0/v0_1/l1.rs#L252-L265) so that we can configure multiple (fallback) l1 urls for the state-prover. Assuming that the `SwitchingTransport` can auto choose one of the available l1 url from the configured list.

Configurations are the same with the ones for sequencer.

### Key places to review:

---
  - https://app.asana.com/0/0/1210126644705051


alxiong (approved):
This is a good idea. Thanks Chengyu!

PR #3378, opened 2025-06-11
Add upgraded block header type

ss-es:


sveitser:
I don't know if this would have some knock-on effects but instead of changing the type of a field should we make a new field like `timestamp_ns` instead, or add a separate fields for the nanoseconds to maintain backwards compatibility for the original timestamp? I feel like that might be a bit less confusing than changing the meaning of timestamp values by changing their unit. I'm afraid this will be a subtle breakage for all consumers that may go unnoticed for a while and potentially cause downtime.

Also do we need it to be nanoseconds and u128? Unless I made a mistake we still have only used a 1/10 of the u64 nanoseconds since 1970. In system with about 2 seconds latency it seems to me that more than 1s resolution isn't very meaningful and we don't have a way to enforce the value to be very accurate, even now with 1 second resolution. I forgot why exactly we want more resolution do we really need to do this in the header?

cc @ss-es @bfish713

sveitser:
We should also add reference tests in `data/v4`.

bfish713:
> I don't know if this would have some knock-on effects but instead of changing the type of a field should we make a new field like `timestamp_ns` instead, or add a separate fields for the nanoseconds to maintain backwards compatibility for the original timestamp? I feel like that might be a bit less confusing than changing the meaning of timestamp values by changing their unit. I'm afraid this will be a subtle breakage for some consumers that may go unnoticed for a while and potentially cause downtime.
> 
> Also do we need it to be nanoseconds and u128? Unless I made a mistake we still have only used a 1/10 of the u64 nanoseconds since 1970. In system with about 2 seconds latency it seems to me that more than 1s resolution isn't very meaningful and we don't have a way to enforce the value to be very accurate, even now with 1 second resolution. I forgot why exactly we want more resolution and just wondering if we really need to do it via the header.
> 
> cc @ss-es @bfish713

The reason we want more resolution is for benchmarking.  We often see rounds taking < 1s, especially in devnets.  It's very possible we introduce a change that reduces total latency by 300ms and we would have a hard time seeing it with 1s resolution.  But the main reason is that we want to get latency down to sub 1s anyway, at that point the 1s resolution will really not be enough. We want to have this via the header because we want to at least estimate the total time it takes a proposal from creation to decide. There could be some timestamp drift but that should be quite low for most running in the cloud.

Completely agree with the first part, we should add the new timestamp rather than replace the old.

ss-es:
thank you very much for the suggestion @sveitser, I've updated the PR to add a new field and I think this solved a lot of the trouble I was having! I was also very uncomfortable with the way I was approaching it previously -- it was too easy to make undetectable mistakes

I've also added reference tests for the new header type

For the type, I was using `u128` initially to be closer to the type from the crate: https://docs.rs/time/latest/time/struct.OffsetDateTime.html#method.unix_timestamp_nanos. I was mostly trying to avoid having to do a conversion here and I've updated the type to `i128` to avoid the conversion entirely. but if there are strong opinions on this, I'm open to changing it

types/src/v0/v0_4/header.rs
```diff
@@ -0,0 +1,70 @@
...
+use serde::{Deserialize, Serialize};
+
+/// A header is like a [`Block`] with the body replaced by a digest.
+#[derive(Clone, Debug, Deserialize, Serialize, Hash, PartialEq, Eq)]
+pub struct Header {
+    /// A commitment to a ChainConfig or a full ChainConfig.
+    pub(crate) chain_config: ResolvableChainConfig,
+    pub(crate) height: u64,
+    pub(crate) timestamp: u64,
+    pub(crate) timestamp_nanos: i128,
```
imabdulbasit: why its i128 and not u128?

types/src/v0/impls/header.rs
```diff
@@ -85,6 +85,11 @@ impl Committable for Header {
                 .u64_field("version_minor", 3)
                 .field("fields", fields.commit())
                 .finalize(),
+            Self::V4(fields) => RawCommitmentBuilder::new(&Self::tag())
+                .u64_field("version_major", 0)
+                .u64_field("version_minor", 3)
```
imabdulbasit:
```suggestion
                .u64_field("version_minor", 4)
```

imabdulbasit (commented):
Could you please include the genesis.toml file for the new v4 version, as well as the upgrade from v3 to v4, so that we can run the native demo locally?

types/src/v0/v0_4/header.rs
ss-es: > For the type, I was using u128 initially to be closer to the type from the crate: https://docs.rs/time/latest/time/struct.OffsetDateTime.html#method.unix_timestamp_nanos. I was mostly trying to avoid having to do a conversion here and I've updated the type to i128 to avoid the conversion entirely. but if there are strong opinions on this, I'm open to changing it

types/src/v0/impls/state.rs
```diff
@@ -407,6 +412,18 @@ impl<'a> Proposal<'a> {
         Ok(())
     }
 
+    /// The `timestamp` and `timestamp_nanos` fields must be coherent
+    fn validate_timestamp_consistency(&self) -> Result<(), ProposalValidationError> {
+        if self.header.timestamp() as i128 * 1_000_000_000 != self.header.timestamp_nanos() {
```
ss-es: oops -- this should be `timestamp_nanos / 10^9` on the other side

sequencer/src/run.rs
```diff
@@ -29,6 +29,35 @@ pub async fn main() -> anyhow::Result<()> {
     let upgrade = genesis.upgrade_version;
 
     match (base, upgrade) {
+        #[cfg(all(feature = "fee", feature = "pos"))]
```
imabdulbasit: There should be a new feature for this upgrade?

types/src/v0/v0_4/header.rs
imabdulbasit: I don't think its a big deal but IMO u128 would be better because the timestamp_ns can never be negative , and the timestamp field already uses u64 and not i64. WDYT @sveitser

crates/hotshot/example-types/src/block_types.rs
```diff
@@ -291,6 +293,12 @@ impl TestBlockHeader {
             timestamp = parent.timestamp;
         }
 
+        let mut timestamp_nanos = OffsetDateTime::now_utc().unix_timestamp_nanos();
```
sveitser: I think we should derive timestamp from timestamp_nanos to ensure the seconds portion always matches.

types/src/v0/v0_4/header.rs
sveitser: I would prefer to use the "simplest" type that works. IMO that is u64 (current timestamp in nanoseconds is at 10% of the range) which is easier to work with and natively supported in every language I know. I also think milliseconds would be enough because the extra resolution (the micro and nanoseconds parts) is IMO mostly noise because clocks of nodes in a distributed system can't be synced to that level of accuracy (I think). Therefore I don't see any advantage of having extra resolution (more than milliseconds) in the header.

sequencer/src/run.rs
ss-es: added the `drb-and-header` feature

imabdulbasit: the builder binary would also need this feature because it uses genesis.toml file

justfile
```diff
@@ -44,6 +44,9 @@ demo-native-pos *args: (build "test" "--features fee,pos")
 demo-native-pos-base *args: (build "test" "--features pos")
     ESPRESSO_SEQUENCER_PROCESS_COMPOSE_GENESIS_FILE=data/genesis/demo-pos-base.toml scripts/demo-native -f process-compose.yaml {{args}}
 
+demo-native-drb-header-upgrade *args: (build "test" "--features pos")
```
imabdulbasit:
```suggestion
demo-native-drb-header-upgrade *args: (build "test" "--features pos,drb-and-header")
```

sequencer/src/run.rs
imabdulbasit: and also sequencer-sqlite

PR #3379, opened 2025-06-11
Improved logs

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Prints events received by the dependency task when failing to propose or vote.

### Key places to review:

---
  - https://app.asana.com/0/0/1210186607314340


crates/hotshot/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -637,6 +637,11 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
 
         Ok(())
     }
+
+    fn print_proposal_events(&self, res: &[Vec<Vec<Arc<HotShotEvent<TYPES>>>>]) {
+        let events: Vec<_> = res.iter().flatten().flatten().map(Arc::as_ref).collect();
+        tracing::error!("Failed to propose, events: {:#?}", events);
```
bfish713: Lets make this a warn!

bfish713 (approved):
This is fine, but seems like maybe refactoring the handle_dep_result functions to call some inner function that returns a result or bool that way we can't forget to call `print_vote_events`

If you agree, we can create a follow up to do this, or just amend this PR

bfish713 (approved):

PR #3380, opened 2025-06-11
move sequencer-sqlite crate into workspace

sveitser:
Exclude it from default members to avoid always turning on the embedded-db feature.

This still requires compiling twice but at least we don't need to handle two lock files and target directories anymore.

Note: Marking this as draft until CI is green.


sequencer-sqlite/Cargo.lock
sveitser, line 1: This hasn't been working for some time, therefore removing it.

.github/workflows/build_static.yml
sveitser, line 1: Hasn't been working for a while.

scripts/build-docker-images-static
sveitser, line 1: Hasn't been working for a while.

PR #3381, opened 2025-06-12
use sqlite by default for sequencer docker image

imabdulbasit:
addresses https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1209466337764166


pls148 (approved):

sveitser (dismissed):
Doesn't this create an issue for operators in case they previously started without setting this env var, they would now switch storage.

imabdulbasit:
> Doesn't this create an issue for operators in case they previously started without setting this env var, they would now switch storage.

that's true so should we wait for switch from fs to sqlite and then merge this?

PR #3382, opened 2025-06-13
print keygen keys to stdout if file path not provided

imabdulbasit:


sequencer/src/bin/keygen.rs
```diff
@@ -128,29 +125,41 @@ fn main() -> anyhow::Result<()> {
...
-
     let seed = opts.seed.unwrap_or_else(|| {
         tracing::debug!("No seed provided, generating a random seed");
         gen_default_seed()
     });
-    fs::write(opts.out.join(".seed"), hex::encode(seed))?;
+
+    if let Some(ref out_dir) = opts.out {
+        fs::create_dir_all(out_dir)?;
+        fs::write(out_dir.join(".seed"), hex::encode(seed))?;
```
sveitser: I feel like it's a bad idea to quietly store the seed in a hidden file. It's mentioned in the docs/help but I doubt everyone is going to read that. So there's the risk of leaving behind the seed file and leaking private keys. If anything we should probably put it into the `.env` file as a comment because it's more obvious that the key file should be kept a secret. What do you think?

It's also weird that we save it now only if an out_dir is provided.

sveitser (approved):
LGTM. Just a comment.

sequencer/src/bin/keygen.rs
imabdulbasit: Yeah I also agree

Issue #1501, opened 2024-05-24
Use an argument parsing library like for example `yargs` for typescript scripts


sveitser:
Use an argument parsing library like for example `yargs` for typescript scripts

_Originally posted by @sveitser in https://github.com/EspressoSystems/espresso-sequencer/pull/1454#discussion_r1613388437_

PR #3388, opened 2025-06-16
WIP Transfer ownership proposal

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
WIP add the ability to create a transfer ownership proposal so that we can transfer ownership of relevant contracts from the multisig to the timelock

### DIFF
--- a/contracts/rust/deployer/src/lib.rs
+++ b/contracts/rust/deployer/src/lib.rs
@@ -561,12 +561,9 @@ pub async fn upgrade_light_client_v2_multisig_owner(
     let proxy = LightClient::new(proxy_addr, &provider);
     let owner_addr = proxy.owner().call().await?._0;
 
-    if !dry_run {
-        tracing::info!("Checking if owner is a contract");
-        assert!(
-            is_contract(&provider, owner_addr).await?,
-            "Owner is not a contract so not a multisig wallet"
-        );
+    if !dry_run && !is_contract(&provider, owner_addr).await? {
+        tracing::error!("Proxy owner is not a contract. Expected: {owner_addr:#x}");
+        anyhow::bail!("Proxy owner is not a contract");
     }
 
     // Prepare addresses
--- a/contracts/script/multisigTransactionProposals/safeSDK/transferOwnership.ts
+++ b/contracts/script/multisigTransactionProposals/safeSDK/transferOwnership.ts
@@ -1,0 +1,188 @@
+import * as dotenv from "dotenv";
+import { ethers } from "ethers";
+import { EthersAdapter } from "@safe-global/protocol-kit";
+import SafeApiKit from "@safe-global/api-kit";
+import Safe from "@safe-global/protocol-kit";
+import { getEnvVar, validateEthereumAddress, getSigner, createAndSignSafeTransaction } from "./utils";
+const TRANSFER_OWNERSHIP_CMD = "transferOwnership" as const;
+
+export interface TransferOwnershipData {
+  proxyAddress: string;
+  rpcUrl: string;
+  newOwner: string;
+  safeAddress: string;
+  useHardwareWallet: boolean;
+}
+
+async function main() {
+  dotenv.config();
+
+  try {
+    const [transferOwnershipData, dryRun] = processCommandLineArguments();
+    console.log(JSON.stringify(transferOwnershipData));
+
+    if (!dryRun) {
+      // Initialize web3 provider using the RPC URL from environment variables
+      const web3Provider = new ethers.JsonRpcProvider(transferOwnershipData.rpcUrl);
+
+      // Get the signer, this signer must be one of the signers on the Safe Multisig Wallet
+      const orchestratorSigner = getSigner(web3Provider, transferOwnershipData.useHardwareWallet);
+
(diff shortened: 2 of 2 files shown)


contracts/rust/deployer/src/lib.rs
```diff
@@ -2519,4 +2525,261 @@ mod tests {
...
+        let mut sepolia_rpc_url = "http://127.0.0.1:8545".to_string();
+        let mut multisig_admin = Address::random();
+        let mut timelock = Address::random();
+        let mut mnemonic =
+            "test test test test test test test test test test test junk".to_string();
+        let mut account_index = 0;
+        let anvil = Anvil::default().spawn();
+        let dry_run = matches!(options.run_mode, RunMode::DryRun);
+        if !dry_run {
+            dotenvy::from_filename_override(".env.deployer.rs.test").ok();
```
sveitser, line 2546: I'm wondering why we don't put these values into the code directly instead of reading them from an .env file.

sequencer/src/bin/deploy.rs
```diff
@@ -504,6 +511,21 @@ async fn main() -> anyhow::Result<()> {
...
     }
 
+    // Execute ownership transfer proposal if requested
+    if opt.propose_transfer_ownership_to_timelock_fee_contract {
+        let timelock_controller = opt.timelock_controller.expect(
+            "Must provide --timelock-controller when transferring ownership from multisig to \
+             timelock",
+        );
+
+        args.propose_transfer_ownership_from_multisig_to_timelock(
```
sveitser, line 521: Shouldn't we force this to use the ops timelock?

Edit: Maybe I misunderstand but it seems currently it would be totally the responsibility of the caller to decide which contract is owner by which timelock. Since the deploy binary already has 57 (I think) arguments right now I think we should make efforts to reduce burden on the user to pick the right things. Now the user needs to know which contract uses which timelock. If we instead chose the right timelock in the code we can throw a descriptive error if the address for that timelock is unset. Additionally the user can configure env vars once and doesn't need to set the env var for the timelock controller to different values between runs, this also reduces cognitive load and room for error.

contracts/script/multisigTransactionProposals/safeSDK/transferOwnership.ts
```diff
@@ -0,0 +1,194 @@
...
+    rpcUrl: rpcUrl,
+    newOwner: newOwner,
+    safeAddress: safeAddress,
+    useHardwareWallet: useHardwareWallet,
+  };
+
+  return [transferOwnershipData, dryRun];
+}
+
+function processCommandLineArguments(): [TransferOwnershipData, boolean] {
```
sveitser, line 114: Can we address #1501 soon to avoid adding more custom argument parsing?

sequencer/src/bin/deploy.rs
sveitser: I'm also wondering if this should be a subcommand instead. I there any value in doing this proposal and doing deployment of contracts at the same time?

sveitser (approved):
LGTM but would like to see comments addressed.

sequencer/src/bin/deploy.rs
alysiahuggins: > I'm also wondering if this should be a subcommand instead. I there any value in doing this proposal and doing deployment of contracts at the same time?

Here I'm assuming that the contract is already deployed, as is the case, but it has a multisig owner and we'd like to transfer the ownership of that contract from the multisig to the timelock contract

alysiahuggins: > Shouldn't we force this to use the ops timelock?

Edit: Maybe I misunderstand but it seems currently it would be totally the responsibility of the caller to decide which contract is owner by which timelock. Since the deploy binary already has 57 (I think) arguments right now I think we should make efforts to reduce burden on the user to pick the right things. Now the user needs to know which contract uses which timelock. If we instead chose the right timelock in the code we can throw a descriptive error if the address for that timelock is unset. Additionally the user can configure env vars once and doesn't need to set the env var for the timelock controller to different values between runs, this also reduces cognitive load and room for error.

I actually do this in my next PR, https://github.com/EspressoSystems/espresso-network/pull/3417, all are owned by the OpsTimelock except for the token which is owned by the SafeExitTimelock. This PR was created to add support only for transferring ownership to a timelock if the contract is already owned by a multisig.

I also agree that there are a ton of arguments to consider. Also added a readme the next PR, https://github.com/EspressoSystems/espresso-network/pull/3417, so that there are clearer instructions.

contracts/rust/deployer/src/lib.rs
alysiahuggins: in this case, when it's not a dry run, you have to use a real safe multisig wallet deployed on a real network and therefore you'd need real sepolia eth, so didn't want to expose any phrases in the test and left it up to the dev's setup

contracts/script/multisigTransactionProposals/safeSDK/transferOwnership.ts
alysiahuggins: ok noted, though i'm anticipating that we won't need to support many more multisig operations as we'd have all contracts owned by a timelock soon. but will keep this in mind

PR #3389, opened 2025-06-17
fix: fixes for docker demo UX

sveitser:
- fix: deployer polling interval ignored
- fix: silently ignore building images if multiple images specified
- feat: lazydocker TUI for docker demo

We ran into these working on https://github.com/EspressoSystems/espresso-network/pull/3306

### DIFF
--- a/justfile
+++ b/justfile
@@ -7,7 +7,14 @@ doc *args:
     cargo doc --no-deps --document-private-items {{args}}
 
 demo *args:
-    docker compose up {{args}}
+    #!/usr/bin/env bash
+    trap "exit" INT TERM
+    trap cleanup EXIT
+    cleanup(){
+        docker compose down -v
+    }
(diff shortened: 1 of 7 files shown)


justfile
```diff
@@ -7,7 +7,21 @@ doc *args:
...
+    CI=${CI:-false}
+    if [ "$CI" = "true" ]; then
+        docker compose up {{args}}
+    else
+        trap "exit" INT TERM
+        trap cleanup EXIT
+        cleanup(){
+            docker compose down -v
+        }
+        >/dev/null 2>&1 docker compose up {{args}} &
```
imabdulbasit, line 21: isn't the redirection wrong here?

imabdulbasit: or its ignoring the `docker compose down -v` output?

sveitser: You mean it shouldn't be a the start of the line? This also works and I think it's easier to read.

PR #3390, opened 2025-06-17
assert decided leaves for restarted nodes

imabdulbasit:


sequencer/src/restart_tests.rs
```diff
@@ -725,25 +709,38 @@ impl TestNetwork {
...
+        for node in self.da_nodes.iter().chain(self.regular_nodes.iter()) {
+            states.push((
+                node.node_id().expect("Node id not found"),
+                node.reference_state.read().await.clone(),
+            ));
+        }
+
+        // assert that all the nodes have same leaves from their event streams
+        // this also ensures validated state consistency
+        let (ref_id, ref_state) = &states[0];
```
sveitser: We could get the reference state by calling next on the iterator `self.da_nodes.iter().chain(self.regular_nodes.iter())` then iterate over the iterator to process the remaining nodes. This would avoid iterating over the nodes twice.

PR #3391, opened 2025-06-17
Print missed dependencies

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Adds a receiver to `EventDependency` which waits for the cancelling signal in parallel to the dependency event. When the cancelling signal is received, the name of the dependency is printed. This enables debugging missed dependencies.

The cancelling signal is emitted in the same places where the dependency task used to be aborted, e.g. view change, timeout, update last proposal / vote.

### Key places to review:

---
  - https://app.asana.com/0/0/1210568790417442


crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -579,9 +601,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> QuorumVoteTaskS
 
             // Cancel the old dependency tasks.
             for view in *self.latest_voted_view..(*new_view) {
-                if let Some(dependency) = self.vote_dependencies.remove(&TYPES::View::new(view)) {
-                    dependency.abort();
```
ss-es, line 574: so this is really the main change right? rather than `abort` we instead request that the dependency task shut itself down via the new `cancel_receiver`

ss-es (commented):
I think I'm a bit confused -- does this PR print the missing dependencies or will that be a separate PR? I'm not sure where we trigger "print missed vote dependencies when we cancel the vote task for a view"

crates/hotshot/task-impls/src/quorum_vote/mod.rs
lukaszrzasik: yes, exactly. We signal the event dependency to cancel itself but before it does so, it prints its name and view so that we know which event dependency was cancelled.

ss-es (approved):
the idea makes sense. I'm not sure I understand the flow entirely, but if you've verified this logs as expected I think it looks fine to me

PR #3392, opened 2025-06-17
Add Stats For leader Participation

bfish713:
Closes #<ISSUE_NUMBER>

### Key places to review:

---
  - https://app.asana.com/0/0/1210344188788301


crates/hotshot/task-impls/src/consensus/mod.rs
```diff
@@ -220,11 +220,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
...
+                let next_epoch = high_qc.data.epoch().map(|x| x + 1);
+
                 let mut consensus_writer = self.consensus.write().await;
                 let high_qc_updated = consensus_writer.update_high_qc(high_qc.clone()).is_ok();
                 let next_high_qc_updated = consensus_writer
                     .update_next_epoch_high_qc(next_epoch_high_qc.clone())
                     .is_ok();
+                consensus_writer.update_validator_participation_epoch(
+                    next_epoch.ok_or(warn!("No next epoch"))?,
+                );
```
lukaszrzasik: I think we should do the same as here when handling `ExtendedQc2Formed`.

crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -372,6 +372,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
         }
 
+        let leader = epoch_membership.leader(self.view_number).await;
+        if let (Ok(leader_key), Some(cur_epoch)) = (leader, cur_epoch) {
+            self.consensus
+                .write()
+                .await
+                .update_validator_participation(leader_key, cur_epoch, true);
+        } else {
+            tracing::warn!("No leader for view {:?}", self.view_number);
```
lukaszrzasik: Not sure about this. This will print the warning also when `cur_epoch` is `None`. This means every time in pre-epoch version. Is this what we want?

crates/hotshot/types/src/consensus.rs
```diff
@@ -281,6 +281,84 @@ impl<T: ConsensusTime> HotShotActionViews<T> {
...
+        self.current_epoch_participation = HashMap::new();
+    }
+
+    fn get_participation(&self, key: TYPES::SignatureKey) -> (f64, Option<f64>) {
+        let current_epoch_participation = self
+            .current_epoch_participation
+            .get(&key)
+            .unwrap_or(&(0, 0));
+        let current_epoch_participation_ratio =
+            current_epoch_participation.0 as f64 / current_epoch_participation.1 as f64;
```
lukaszrzasik: Two things:
1. I think there's high risk of 'divide by 0' error here.
2. If I'm not mistaken, this ratio gives us # of views in which a node was a leader divided by # of views in which the node actually proposed, e.g. the node was a leader in 20 views but proposed only 10 times, gives us 2. Maybe it's only me but wouldn't it be easier to interpret when the number was inversed? For the example above it would be 0.5 meaning 50% participation.

PR #3394, opened 2025-06-17
Clean up some logs, enforce Display not Debug for VerKey and aliases

pls148:
Went over a bunch of logs. Cleaned up some things which weren't easy to search for when moving from DD. Converted as many uses of VerKey from Debug to Display as I could find.

---
  - https://app.asana.com/0/0/1210564264311140

### DIFF
--- a/crates/hotshot/hotshot/src/lib.rs
+++ b/crates/hotshot/hotshot/src/lib.rs
@@ -1233,7 +1233,7 @@ async fn load_start_epoch_info<TYPES: NodeType>(
     let first_epoch_number =
         TYPES::Epoch::new(epoch_from_block_number(epoch_start_block, epoch_height));
 
-    tracing::warn!("Calling set_first_epoch for epoch {first_epoch_number:?}");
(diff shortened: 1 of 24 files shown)


crates/hotshot/hotshot/src/lib.rs
```diff
@@ -1233,7 +1233,7 @@ async fn load_start_epoch_info<TYPES: NodeType>(
     let first_epoch_number =
         TYPES::Epoch::new(epoch_from_block_number(epoch_start_block, epoch_height));
 
-    tracing::warn!("Calling set_first_epoch for epoch {:?}", first_epoch_number);
+    tracing::warn!("Calling set_first_epoch for epoch {first_epoch_number:?}");
```
mrain: Should we also implement `Display` for `EpochNumber`, `ViewNumber`, etc?

pls148: we do, it prints the inside value. I'll go over these again and where they're redundant go down to Display instead of Debug

PR #3396, opened 2025-06-18
rewrite stake table state

imabdulbasit:


types/src/v0/impls/stake_table.rs
```diff
@@ -2233,4 +2304,233 @@ mod tests {
...
+        }));
+        assert!(matches!(result, Err(StakeTableError::InsufficientStake)));
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_decaf_stake_table() {
+        setup_test();
+
+        let events_json =
+            std::fs::read_to_string("../data/v3/decaf_stake_table_events.json").unwrap();
```
sveitser, line 2583: When checking in test vectors it's nice to leave some notes about how the data was constructed. In case we for example ever want to regenerate. Not sure exactly what would be best here. But for example we could add a commented or ignored test with the code that creates the json file, or at least shows how that was done.

types/src/v0/impls/stake_table.rs
```diff
@@ -2233,4 +2304,233 @@ mod tests {
...
+
+        // Second registration with the same validator should fail
+        let result = state.apply_event(StakeTableEvent::Register((&validator).into()));
+        assert!(
+            matches!(result, Err(StakeTableError::AlreadyRegistered(addr)) if addr == validator.account)
+        );
+    }
+
+    #[test]
+    fn test_register_v2_validator() {
```
sveitser: We could consider using rstest fixtures to reduce the amount of setup code and also get more coverage. For example all events where we have V2 events should run the same tests for V1 and V2. And for example for registering validators we should also check that it fails if a V1 registered validator is registered again with the V2 event.

I'm also fine merging this PR first but I think there's some room for improvement in terms of test coverage.

types/src/v0/impls/stake_table.rs
```diff
@@ -2233,4 +2304,233 @@ mod tests {
...
+
+    #[test]
+    fn test_register_validator_v2_auth_fails() {
+        let mut state = StakeTableState::new();
+        let mut val = TestValidator::random();
+        val.bls_sig = Default::default();
+        let event = StakeTableEvent::RegisterV2((&val).into());
+
+        let result = state.apply_event(event);
+        assert!(matches!(
```
sveitser, line 2348: This is kind of annoying because I think it won't provide a good error message. But [assert_matches!](https://doc.rust-lang.org/nightly/std/assert_matches/macro.assert_matches.html) is nightly only :(

sveitser: We could enable the "unstable" feature of `pretty_assertions` then use `assert_matches!` from there. The "unstable" feature just doesn't have semver guarantee but I think that's fine here.

types/src/v0/impls/stake_table.rs
```diff
@@ -239,168 +215,220 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
 
                 let stake_table_key: BLSPubKey = blsVK.into();
                 let state_ver_key: SchnorrPubKey = schnorrVK.into();
+
                 // The stake table contract enforces that each bls key is only used once.
-                if bls_keys.contains(&stake_table_key) {
-                    bail!("bls key already used: {}", stake_table_key.to_string());
-                };
+                if !self.used_bls_keys.insert(stake_table_key) {
+                    tracing::warn!(
```
sveitser:
```suggestion
                    tracing::error!(
```
This should really never happen, so I think we should at least log an error.

types/src/v0/impls/stake_table.rs
```diff
@@ -239,168 +215,220 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
+
+                val.stake += amount;
                 // Insert the delegator with the given stake
                 // or increase the stake if already present
-                validator_entry
-                    .delegators
+                val.delegators
                     .entry(delegator)
-                    .and_modify(|stake| *stake += amount)
+                    .and_modify(|s| *s += amount)
```
sveitser: `stake` is too many characters? :D

types/src/v0/impls/stake_table.rs
```diff
@@ -239,168 +215,220 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
+                    return Err(StakeTableError::BlsKeyAlreadyUsed(
+                        stake_table_key.to_string(),
+                    ));
+                }
 
                 // The contract does *not* enforce that each schnorr key is only used once.
-                if schnorr_keys.contains(&state_ver_key) {
-                    tracing::warn!("schnorr key already used: {}", state_ver_key.to_string());
-                };
+                if !self.used_schnorr_keys.insert(state_ver_key.clone()) {
```
sveitser, line 283: Maybe we should just enforce these keys to be unique in the contract too. I think we discussed that it makes sense but we're also thinking about totally rewriting the stake table. :thinking: let's discuss outside this PR.

types/src/v0/v0_3/stake_table.rs
```diff
@@ -103,3 +103,79 @@ pub enum StakeTableEvent {
     KeyUpdate(ConsensusKeysUpdated),
     KeyUpdateV2(ConsensusKeysUpdatedV2),
 }
+
+
+#[derive(Debug, Error)]
+pub enum StakeTableError {
```
sveitser, line 109: Sorry this is going to blow up a bit. 

I think we need to distinguish 2 types of errors. One type are errors that are expected because the contract allows it. For example unauthenticated events, or (currently) duplicate schnorr keys. These are errors we need to handle correctly and continue (ignoring the offending events).

The other types of errors indicate a bug in the contract. They should never happen. If they do happen the GCL nodes should not continue running at all.

types/src/v0/impls/stake_table.rs
```diff
@@ -409,41 +437,64 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
     }
+}
 
-    Ok(validators)
+pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
+    events: I,
+) -> Result<ValidatorMap, StakeTableError> {
+    let mut state = StakeTableState::new();
+    for event in events {
+        state.apply_event(event)?;
```
sveitser: Here we need to handle expected and unexpected errors differently. Only unexpected errors should propagate.

I think we can use the nested result pattern we use in crypto. 

https://github.com/EspressoSystems/jellyfish/blob/4ba71c5d381667ab7c02e64537494502b3e582ea/vid/src/lib.rs#L67

types/src/v0/v0_3/stake_table.rs
sveitser: Correction, since we pre-filter unauthenticated events that is also an unexpected error. But duplicate schnorr keys are possible with the current stake table contract.

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ae5f382583faf8c20dc91336be5ffe1db95cc0f3

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ae5f382583faf8c20dc91336be5ffe1db95cc0f3

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ae5f382583faf8c20dc91336be5ffe1db95cc0f3 and https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ba027c88b5fcc7226a6afaa0557b0e401058f097

types/src/v0/impls/stake_table.rs
imabdulbasit: I have removed this log. we are applying events in. validators_from_l1_events() and there I have added error level log for unexpected error and warn  level for expected error 
https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ba027c88b5fcc7226a6afaa0557b0e401058f097

types/src/v0/impls/stake_table.rs
imabdulbasit: reverted https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ae5f382583faf8c20dc91336be5ffe1db95cc0f3

types/src/v0/v0_3/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3396/commits/ba027c88b5fcc7226a6afaa0557b0e401058f097

PR #3399, opened 2025-06-19
Fix Libp2p DHT

rob-maron:
### This PR:

Fixes an issue with the DHT caused by a change to the Libp2p KDF not fully being propagated. It was caused by the fact that we did not add proper expiries to our keys, meaning even if we overwrote a value in the DHT (as happened in this case), the old value would eventually reappear (as almost all nodes stored the old copy, and less than that had the new one)

It does this by:
- When loading the DHT from persistence, getting, and setting: we ignore keys without expiries
- When performing a get request, we prioritize the key with the highest expiration (likely to be newer)
- Removes `NUM_REPLICATED_TO_TRUST` since the values themselves are authenticated on the storage side
- Since we now already take the minimum of  (default Kademlia replication factor, 2/3 the number of nodes) to determine our replication factor, I updated the default one to be 20. This is more in line with Libp2p defaults, and may improve things as well.
- In order to improve routing before everybody is upgraded, we now use the previously-used `StakeTableAuthentication` (which still exists at the protocol level) to build a map from consensus key -> Libp2p PID. Then, when looking up the mapping to send a message, we now use the validated, local map first before checking the DHT. This gives it a self-healing property. Values are also removed on disconnect to prevent an attack where a bunch of fake nodes are created and stored in the map, taking up space indefinitely.

### Key places to review:

`dht/mod.rs`


rob-maron:
Drafting while I fix some of these test errors

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -265,33 +259,29 @@ impl<K: SignatureKey + 'static, D: DhtPersistentStorage> DHTBehaviour<K, D> {
...
-                                *num_entries
-                            },
-                            std::collections::hash_map::Entry::Vacant(v) => {
-                                v.insert(1);
-                                1
-                            },
+                        // Make sure the record has an expiration time
+                        if record.record.expires.is_some() {
+                            query.records.push(record.record);
+                            true
```
bfish713, line 269: This is to purge records from the old impl?

crates/hotshot/libp2p-networking/src/network/transport.rs
```diff
@@ -106,33 +109,33 @@ impl<T: Transport, Types: NodeType, C: StreamMuxer + Unpin> StakeTableAuthentica
     /// - The signature is invalid
     pub async fn verify_peer_authentication<R: AsyncReadExt + Unpin>(
         stream: &mut R,
-        membership: Arc<Option<Arc<RwLock<Types::Membership>>>>,
         required_peer_id: &PeerId,
+        consensus_key_to_pid_map: Arc<Mutex<BiMap<S, PeerId>>>,
     ) -> AnyhowResult<()> {
-        // If we have a membership, read message and validate
-        if membership.is_some() {
```
bfish713, line 113: We are getting rid of auth using the membership, and requiring multiple peers to return us the record.  What's stopping spoofing now (some random node claiming a HS key, which it does not control, is associated with it's PID?

crates/hotshot/libp2p-networking/src/network/transport.rs
```diff
@@ -106,33 +109,33 @@ impl<T: Transport, Types: NodeType, C: StreamMuxer + Unpin> StakeTableAuthentica
...
+        let message = read_length_delimited(stream, MAX_AUTH_MESSAGE_SIZE).await?;
+
+        // Deserialize the authentication message
+        let auth_message: AuthMessage<S> =
+            bincode::deserialize(&message).with_context(|| "Failed to deserialize auth message")?;
+
+        // Verify the signature on the public keys
+        let public_key = auth_message
+            .validate()
+            .with_context(|| "Failed to verify authentication message")?;
```
bfish713, line 125: I guess we just are checking the signature and it doesn't really matter if the key is in the Membership/stake table, seems fine to me.  Just want to be sure I'm understanding it

bfish713 (approved):
I think this makes sense, but I struggled to mentally model how this will change everything.  I am really thinking we should try to pare back the DHT impl to be much closer to stock and only add custom logic for key authentication for HS keys.  IIRC a lot of this code is basically the same as what the underlying impl is doing and we're just opening ourselves up for mistakes, missing bug fixes, or future incompatibility.

Unfortunately it's probably too big a change to do right now so we can just incrementally improve like this

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
rob-maron: Yeah, anything without an expiry should be removed. This is so that the new records can be propagated more easily

crates/hotshot/libp2p-networking/src/network/transport.rs
rob-maron: The DHT records are validated through: https://github.com/EspressoSystems/espresso-network/blob/1b03728d5029fac341f3a63956345262a1357158/crates/hotshot/libp2p-networking/src/network/behaviours/dht/store/validated.rs#L73-L83

The "auth on new connection" mapping is validated through: https://github.com/EspressoSystems/espresso-network/blob/1b03728d5029fac341f3a63956345262a1357158/crates/hotshot/libp2p-networking/src/network/transport.rs#L107

crates/hotshot/libp2p-networking/src/network/transport.rs
rob-maron: Yeah, the mapping is still authenticated (see above)

PR #3400, opened 2025-06-19
Safe Exit timelock

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
Adds two separate timelock contracts for enhanced security and governance:

1. **OpsTimelock** - A short-delay timelock for operational control during early protocol phases, allowing core team to perform upgrades and configuration changes with minimal delay
2. **SafeExitTimelock** - A longer-delay timelock specifically for the ESP token contract, providing users with a safe exit period before potentially dangerous operations

The implementation includes:
- New `OpsTimelock.sol` and `SafeExitTimelock.sol` contracts extending OpenZeppelin's `TimelockController`
- Updated deployment infrastructure to support both timelock types
- Comprehensive test coverage for timelock functionality
- Updated environment configuration with separate timelock parameters
- Rust bindings and deployment scripts for both contracts

### This PR does not:
- New `OpsTimelock.sol` and `SafeExitTimelock.sol` contracts extending OpenZeppelin's `TimelockController`
- Rust bindings and deployment scripts for both contracts and updated deployment infrastructure to support both timelock types

### Key places to review:

cargo test
```bash
cargo test -p espresso-contract-deployer
```

docker demo
```bash
just pull
./scripts/build-docker-images-native --image deploy
just demo deploy-sequencer-contracts
```

docker service
```bash
just pull
./scripts/build-docker-images-native --image deploy
 docker compose --env-file .env run deploy-sequencer-contracts deploy --deploy-ops-timelock --deploy-safe-exit-timelock
```

### Things tested

* Unit tests for both timelock contracts including delay enforcement, role-based access control, and operation scheduling/execution
* Docker deployment workflow with both timelock types

---
  - https://app.asana.com/0/0/1210591958562328


contracts/src/SafeExitTimelock.sol
```diff
@@ -5,7 +5,7 @@ import "@openzeppelin/contracts/governance/TimelockController.sol";
 
 /// @title TimelockController
 /// @notice A timelock controller for the StakeTable contract
-contract Timelock is TimelockController {
+contract SafeExitTimelock is TimelockController {
```
sveitser, line 11: It seems a bit strange to have two contracts that are exactly the same just with two different names and create bindings for both. Is there a downside to re-using the contract? If we want different rust types we could create newtype wrappers around the Timelock bindings. @alxiong wdyt?

contracts/test/EspToken.t.sol
```diff
@@ -59,4 +61,215 @@ contract EspTokenUpgradabilityTest is Test {
...
+        timelock.schedule(
+            address(token), 0, transferOwnershipData, bytes32(0), bytes32(0), minDelaySeconds
+        );
+        vm.stopPrank();
+
+        vm.warp(block.timestamp + minDelaySeconds + 1);
+
+        address nonExecutor = makeAddr("nonExecutor");
+        vm.startPrank(nonExecutor);
+        vm.expectRevert();
```
sveitser: Suggestion: check specific revert?

sveitser (approved):
LGTM but if possible we should always check for specific reverts to make sure the calls fail for the right reasons. And one concern with duplication when using the exact same contract.

contracts/test/EspToken.t.sol
sveitser: Also for other cases of bare `expectRevert()`.

contracts/src/SafeExitTimelock.sol
alysiahuggins: i agree that using one would be better from a code maintenance pov. I was optimizing for explorers and multisig UIs so that the viewer can easily distinguish between contracts via the deployed name (i.e. seeing `SafeExitTimelock` on the explorer and `OpsTimelock` for their respective contract addresses). Also this allows for any implementation divergence in the future.

alysiahuggins: tagging @alxiong for visibility

sveitser: Ok. You mean when someone looks at the verified source code on etherscan, for example? Or where would this name show up? It's not part of the contract binary.

alxiong: >If we want different rust types we could create newtype wrappers around the Timelock bindings. @alxiong wdyt?

yes, that's possible. i also tend to agree with reducing solidity duplication, esp they are exactly the same bytecode. 

```rust
// alloy generated bindings
pub struct TimeLock {}

// in our sol_types.rs, we can define wrapper
#[derive(Deref)]
pub struct SafeExitTimeLock(TimeLock);
```

then all of `TimeLock`'s methods should be available directly from the outer wrapper. 

since we manage deployments in rust now, this rust-level name/type differentiation is sufficient.

alysiahuggins: we spoke on slack and agreed to have the separate contracts so that they are more easily distinguished on contract UIs such as blockchain explorers etc.

PR #3403, opened 2025-06-22
feat: ledger support in deploy rust binary

sveitser:
This PR adds ledger signer support to the rust deploy binary. With a script for interactive testing.

### DIFF
--- a/contracts/rust/deployer/src/lib.rs
+++ b/contracts/rust/deployer/src/lib.rs
@@ -17,7 +17,10 @@ use alloy::{
         Provider, ProviderBuilder, RootProvider,
     },
     rpc::{client::RpcClient, types::TransactionReceipt},
-    signers::local::{coins_bip39::English, MnemonicBuilder, PrivateKeySigner},
+    signers::{
+        ledger::LedgerSigner,
+        local::{coins_bip39::English, MnemonicBuilder, PrivateKeySigner},
+    },
     transports::http::reqwest::Url,
 };
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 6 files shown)


contracts/rust/deployer/src/provider.rs
```diff
@@ -0,0 +1,37 @@
+use alloy::signers::ledger::{HDPath, LedgerError, LedgerSigner};
+use anyhow::{bail, Result};
+
+/// Try to obtain a ledger signer
+///
+/// Handles some common errors by prompting the user.
+pub async fn connect_ledger(account_index: usize) -> Result<LedgerSigner> {
```
sveitser, line 7: This code is moved here from the staking-cli

contracts/rust/deployer/src/lib.rs
```diff
@@ -289,6 +316,7 @@ pub async fn verify_node_js_files() -> Result<()> {
         Some(true),
     )
     .await?;
+    tracing::info!("Ok!");
```
imabdulbasit: can we add a better info message here?

imabdulbasit (approved):

contracts/rust/deployer/src/lib.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/3403/commits/8d63d107431e8fc5bf689efb5aa78422bd8c4d9f

alxiong (commented):
Changes also looks good to me. just some relocating and a straightforward script.

PR #3404, opened 2025-06-22
WIP: forge script with ledger support

sveitser:
UX of our deploy script isn't great with ledger. We don't have descriptive logs about what we're asking to sign. I wondered how forge does in this regard so made this deploy script.

### DIFF
--- a/contracts/script/EspToken.s.sol
+++ b/contracts/script/EspToken.s.sol
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import { Script } from "forge-std/Script.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import { EspToken } from "../src/EspTokenV2.sol";
+
+/// @notice Deploys an upgradeable Fee Contract using the OpenZeppelin Upgrades plugin.
+///
+/// To deploy with the ledger, add the --ledger flag to the forge script command:
+///
+/// forge script --ledger DeployEspTokenScript --broadcast
+///
+/// To use a mnemonic
+/// echo test test test test test test test test test test test junk | tr -d '\n' > .mnemonic
+/// forge script --mnemonics .mnemonic DeployEspTokenScript --broadcast
+contract DeployEspTokenScript is Script {
+    string internal contractName = "EspTokenV2.sol";
+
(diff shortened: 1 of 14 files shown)


contracts/script/EspToken.s.sol
```diff
@@ -0,0 +1,99 @@
...
+///
+/// To deploy with the ledger, add the --ledger flag to the forge script command:
+///
+/// forge script --ledger DeployEspTokenScript --broadcast
+///
+/// To use a mnemonic
+/// echo test test test test test test test test test test test junk | tr -d '\n' > .mnemonic
+/// forge script --mnemonics .mnemonic DeployEspTokenScript --broadcast
+contract DeployEspTokenScript is Script {
+    string internal contractName = "EspTokenV2.sol";
```
alysiahuggins: i guess this will change pending on the decision from the thread

contracts/script/EspToken.s.sol
```diff
@@ -0,0 +1,99 @@
...
+            address owner,
+            address initialRecipient,
+            uint256 initialSupply,
+            string memory name,
+            string memory symbol
+        )
+    {
+        vm.startBroadcast();
+
+        owner = vm.envAddress("ESP_TOKEN_INITIAL_OWNER");
```
alysiahuggins: is this initial owner expected to be an EOA or the multisig? if the former then we have to add transfer ownership code, if the latter we should make it clear that it should be a secure owner like the safe multisig

alysiahuggins (approved):
left two comments but looks good otherwise

contracts/script/EspToken.s.sol
sveitser: Yeah this is from yesterday. Thanks for catching. I will change it.

contracts/script/EspToken.s.sol
sveitser: Yes I think I should add the ownership transfer

sveitser: We can  just set the multisig owner during initialization: https://github.com/EspressoSystems/espresso-network/pull/3404/commits/13db97bf8af1f3783d3d7e59d705d56cff5d4e10

contracts/script/EspToken.s.sol
sveitser: The variable wasn't used but I was deploying the V2. Fixed here: https://github.com/EspressoSystems/espresso-network/pull/3404/commits/13db97bf8af1f3783d3d7e59d705d56cff5d4e10

sveitser:
don't need this anymore

PR #3405, opened 2025-06-23
Upgrade esp v2

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR does not:
* Makes ESP token configuration explicit by removing default values for token name, symbol, and supply
* Adds ESP token V2 upgrade support to the deployment pipeline
* Adds example environment file for ESP token deployment
* Updates Docker Compose and process-compose configs to include ESP token V2 upgrade

### Key places to review:
https://www.notion.so/espressosys/ESP-Token-Deployment-Instructions-2182431b68e98016a66ac7383a0648f0

---
  - https://app.asana.com/0/0/1210253168280786


process-compose.yaml
```diff
@@ -38,12 +38,16 @@ processes:
   deploy-sequencer-contracts:
     # The contract addresses are implicitly inherited from .env. We need to unset these or else the
     # script will think they're already deployed.
+    environment:
+      - ESP_TOKEN_NAME="Espresso"
```
sveitser: process compose should pass all values from `.env` to all services already. So I think these shouldn't be needed.

sequencer/src/bin/deploy.rs
```diff
@@ -209,16 +212,16 @@ struct Options {
     #[clap(long, env = "ESPRESSO_SEQUENCER_EPOCH_START_BLOCK")]
     epoch_start_block: Option<u64>,
     /// The initial supply of the tokens.
-    #[clap(long, env = "ESP_TOKEN_INITIAL_SUPPLY", default_value_t = U256::from(3590000000u64))]
-    initial_token_supply: U256,
+    #[clap(long, env = "ESP_TOKEN_INITIAL_SUPPLY")]
+    initial_token_supply: Option<U256>,
```
sveitser, line 219: I think I'm a bit lost as to why we're making this change to remove the defaults.

alysiahuggins: The goal behind removing the default value is to force explicit configuration, either via CLI or environment variable. The idea is that the config should always be correct and intentional, and we shouldn’t fall back to a hardcoded default that might silently cause incorrect behavior. Can revert if you don't agree

sveitser: I think that's fine. In general I have a bit of a hard time imagining needing configurability for the token name and symbol. If it's not needed it's usually better to not have it.

I think I might be more lost with the purpose of the current V2 contract, since it only overrides the token name with a hardcoded value in the contract. W haven't deployed V2 on decaf and will deploy V1 on mainnet

So I'm wondering if we should make the V2 a placeholder contract for now and develop the rewards withdrawal into it that contract later,

The PR can just mostly stay as it is, it's nice that we have the infra to deploy the new contract already.

alysiahuggins: there was a conversation in zulip about the v2 contract and how to ensure that we can have the correct name in both environments without the contract code being different across two envs (as the sepolia contract was already deployed and the name change was requested after). So this was the solution discussed and accepted.

will message you on slack to understand what changes you'd like to see before this is merged

sveitser (approved):

process-compose.yaml
alysiahuggins: thanks, adjusted

sequencer/src/bin/deploy.rs
alysiahuggins: we confirmed that it's ok to leave as is

alysiahuggins:
note, we will not be deploying esp v2 but we have the upgrade code available via this PR

PR #3407, opened 2025-06-23
add transaction stream endpoints

imabdulbasit:
This PR adds two endpoints to stream transactions: one streams all transactions, and the other streams transactions for a specific namespace. Internally, it uses the block stream to get transactions from each block and filters them by namespace if specified. It also includes a test that submits transactions for multiple namespaces and asserts that both endpoints return the expected results.

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -2051,7 +2051,10 @@ mod test {
     use hotshot_contract_adapter::sol_types::{EspToken, StakeTableV2};
     use hotshot_example_types::node_types::EpochsTestVersions;
     use hotshot_query_service::{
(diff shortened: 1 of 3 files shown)


sequencer/src/api.rs
```diff
@@ -4982,4 +4985,125 @@ mod test {
...
+                break;
+            }
+        }
+
+        // Validate namespace-specific stream endpoint
+
+        for (namespace, expected_ns_txns) in &namespace_tx {
+            let mut api_namespace_txns = client
+                .socket(&format!(
+                    "availability/stream/transactions/0/namespace/{}",
```
pls148: "/namespace/{namespace}" instead of separate parameter

PR #3409, opened 2025-06-23
Add Validator tracking, adjust Stake Table tracking, add support to retrieve both to Client

Ayiga:
### This PR:

Adds a few features to the `node-metrics` (`node-validator` service)

* Adds ability to retrieve `StakeTable` snapshot
* Adds ability to subscribe to `StakeTable` updates
* Adds ability to retrieve `Validator` list snapshot
* Adds ability to subscribe to `Validator` updates

In order to support the ability to Delegate Stake to validators the client needs to be able to reliably retrieve information about the current staking details, as well as to be able to see the current Epcoh state.  This applies, generally, to the `StakeTable` (for the current and potential upcoming epochs), as well as to the Total Validator Set (Current Epoch's Active Set, upcoming Epoch's active set, and current L1 Total Active Set).

As of this implementation, we only have the ability to track the Current Epoch's Active Validator Set, and the Current epoch's Stake Table.  Retrieving / deriving the other states is left as future work.

### This PR does not:

Affect the existing implementation in any meaningful way.

### Key places to review:

* node-metrics/src/service/data_state/mod.rs

Changes to the `DataState` have been made to broadcast the current epcoh `StakeTable`.  Additionally, changes have been made to retrieve the active Validator state.

### DIFF
--- a/node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
+++ b/node-metrics/src/api/node_validator/v0/create_node_validator_api.rs
@@ -123,13 +123,12 @@ pub async fn create_node_validator_processing(
     let hotshot_config = get_config_stake_table_from_sequencer(hotshot_client.clone())
         .await
         .map_err(CreateNodeValidatorProcessingError::FailedToGetStakeTable)?;
-    let mut stake_table = hotshot_config.known_nodes_with_stake.clone();
+    let mut stake_table = hotshot_config.known_nodes_with_stake();
     let mut validator_map = IndexMap::new();
 
-    if let (Some(epoch_starting_block), Some(num_blocks_per_epoch)) = (
-        hotshot_config.epoch_start_block,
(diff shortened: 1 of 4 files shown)


imabdulbasit (approved):
I had few questions which @Ayiga and I discussed over a call. 

There could be some future work, such as tracking all validators instead of just the active ones, which would require a new endpoint that returns the unfiltered validators set. We also discussed whether validator exits should be tracked, which would be possible once we have this endpoint. 

Additionally, it might be better for node metrics to listen to an endpoint that streams all stake table events, so that it can display all the info without waiting for consensus side validators set as it is delayed by 2 epochs, and storing the whole validator set for every epoch can take up lot of memory

It might also be helpful to  display the leader for a block, which would require adding a new endpoint( I think we can just look the membership lookup_leader()). However, this could be addressed in separate PRs, as it requires changes on the confirmation layer side.

PR #3413, opened 2025-06-24
Add tests for restarting after epoch upgrade

ss-es:
Adds HotShot integration tests for restarting:

- after the epoch upgrade, but before the first epoch
- after the first epoch, but before the second epoch


crates/hotshot/testing/tests/test_upgrade_restart.rs
```diff
@@ -0,0 +1,118 @@
...
+        for i in 0..7 {
+            catchup_nodes.push(ChangeNode {
+                idx: i,
+                updown: NodeAction::RestartDown(0),
+            })
+        }
+
+        let spinning_properties = SpinningTaskDescription {
+            // Restart all the nodes in view 40, i.e. after the upgrade is decided but before the first epoch
+            node_changes: vec![(40, catchup_nodes)],
```
bfish713: Can we restart between decided and applied, if this isn't already

PR #3414, opened 2025-06-24
Fix HotShotConfig migration

ss-es:
Fixes internal hotshot config migration in storage

### DIFF
--- a/sequencer/src/lib.rs
+++ b/sequencer/src/lib.rs
@@ -317,12 +317,12 @@ where
         network_params.config_peers,
     ) {
         (Some(config), _) => {
-            tracing::info!("loaded network config from storage, rejoining existing network");
+            tracing::warn!("loaded network config from storage, rejoining existing network");
             (config, false)
         },
         // If we were told to fetch the config from an already-started peer, do so.
         (None, Some(peers)) => {
-            tracing::info!(?peers, "loading network config from peers");
+            tracing::warn!(?peers, "loading network config from peers");
             let peers = StatePeers::<SequencerApiVersion>::from_urls(
                 peers,
                 network_params.catchup_backoff,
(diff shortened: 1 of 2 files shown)


rob-maron (approved):

sequencer/api/migrations/postgres/V1001__add_drb_params.sql
```diff
@@ -0,0 +1,12 @@
+-- HotShotConfig was upgraded to include parameters for stake table capacity and DRB difficulty. Configs
+-- which were persisted before this upgrade may be missing these parameters. This migration
+-- initializes them with a default. We use the `||` operator to merge two JSON objects, one
+-- containing default values for the new config parameters and one containing the existing config.
+-- When keys are present in both, the rightmost operand (the existing config) will take precedence.
+UPDATE network_config SET
+    config = jsonb_set(config, '{config}', '{
+        "stake_table_capacity": 200,
+        "drb_difficulty": 0,
+        "drb_upgrade_difficulty": 0,
```
imabdulbasit:
```suggestion
        "drb_upgrade_difficulty": 0
```

sequencer/src/catchup.rs
```diff
@@ -236,10 +236,14 @@ impl<ApiVer: StaticVersionType> StatePeers<ApiVer> {
             .retry(self, move |provider, retry| {
                 let my_own_validator_config = my_own_validator_config.clone();
                 async move {
-                    let cfg = provider
+                    let cfg: PublicNetworkConfig = provider
                         .fetch(retry, |client| {
-                            client.get::<PublicNetworkConfig>("config/hotshot").send()
```
imabdulbasit, line 241: I also think that this is not necessary now that we have migration, but it doesn't make any difference

PR #3416, opened 2025-06-27
rust: edition: 2021 -> 2024

sveitser:
I don't know how to update the edition but have cargo fmt stay on 2021 for easier review of the edition changes. 

Was also planning on just upgrading the edition but because the CI is already on 1.88 needed to fix clippy for 1.88 anyway.

My suggestion is to review commits excluding `cargo fmt` individually. If you want to verify `cargo fmt` changes, checking out the commit before and running `cargo fmt` then diffing against the next commit should be much faster than actually reading the diff.

Closes #2663


sequencer/src/bin/nasty-client.rs
```diff
@@ -528,7 +528,7 @@ impl<T: Queryable> ResourceManager<T> {
                     // be sure the endpoint is healed.
                     tracing::info!("succeeded after {i} retries; checking health");
                     for _ in 0..self.cfg.min_retries {
-                        f().into_future().await.context(
+                        TryFutureExt::into_future(f()).await.context(
```
sveitser, line 531: This an similar changes are because there are 3 traits that implement `into_future` in scope now.

crates/hotshot/hotshot/src/types/handle.rs
```diff
@@ -87,7 +87,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions>
     }
 
     /// obtains a stream to expose to the user
-    pub fn event_stream(&self) -> impl Stream<Item = Event<TYPES>> {
+    pub fn event_stream(&self) -> impl Stream<Item = Event<TYPES>> + use<TYPES, I, V> {
```
sveitser, line 90: This change (and similar changes) are due to this: https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

sequencer/src/api/options.rs
```diff
@@ -541,7 +541,7 @@ impl Options {
         port: u16,
         app: App<S, E>,
         bind_version: ApiVer,
-    ) -> impl Future<Output = anyhow::Result<()>>
+    ) -> impl Future<Output = anyhow::Result<()>> + use<S, E, ApiVer>
```
imabdulbasit, line 544: why is this needed now?

sequencer/src/bin/utils/keygen.rs
```diff
@@ -27,11 +27,11 @@ enum Scheme {
 }
 
 impl Scheme {
-    fn gen(self, seed: [u8; 32], index: u64, env_file: &mut impl Write) -> anyhow::Result<()> {
+    fn r#gen(self, seed: [u8; 32], index: u64, env_file: &mut impl Write) -> anyhow::Result<()> {
```
imabdulbasit: maybe we rename this function to generate()?

sequencer/src/api/options.rs
sveitser: See my previous comment. 

https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

sequencer/src/bin/utils/keygen.rs
sveitser: 6fa1213b9a986a2ce53ad474bcaa2e452bffe45c

crates/hotshot/utils/src/anytrace/macros.rs
```diff
@@ -23,13 +23,13 @@ macro_rules! trace {
         message: format!("{}: {}", line_info!(), format!($message))
       }
   };
-  ($error:expr) => {
+  ($error:expr_2021) => {
```
imabdulbasit: why the `expr` doesn't work here? Why do we need to use the `expr_2021` explicitly?

vid/src/utils/bytes_to_field.rs
```diff
@@ -105,11 +105,11 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         let mut elem_bytes = Vec::with_capacity(self.elem_byte_capacity);
         for _ in 0..elem_bytes.capacity() {
-            if let Some(byte) = self.bytes_iter.next() {
+            match self.bytes_iter.next() { Some(byte) => {
                 elem_bytes.push(*byte.borrow());
-            } else {
+            } _ => {
```
imabdulbasit: using `None` would be better here than `_`?

sveitser: It seems we can still compile even if we revert that automatic fix. I don't think it should change the behavior if we leave it as it was, so I reverted the change.

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -180,41 +180,46 @@ impl<K: SignatureKey + 'static, D: DhtPersistentStorage> DHTBehaviour<K, D> {
...
-                    retry_count: retry_count - 1,
-                    records: Vec::new(),
-                };
-
-                // Add the key to the outstanding queries and in-progress queries
-                self.outstanding_dht_query_keys.insert(key, qid);
-                self.in_progress_record_queries.insert(qid, query);
-            }
+            },
+            _ => {
```
imabdulbasit: replace it with `None`?

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -317,36 +322,43 @@ impl<K: SignatureKey + 'static, D: DhtPersistentStorage> DHTBehaviour<K, D> {
...
-                            progress: DHTProgress::NotStarted,
-                            notify,
-                            key,
-                            retry_count: new_retry_count,
-                            records: Vec::new(),
-                        });
-                    }
-                    warn!("Get DHT: Internal disagreement for get dht request {progress:?}! Giving up because out of retries. ");
+                    },
+                    _ => {
```
imabdulbasit: here as well

crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -776,7 +795,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> QuorumVoteTaskS
                 }
                 self.vote_dependencies = current_tasks;
             },
-            HotShotEvent::ViewChange(mut view, _) => {
+            &HotShotEvent::ViewChange(mut view, _) => {
```
imabdulbasit, line 798: why it requires the reference now?

imabdulbasit (approved):

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
sveitser: I undid this whole change in 62ca4691ad393dec7012ed794cea57c7a0e44a8a

crates/hotshot/task-impls/src/quorum_vote/mod.rs
sveitser: Because we want `mut view`?

crates/hotshot/libp2p-networking/src/network/behaviours/dht/mod.rs
sveitser: also reverted this if let -> match change

crates/hotshot/utils/src/anytrace/macros.rs
sveitser: everything does seem to work with `expr`, thanks

sveitser:
`slow-tests-sqlite` timeout after 40 minutes

sveitser:
done now

PR #3417, opened 2025-06-27
Timelock proposals

alysiahuggins:
Closes #<ISSUE_NUMBER>


contracts/rust/deployer/src/builder.rs
```diff
@@ -77,6 +96,22 @@ pub struct DeployerArgs<P: Provider + WalletProvider> {
...
+    #[builder(default)]
+    timelock_operation_delay: Option<U256>,
+    #[builder(default)]
+    timelock_operation_function_signature: Option<String>,
+    #[builder(default)]
+    timelock_operation_function_values: Option<Vec<String>>,
+    #[builder(default)]
+    timelock_operation_salt: Option<String>,
+    #[builder(default)]
+    timelock_owner: Option<bool>,
```
sveitser: Something like 

```suggestion
    use_timelock_owner: Option<bool>,
```

? 

I was reading code and thought this is an address.

alysiahuggins: updated the code [change flag from timelock_owner to use_timelock_owner](https://github.com/EspressoSystems/espresso-network/pull/3417/commits/18dc140413d8c7db9dc417e6e442c201b96e2ebd)

contracts/rust/deployer/README.md
```diff
@@ -0,0 +1,569 @@
...
+The code sets the OpsTimelock as the owner of the FeeContract
+
+### Deploying with Cargo
+
+```bash
+set -a
+source .env
+set +a
+unset ESPRESSO_SEQUENCER_FEE_CONTRACT_PROXY_ADDRESS
+RUST_LOG=info cargo run --bin deploy -- --deploy-ops-timelock --deploy-fee --timelock-owner --rpc-url=$RPC_URL
```
sveitser: I think here and for other invocations it should now be

```suggestion
RUST_LOG=info cargo run --bin deploy -- --deploy-ops-timelock --deploy-fee --use-timelock-owner --rpc-url=$RPC_URL
```

contracts/rust/deployer/src/builder.rs
```diff
@@ -116,13 +176,24 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
                 )
                 .await?;
 
-                // NOTE: we don't transfer ownership to multisig, we only do so after V2 upgrade
+                if let Some(use_timelock_owner) = self.use_timelock_owner {
+                    tracing::info!("Transferring ownership to SafeExitTimelock");
+                    // deployer is the timelock owner
+                    if use_timelock_owner {
+                        let timelock_addr = contracts
+                            .address(Contract::SafeExitTimelock)
```
sveitser, line 192: Maybe a brief comment as to why the token one uses the `SafeExitTimelock` ?

contracts/rust/deployer/src/builder.rs
```diff
@@ -363,6 +478,122 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+            B256::from_hex(stripped).context("Invalid salt hex format")?
+        } else {
+            B256::from_hex(&salt).context("Invalid salt hex format")?
+        };
+
+        let timelock_operation_data = TimelockOperationData {
+            target: target_addr,
+            value,
+            data: function_calldata,
+            predecessor: B256::ZERO, // Default to no predecessor
```
sveitser, line 576: The `predecessor` is for chaining operations, or?

alysiahuggins: yes for chaining operations

contracts/rust/deployer/src/lib.rs
```diff
@@ -1494,6 +1105,48 @@ pub async fn deploy_safe_exit_timelock(
...
 
+/// Encode a function call with the given signature and arguments
+///
+/// Parameters:
+/// - `signature`: e.g. `"transfer(address,uint256)"`
+/// - `args`: Solidity typed arguments as `Vec<&str>`
+///
+/// Returns:
+/// - Full calldata: selector + encoded arguments
+pub fn encode_function_call(signature: &str, args: Vec<String>) -> Result<Bytes> {
```
sveitser, line 1110: I read this function and was a bit surprised we have to manually dig into the function string. So was wondering if this is yet another hard to find thing in alloy.

It seems that if we go via `Function::parse` which we use later anyway we can do it without looking into the arguments. I haven't tested this but it looks okay and compiles.

```rust
pub fn encode_function_call(signature: &str, args: Vec<String>) -> anyhow::Result<Bytes> {
    let func = Function::parse(signature)?;

    // Parse argument types from the function signature
    let arg_types: Vec<DynSolType> = func
        .inputs
        .iter()
        .map(|param| param.resolve())
        .collect::<Result<Vec<_>, _>>()
        .map_err(|e| anyhow!("Failed to resolve parameter type: {e}"))?;

    // Validate argument count
    if args.len() != arg_types.len() {
        anyhow::bail!(
            "Mismatch between argument count ({}) and type count ({})",
            args.len(),
            arg_types.len()
        );
    }

    // Parse and encode arguments in one step
    let dyn_values: Vec<DynSolValue> = args
        .iter()
        .zip(&arg_types)
        .map(|(arg, ty)| ty.coerce_str(arg))
        .collect::<Result<_, _>>()
        .map_err(|e| anyhow!("Failed to coerce argument: {e}"))?;

    Ok(Bytes::from(func.abi_encode_input(&dyn_values)?))
}

```

contracts/rust/deployer/src/lib.rs
```diff
@@ -2526,6 +2190,96 @@ mod tests {
...
+        assert!(timelock.isOperationDone(operation_id).call().await?._0);
+        assert!(!timelock.isOperationPending(operation_id).call().await?._0);
+        assert!(!timelock.isOperationReady(operation_id).call().await?._0);
+        // check that the new owner is the provider_wallet
+        let fee_contract = FeeContract::new(operation.target, &provider);
+        assert_eq!(fee_contract.owner().call().await?._0, provider_wallet);
+
+        operation.value = U256::from(1);
+        //transfer ownership back to the timelock
+        let _ = fee_contract.transferOwnership(timelock_addr).send().await?;
```
sveitser: I believe this doesn't wait for the transaction to be mined so the next tx might sometimes fail due to race conditions. At least we've run into this many times with ethers before. To avoid those race conditions we can fetch the tx receipt.

contracts/rust/deployer/README.md
alysiahuggins: ooh thanks for that https://github.com/EspressoSystems/espresso-network/pull/3417/commits/25cb729dcd1e0becacbf5862da312715a1dbe6cd

contracts/rust/deployer/src/builder.rs
alysiahuggins: good idea, https://github.com/EspressoSystems/espresso-network/pull/3417/commits/ff03381d4c77281a8a0d2ea8670ba59d6eef534c

contracts/rust/deployer/src/lib.rs
alysiahuggins: good catch! you're right that Function::parse already gives us the inputs, so we can avoid manually parsing the argument types. I initially split the signature to get the types for DynSolType construction before realizing Function::parse could work there too. I'll update!

contracts/rust/deployer/src/lib.rs
alysiahuggins: ahhh ok thanks for that https://github.com/EspressoSystems/espresso-network/pull/3417/commits/6991b5e225b76b964acd16ae9cf47b8f2272890f

sveitser (approved):
I think this is fine but it was really hard to see what were changes vs what code was just moved around. Github also fails to show that. I also think we shouldn't create extra modules so made a PR to remove that along with the wildcard imports.

contracts/rust/deployer/src/lib.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3417/commits/4ffa9e0a21d76de724f428bd2652216ff2be50c3

PR #3423, opened 2025-06-30
Limit retries when fetching stake table events and panic after max duration

imabdulbasit:
addresses: https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1209524344994408

This PR replaces the infinite retry loop used when fetching stake table events with a maximum retry duration (defaulting to 15 minutes). If the operation continues to fail beyond this limit, it panics with a message explaining the potential caus such as exceeding the RPC  range limits due to large block ranges or a large number of events returned. It also suggests a possible solution which is reducing the block range


types/src/v0/impls/stake_table.rs
```diff
@@ -1260,6 +1277,41 @@ impl Fetcher {
...
+                    panic!(
+                        "\nFailed to complete operation `{operation_name}` after 15 minutes.\n\
+                        error: {err}\n\n\
+                         This might be caused by:\n\
+                           - The current block range being too large for your RPC provider.\n\
+                           - The event query returning more data than your RPC allows as some RPC providers limit the number of events returned.\n\n\
+                         Suggested solution:\n\
+                           - Reduce the value of the environment variable `ESPRESSO_SEQUENCER_L1_EVENTS_MAX_BLOCK_RANGE` \
+                             to query smaller ranges.\n\
+                           - Use a different RPC provider with higher rate limits."
```
sveitser: Another option is that the RPC is having an outage. Let's also mention that multiple RPC providers can be configured.

types/src/v0/impls/stake_table.rs
```diff
@@ -1260,6 +1277,41 @@ impl Fetcher {
...
+    E: std::fmt::Display,
+{
+    let start = Instant::now();
+    loop {
+        match operation().await {
+            Ok(result) => return result,
+            Err(err) => {
+                if start.elapsed() >= max_duration {
+                    panic!(
+                        "\nFailed to complete operation `{operation_name}` after 15 minutes.\n\
```
sveitser: 15 minutes should be `max_duration`.

Also small thing but why not use a raw string for this panic? To avoid all the newlines.

imabdulbasit: have to manually set the indentation

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3423/commits/7171d87c6553ec37c73166b76a82cb6d7e590d70

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3423/commits/7171d87c6553ec37c73166b76a82cb6d7e590d70

types/src/v0/impls/stake_table.rs
```diff
@@ -777,87 +781,94 @@ impl Fetcher {
...
+        stream::iter(events.into_iter().filter(|(event, log)| {
+            if let Err(e) = event.authenticate() {
+                tracing::warn!(%e, "Failed to authenticate ValidatorRegisteredV2 event: {}", log.display());
+                return false;
             }
-        });
+            true
+        }))
+    }
+});
```
sveitser: This looks weird.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3423/commits/52566539a7245ac6760923025f5b48efa769836b

types/src/v0/impls/stake_table.rs
```diff
@@ -2596,4 +2667,34 @@ mod tests {
             "Stake table reconstructed from events does not match the expected stake table "
         );
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[should_panic]
```
sveitser, line 2677: It feels a bit wrong that we have to connect to a third party RPC to check if this works. But it's better than what we have now.

types/src/v0/impls/stake_table.rs
```diff
@@ -777,63 +781,72 @@ impl Fetcher {
...
-                        .await
-                    {
-                        Ok(events) => break stream::iter(events),
-                        Err(err) => {
-                            tracing::warn!(from, to, %err, "ValidatorRegistered Error");
-                            sleep(retry_delay).await;
-                        },
-                    }
-                }
+                let events = retry(
```
sveitser, line 797: We still have an infinite loop on the `initializedAtBlock` call. Should we panic there too? It doesn't involve the L1 events range but if the RPC is misconfigured or broken it would now hang forever.

imabdulbasit: I think nothing would work if rpc is broken or misconfigured because we need it to propose and validate as well

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3423/commits/1960a75696840da31d83195908f1d5a6b70a7754

types/src/v0/impls/stake_table.rs
imabdulbasit: yeah, the decaf stake table events test also uses public node so I just decided to use it here as well. I am not sure if Anvil has the limits for querying data

PR #3426, opened 2025-07-01
Fetch block reward from contract on stake tables reload

imabdulbasit:
This PR adds block reward fetching when reloading stake tables at startup.


ss-es (commented):
I need to think about this a bit more, but I think overall it makes sense

however, I think we should make `block_reward` in `EpochCommittees` an `Option`, just like `first_epoch`, and `has_randomized_stake_table` should ensure that `block_reward` is not `None`

ss-es (changes requested):
unless I'm missing something, I think `has_randomized_stake` should require that `self.block_reward` is not `None`

PR #3428, opened 2025-07-01
Fixes from release-rc-202560623

ss-es:
Merges fixes from testing into main


crates/hotshot/task-impls/src/transactions.rs
```diff
@@ -648,14 +648,6 @@ impl<TYPES: NodeType, V: Versions> TransactionTaskState<TYPES, V> {
                     continue;
                 }
 
-                // verify the message signature and the fee_signature
-                if !header_input.validate_signature(block_info.offered_fee, &block_data.metadata) {
```
sveitser, line 662: Wonder why this has been removed since the change looks potentially dangerous.

ss-es: note that the check already happens earlier in `validate_signature_and_get_input` (take a look at the implementation of this method), *before* the headerinput gets cast to the newer non-legacy version. the effect of this extra check here is that we're possibly validating a legacy signature as a non-legacy signature, which will pretty much always fail

sveitser: Thanks.

PR #3430, opened 2025-07-02
fix: rustfmt not formatting some files

sveitser:
It's a [known issue](https://github.com/rust-lang/rustfmt/issues/3863#issuecomment-678632593) that rustfmt sometimes gives up. It seems that reducing the number of occurences where it can't apply formatting due to long strings helps a lot with this. 

This PR also removes formatting the bindings which are auto-generated and barely human-readable so IMO not necesary to format with rustfmt.


justfile
```diff
@@ -159,9 +159,6 @@ gen-bindings:
       --libraries contracts/src/libraries/PlonkVerifier.sol:PlonkVerifier:0xffffffffffffffffffffffffffffffffffffffff \
       --libraries contracts/src/libraries/PlonkVerifierV2.sol:PlonkVerifierV2:0xffffffffffffffffffffffffffffffffffffffff
 
-    cargo fmt --all
-    cargo sort -g -w
-
```
alxiong, line 164: since we ignore the bindings, why remove these?

sveitser: Just because it's now unrelated code. As in, I think generating the bindings has nothing to do with formatting or sorting cargo files.

alxiong: I see, fair enough.

imabdulbasit (approved):

PR #3431, opened 2025-07-02
availability types reference tests

imabdulbasit:
---
  - https://app.asana.com/0/0/1210698118936746


data/v1/leaf_query_data_legacy.json
```diff
@@ -0,0 +1,60 @@
...
+      "signatures": null,
+      "view_number": 0,
+      "vote_commitment": "COMMIT~u_V3iRAPFTMyDkPw5CZvLnWzDrnnl68rtJfYSaaM_2iY"
+    },
+    "parent_commitment": "COMMIT~AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB",
+    "upgrade_certificate": null,
+    "view_number": 0
+  },
+  "qc": {
+    "_pd": null,
```
sveitser, line 52: Wonder why we serialize PhantomData fields but it would probably break some stuff if we removed it now?

types/src/reference_tests.rs
```diff
@@ -509,3 +525,93 @@ async fn test_reference_ns_proof_enum_advz() {
...
+    let instance_state = NodeState::default();
+    let leaf =
+        LeafQueryDataLegacy::<SeqTypes>::genesis::<TestVersions>(&validated_state, &instance_state)
+            .await;
+    reference_test_without_committable("v1", "leaf_query_data_legacy", &leaf);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_leaf_query_data_legacy_v2() {
+    let validated_state = ValidatedState::default();
```
sveitser, line 542: There are quite a few fields where optional types are not present (so we get a `null`), or values are otherwise empty (like the commitments). I think it would be worth it to have non-default types here otherwise I think there are cases where changes might go unnoticed. Wdyt @imabdulbasit ?

imabdulbasit: I did think about it, but its not easy to create a new non- genesis`Leaf`. The only way I could create it is from `QuorumProposal` which has public fields. but then I looked at message_compat tests and it already is testing some fields I was adding. Anyways I still think it is better to just add those fields even if message compat tests already test them such as `UpgradeCertificate`

data/v1/leaf_query_data_legacy.json
imabdulbasit: yeah, its going to break stuff now. I agree we could have skipped serialization

types/src/reference_tests.rs
sveitser: I just noticed that the `...2` types are missing from message compat tests. We should probably invest some time into creating some tooling to build "non-default" test data. But this PR is already a good addition so merge when you think it's ready.

sveitser (approved):

types/src/reference_tests.rs
imabdulbasit: do you mean the v2 types?

imabdulbasit: I think yeah now we should have 2 message compat tests

imabdulbasit: oh we already have 3 message compat tests for different versions , but the v2 message types are missing

PR #3433, opened 2025-07-03
fix: staking-cli needlessly require token address

sveitser:
solution: fetch it from the stake table contract

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -1029,14 +1029,10 @@ pub mod test_helpers {
             let stake_table_address = contracts
                 .address(Contract::StakeTableProxy)
                 .expect("StakeTableProxy address not found");
-            let token_addr = contracts
-                .address(Contract::EspTokenProxy)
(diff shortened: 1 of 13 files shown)


staking-cli/tests/cli.rs
```diff
@@ -185,12 +185,59 @@ async fn test_cli_contract_revert(#[case] version: StakeTableContractVersion) ->
...
+        )
+        .arg("--commission")
+        .arg("12.34")
+        .output()?;
+    match signer {
+        Signer::Mnemonic => {
+            result.assert_success();
+        },
+        Signer::BrokeMnemonic => {
+            result.assert_failure();
```
sveitser, line 225: I guess I should inspect what error we throw here

sveitser: https://github.com/EspressoSystems/espresso-network/pull/3433/commits/3897132e12648d9fe9032b5f7e8b38df9a9c09a4

PR #3439, opened 2025-07-08
View Sync split byzantine test

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Implements [a byzantine scenario described here](https://www.notion.so/espressosys/Split-view-sync-II-2242431b68e98071bde7fee0b6b0d69c).

### Key places to review:

---
  - https://app.asana.com/0/0/1210134443040116
  - https://app.asana.com/0/0/1210797881811039

### DIFF
--- a/crates/hotshot/task-impls/src/view_sync.rs
+++ b/crates/hotshot/task-impls/src/view_sync.rs
@@ -571,28 +571,36 @@ impl<TYPES: NodeType, V: Versions> ViewSyncTaskState<TYPES, V> {
         Ok(())
     }
 
+    /// Garbage collect tasks for epochs older than the highest finalized epoch
+    /// or older than the previous epoch, whichever is greater.
+    /// Garbage collect views older than the highest finalized view including the highest finalized view.
     async fn garbage_collect_tasks(&self) {
+        let previous_epoch = self
+            .cur_epoch
+            .map(|e| e.saturating_sub(1))
+            .map(TYPES::Epoch::new);
+        let gc_epoch = self.highest_finalized_epoch_view.0.max(previous_epoch);
         Self::garbage_collect_tasks_helper(
             &self.replica_task_map,
-            &self.highest_finalized_epoch_view.0,
+            &gc_epoch,
             &self.highest_finalized_epoch_view.1,
         )
         .await;
         Self::garbage_collect_tasks_helper(
             &self.pre_commit_relay_map,
-            &self.highest_finalized_epoch_view.0,
+            &gc_epoch,
             &self.highest_finalized_epoch_view.1,
         )
         .await;
         Self::garbage_collect_tasks_helper(
             &self.commit_relay_map,
-            &self.highest_finalized_epoch_view.0,
+            &gc_epoch,
             &self.highest_finalized_epoch_view.1,
         )
         .await;
         Self::garbage_collect_tasks_helper(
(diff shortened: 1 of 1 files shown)


crates/hotshot/task-impls/src/view_sync.rs
```diff
@@ -239,7 +227,22 @@ impl<TYPES: NodeType, V: Versions> ViewSyncTaskState<TYPES, V> {
...
+                    self.highest_finalized_epoch_view = (epoch, view);
+                } else if view > self.highest_finalized_epoch_view.1 {
+                    tracing::error!(
+                        "We finalized a higher view but the epoch is lower. This should never \
+                         happen. Current highest finalized epoch view: {:?}, new highest \
+                         finalized epoch view: {:?}",
+                        self.highest_finalized_epoch_view,
+                        (epoch, view)
+                    );
+                    return;
```
bfish713: Don't we still want to clear the map after this since this view is for the wrong epoch, and it's finished?

crates/hotshot/task-impls/src/view_sync.rs
```diff
@@ -498,28 +516,45 @@ impl<TYPES: NodeType, V: Versions> ViewSyncTaskState<TYPES, V> {
...
-                    self.last_garbage_collected_view = self.cur_view - 1;
+                    let mut replica_task_map = self.replica_task_map.write().await;
+                    replica_task_map.retain(|e, _| e >= &self.highest_finalized_epoch_view.0);
+                    if let Some(view_map) =
+                        replica_task_map.get_mut(&self.highest_finalized_epoch_view.0)
+                    {
+                        view_map.retain(|v, _| v > &self.highest_finalized_epoch_view.1)
+                    };
+                    replica_task_map.retain(|_, view_map| !view_map.is_empty());
+                    drop(replica_task_map);
```
bfish713: nit: lets make a function to do this since all these maps are the same type and have the same key (type of the task shouldn't matter).

crates/hotshot/task-impls/src/view_sync.rs
```diff
@@ -498,28 +516,45 @@ impl<TYPES: NodeType, V: Versions> ViewSyncTaskState<TYPES, V> {
...
-                            .remove_entry(&(TYPES::View::new(i), epoch));
-                        self.finalize_relay_map
-                            .write()
-                            .await
-                            .remove_entry(&(TYPES::View::new(i), epoch));
-                    }
-
-                    self.last_garbage_collected_view = self.cur_view - 1;
+                    let mut replica_task_map = self.replica_task_map.write().await;
+                    replica_task_map.retain(|e, _| e >= &self.highest_finalized_epoch_view.0);
```
bfish713: It seems like this logic makes more sense to be handled on the ViewSyncFinalizedCertRecv or Formed event

bfish713 (commented):
One thing not covered here, is we need a way to cancel view sync for old views that we're just never going to form a cert.  For example might miss a couple views, enter view sync, then get a QC that pulls them to a newer view.  Assuming no timeouts happen this task will stay in their map forever because nothing gets finalized.

crates/hotshot/task-impls/src/view_sync.rs
lukaszrzasik: Yes, you're right. I'll remove `return`.

PR #3441, opened 2025-07-09
Fix aborting on `handle_quorum_proposal_validated`

ss-es:


bfish713 (approved):

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -169,6 +166,87 @@ pub async fn handle_drb_result<TYPES: NodeType, I: NodeImplementation<TYPES>>(
...
+
+        if has_stake_current_epoch {
+            let computed_result = current_epoch_membership
+                .next_epoch()
+                .await
+                .context(warn!("No stake table for epoch {}", epoch_val + 1))?
+                .get_epoch_drb()
+                .await
+                .clone()
+                .context(warn!("DRB result not found"))?;
```
ss-es, line 235: this should be the only difference with the old version of this function (previously in quorum_vote::handlers)

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -46,90 +43,6 @@ use crate::{
...
-                     proposed DRB result of {proposal_result:?}"
-                )
-            );
-        }
-
-        Ok(())
-    } else {
-        Err(error!("Epochs are not available"))
-    }
-}
```
ss-es, line 131: this function was moved (with minor modifications) to task_impls::helpers

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -181,11 +94,6 @@ pub(crate) async fn handle_quorum_proposal_validated<
         .version(proposal.view_number())
         .await?;
 
-    if version >= V::Epochs::VERSION {
-        // Don't vote if the DRB result verification fails.
-        verify_drb_result(proposal, task_state).await?;
-    }
-
```
ss-es, line 188: we do not abort proposing on `handle_quorum_proposal_validated`, so this check previously did not cause proposing to fail

crates/hotshot/types/src/drb.rs
```diff
@@ -167,6 +167,21 @@ pub async fn compute_drb_result(
...
+        difficulty_level: drb_input.difficulty_level,
+    };
+
+    let store_drb_progress = store_drb_progress.clone();
+    tokio::spawn(async move {
+        if let Err(e) = store_drb_progress(final_drb_input).await {
+            tracing::warn!("Failed to store DRB progress during calculation: {}", e);
+        }
+    });
+
```
ss-es: we now store the final result as a checkpoint

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -437,13 +437,11 @@ where
...
-        // or compute it if it's not available
+        // get the drb result if it's available
         let drb = if let Ok(drb) = drb_membership
-            .get_epoch_drb(transition_block_for_epoch(
-                *(root_epoch + 1),
-                self.epoch_height,
-            ))
+            .next_epoch_stake_table()
+            .await?
+            .get_epoch_drb()
```
ss-es: note that `drb_membership` is the epoch whose root leaf has the DRB result we need (as `next_drb_result` on the transition leaf)

therefore, we need to call `get_epoch_drb` on the *subsequent* epoch

crates/hotshot/types/src/traits/election.rs
```diff
@@ -156,11 +156,8 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
     /// Gets the DRB result for the given epoch
     fn get_epoch_drb(
         _membership: Arc<RwLock<Self>>,
-        _block_height: u64,
```
ss-es, line 159: I think this is weird to call with `block_height`

if the membership implementation tracks the DRB results via block height (which it certainly can) or needs it to perform catchup etc, it should do the calculation itself. fundamentally, I think the interface for `Membership` should use epochs not block height

types/src/v0/impls/stake_table.rs
```diff
@@ -1900,13 +1904,34 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+                    .context(format!("Missing randomized committee for epoch {}", epoch))
+            },
+        };
+
         let stake_table = membership_reader.stake_table(Some(epoch)).clone();
         let success_threshold = membership_reader.success_threshold(Some(epoch));
+
+        let block_height =
+            transition_block_for_epoch(previous_epoch, membership_reader.epoch_height);
+
```
ss-es, line 2213: **PLEASE DOUBLE CHECK THIS**

is `previous_epoch` correct here?

builder/src/non_permissioned.rs
```diff
@@ -66,12 +66,15 @@ pub fn build_instance_state<V: Versions>(
         chain_config,
     );
 
+    let epoch_height = 100;
```
bfish713: I assume this won't cause any problem since we don't actually need the epoch height in the builder but want to double check

crates/hotshot/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -596,13 +596,16 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
         let next_drb_result = if is_epoch_transition(block_header.block_number(), self.epoch_height)
         {
             if let Some(epoch_val) = &epoch {
-                self.consensus
-                    .read()
+                let drb_result = epoch_membership
```
bfish713, line 599: I feel like it might be slightly better to directly call `Membership::get_epoch_drb(epoch)` instead.  I don't think at the moment it matters but I can imagine some case where a node going offline calculates the next DRB but doesn't have the stake table for some reason.  Not a big deal either though

types/src/v0/impls/stake_table.rs
bfish713: Looks correct, my logic is:

- This function is called with the epoch we want a DRB for (the epoch which the DRB result is used)
- The transition blocks for the epoch before this epoch should have the DRB result
- I checked the logic in `EpochMembership::fetch_or_calc_drb_results`.  It starts from the root epoch (epoch we want a DRB result for - 2) increments one time to get DRB membership (membership where the DRB was decided) and then calls this function for the epoch after DRB membership (root + 2, or epoch we want DRB result for - 0). Seems to be expected.

builder/src/non_permissioned.rs
ss-es: I assumed so since we were already using this here -- actually, I meant to replace the initial `100` on line 79 with `epoch_height`; updated

let me see if CI passes if I set this to `genesis.epoch_height.unwrap()`

bfish713 (approved):

PR #3446, opened 2025-07-10
test: stateful property based stake table tests

sveitser:
- Closes https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1210762464879378

### This PR:

- Adds invariant (stateful fuzz) testing for the StakeTableV2 contract using both Foundry and Echidna.
- Splits the contract CI workflow into parallel linting and testing jobs for faster feedback.
- Add summary output to the contract CI workflow: https://github.com/EspressoSystems/espresso-network/actions/runs/16326462885?pr=3446

### This PR does not:
- Make any changes to non-test contracts.

###  Key places to review:
 - `contracts/test/README.md`
 - The newly added test contracts.
 - There is a utils contract for outputing to terminal. IMO it doesn't need to be reviewed carefully.

### How to test

```
env FOUNDRY_PROFILE=quick just contracts-test-invariant
just contracts-test-echidna
```

Introduce bugs in the ST contract (for example, remove some revert) and check if the tests can catch it.

---
  - https://app.asana.com/0/0/1210762464879378


contracts/src/StakeTable.sol
```diff
@@ -241,7 +241,7 @@ contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpg
             revert ZeroAddress();
         }
         token = ERC20(_tokenAddress);
-        lightClient = LightClient(_lightClientAddress);
+        lightClient = ILightClient(_lightClientAddress);
```
sveitser, line 244: @alxiong @alysiahuggins another practical argument why interfaces are better. It seems that this is handled much better by testing tools. Still need to figure out what this does to the bytecode and if this is upgrade safe, and if not, how we should handle this.

sveitser: Nice. It does not change the bytecode of the stake table contract :partying_face:

alxiong: > another practical argument why interfaces are better. It seems that this is handled much better by testing tools.

didn't understand. 
1. wdym by "handled better by testing tool" here? what are we trying to achieve?
2. what other benefits you have in mind? 
3. what's the take away here?

sveitser: Oh. The echidna tool didn't work with the LC contract dependency. Certora also had issues. These go away if we use an interface instead.

alysiahuggins: awesome! nice work @sveitser

sveitser:
@alxiong @alysiahuggins I think this is ready for review now.

contracts/src/interfaces/ILightClient.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface ILightClient {
```
sveitser, line 4: Should we add a more complete interface? That's currently not needed anywhere, but would probably be nice to have for consumers.

contracts/test/MockStakeTableV2.sol
```diff
@@ -0,0 +1,50 @@
...
+            revert InvalidCommission();
+        }
+
+        blsKeys[_hashBlsKey(blsVK)] = true;
+        validators[validator] = Validator({ status: ValidatorStatus.Active, delegatedAmount: 0 });
+
+        emit ValidatorRegisteredV2(validator, blsVK, schnorrVK, commission, blsSig, schnorrSig);
+    }
+
+    function updateConsensusKeysV2(
```
sveitser: This function is not used. Can consider removing.

contracts/test/StakeTableV2PropTestBase.sol
```diff
@@ -0,0 +1,740 @@
...
+        mapping(address validator => EnumerableSet.AddressSet delegators) delegators;
+        mapping(address validator => EnumerableSet.AddressSet actors) pendingWithdrawals;
+    }
+
+    MockStakeTableV2 public stakeTable;
+    MockERC20 public token;
+    MockLightClient public lightClient;
+    IVM public ivm = IVM(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+
+    uint256 public constant INITIAL_BALANCE = 1000000000e18;
```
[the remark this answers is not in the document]
sveitser: No.

foundry.toml
```diff
@@ -26,6 +26,11 @@ ignored_warnings_from = ['contracts/lib']
 optimizer = true
 
 # See more config options https://github.com/foundry-rs/foundry/tree/master/config
+#
+[invariant]
+show_metrics = true
+fail_on_revert = true
```
sveitser, line 32: This is important. It means that if we have an uncaught error in the fuzz tests it would revert. I guess same caveats as with normal foundry testing reverts apply and it would not catch reverts from internal function calls.

contracts/src/interfaces/ILightClient.sol
alxiong: This interface looks too minimal to be useful?

contracts/test/README.md
```diff
@@ -0,0 +1,46 @@
...
+   foundry invariant test, because it's unclear how to achieve it with echidna.
+
+The tests track detailed statistics about function calls, reverts, and state changes to provide insight into test
+coverage and state space exploration.
+
+Key challenges are:
+
+1. Finding good invariants to track. More should be added.
+1. Evolving the state space: a lot of extra tracking is needed to have a reasonable fraction of non-reverting
+   transactions.
```
alxiong, line 31: what does this mean? 
so most fuzz inputs to the invariant tests would cause revert? 
what are these "extra tracking" required?

contracts/echidna.yaml
```diff
@@ -0,0 +1,4 @@
+corpusDir: "contracts/corpus"
```
alxiong, line 1: btw, should we add this to .gitignore?

contracts/test/MockStakeTableV2.sol
alysiahuggins: do you mean not used by the confirmation layer?

contracts/src/StakeTable.sol
```diff
@@ -176,7 +176,7 @@ contract StakeTable is Initializable, InitializedAt, OwnableUpgradeable, UUPSUpg
     ///
     /// @dev Currently unused but will be used for slashing therefore already included in the
     /// contract.
-    LightClient public lightClient;
+    ILightClient public lightClient;
 
     /// The staking token contract.
     ERC20 public token;
```
alysiahuggins, line 182: just curious, why did the LightClient object need to be an interface but not the ERC20 one?

contracts/src/interfaces/ILightClient.sol
alysiahuggins, line 4: This looks good! One small suggestion: since ILightClient was introduced for compatibility with the StakeTable contract and testing tools, it would be helpful to have LightClient explicitly inherit from ILightClient. I'm thinking that this way, the interface doesn't end up as a disconnected piece of code.

```solidity
contract LightClient is ILightClient {
///...
}
```

justfile
```diff
@@ -231,3 +231,15 @@ test-go:
...
+    nix develop .#echidna -c echidna contracts/test/StakeTableV2.echidna.sol --contract StakeTableV2EchidnaTest --config contracts/echidna.yaml {{args}}
+
+contracts-test-forge *args='-vv':
+    forge test --no-match-test "testFuzz_|invariant_" {{args}}
+
+contracts-test-fuzz *args='-vv':
+    forge test --match-test testFuzz {{args}}
+
+contracts-test-invariant *args='-vv':
+    forge test --match-test invariant_ {{args}}
```
alysiahuggins, line 249: do we want to continue with the same naming convention or no need? e.g. testInvariant

contracts/test/README.md
```diff
@@ -0,0 +1,46 @@
+# Contract Tests
+
```
alysiahuggins, line 2: thanks for the readme :). When do you choose to write fuzz vs invariant vs echidna tests?

alxiong: >fuzz vs invariant

fuzz is usually random inputs against a single function; 
invariant is random sequence of calls on a predefined list of functions to ensure invaraints hold afterward.

now, in our fuzz test function, we could invoke multiple contract functions, but this is a fixed flow/sequence of invocation. fuzzing is only mutating the input to our `testFuzz_xx()`

> fuzz vs echidna tests

see [this](https://www.youtube.com/watch?v=QofNQxW_K08&t=2506s), short answer is echidna has more features and smarter corpse-guided fuzzing strategy (?), foundry is more dev friendly and better hooked into normal dev workflow. 

iiuc, [echidna's fuzzing](https://secure-contracts.com/program-analysis/echidna/introduction/how-to-test-a-property.html) is actually closer to foundry's invariant testing. 
you specify a property but echidna will randomly walk through your available `public` functions, you don't have to define a fixed sequence of interaction.

contracts/echidna.yaml
sveitser: yeah, good idea

contracts/test/README.md
sveitser: @alxiong  For example, If you want to do a delegation successfully you need to

1. pick a validator that is registered, but not exited
2. have some tokens
3. delegate less than the amount of tokens you have

If you want to sample efficiently all the tracking needs to be O(1) and all the sampling needs to be O(1). Otherwise we massively slow down the progress as the number of participants grow.

contracts/src/interfaces/ILightClient.sol
sveitser: Add complete ILightClient interface

contracts/test/MockStakeTableV2.sol
sveitser: 07cddf7f21e823fe98f60c14b45cf048f2997ac0

contracts/src/StakeTable.sol
sveitser: I think it's because the contract, resp. its dependencies are a lot more complicated

justfile
sveitser: This `invariant_` prefix is from foundry: https://getfoundry.sh/forge/advanced-testing/invariant-testing/

contracts/src/interfaces/ILightClient.sol
sveitser: I opened an issue for the interface #3482 . I'm not quite sure how we should deal with versions for the interface.

PR #3449, opened 2025-07-11
Add endpoint to get transaction without proof

jbearer:
See https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1210765337377531?focus=true

### DIFF
--- a/hotshot-query-service/api/availability.toml
+++ b/hotshot-query-service/api/availability.toml
@@ -88,7 +88,11 @@ Opens a WebSockets connection and sends a stream of the same data type returned 
 """
 
 [route.get_header]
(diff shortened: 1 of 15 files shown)


hotshot-query-service/api/availability.toml
```diff
@@ -263,8 +275,26 @@ Returns
 ```
 """
 
+[route.get_transaction_no_proof]
```
Sneh1999: nit: should we rename this to `get_transaction_without_proof`

Sneh1999 (approved):
Thanks Jeb for working on this 🙏 
added one nit comment but otherwise looks good, someone from POS team should also review probably

PR #3450, opened 2025-07-14
Dynamic block reward

imabdulbasit:
This PR calculates the block reward based on the reward formula ([details](https://www.notion.so/espressosys/Staking-rewards-node-operator-delegations-2152431b68e980d19d52cdbf3e498269)), and also calculates the average block time that the epoch takes using the header timestamp. It also adds the total_reward_distributed field to the v4 header. This is used to calculate the correct total supply as total supply is now` initial_supply + rewards distributed. `

Since the new formula calculates the stake ratio as **total staked** / **total supply.** If the ratio is lower than **0.01%**,, the block reward is zero as the commission basis point has **0.01%** precision. 

We use 2s block time for nodes upgrading  from V2 to V4 or directly starting on V4. This is because we need two epoch roots to calculate the average block time. The first block in which rewards are distributed would not have the previous epoch root

When the node upgrades from v3 to v4, the total reward distributed in the first block of the new version is calculated as the sum of all fixed block rewards issued since rewards began (starting from the 3rd epoch). This ensures continuity in the reward total even though earlier block header did not have `total_rewards_distributed` field.
Calculation [here](https://github.com/EspressoSystems/espresso-network/pull/3450/files#diff-fec8912c12800f6a2b332856889dde028381ccd1f9082b105143535f43bf9358R563)

---
  - https://app.asana.com/0/0/1210687930753887


data/v4/header.json
```diff
@@ -37,7 +37,8 @@
     "payload_commitment": "AvidMCommit~GVev7Msgzxsj2Kfx6HfqnNoK4USAnZ602eXda3Ic0zrY",
     "reward_merkle_tree_root": "MERKLE_COMM~AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAUAAAAAAAAAAAAAAAAAAAAKA",
     "timestamp": 789,
-    "timestamp_millis": 789000000000
+    "timestamp_millis": 789000000000,
+    "total_reward_distributed": "0x0"
```
sveitser: We don't use hexadecimals for other numbers (like fee_info.amount). Is this choice here on purpose?

sequencer/src/api.rs
```diff
@@ -3643,7 +3649,7 @@ mod test {
         let coordinator = node_state.coordinator;
 
         let membership = coordinator.membership().read().await;
-        let block_reward = membership.block_reward();
+        let block_reward = membership.block_reward(None).expect("block reward is None");
```
sveitser: nit: for expect I think the idea is to put something that was expected, so in this case "not None"

sequencer/src/api.rs
```diff
@@ -3738,6 +3745,276 @@ mod test {
...
+            let total_reward = block_reward.0;
+            let leader_commission_basis_points = U256::from(leader_validator.commission);
+            let calculated_leader_commission_reward = leader_commission_basis_points
+                .checked_mul(total_reward)
+                .context("overflow")?
+                .checked_div(U256::from(COMMISSION_BASIS_POINTS))
+                .context("overflow")?;
+
+            assert!(
+                computed_rewards.leader_commission().0 - calculated_leader_commission_reward
```
sveitser, line 3953: Can this underflow? In practice because we give the round off error funds to the leader probably not, I guess?

sequencer/src/api.rs
```diff
@@ -3738,6 +3745,276 @@ mod test {
...
+            if let EventType::Decide { leaf_chain, .. } = event.event {
+                let height = leaf_chain[0].leaf.height();
+                tracing::info!("Node 0 decided at height: {height}");
+                if height > EPOCH_HEIGHT * 3 {
+                    break;
+                }
+            }
+        }
+
+        // Verify that there are no validators for epoch # 1 and epoch # 2
```
sveitser, line 3821: Not the topic of this PR but it's a bit weird that this API path returns nothing for the first 2 epochs. Should probably have used a more descriptive path.

sequencer/src/persistence/fs.rs
```diff
@@ -1369,7 +1370,7 @@ impl SequencerPersistence for Persistence {
 
         let drb_result_bytes = bincode::serialize(&drb_result).context("serialize drb result")?;
 
-        let file_path = dir_path.join(epoch.to_string()).with_extension("txt");
+        let file_path = dir_path.join(epoch.to_string()).with_extension("bin");
```
sveitser: Do nodes on the old version handle this change gracefully?

types/src/v0/impls/header.rs
```diff
@@ -396,7 +399,7 @@ impl Header {
         mut state: ValidatedState,
         chain_config: ChainConfig,
         version: Version,
-        reward_distributor: Option<RewardDistributor>,
+        total_reward_distributed: Option<RewardDistributor>,
```
sveitser: Feels like the old name was better? If we only need the total passing in just that value would reduce coupling.

types/src/v0/impls/reward.rs
```diff
@@ -478,29 +486,113 @@ impl RewardDistributor {
...
+
+        block_reward
+    } else {
+        instance_state
+            .block_reward(None)
+            .await
+            .with_context(|| format!("fixed block reward is None for epoch {epoch}"))?
+    };
+
+    total_distributed += block_reward.0;
```
sveitser: Maybe something like `previously_distributed`? To better distinguish it from the total after the distribution.

types/src/v0/impls/stake_table.rs
```diff
@@ -1069,6 +1081,21 @@ impl Fetcher {
     /// contract address. We use the stake table contract initialization block as a safe upper bound when scanning
     ///  backwards for the token contract initialization event
     pub async fn fetch_block_reward(&self) -> Result<RewardAmount, FetchRewardError> {
```
sveitser: The comment above is outdated.

types/src/v0/impls/stake_table.rs
```diff
@@ -2708,4 +2957,84 @@ mod tests {
...
+    }
+
+    #[test]
+    fn sanity_check_p_and_rp() {
+        let total_stake = BigDecimal::from_str("1000").unwrap();
+        let total_supply = BigDecimal::from_str("10000").unwrap(); // p = 0.1
+
+        let (p, rp) = calculate_p_and_rp(&total_stake, &total_supply).unwrap();
+
+        assert!(p > BigDecimal::from(0));
```
sveitser, line 3015: These checks seem to be a bit loose. Can we check we correctly compute the values from the original spec?

types/src/v0/impls/stake_table.rs
```diff
@@ -1582,6 +1763,43 @@ impl EpochCommittees {
...
+
+    if p < BigDecimal::from(0) || p > BigDecimal::from(1) {
+        return Err(anyhow::anyhow!("Stake ratio p must be in the range [0, 1]"));
+    }
+
+    let threshold = BigDecimal::from_str("0.01")?;
+    let numerator = BigDecimal::from_str("0.03")?;
+    let two = BigDecimal::from_u32(2).unwrap();
+
+    let inner = two * if p <= threshold { threshold } else { p.clone() };
```
sveitser: This is a bit hard to read... `2 * max(threshold, p)` ?

Also I think it's better to write out `proportion_staked` than use `p` (and `rp`) unless we want to write larger formulas on a single line, but the computation is already broken up. The longer names provide some context.

types/src/v0/impls/stake_table.rs
```diff
@@ -1379,12 +1410,144 @@ impl EpochCommittees {
...
+                .checked_div(epoch_height)
+                .context("Epoch height is zero. cannot compute average block time")?
+        };
+        tracing::debug!(?epoch, "avg_block_time={avg_block_time:?}");
+
+        let blocks_per_year = MILLISECONDS_PER_YEAR / avg_block_time as u128;
+        tracing::debug!(?epoch, "blocks_per_year={blocks_per_year:?}");
+
+        let block_reward = ((total_supply * U256::from(inflation_rate_bp))
+            / U256::from(blocks_per_year))
```
sveitser: checked division? I know it's strictly speaking not necessary unless epoch_height > time_diff but yeah, seems safer for the future

types/src/v0/v0_1/state.rs
```diff
@@ -93,6 +93,7 @@ pub struct RewardAmount(pub U256);
 pub(crate) const INFLATION_RATE: u128 = 300; // 3% in basis points
 pub(crate) const ASSUMED_BLOCK_TIME_SECONDS: u128 = 2;
 pub(crate) const SECONDS_PER_YEAR: u128 = 60 * 60 * 24 * 365;
+pub(crate) const MILLISECONDS_PER_YEAR: u128 = 86_400_000 * 365;
```
sveitser: seconds_per_year * 1000?

sveitser (approved):
LGTM. I would feel better if we had more explicit tests for the core calculation though. So that we're certain it matches the spec. This gives a better guarantee than reviewing the code and will reduce the changes for bugs being introduced in the future.

sequencer/src/api.rs
imabdulbasit: do you think we should return an error from this API for first two epochs?

sequencer/src/api.rs
imabdulbasit: yeah, the leader receives commission reward greater than what it should because of round offf error funds, so it won't underflow

types/src/v0/impls/reward.rs
imabdulbasit: this actually includes previous + current block reward

data/v4/header.json
imabdulbasit: No, I am not sure why it is using hex here even though `FeeAmount` is also using alloy `U256`

sveitser: We have some custom serde code for some types.

types/src/v0/impls/stake_table.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3450/commits/d1ac4713e31cce6b6f089f72b7573f003527dbd5
https://github.com/EspressoSystems/espresso-network/pull/3450/commits/04451bc84dacaedc722fc32292d9f0129df39146

sequencer/src/persistence/fs.rs
imabdulbasit: I think it's okay to change the extension, provided that we delete the old .txt files, as nodes will catch up for the stake tables. However, I’ve reverted it for now and added fallback deserialization.

data/v4/header.json
imabdulbasit: changed it to use `RewardAmount` type

PR #3456, opened 2025-07-16
feat: TxProof for AvidM

### LINKED ISSUE #3455: TxProof is missing for AvidM scheme

mrain:
Closes #3455


alxiong:
> Encountered [this tech debt](https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/types/src/v0/impls/block/full_payload/payload.rs#L232-L249). Needs to make decision before moving on.

I think we should update `trait QueryablePayload::transaction_with_proof()` to accept an additional `aux: Self::VidCommon` field? would that be enough? 

also consult the owner of the query-service (I'm not sure who)

mrain:
decision: we'll have separate APIs for transaction only and proof only, waiting for this PR https://github.com/EspressoSystems/espresso-network/pull/3449 before proceeding

types/src/v0/impls/block/test.rs
```diff
@@ -73,7 +76,10 @@ async fn basic_correctness() {
             "not all test txs consumed by block.iter"
         );
 
-        let vid_commit = VidCommitment::V0(vid_commit);
+        // TODO(Chengyu): these APIs are troublemakers
+        let VidCommon::V0(vid_common) = vid_common else {
+            unreachable!();
```
alxiong: at least with a panic message?

types/src/v0/impls/block/namespace_payload/tx_proof/avidm.rs
```diff
@@ -0,0 +1,89 @@
...
+        let VidCommon::V1(common) = common else {
+            tracing::info!("VID version mismatch");
+            return None;
+        };
+
+        let (txs, _) = self.ns_proof.verify(ns_table, commit, common)?;
+        if self.tx_index.0 > txs.len() {
+            return Some(false);
+        }
+        Some(&txs[self.tx_index.0] == tx)
```
alxiong, line 87: not sure i follow, the tx proof only contains a namespace proof and an index?

i thought it's a nested merkle proof or two separate tx proofs -- one NsProof, one for tx being the segment of leaves in the systematic RS codeword that merkle-commit to `vid_commit`?

types/src/v0/impls/block/full_payload/payload.rs
```diff
@@ -242,21 +238,7 @@ impl QueryablePayload<SeqTypes> for Payload {
...
-                // Need a `ADVZCommon` to proceed. Need to modify return type to accept either
-                // version of VID common and produce a proof from the corresponding version.
-                &hotshot_types::vid::advz::advz_scheme(10)
-                    .disperse(&self.raw_payload)
-                    .unwrap()
-                    .common
-            },
-        };
-        let proof = TxProof::new(index, self, common)?.1;
+        let proof = TxProof::new(index, self, vid.common())?.1;
```
alxiong, line 241: what about we do the version dispatch centrally here, so that `tx_proof/avidm.rs`'s method will directly takes in `common: avidm::AvidMCommon` and don't have to add separate version detection both in advz and avidm.

types/src/v0/impls/block/namespace_payload/tx_proof/avidm.rs
mrain: No we are implementing the most naive one here. The NsProof struct somehow contains the whole namespace payload.

types/src/v0/impls/block/full_payload/payload.rs
mrain: version detection is centralized here: https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/types/src/v0/txproof.rs#L14-L26

mrain:
@alxiong 
> 1. is this endpoint tested anywhere? how do i know this TxProof functionality works?

APIs are tested here: https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/hotshot-query-service/src/availability.rs#L1046-L1097
> 2. what happens to the conversion about "separating tx getter and tx proof getter", tx only and proof only API are still the decision? right now, the `TxProof::new()` seems to return both.

You can find the separation here: https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/types/src/v0/impls/block/full_payload/payload.rs#L231-L243
Also we have separated endpoints here: https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/hotshot-query-service/api/availability.toml#L241-L298

types/src/v0/impls/block/test.rs
mrain: this is removed: [d58dc14](https://github.com/EspressoSystems/espresso-network/pull/3456/commits/d58dc141b124ee94b32bd59f55f196e239b89967)

alxiong (approved):

hotshot-query-service/api/availability.toml
```diff
@@ -244,7 +244,7 @@ PATH = ["transaction/:height/:index", "transaction/hash/:hash"]
 ":index" = "Integer"
 ":hash" = "TaggedBase64"
 DOC = """
-Get a transaction by its `index` in the block at `height` or by its hash.
+Get a transaction inclusion proof by its `index` in the block at `height` or by its hash.
```
jbearer: Since we deleted the `proof` field from the return type, this addition seems out of place?

jbearer (approved):

PR #3459, opened 2025-07-21
Add Leader and replica stats

bfish713:
This PR adds leader and replica stats by listening to the hotshot events, and adding timestamps. It logs the stats in csv format after every epoch. 
It also includes a binary which takes in leader and replica csv file path,  generates graphs, and prints the stats.

Replica Stats:
- **deltas from View change** (in milliseconds)
    - `DA Cert Δ`– Time from `view_change` to `da_certificate_recv`.
    - `VID Δ` – Time from `view_change` to `vid_share_recv`.
    - `Proposal Δ` – Time from `view_change` to `proposal_recv`.
    - For each: prints **Count**, **Min**, **Max**, and **Average**.
- Plots
    - Scatter plot for Proposal received timestamp  by View
        - marks timed out views in red
    - Bar chart for First event frequency
    - Scatter plots for DAC, VID, and Proposal Timestamps (s) by view
    

Leader Stats:
- **Deltas from Block Built** (in milliseconds)
    - `DA Cert Δ` – Time from `block_built` to `da_cert_send`.
    - `VID Disperse Δ` – Time from `block_built` to `vid_disperse_send`.
    - For each: prints **Count**, **Min**, **Max**, and **Average**.
- **Delta from Previous Proposal Sent**
    - `Block built Δ` – Time from `prev_proposal_send` to `block_built`.
        
        Prints **Count**, **Min**, **Max**, and **Average**.
        
- Plots:
    - Scatter plots for deltas ( DA, VID, QC ) from Block Built  by View
    - Scatter plots for DA,QC,VID  timestamps against view
    - Scatter plot for block built - prev proposal

---
  - https://app.asana.com/0/0/1210848137472319


benchmark-stats/src/main.rs
```diff
@@ -0,0 +1,542 @@
...
+    subcommand: SubCommands,
+}
+
+#[derive(Subcommand)]
+enum SubCommands {
+    /// Analyze replica stats from a CSV file
+    Replica {
+        /// Path to the replica stats CSV file
+        path: PathBuf,
+        /// Output HTML file (default: replica_stats.html)
```
sveitser, line 29: @imabdulbasit Add files to .gitignore?

crates/hotshot/task-impls/src/stats.rs
```diff
@@ -0,0 +1,346 @@
...
+        match event.as_ref() {
+            HotShotEvent::BlockRecv(block_recv) => {
+                self.leader_entry(block_recv.view_number).block_built = Some(now);
+            },
+            HotShotEvent::QuorumProposalRecv(proposal, _) => {
+                self.replica_entry(proposal.data.view_number())
+                    .proposal_recv = Some(now);
+            },
+            HotShotEvent::QuorumVoteRecv(_vote) => {
+                // self.leader_entry(vote.view_number()).vote_recv = Some(now);
```
sveitser: Why is this commented out?

crates/hotshot/task-impls/src/stats.rs
```diff
@@ -0,0 +1,346 @@
...
+            da_cert_send: None,
+        }
+    }
+}
+
+impl<TYPES: NodeType> ReplicaViewStats<TYPES> {
+    fn new(view: TYPES::View) -> Self {
+        Self {
+            view,
+            view_change: None,
```
sveitser, line 76: Feels like we could  save many lines of code by implementing Default?

crates/hotshot/task-impls/src/stats.rs
```diff
@@ -0,0 +1,346 @@
...
+    anytrace::{Error, Level, Result},
+    line_info, warn,
+};
+use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
+
+use crate::events::HotShotEvent;
+
+#[derive(Serialize, Deserialize)]
+pub struct LeaderViewStats<TYPES: NodeType> {
```
sveitser, line 23: I'm not sure but I think if we had an enum for leader and replica stats the handling of hotshot events would be simpler and more type safe. IIUC it relies on using matching field names.

crates/hotshot/task-impls/src/stats.rs
```diff
@@ -0,0 +1,346 @@
...
+                    let _ = self.dump_stats();
+                    self.garbage_collect(*view - 1);
+                }
+
+                let leader = self
+                    .membership_coordinator
+                    .membership_for_epoch(*epoch)
+                    .await?
+                    .leader(*view)
+                    .await?;
```
sveitser, line 310: Not sure if that matters but we have 2 async calls here and we take the timestamp at the start of the function, so the timestamp set as builder_start might be a bit stale.

crates/hotshot/task-impls/src/stats.rs
```diff
@@ -0,0 +1,346 @@
...
+                if leader == self.public_key {
+                    self.leader_entry(*view).builder_start = Some(now);
+                }
+            },
+            HotShotEvent::Timeout(..) => {
+                self.replica_entry(self.view).timeout_triggered = Some(now);
+            },
+            HotShotEvent::TransactionsRecv(_txns) => {
+                // TODO: Track transactions by time
```
sveitser, line 319: Link to a ticket?

crates/hotshot/task-impls/src/stats.rs
imabdulbasit: view does not implement default trait

crates/hotshot/task-impls/src/stats.rs
imabdulbasit: cc @bfish713

crates/hotshot/task-impls/src/stats.rs
imabdulbasit: would require match statements everywhere, not sure if there is lot of benefit of creating an enum

crates/hotshot/task-impls/src/stats.rs
bfish713: Oh we need to do some other kind of handling on this.  This will overwrite the vote_recv stat each time a vote is received, so the stat would only tell us when the last vote was received.  That's not very useful.  I think we can delete this line.

crates/hotshot/task-impls/src/stats.rs
bfish713: I think we want to take the stat time before we do the async calls on purpose.  The stats task isn't doing the actual processing so if it takes a long time we don't want to make it seem like the Hotshot did something faster.

PR #3461, opened 2025-07-22
Add new reward merkle tree with keccak hash algorithm

imabdulbasit:
Closes https://app.asana.com/0/0/1210789643283194

This PR adds a new reward merkle tree which uses the keccak hashing algorithm. The old reward merkle tree (SHA3) is now renamed to legacy reward merkle tree. The V3 version uses the legacy reward merkle tree, and the V4 uses the new reward merkle tree.

This PR does not migrate the rewards from the legacy to the new one, as this requires catching up for all the accounts.

---
  - https://app.asana.com/0/0/1210789643283194


sequencer/api/migrations/postgres/V1003__rename_reward_tree_legacy.sql
```diff
@@ -0,0 +1,35 @@
+-- Rename the existing reward_merkle_tree table to reward_merkle_tree_legacy
```
sveitser: I wonder why doing the rename vs just adding a v2 table. Isn't it clearer to migrate forwards without changing things that currently exist?

types/src/v0/v0_3/state.rs
```diff
@@ -0,0 +1,105 @@
...
+
+ 
+pub(crate) const INFLATION_RATE: u128 = 300; // 3% in basis points
+pub(crate) const ASSUMED_BLOCK_TIME_SECONDS: u128 = 2;
+pub(crate) const SECONDS_PER_YEAR: u128 = 60 * 60 * 24 * 365;
+pub(crate) const MILLISECONDS_PER_YEAR: u128 = 86_400_000 * 365;
+pub(crate) const BLOCKS_PER_YEAR: u128 = SECONDS_PER_YEAR / ASSUMED_BLOCK_TIME_SECONDS;
+pub const COMMISSION_BASIS_POINTS: u16 = 10_000;
+ 
+
```
sveitser: rustfmt gave up on this file?

sequencer/api/migrations/postgres/V1003__rename_reward_tree_legacy.sql
imabdulbasit: Yeah, I thought to keep the table name as the struct but I think you are right

types/src/v0/v0_3/state.rs
imabdulbasit: fixed manually

sequencer/src/catchup.rs
```diff
@@ -408,6 +411,44 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
+                .post::<RewardMerkleTreeLegacy>(&format!(
+                    "catchup/{height}/{}/reward-accounts",
+                    view.u64()
+                ))
+                .body_binary(&accounts.to_vec())?
+                .send()
+                .await?;
+
+            // Verify proofs.
+            // Verify proofs.
```
sveitser:
```suggestion
```

sequencer/api/catchup.toml
```diff
@@ -100,7 +100,34 @@ DOC = """
 Bulk version of `/:height/:view/account`. The request body should be a JSON array consisting of
 TaggedBase64-encoded reward accounts.
 
-The response is a `RewardMErkleTree` containing sub-trees for each of the requested accounts, which is
+The response is a `RewardMerkleTreeV1` containing sub-trees for each of the requested accounts, which is
```
sveitser, line 103:
```suggestion
The response is a `RewardMerkleTreeV2` containing sub-trees for each of the requested accounts, which is
```

?

sveitser: ah no, that's for the above

sequencer/src/api.rs
```diff
@@ -5449,4 +5575,91 @@ mod test {
...
     }
+
+    #[rstest]
+    #[case(PosVersionV3::new())]
+    #[case(PosVersionV4::new())]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_v3_and_v4_reward_tree_updates<Ver: Versions>(
+        #[case] versions: Ver,
+    ) -> anyhow::Result<()> {
+        // This test checks that the correct merkle tree is updated based on version
```
sveitser, line 5587: nice

sequencer/src/api/sql.rs
```diff
@@ -91,13 +93,50 @@ impl SequencerDataSource for DataSource {
...
+        );
+
+        // Check if we have the desired state snapshot. If so, we can load the desired accounts
+        // directly.
+        if height < block_height {
+            load_v1_reward_accounts(&mut tx, height, accounts).await
+        } else {
+            let accounts: Vec<_> = accounts
+                .iter()
+                .map(|acct| RewardAccountV2::from(*acct))
```
sveitser, line 122: @imabdulbasit I'm a bit confused what's going on here. Why do we make V2 accounts here to later return a V1 tree?

sveitser: Ah 

https://github.com/EspressoSystems/espresso-sequencer/blob/8afd04c4d4fa053baf2ddf29ab338374ef2cd8d3/sequencer/src/api/sql.rs#L562

imabdulbasit: This is to reuse the reconstruct_state() which takes in RewardAccountV2

sequencer/src/api/sql.rs
```diff
@@ -91,13 +93,50 @@ impl SequencerDataSource for DataSource {
...
+            .context("getting block height")? as u64;
+        ensure!(
+            block_height > 0,
+            "cannot get accounts for height {height}: no blocks available"
+        );
+
+        // Check if we have the desired state snapshot. If so, we can load the desired accounts
+        // directly.
+        if height < block_height {
+            load_v1_reward_accounts(&mut tx, height, accounts).await
```
sveitser, line 118: This function name is a bit non-intuitive here. Should we call this `authenticate_reward_amounts` or something like that?

And is the reason why we can call this here because if `height < block_height` we will have the leave in storage?

imabdulbasit: its loading accounts from database so I don't understand why call this `authenticate_reward_amounts`

sequencer/src/api/data_source.rs
```diff
@@ -247,8 +248,34 @@ pub(crate) trait CatchupDataSource: Sync {
         instance: &NodeState,
         height: u64,
         view: ViewNumber,
-        accounts: &[RewardAccount],
-    ) -> impl Send + Future<Output = anyhow::Result<RewardMerkleTree>>;
+        accounts: &[RewardAccountV2],
+    ) -> impl Send + Future<Output = anyhow::Result<RewardMerkleTreeV2>>;
+
+    fn get_reward_account_v1(
```
sveitser, line 254: any particular reason to not rename the `get_reward_account{,s}` above to `..._v2`?

sequencer/src/api/sql.rs
imabdulbasit: yeah, if the height < block_height then we should load the  merklized state for that height  stored in db

sequencer/src/api/data_source.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3461/commits/a36d6327c194b833e1aa5ff8a75be875ea74ee44

types/src/v0/impls/reward.rs
```diff
@@ -213,105 +267,244 @@ impl CanonicalDeserialize for RewardAccount {
...
+                tree.non_membership_remember(RewardAccountV2(self.account), proof)?;
+                Ok(())
+            },
+        }
     }
 }
 
 #[allow(dead_code)]
-impl RewardAccountProof {
+impl RewardAccountProofV1 {
```
sveitser, line 415: Do you know why we need this `allow(dead_code)`?

types/src/v0/impls/state.rs
```diff
@@ -1025,16 +1078,34 @@ impl HotShotState<SeqTypes> for ValidatedState {
...
 
-        let reward_merkle_tree = if block_header.reward_merkle_tree_root().size() == 0 {
-            RewardMerkleTree::new(REWARD_MERKLE_TREE_HEIGHT)
-        } else {
-            RewardMerkleTree::from_commitment(block_header.reward_merkle_tree_root())
+        let (reward_merkle_tree_v1, reward_merkle_tree_v2) = match block_header
+            .reward_merkle_tree_root()
+        {
+            Either::Left(reward_legacy) => {
+                let reward_merkle_tree = RewardMerkleTreeV2::new(REWARD_MERKLE_TREE_V2_HEIGHT);
```
sveitser:
```suggestion
                let reward_merkle_tree_v2 = RewardMerkleTreeV2::new(REWARD_MERKLE_TREE_V2_HEIGHT);
```

Just for consistency?

types/src/v0/impls/state.rs
```diff
@@ -101,10 +107,17 @@ pub enum ProposalValidationError {
         expected_root: FeeMerkleCommitment,
         proposal_root: FeeMerkleCommitment,
     },
+    #[error(
+        "Invalid legacy Reward Root Error: expected={expected_root:?}, proposal={proposal_root:?}"
+    )]
+    InvalidRewardRootLegacy {
```
sveitser: This also still uses the Legacy terminology.

types/src/v0/impls/reward.rs
imabdulbasit: This was needed before, not sure why but I tried removing it now and clippy seems happy.

sequencer/src/api/sql.rs
sveitser: I think it's a bit weird because it takes the accounts as input and loads the proofs instead to build the MT. Isn't the point of loading the proofs that we authenticate them? Otherwise we could just build a tree from the accounts directly.

sveitser: But it's fine.

sveitser (approved):
LGTM. I will work on porting over the prototype solidity verifier tomorrow and a test to check that we can verify in the EVM. If there's any problem we'll see it and fix it then.

PR #3464, opened 2025-07-24
Rename builder directories

shenkeyao:
Closes https://app.asana.com/1/1208976916964769/project/1209976130071762/task/1210873516371478?focus=true.

### This PR:
* Moves `builder` to `crates`, where other previous HotShot directories are.
* Renames `marketplace-builder-shared` to `shared` because there's no marketplace builder anymore.
* Renames `hotshot-builder-core` to `legacy` to make it clear it's for the legacy builder.
* Renames `hotshot-builder-core-refactored` to `refactored` for simplicity.
* Moves `shared`, `legacy`, and `refactored` to `crates/hotshot-builder` to align with the plan to reorganize directories in the repo.

### This PR does not:
* Integrate the refactored version into the codebase yet, which we'll do next.

### Key places to review:
* Whether the renamings make sense.

**Note**: It would be nice to delete the `builder-shared` directory and move all its content to `hotshot-builder-refactored`, since it's technically not a "shared" directory. But this would break the legacy builder and make it hard to review changes, so it's probably easier to do this after we are ready to remove `hotshot-builder-legacy`. I'll leave it to @QuentinI to decide when to do this. Follow-up task: https://app.asana.com/1/1208976916964769/project/1210873516302608/task/1210873516371500?focus=true.


QuentinI:
> Renames builder to hotshot-builder because it originally was the builder directory in the hotshot repo

`builder` was always a directory in the sequencer repo, it depends on sequencer types and builds a binary for the builder.
This would also be confusing, because we have "hotshot-*" directories for things that don't depend on the sequencer types, like "hotshot-query-service" and "hotshot-events-service", and this PR flips it by naming sequencer-independent builder libraries without hotshot- prefix and giving the prefix only to the one crate that actually brings in the sequencer types.

PR #3470, opened 2025-07-29
clippy: enable significant_drop_in_scrutinee 

### LINKED ISSUE #3469: Turn on `clippy::significant_drop_in_scrutinee`
We recently had a deadlock bug that took a while to find in #3467 . Having this clippy lint would have likely prevented it.

sveitser:
Closes #3469

### DIFF
--- a/crates/hotshot/hotshot/src/traits/networking/push_cdn_network.rs
+++ b/crates/hotshot/hotshot/src/traits/networking/push_cdn_network.rs
@@ -571,7 +571,8 @@ impl<K: SignatureKey + 'static> ConnectedNetwork<K> for PushCdnNetwork<K> {
     /// - If we fail to receive messages. Will trigger a retry automatically.
     async fn recv_message(&self) -> Result<Vec<u8>, NetworkError> {
         // If we have a message in the internal queue, return it
-        if let Some(message) = self.internal_queue.lock().pop_front() {
+        let queued_message = self.internal_queue.lock().pop_front();
+        if let Some(message) = queued_message {
             return Ok(message);
         }
 
--- a/crates/hotshot/task-impls/src/da.rs
+++ b/crates/hotshot/task-impls/src/da.rs
@@ -473,12 +473,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
                     metadata: metadata.clone(),
                 });
                 // Save the payload early because we might need it to calculate VID for the next epoch nodes.
-                if let Err(e) = self
+                let update_result = self
                     .consensus
                     .write()
                     .await
-                    .update_saved_payloads(view_number, payload_with_metadata)
-                {
+                    .update_saved_payloads(view_number, payload_with_metadata);
+                if let Err(e) = update_result {
                     tracing::trace!("{e:?}");
                 }
             },
--- a/crates/hotshot/testing/src/block_builder/random.rs
+++ b/crates/hotshot/testing/src/block_builder/random.rs
@@ -139,11 +139,11 @@ impl<TYPES: NodeType<Transaction = TestTransaction>> RandomBuilderTask<TYPES> {
             // Let new VID scheme ship with Epochs upgrade.
             let block = build_block::<TYPES>(transactions, pub_key.clone(), priv_key.clone()).await;
 
-            if let Some((hash, _)) = blocks
+            let push_result = blocks
                 .write()
                 .await
-                .push(block.metadata.block_hash.clone(), block)
-            {
+                .push(block.metadata.block_hash.clone(), block);
+            if let Some((hash, _)) = push_result {
                 tracing::warn!("Block {hash} evicted");
             };
             if time_per_block < start.elapsed() {
--- a/hotshot-builder-core/src/service.rs
+++ b/hotshot-builder-core/src/service.rs
@@ -840,7 +840,8 @@ impl<Types: NodeType> ProxyGlobalState<Types> {
         if let Some((vid_trigger, block_payload, metadata)) = extracted_block_info_option {
             tracing::info!("Trying sending vid trigger info for {block_id}",);
 
-            if let Some(trigger_writer) = vid_trigger.write().await.take() {
+            let trigger_writer = vid_trigger.write().await.take();
+            if let Some(trigger_writer) = trigger_writer {
                 tracing::info!("Sending vid trigger for {block_id}");
                 let _ = trigger_writer.send(TriggerStatus::Start);
                 tracing::info!("Sent vid trigger for {block_id}");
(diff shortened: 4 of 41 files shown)


hotshot-builder-core/src/service.rs
```diff
@@ -840,7 +840,8 @@ impl<Types: NodeType> ProxyGlobalState<Types> {
         if let Some((vid_trigger, block_payload, metadata)) = extracted_block_info_option {
             tracing::info!("Trying sending vid trigger info for {block_id}",);
 
-            if let Some(trigger_writer) = vid_trigger.write().await.take() {
+            let trigger_writer = vid_trigger.write().await.take();
+            if let Some(trigger_writer) = trigger_writer {
```
sveitser, line 844: I'm reviewing the lint fixes to check if we don't implicitly depend on the un-intuitive dropping behaviour that the new lint prevents.   

This was first added here: https://github.com/EspressoSystems/hotshot-builder-core/commit/6198d767ce96dedde4ce4f3648b661ba574176bf

I think this should be fine because the `take` is atomic and nothing else can get the trigger writer. 

cc @Ayiga since you ventured into this code at some point in your past.

crates/hotshot/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -571,7 +571,8 @@ impl<K: SignatureKey + 'static> ConnectedNetwork<K> for PushCdnNetwork<K> {
     /// - If we fail to receive messages. Will trigger a retry automatically.
     async fn recv_message(&self) -> Result<Vec<u8>, NetworkError> {
         // If we have a message in the internal queue, return it
-        if let Some(message) = self.internal_queue.lock().pop_front() {
+        let queued_message = self.internal_queue.lock().pop_front();
+        if let Some(message) = queued_message {
```
sveitser, line 575: I think this is fine, but cc @rob-maron since I'm messing with your locks

crates/hotshot/task-impls/src/da.rs
```diff
@@ -473,12 +473,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
...
                 // Save the payload early because we might need it to calculate VID for the next epoch nodes.
-                if let Err(e) = self
+                let update_result = self
                     .consensus
                     .write()
                     .await
-                    .update_saved_payloads(view_number, payload_with_metadata)
-                {
+                    .update_saved_payloads(view_number, payload_with_metadata);
+                if let Err(e) = update_result {
```
sveitser, line 481: Only logging the error in the if let clause. Ok.

crates/hotshot/testing/src/block_builder/random.rs
```diff
@@ -139,11 +139,11 @@ impl<TYPES: NodeType<Transaction = TestTransaction>> RandomBuilderTask<TYPES> {
...
             let block = build_block::<TYPES>(transactions, pub_key.clone(), priv_key.clone()).await;
 
-            if let Some((hash, _)) = blocks
+            let push_result = blocks
                 .write()
                 .await
-                .push(block.metadata.block_hash.clone(), block)
-            {
+                .push(block.metadata.block_hash.clone(), block);
+            if let Some((hash, _)) = push_result {
```
sveitser, line 146: Just logging, okay.

sequencer/src/lib.rs
```diff
@@ -312,8 +312,9 @@ where
     // Print the libp2p public key
     info!("Starting Libp2p with PeerID: {libp2p_public_key}");
 
+    let loaded_network_config_from_persistence = persistence.load_config().await?;
```
sveitser, line 315: This is a bit of a long name, but prevents rustfmt to reformat the entire let statement.

sveitser: I think here the intent is not to hold the read lock taken out in `load_config()` over the whole match statement so this change looks good to me. I'm not sure if this is what currently happens though because if yes, we should be observing a deadlock when we need the write lock to save the config later inside the match arm.

cc @imabdulbasit

types/src/v0/impls/l1.rs
```diff
@@ -325,7 +325,8 @@ impl Service<RequestPacket> for SwitchingTransport {
             }
 
             // If we've been rate limited, back off until the limit (hopefully) expires.
-            if let Some(t) = current_transport.status.read().rate_limited_until {
+            let rate_limit_until = current_transport.status.read().rate_limited_until;
+            if let Some(t) = rate_limit_until {
```
sveitser, line 329: Here it looks to me like the intention is to drop the read lock. So I think this change makes sense.

cc @rob-maron

types/src/v0/impls/l1.rs
```diff
@@ -468,7 +469,8 @@ impl L1Client {
     /// The L1 client will still be usable, but will stop updating until [`start`](Self::start) is
     /// called again.
     pub async fn shut_down_tasks(&self) {
-        if let Some(update_task) = self.update_task.0.lock().await.take() {
+        let update_task = self.update_task.0.lock().await.take();
```
sveitser, line 472: This is dropping the lock after `take()` so fine.

types/src/v0/impls/stake_table.rs
```diff
@@ -629,7 +629,8 @@ impl Fetcher {
             // It keeps retrying until the chain config is upgraded
             // after a successful upgrade to an epoch version.
             let stake_contract_address = loop {
-                match chain_config.lock().await.stake_table_contract {
+                let contract = chain_config.lock().await.stake_table_contract;
+                match contract {
```
sveitser, line 633: Dropping the lock before returning or logging seems correct.

types/src/v0/impls/stake_table.rs
```diff
@@ -643,7 +644,8 @@ impl Fetcher {
             // Begin the main polling loop
             loop {
                 let finalized_block = loop {
-                    if let Some(block) = state.lock().await.last_finalized {
+                    let last_finalized = state.lock().await.last_finalized;
+                    if let Some(block) = last_finalized {
```
sveitser, line 648: Dropping lock before break seems fine.

crates/hotshot/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: Yeah this is fine 👍

types/src/v0/impls/l1.rs
rob-maron: Looks good

rob-maron (approved):
All of the changes look good

sequencer/src/lib.rs
imabdulbasit: I think the lock gets dropped because the function is executed and we acquire lock inside the function
example code to test this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4883add3c875d85197c23cf778e12e1c

sveitser: Thanks @imabdulbasit . Based on discussion it seems this is either a bug in clippy or something in NetworkConfig or SeqTypes that has the significant drop annotation 🤷

PR #3478, opened 2025-08-01
feat: solidity prototype reward MT verifier

sveitser:
- add reward MT verifier lib
- add reward MT verifier mock contract
- convert jellyfish proof to solidity
- add integration test
- add TryToSol and ToSol traits

---
  - https://app.asana.com/0/0/1210974169787170

### DIFF
--- a/contracts/rust/adapter/src/lib.rs
+++ b/contracts/rust/adapter/src/lib.rs
@@ -25,17 +25,3 @@ pub fn u256_to_field<F: PrimeField>(x: U256) -> F {
     let bytes: [u8; 32] = x.to_le_bytes();
     F::from_le_bytes_mod_order(&bytes)
 }
-
-/// Fallible type conversion from native rust types to Solidity-compatible types
-pub trait TryToSol<T>: Sized {
-    type Error;
-
-    /// Convert to a Solidity-compatible representation
-    fn try_to_sol(&self) -> Result<T, Self::Error>;
-}
-
-/// Infallible type conversion from native rust types to Solidity-compatible types
-pub trait ToSol<T>: Sized {
-    /// Convert to a Solidity-compatible representation
-    fn to_sol(&self) -> T;
-}
--- a/types/src/v0/impls/reward.rs
+++ b/types/src/v0/impls/reward.rs
@@ -9,7 +9,7 @@ use ark_serialize::{
     CanonicalDeserialize, CanonicalSerialize, Compress, Read, SerializationError, Valid, Validate,
 };
 use hotshot::types::BLSPubKey;
-use hotshot_contract_adapter::{sol_types::AccruedRewardsProofSol, ToSol, TryToSol};
+use hotshot_contract_adapter::sol_types::AccruedRewardsProofSol;
 use hotshot_types::{
     data::{EpochNumber, ViewNumber},
     traits::{election::Membership, node_implementation::ConsensusTime},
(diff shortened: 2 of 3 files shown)


contracts/test/mocks/RewardClaimPrototypeMock.sol
```diff
@@ -0,0 +1,30 @@
...
+import "../../src/libraries/RewardMerkleTreeVerifier.sol";
+
+/**
+ * @title RewardClaimPrototypeMock
+ * @dev Mock contract for testing reward merkle tree verification
+ *
+ * This contract currently only exists to make it possible to call the verifier
+ * library.
+ */
+contract RewardClaimPrototypeMock {
```
alxiong, line 13: why don't we just turn the RMTVerifier into a contract that we can directly call? Not much benefits to make it a library anyway, right?

types/src/v0/v0_4/state.rs
```diff
@@ -43,7 +45,7 @@ pub type RewardMerkleTreeV2 = UniversalMerkleTree<
     From,
     Into,
 )]
-#[display("{_0:x}")]
+#[display("{_0}")]
```
alxiong, line 48: why remove the hex string format?

contracts/rust/adapter/src/lib.rs
```diff
@@ -25,3 +25,17 @@ pub fn u256_to_field<F: PrimeField>(x: U256) -> F {
...
+
+    /// Convert to a Solidity-compatible representation
+    fn try_to_sol(&self) -> Result<T, Self::Error>;
+}
+
+/// Infallible type conversion from native rust types to Solidity-compatible types
+pub trait ToSol<T>: Sized {
+    /// Convert to a Solidity-compatible representation
+    fn to_sol(&self) -> T;
+}
```
alxiong: Q: currently we have so many conversion `From<T> for TSol`, are you suggesting that we should implement this trait instead?

types/src/v0/v0_4/state.rs
sveitser: it formats as `0xabc` this way (I think because alloy's ToString does this), previously as `abc`. I like to have the prefix.

contracts/rust/adapter/src/lib.rs
sveitser: I'm not sure. I think there is some benefit of having a trait for explicitly converting to `sol` even though in many cases we can use `From` to do it. However for `From` the type needs to be known to use the ergonomic `into` variant. Since the family of EVM compatible types is kind of special I think having a trait for just this would be kind of nice, but I don't ask for and don't want to do this refactor right now 😅 

Why did I add it here? I want the solidity type to live in the contract adapter like the others. The implementation kind of has to live in espresso types (with the type we are converting from) unless we want to do a big refactor. So I couldn't implement `From` because we don't own the type or the trait in espresso-types. I could just use `TryInto` instead of that `TryToSol` trait, which might not have occurred to me when I implemented this.

What do you think?

contracts/test/mocks/RewardClaimPrototypeMock.sol
sveitser: It's a library without external functions so during compilation it is inlined. I think it's nice to have a stateless library that we can use in mock and real contracts. For testing in mock contracts we can pass in the root as calldata and for real contract we pass it in from storage. I think the library is most suitable for this.

PR #3479, opened 2025-08-01
bound exitEscrowPeriod in initialize method

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
this pr adds bounds checking for the exit escrow period in the staketable contract to prevent invalid values during initialization. it also adds tests to ensure the contract reverts when the period is out of bounds and succeeds when within bounds.

### Key places to review:
staketable.sol initializeState, staketable.t.sol new tests for escrow period bounds

### How to test this PR: 
forge test --match-contract StakeTable

### Things tested 
- exitEscrow period too short, too long and valid

---
  - https://app.asana.com/0/0/1210948765340118


sveitser:
@alysiahuggins the function you linked doesn't seem to read from the stake table contract. It's supposed to get the static stake table from an espresso node. Conceptually we start with a static stake table that isn't on L1 and should be able to deploy the LC contract without needing the L1 stake table which was added later.

contracts/rust/deployer/src/builder.rs
```diff
@@ -284,12 +284,11 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
 
-                // TEST-ONLY: if this config is not yet set, we use a large default value
+                // TEST-ONLY: if this config is not yet set, we use a reasonable default value
                 // to avoid contract complaining about invalid zero-valued blocks_per_epoch.
-                // This large value will act as if we are always in epoch 1, which won't conflict
-                // with the effective purpose of the real `PublicNetworkConfig`.
+                // This value will allow tests to proceed with realistic epoch behavior.
                 if use_mock && blocks_per_epoch == 0 {
-                    blocks_per_epoch = u64::MAX;
+                    blocks_per_epoch = 10;
```
sveitser: @alxiong do you remember why we set `u64::max` here? I recall there was some kind of hack for the prover to make sure we stay in epoch 0.

cc @alysiahuggins

sveitser:
@alysiahuggins I think the problem is that sending proofs to L1 reverts with `MissingEpochRootUpdate` when running `just demo-native`. I think this is likely due to removing the hack I commented on above.

contracts/rust/deployer/src/builder.rs
alysiahuggins: ah so is there a scenario in a real env where we'd want the blocks_per_epoch to be u64::max? if so, i'll have to remove the upper bound here https://github.com/EspressoSystems/espresso-network/blob/536ad560aefd9d8b42e6033686fe817b3036486d/contracts/src/StakeTable.sol#L252-L254

sveitser: I don't think we should need to set this in a real env. @mrain @alxiong?

alxiong: Yes, like Mathis said, it was because when we don't have epoch change we want to always stay in epoch 1. 

But maybe it's safe to remove now, I think?

alysiahuggins: ok so you're saying that we want to allow the setting of the `exitEscrowPeriod` to u64::max / uint64.max (during initialization) to prevent epoch changes in the contract? Sure, can revert.

The max exitEscrowPeriod is also limited in the `updateExitEscrowPeriod()` method, I think that makes sense because otherwise that can halt epoch transitions (and i think there are other ways to halt withdrawals etc - example via pausing which is already implemented and a lot clearer as a governance operation). 

https://github.com/EspressoSystems/espresso-network/blob/536ad560aefd9d8b42e6033686fe817b3036486d/contracts/src/StakeTableV2.sol#L182-L193

contracts/broadcast/LightClientArbitrumStaging.s.sol/421614/run-1744734816.json
```diff
@@ -0,0 +1,165 @@
+{
```
sveitser: @alysiahuggins can we remove these files? Maybe add the broadcast directory to .gitignore?

alysiahuggins: good point as we no longer deploy via the foundry scripts
done here https://github.com/EspressoSystems/espresso-network/pull/3479/commits/e26be7e733bdca3a4ca5aca2b81a10b3df6f4803

contracts/rust/deployer/src/builder.rs
alysiahuggins: the decision here was to remove the upper bound in the initialization method but keep it in the update method https://github.com/EspressoSystems/espresso-network/pull/3479/commits/6c3b725bd450422c75cec9b26c552c32795aeb24

contracts/broadcast/LightClientArbitrumStaging.s.sol/421614/run-1744734816.json
sveitser: We also need to un-commit the files.

.env
```diff
@@ -63,6 +63,7 @@ ESPRESSO_SEQUENCER_PERMISSIONED_PROVER=0x14dc79964da2c08b23698b3d3cc7ca32193d995
 ESP_TOKEN_INITIAL_SUPPLY=3590000000
 ESP_TOKEN_NAME=Espresso
 ESP_TOKEN_SYMBOL=ESP
+ESPRESSO_SEQUENCER_BLOCKS_PER_EPOCH=10
```
sveitser: I first thought this should break the demos that do upgrades (those require at least 150 block per epoch) due to guardrails in hotshot upgrades. I think the reason why it works anyway is because before deployment we fetch `blocks_per_epoch` from the sequencer API instead.

Do we actually need this to be configurable via CLI argument / env var?

sveitser: https://github.com/EspressoSystems/espresso-network/blob/e26be7e733bdca3a4ca5aca2b81a10b3df6f4803/sequencer/src/bin/deploy.rs#L444-L493

IIUC we only use the value for LC v2 and v3 upgrades and for those we either set a mock value or fetch it from the sequencer API.

alysiahuggins: that's fair, addressed here https://github.com/EspressoSystems/espresso-network/pull/3479/commits/5d84a89c1d29d2e5c23f596027622e948c5c7dc7

PR #3483, opened 2025-08-05
feat: staking-cli export signed node signatures

### LINKED ISSUE #3476: staking-cli: decouple Espresso and Ethereum keys
Currently to register a validator and rotate-keys the staking-cli needs to run on machine that has access to both Espresso and Ethereum keys. This makes it difficult to have the keys in separate systems which is bad for signing security.

- Add a command `sign-address` to the staking-cli.

```
staking-cli sign-address --output sig.json # save sig to json
staking-cli sign-address --output sig.toml # ... to toml
staking-cli sign-address # print json signature to stdout
```

Required arguments are bls and schnorr private key.

The `register-validator` and `rotate-keys` commands will accepts `--address-signatures` arguments as mutually exclusive alternatives to accepting the espresso private keys.

The value of `--address-signatures` can be a string (path) in which case it will load that file, `-` will read json from stdin. If the file is not found or can't de de-serialized exit with an error.

### TODO
- [ ] Add integration tests for all cases, ideally with rstest to avoid unnecessary code duplication
- [ ] Define a type `AddressSignaturePayload` holding the two signatures, and the public keys.
- [ ] Accept and process the new command line flags.

sveitser:
Allows de-coupling of Espresso and Ethereum keys.

- export node signatures
- load node signatures for validator registration and consensus keys rotation
- support json/toml files and stdin/out
- verify signatures on import to avoid errors

Closes #3476

### DIFF
--- a/staking-cli/tests/cli.rs
+++ b/staking-cli/tests/cli.rs
@@ -1,18 +1,21 @@
-use std::{
-    path::PathBuf,
-    process::{Command, Output, Stdio},
-};
-
 use alloy::primitives::{
     utils::{format_ether, parse_ether},
     Address, U256,
 };
 use anyhow::Result;
+use common::{base_cmd, Signer, TestSystemExt};
 use hotshot_contract_adapter::stake_table::StakeTableContractVersion;
+use predicates::str;
 use rand::{rngs::StdRng, SeedableRng as _};
-use staking_cli::{demo::DelegationConfig, deploy, deploy::Signer, Config};
+use staking_cli::{
+    demo::DelegationConfig,
+    deploy::{self},
+    Config,
+};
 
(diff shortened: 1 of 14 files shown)


staking-cli/tests/cli.rs
```diff
@@ -23,80 +26,9 @@ async fn stake_table_versions(#[case] _version: StakeTableContractVersion) {}
 
 const TEST_MNEMONIC: &str = "wool upset allow cheap purity craft hat cute below useful reject door";
 
-trait AssertSuccess {
```
sveitser, line 26: This trait's functionality is no longer needed because `assert_cmd` covers it, in a much better way.

imabdulbasit (approved):
LGTM!

PR #3484, opened 2025-08-05
fix: build without lockfile

sveitser:
Started failing recently on `main`.

I updated our lockfile and had to fix a few things.

Somehow the `derive_more` `Into` on `RewardAccount` is not working anymore in one of our tests. I looked a bit into this but couldn't figure out why. We can access the inner type directly in this test so went with that.

The remaining changes should be uncontroversial. Some changes are from `cargo fmt` where I previously failed to format a file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,9 +23,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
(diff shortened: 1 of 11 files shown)


ss-es (approved):
lgtm

PR #3485, opened 2025-08-05
Modify sdk management script to remove potential security vulnerabilities

Ayiga:
### This PR:

Renames `download` to `espresso-network-lib-utility` and moves it under a `cmd` folder to indicate that it is meant to be a binary compilation unit.

Removes the need for the file to be run remotely as it currently is utilized in the [nitro-espresso-integration](https://github.com/EspressoSystems/nitro-espresso-integration/blob/afe73e0133061d6780d86053576bd1d2d65dc09a/scripts/prepare-espresso-crypto-helper#L31).

Addresses the need for it to be run with [elevated permissions](https://github.com/EspressoSystems/nitro-espresso-integration/blob/afe73e0133061d6780d86053576bd1d2d65dc09a/scripts/prepare-espresso-crypto-helper#L6) to [create directories](https://github.com/EspressoSystems/nitro-espresso-integration/blob/afe73e0133061d6780d86053576bd1d2d65dc09a/scripts/prepare-espresso-crypto-helper#L53), [symlinks](https://github.com/EspressoSystems/nitro-espresso-integration/blob/afe73e0133061d6780d86053576bd1d2d65dc09a/scripts/prepare-espresso-crypto-helper#L53), and [cleanup](https://github.com/EspressoSystems/nitro-espresso-integration/blob/afe73e0133061d6780d86053576bd1d2d65dc09a/scripts/prepare-espresso-crypto-helper#L31) existing implementations as it current is needed in nitro-espresso-integration.

This is done by eliminating the [need to determine the directory](https://github.com/EspressoSystems/espresso-network/blob/ffd98bfdb6947ba074e61d1c9c3d17075d9796ec/sdks/go/download/main.go#L244) by running the script remotely using `go run`.  Instead, we aim to derive the installation location using other utilities.

We also recognize that go manages it's module downloads in a directory that is intentionally given the permissions to make it [read-only by default](https://go.dev/ref/mod), yet is still owned by the user.  So we are able to modify the permissions of the directories to make them writable temporarily so we can make the adjustments as needed.


ImJeremyHe (approved):
Nice! Thank you so much @Ayiga

PR #3489, opened 2025-08-06
feat(state-relay): for light client V3

mrain:
Closes #<ISSUE_NUMBER>

### This PR:
- Add relay support for light client V3
- Rewrite the relay server logic
- Lock guarding the access to the stake table


crates/hotshot/types/src/light_client.rs
```diff
@@ -59,19 +70,26 @@ pub struct StateSignatureRequestBody {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct LCV3StateSignatureRequestBody {
     /// The public key associated with this request
     pub key: StateVerKey,
     /// The associated light client state
     pub state: LightClientState,
+    /// The stake table used for the next HotShot block
+    pub next_stake: StakeTableState,
+    /// The auth root
+    // TODO(Chengyu): FixedBytes doesn't implement Canonical(De)Serialize. Is it a problem?
```
alxiong: shoudn't be a problem. we only need serde's `Serialize`.

crates/hotshot/types/src/light_client.rs
```diff
@@ -59,19 +70,26 @@ pub struct StateSignatureRequestBody {
...
+    pub auth_root: FixedBytes<32>,
     /// The associated signature of the light client state
     pub signature: StateSignature,
+    /// The associated signature of the light client state for LCV2
+    pub v2_signature: StateSignature,
```
alxiong, line 107: trying to understand, this is part of the strategy for backward compatibility: simply attach the last version signature?

sequencer/api/state_relay_server.toml
```diff
@@ -10,10 +10,19 @@ DOC = """
...
+DOC = """
+Fetch the latest light client state who has enough corresponding Schnorr signatures collected,
+as well as a list of those signatures.
+"""
+
 [route.getlateststate]
 PATH = ["state"]
 METHOD = "GET"
 DOC = """
+DEPRECATED!
```
alxiong, line 25: since we can completely restart new relay-server, why don't we just remove these apis completely?

sequencer/src/state_signature/relay_server/lcv1_relay.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn get_latest_signature_bundle(&self) -> Result<LCV1StateSignaturesBundle, ServerError>;
+
+    /// Post a signature to the relay server
+    /// # Errors
+    /// Errors if the signature is invalid, already posted, or no longer needed.
+    async fn post_signature(
+        &mut self,
+        req: LCV1StateSignatureRequestBody,
+    ) -> Result<(), ServerError>;
+}
```
alxiong, line 31: instead of having three traits, each having a single implementaor, can we have a single trait with associated types and have three implementors?

```rust
trait DataSource<const V: usize> {
    type ReqBody;
    
    fn post_signature(&self, req: Self::ReqBody);
}

struct ServerStateV1;
struct ReqBodyV1;
impl DataSource<1> for ServerStateV1 {
    
    type ReqBody = ReqBodyV1;
    fn post_signature(&self, req: ReqBodyV1) {}
}

struct ServerStateV2;
struct ReqBodyV2;
impl DataSource<2> for ServerStateV2 {
    
    type ReqBody = ReqBodyV2;
    fn post_signature(&self, req: ReqBodyV2) {}
}

struct ServerState;
impl DataSource<1> for ServerState {
    type ReqBody = ReqBodyV1;
    fn post_signature(&self, req: ReqBodyV1) {}
}
impl DataSource<2> for ServerState {
    type ReqBody = ReqBodyV2;
    fn post_signature(&self, req: ReqBodyV2) {}
}
```

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,166 @@
...
+                epoch_start_block: None,
+                stake_table_infos: HashMap::new(),
+                genesis_stake_table_info: None,
+                gc_queue: BTreeSet::new(),
+            })),
+        }
+    }
+
+    /// Return the genesis stake table info
+    pub async fn get_genesis_stake_table_info(&self) -> anyhow::Result<StakeTableInfo> {
```
alxiong: nit: getter [naming convention](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) in rust is without the `get_` prefix

sequencer/src/state_signature/relay_server.rs
```diff
@@ -610,37 +201,50 @@ where
...
     let state = RwLock::new(
         StateRelayServerState::new(sequencer_url).with_shutdown_signal(shutdown_listener),
     );
     let mut app = App::<RwLock<StateRelayServerState>, ServerError>::with_state(state);
 
-    app.register_module("api", api).unwrap();
+    let v1_api = define_api(&options, bind_version, "1.0.0".parse().unwrap()).unwrap();
+    let v2_api = define_api(&options, bind_version, "2.0.0".parse().unwrap()).unwrap();
+    let v3_api = define_api(&options, bind_version, "3.0.0".parse().unwrap()).unwrap();
+    app.register_module("api", v1_api)?
```
alxiong, line 248: curious, should we give them different module name?

sequencer/src/state_signature.rs
alxiong, line 1: why this state-signer only has changes to V2, but never sign V3?

hotshot-state-prover/src/v3/service.rs
```diff
@@ -88,10 +88,10 @@ pub fn load_proving_key(stake_table_capacity: usize) -> ProvingKey {
 /// Get the latest LightClientState and signature bundle from Sequencer network
 pub async fn fetch_latest_state<ApiVer: StaticVersionType>(
     client: &Client<ServerError, ApiVer>,
-) -> Result<StateSignaturesBundle, ProverError> {
+) -> Result<LCV2StateSignaturesBundle, ProverError> {
```
alxiong, line 91: why not V3 here?

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -666,11 +666,11 @@ async fn main() -> anyhow::Result<()> {
...
-        )
-        .with_blocks_per_epoch(blocks_per_epoch)
-        .with_epoch_start_block(epoch_start_block)
-        .with_thresholds(thresholds)
-        .with_known_nodes(known_nodes);
+        );
+        // .with_blocks_per_epoch(blocks_per_epoch)
+        // .with_epoch_start_block(epoch_start_block)
+        // .with_thresholds(thresholds)
+        // .with_known_nodes(known_nodes);
```
alxiong: remove this.

just so i understand, so currently the Relayer server get these info via the tracker,
but all those `async fetch_xx_info()` has to be triggered in runtime, not at startup?

hotshot-state-prover/src/v3/service.rs
mrain: not touched in this PR

sequencer/src/state_signature.rs
mrain: will do in next PR

sequencer/api/state_relay_server.toml
mrain: we need to take care of old prover and sequencers.

crates/hotshot/types/src/light_client.rs
mrain: yes, signing logic is not included in this PR

sequencer/src/bin/espresso-dev-node.rs
mrain: yes, when it receives signatures from the replicas

sequencer/src/state_signature/relay_server.rs
mrain: that's also an option, one module for each version
cc @jbearer

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: [3a04bd1](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/3a04bd15aef7630618f6c280289a50ab0bbfd416)

sequencer/src/bin/espresso-dev-node.rs
mrain: removed in [2a6dc02](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/2a6dc02ba95dee7f8fcb2b35cca7a8eeb6e16840)

crates/hotshot/types/src/light_client.rs
mrain: removed in [2a6dc02](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/2a6dc02ba95dee7f8fcb2b35cca7a8eeb6e16840)

sequencer/src/state_signature/relay_server/lcv1_relay.rs
mrain: I just feel weird about having a abstract generic.

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
alxiong: right, but the rest of them (such as all the `DataSource`) are not fixed.
up to you, I'm not insisting, only suggesting. not a big deal.

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+        };
+
+        write_guard
+            .stake_table_infos
+            .insert(epoch, stake_table_info.clone());
+        write_guard.gc_queue.insert(epoch);
+        tracing::debug!("Stake table info for epoch {} updated", epoch);
+        // Remove the stake table info if it's older than 2 epochs
+        while let Some(&old_epoch) = write_guard.gc_queue.first() {
+            if old_epoch >= epoch - 2 {
```
alxiong: what if `epoch < 2`?

```rust
if epoch < 2 || old_epoch >= epoch - 2 {
    break;
}
```

also can we define this number as a constant, so it has meaning?

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+use url::Url;
+
+/// Stake table info for a specific epoch
+#[derive(Clone, Debug, Default)]
+pub struct StakeTableInfo {
+    /// Minimum weight to form an available state signature bundle
+    pub threshold: U256,
+    /// Stake table: map(vk, weight)
+    pub known_nodes: HashMap<StateVerKey, U256>,
```
alxiong, line 26: can we wrap this in `Arc`, cuz we frequently `.clone()` this info

sequencer/src/state_signature/relay_server/lcv3_relay.rs
```diff
@@ -0,0 +1,176 @@
...
+    async fn post_signature(
+        &mut self,
+        req: LCV3StateSignatureRequestBody,
+    ) -> Result<(), ServerError>;
+}
+
+/// Server state that tracks the light client V3 state and signatures
+pub struct LCV3StateRelayServerState {
+    /// Bundles for light client V3
+    bundles: HashMap<u64, HashMap<LightClientState, LCV3StateSignaturesBundle>>,
```
alxiong, line 37: Similarly, can we wrap this behind `Arc` to avoid unnecessary deep cloning during `self.latest_available_bundle = Some(bundle.clone());` etc.

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+        drop(read_guard);
+        tracing::trace!("Acquire write lock for genesis stake table info");
+        let mut write_guard = self.inner.write().await;
+        let genesis_stake_table =
+            fetch_stake_table_from_sequencer(&write_guard.sequencer_url, None).await?;
+        let genesis_total_stake = genesis_stake_table.total_stakes();
+
+        if let Some(stake_table_info) = &write_guard.genesis_stake_table_info {
+            return Ok(stake_table_info.clone());
+        }
```
alxiong: we should check first before write, namely the "[double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking)"

Meaning, instead of: 
```rust
// 1. Check without lock
if let Some(value) = read_lock.get() { return value; }

// 2. Get write lock
let mut write_lock = write().await;

// 3. WRONG: Do expensive work BEFORE double-check
let expensive_result = network_call().await;

// 4. TOO LATE: Double-check after expensive work
if let Some(value) = write_lock.get() { return value; }
```

Do this instead: 

```rust
// 1. Check without lock 
if let Some(value) = read_lock.get() { return value; }

// 2. Get write lock 
let mut write_lock = write().await;

// 3. Double-check BEFORE expensive work
if let Some(value) = write_lock.get() { return value; }

// 4. Only do expensive work if still needed
let expensive_result = network_call().await;
```

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+                tracing::trace!("Drop read lock for stake table info");
+                drop(read_guard);
+                tracing::trace!("Acquire write lock for stake table info");
+                let mut write_guard = self.inner.write().await;
+                if let Some(blocks_per_epoch) = write_guard.blocks_per_epoch {
+                    (blocks_per_epoch, write_guard.epoch_start_block.unwrap())
+                } else {
+                    tracing::debug!("Fetching epoch config from sequencer");
+                    let (blocks_per_epoch, epoch_start_block) =
+                        fetch_epoch_config_from_sequencer(&write_guard.sequencer_url).await?;
```
alxiong, line 134: similarly in the else branch,wwe should check again before `fetch_epoch_xxx()` network call.

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
```diff
@@ -0,0 +1,198 @@
...
+        tracing::trace!("Drop read lock for stake table info");
+        drop(read_guard);
+        tracing::trace!("Acquire write lock for stake table info");
+        let mut write_guard = self.inner.write().await;
+        if let Some(stake_table_info) = write_guard.stake_table_infos.get(&epoch) {
+            return Ok(stake_table_info.clone());
+        }
+
+        tracing::debug!("Fetching stake table for epoch {} from sequencer", epoch);
+        let stake_table = fetch_stake_table_from_sequencer(
```
alxiong, line 169: here again, should check after obtaining the write lock and before doing network fetch.

sequencer/src/state_signature/relay_server/lcv1_relay.rs
alxiong: i didn't mean to use `usize`, but use a generic `Version` type, like the one we already had. 

but I won't insist, just a suggestion to see if you like it.

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: already checked in L125

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: checked in L160

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: fixed in [a175f83](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/a175f8342a45e6f7399b66056e64849e313d39b2)

sequencer/src/state_signature/relay_server/lcv3_relay.rs
mrain: Wrapping it requires a RwLock or Mutex. Sounds like too much work

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: Done in [a175f83](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/a175f8342a45e6f7399b66056e64849e313d39b2)

sequencer/src/state_signature/relay_server/stake_table_tracker.rs
mrain: Done in [a175f83](https://github.com/EspressoSystems/espresso-network/pull/3489/commits/a175f8342a45e6f7399b66056e64849e313d39b2)

PR #3490, opened 2025-08-06
Test membership catchup

ss-es:


crates/hotshot/example-types/src/membership/fetcher.rs
```diff
@@ -0,0 +1,233 @@
...
+                                (
+                                    proposal.data.block_header().block_number(),
+                                    Leaf2::from_quorum_proposal(&proposal.data.clone()),
+                                )
+                            })
+                            .collect();
+
+                        let Some(leaf) = leaves.get(&requested_height) else {
+                            tracing::error!(
+                                "Block at height {} not found in storage",
```
pls148: Nit: put {requested_height} in log string

pls148: also make both of these one log message for easier correlation

crates/hotshot/example-types/src/membership/fetcher.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    pub async fn fetch_leaf(
+        &self,
+        height: u64,
+        source: TYPES::SignatureKey,
+    ) -> anyhow::Result<Leaf2<TYPES>> {
+        let leaf_request = Message {
+            sender: self.public_key.clone(),
+            kind: MessageKind::<TYPES>::External(
+                bincode::serialize(&(height, self.public_key.clone()))
```
pls148, line 179: Does bincode::serialize not work if the tuple member is a reference?

crates/hotshot/example-types/src/membership/strict_membership.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    async fn get_epoch_root(
+        membership: Arc<RwLock<Self>>,
+        block_height: u64,
+        _epoch: TYPES::Epoch,
+    ) -> anyhow::Result<Leaf2<TYPES>> {
+        let membership_reader = membership.read().await;
+
+        for node in membership_reader.inner.full_stake_table() {
```
pls148: I doubt this could actually happen, but I'd store the full stake table and an Arc::clone() of membership_reader.fetcher, then release the membership_reader lock to avoid the unnecessary hold of that lock across the acquisition of another lock. Could also store the read lock on fetcher as fetcher_reader to avoid multiple lock acquisition but that's less important than avoiding the double lock acquire

crates/hotshot/example-types/src/membership/strict_membership.rs
```diff
@@ -0,0 +1,349 @@
...
+                    anyhow::bail!("Missing initial DRB result for epoch {epoch:?}");
+                },
+            };
+
+            let drb_block_height =
+                transition_block_for_epoch(previous_epoch, membership_reader.epoch_height);
+
+            let mut drb_leaf = None;
+
+            for node in membership_reader.inner.full_stake_table() {
```
pls148: Same thing here with the double lock hold

crates/hotshot/example-types/src/membership/fetcher.rs
ss-es: it doesn't -- this really does need to be `clone` as far as I'm aware

crates/hotshot/example-types/src/membership/strict_membership.rs
ss-es: ah great catch, fixed in fda5896

crates/hotshot/example-types/src/membership/fetcher.rs
ss-es: fixed in 5c2b172 and 47e922c

crates/hotshot/example-types/src/membership/fetcher.rs
```diff
@@ -0,0 +1,232 @@
...
+                match network_receiver.recv_direct().await {
+                    Ok(Event {
+                        view_number: _,
+                        event: EventType::ExternalMessageReceived { sender: _, data },
+                    }) => {
+                        let (requested_height, requester): (u64, TYPES::SignatureKey) =
+                            match bincode::deserialize(&data) {
+                                Ok(message) => message,
+                                Err(e) => {
+                                    tracing::debug!("Failed to deserialize message: {:?}", e);
```
pls148: nit: {e:?}

crates/hotshot/example-types/src/membership/fetcher.rs
```diff
@@ -0,0 +1,232 @@
...
+            loop {
+                match network_receiver.recv_direct().await {
+                    Ok(Event {
+                        view_number: _,
+                        event: EventType::ExternalMessageReceived { sender: _, data },
+                    }) => {
+                        let leaf: Leaf2<TYPES> = match bincode::deserialize(&data) {
+                            Ok(message) => message,
+                            Err(e) => {
+                                tracing::debug!("Failed to deserialize message: {:?}", e);
```
pls148: nit: {e:?}

PR #3491, opened 2025-08-06
WIP: RewardClaim prototype contract and E2E test

sveitser:
*Not* functional yet.

Just opening this for visibility. We haven't finished all the work required.

The error we currently get from the prover in the test is 

```
2025-08-06T15:16:25.120057Z ERROR hotshot_state_prover::v3::service: Cannot sync the light client state, will retry in 2.0s: Internal error when generating the SNARK proof: An error in the Plonk circuit: The 34805-th gate failed: gate: Constant Gate, wire values: [BigInt([0, 0, 0, 0]), BigInt([0, 0, 0, 0]), BigInt([0, 0, 0, 0]), BigInt([0, 0, 0, 0]), BigInt([0, 0, 0, 0])], pub_input: , expected_gate_output: 1, gate_output:
```


alxiong:
Recently @mrain's merged PR should unblock this work.

contracts/src/EspTokenV2.sol
```diff
@@ -2,12 +2,25 @@
...
+    function initializeV2() public reinitializer(2) {
+        __AccessControl_init();
+        _grantRole(DEFAULT_ADMIN_ROLE, owner());
+    }
+
+    function mint(address to, uint256 amount) public onlyRole(MINTER_ROLE) {
+        _mint(to, amount);
+    }
+
     function name() public pure override returns (string memory) {
```
alysiahuggins: feel free to remove this since we decided that we'll accept the diff token name in sepolia

contracts/src/RewardClaim.sol
```diff
@@ -0,0 +1,89 @@
...
+        espToken = EspTokenV2(_espToken);
+        lightClient = LightClientV3(_lightClient);
+    }
+
+    function _authorizeUpgrade(address newImplementation) internal override onlyOwner {
+        emit Upgrade(newImplementation);
+    }
+
+    function claimRewards(
+        uint256 accruedReward,
```
alysiahuggins: why do we have to know the total rewards accrued? why not instead the current claim amount (currently represented by `newClaimAmount`?

Also how does one get the proof and authRootInputs?

It sounds like the user has to depend on a service to retrieve rewards then?

(i'm not familiar with the user journey for claiming rewards)

contracts/src/RewardClaim.sol
```diff
@@ -0,0 +1,89 @@
...
+                authRootInputs[6]
+            )
+        );
+
+        require(uint256(authRoot) == lightClient.authRoot(), InvalidAuthRoot());
+
+        uint256 newClaimAmount = accruedReward - claimedRewards[msg.sender];
+        claimedRewards[msg.sender] = accruedReward;
+
+        espToken.mint(msg.sender, newClaimAmount);
```
alysiahuggins: one thing i want to think about is limiting the cost of an attack, for example if someone is able to exploit the proof, then they can mint a large amount of tokens. However, if we can limit the mint amount the cost of an attack can be minimized. Perhaps there should be a mint limit that governance can modify when needed. 

there can be caps based on how much an address staked or some other metric for a cap

alysiahuggins (commented):
looks great, just let a few questions about the design so far

contracts/src/RewardClaim.sol
alxiong: because the Reward Tree only contains the total accured reward, thus the merkle proof can only testify that value. And we can compute the `newClaimAmount` in this function as Mathis did.

contracts/src/RewardClaim.sol
```diff
@@ -0,0 +1,89 @@
...
+        __UUPSUpgradeable_init();
+        espToken = EspTokenV2(_espToken);
+        lightClient = LightClientV3(_lightClient);
+    }
+
+    function _authorizeUpgrade(address newImplementation) internal override onlyOwner {
+        emit Upgrade(newImplementation);
+    }
+
+    function claimRewards(
```
alxiong: Do you want to add a [ReentranceGuard](https://docs.openzeppelin.com/contracts/5.x/api/utils#ReentrancyGuard)? no harm i think

contracts/src/libraries/RewardMerkleTreeVerifier.sol
```diff
@@ -41,19 +41,17 @@ library RewardMerkleTreeVerifier {
...
-        bytes32 root,
-        address key,
-        uint256 value,
-        AccruedRewardsProof calldata proof
-    ) internal pure returns (bool) {
+    function authRootInput(address key, uint256 value, AccruedRewardsProof calldata proof)
+        internal
+        pure
+        returns (bytes32)
+    {
```
alxiong, line 52: why change? I like `verify()` better.

also I'd rename to `computeRoot` or `recomputeRewardCommitment`. current name `authRootInput()` is non-descriptive.

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,520 @@
...
+    args.deploy(&mut l1_contracts, Contract::LightClientV2)
+        .await?;
+    args.deploy(&mut l1_contracts, Contract::LightClientV3)
+        .await?;
+    args.deploy(&mut l1_contracts, Contract::StakeTableProxy)
+        .await?;
+    args.deploy(&mut l1_contracts, Contract::RewardClaimProxy)
+        .await?;
+    args.deploy(&mut l1_contracts, Contract::StakeTableV2)
+        .await?;
```
alxiong: don't we have a handy function that does them all? i vaguely remember we do.

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,520 @@
...
+        .claimRewards(reward_data.balance, proof_sol.into(), auth_root_inputs)
+        .send()
+        .await;
+    assert!(double_claim_result.is_err(),);
+
+    println!("All reward claim tests passed successfully!");
+
+    relay_server_handle.abort();
+    drop(network);
+    drop(anvil);
```
alxiong: nit: they are auto-drop when go out of scope.

contracts/src/RewardClaim.sol
alysiahuggins: so what's the service that users will have to depend on to get that info so that they can claim?

sveitser: You can look at the rust E2E test in this PR for the flow.

https://github.com/EspressoSystems/espresso-network/blob/2e1c3740bcbee0013b62ec909e0312d4de5aa398/tests/reward_claims_e2e.rs#L387-L391

The UI can get the view number, block height from the LC state in the contract, the address is known. This allows the UI to fetch a proof from our query service.

We will probably add the functionality to fetch proofs for the reward tree to the reward API. Currently it's only available on the catchup API so the e2e test is using that.

contracts/src/libraries/RewardMerkleTreeVerifier.sol
sveitser: We do I ran into some issues with timelock etc. I'll take another look.

contracts/src/RewardClaim.sol
alysiahuggins: ok got it, sounds good - as both those endpoints are permissionless too

contracts/src/libraries/RewardMerkleTreeVerifier.sol
sveitser: Looks like i commented on the wrong comment. Anyway, I renamed it back to `computeRoot`

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,515 @@
...
+    );
+
+    let reward_account = format!("0x{:x}", claimer_address);
+    println!(
+        "Testing reward claim for validator account: {}",
+        reward_account
+    );
+
+    let reward_proof_url = format!(
+        "{}catchup/{}/{}/reward-account-v2/{}",
```
sveitser: This will be changed to use the reward API but we have to first change the return type of that API to something more friendly to convert into EVM compatible types.

tests/reward_claims_e2e.rs
sveitser: I ran into issue with it. Don't remember exactly what it was.

sveitser: https://github.com/EspressoSystems/espresso-network/pull/3491/commits/3bd0be0d6aa941bc74bec1c0cc3dc4e97bcb2003

sveitser:
I changed to by default deploy more of our contracts in tests. I think that's mostly a good idea but might break some things. Let's see what CI says.

contracts/src/RewardClaim.sol
```diff
@@ -0,0 +1,89 @@
...
+        returns (uint8 majorVersion, uint8 minorVersion, uint8 patchVersion)
+    {
+        return (1, 0, 0);
+    }
+
+    function _verifyAuthRoot(
+        uint256 lifetimeRewards,
+        bytes32[160] memory proof,
+        bytes32[7] memory authRootInputs
+    ) internal view virtual returns (bool) {
```
sveitser, line 70: I might remove/inline this but I think it will be handy for overwriting in a mock contract to test the rate limiting. I'll decide when I add the rate limiting.

contracts/rust/adapter/src/reward.rs
```diff
@@ -69,35 +73,43 @@ mod tests {
...
     fn test_reward_claim_input_roundtrip_json() {
-        let original = test_input();
+        let original = RewardClaimInput::random();
         let json = serde_json::to_string(&original).unwrap();
         let decoded: RewardClaimInput = serde_json::from_str(&json).unwrap();
         assert_eq!(decoded, original);
     }
 
     #[test]
     fn test_decode_abi_auth_data() {
```
alysiahuggins, line 111: can we add tests for invalid data too? or harder to construct?

contracts/rust/adapter/src/reward.rs
```diff
@@ -69,35 +73,43 @@ mod tests {
...
-    fn test_input() -> RewardClaimInput {
-        let siblings = RewardProofSiblings([B256::random(); 160]);
-        let auth_root_inputs = RewardAuthRootInputs([B256::random(); 7]);
-        RewardClaimInput {
-            lifetime_rewards: B256::random().into(),
-            auth_data: RewardAuthData::new(siblings, auth_root_inputs).into(),
+    impl RewardAuthData {
+        fn random() -> Self {
+            Self {
+                siblings: RewardProofSiblings([B256::random(); 160]),
```
alysiahuggins: should we make 160 & 7 a code constant?

contracts/rust/deployer/src/builder.rs
```diff
@@ -495,6 +495,38 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+
+                if let Some(use_timelock_owner) = self.use_timelock_owner {
+                    // TODO: (MA) this needs to be discussed before finalizing.
+                    //
+                    // RewardClaim uses SafeExitTimelock because:
+                    // - It is not expected to require urgent upgrades.
+                    // - In emergency situations it can be paused.
+                    // - It can mint ESP tokens, users should have enough time
+                    //   to react if they do not agree with an upgrade.
+                    tracing::info!("Transferring ownership to SafeExitTimelock");
```
alysiahuggins, line 516: do we want the SafeExitTimelock for this one or the OpsTimelock? Perhaps too early to decide

contracts/src/EspTokenV2.sol
```diff
@@ -3,15 +3,40 @@ pragma solidity ^0.8.0;
...
+
     constructor() {
         _disableInitializers();
     }
 
-    function name() public pure override returns (string memory) {
-        return "Espresso";
+    /// @notice Initializes the V2 upgrade with the RewardClaim contract address
+    /// @param _rewardClaim Address of the RewardClaim contract
+    function initializeV2(address _rewardClaim) public reinitializer(2) {
```
alysiahuggins: i missed this but we should add the `onlyOwner` function to reinitializing functions. Functions with the `reinitializer` modifier can be called only once but it doesn't have access control. In V1 versions of contracts, there is no owner so the initializing method cannot be owner controlled, however subsequent reinitializing functions can be. I've created an [issue](https://github.com/EspressoSystems/espresso-network/issues/3654) so that we can update other contracts like LightClientV3. We aren't affected by this since our deployer always initializes within the same transaction as the upgrade call but better to have it than not.

here's a PR for this change that you can merge into this branch https://github.com/EspressoSystems/espresso-network/pull/3655

contracts/src/EspTokenV2.sol
```diff
@@ -3,15 +3,40 @@ pragma solidity ^0.8.0;
...
+        rewardClaim = _rewardClaim;
```
alysiahuggins, line 27: should we add a check for a non-zero address?

contracts/src/EspTokenV2.sol
```diff
@@ -3,15 +3,40 @@ pragma solidity ^0.8.0;
...
+    }
+
+    /// @notice Mints new tokens to a specified address
+    /// @notice Only the RewardClaim contract can mint new tokens
+    ///
+    /// @param to Address to receive the minted tokens
+    /// @param amount Number of tokens to mint
+    function mint(address to, uint256 amount) public {
```
alysiahuggins, line 35: nit: it's simple enough but can add some unit tests

contracts/src/RewardClaim.sol
```diff
@@ -0,0 +1,87 @@
...
+        emit RewardsClaimed(msg.sender, availableToClaim);
+    }
+
+    function getVersion()
+        public
+        pure
+        returns (uint8 majorVersion, uint8 minorVersion, uint8 patchVersion)
+    {
+        return (1, 0, 0);
+    }
```
alxiong, line 64: let's only use `majorVersion` as we have learned elsewhere that the `reinitializer(n)` only takes in a single value anyway. no point having minor and patch version.

also this should be `virtual` so that it's overwritable

contracts/rust/deployer/src/builder.rs
```diff
@@ -495,6 +495,38 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+                let lc_addr = contracts
+                    .address(Contract::LightClientProxy)
+                    .context("no LightClient proxy address")?;
+                let addr = crate::deploy_reward_claim_proxy(
+                    provider, contracts, token_addr, lc_addr, admin,
+                )
+                .await?;
+
+                if let Some(use_timelock_owner) = self.use_timelock_owner {
+                    // TODO: (MA) this needs to be discussed before finalizing.
```
alxiong: using optionally timelock owner looks reasonable to me. can remove this TODO? or do we need more approval from the product side?

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,507 @@
...
+            "--mnemonic",
+            TEST_MNEMONIC,
+            "--accounts",
+            "10",
+            "--balance",
+            "1000000",
+        ])
+        .spawn();
+    let l1_url = anvil.endpoint_url();
+    // TODO: remove, use below for external anvil, to see console.log statements
```
alxiong: remove?

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,507 @@
...
+    println!("Admin address: {}", admin);
+
+    let relay_server_port = pick_unused_port().unwrap();
+    let relay_server_url: Url = format!("http://localhost:{relay_server_port}")
+        .parse()
+        .unwrap();
+    let sequencer_api_port = pick_unused_port().unwrap();
+
+    let network_config = TestConfigBuilder::default()
+        .epoch_height(EPOCH_HEIGHT)
```
alxiong: i feel like we should have (maybe we do already such as `deployer::build_signer()`) some test util to set these up. 

This test fixture setup is likely duplicated with many other test.
we should extract them to helper functions, so that this file is dominated by logic unrelated to reward claims, wdyt?

Generally I find this e2e file too long, would help readability if we group some action into its own unit/function. 
(e.g. the "wait for consensus to decide on new block" and "trigger the state prover once" should definitely be in their own func imo)

it would also help if at the top of this test func, add comment to explain the intended overall workflow to test.

tests/reward_claims_e2e.rs
```diff
@@ -0,0 +1,507 @@
...
+    let anvil = Anvil::new()
+        .args([
+            "--slots-in-an-epoch",
+            "0",
+            "--mnemonic",
+            TEST_MNEMONIC,
+            "--accounts",
+            "10",
+            "--balance",
+            "1000000",
```
alxiong: why are any of these necessary? (rather than just the default)

contracts/rust/deployer/src/builder.rs
sveitser: If we all agree to use the safe exit timelock instead of ops timelock I think we can decide now.

cc @alxiong @alysiahuggins

contracts/src/RewardClaim.sol
sveitser: This is address in:

- #3651

contracts/src/RewardClaim.sol
sveitser: > let's only use majorVersion as we have learned elsewhere that the reinitializer(n) only takes in a single value anyway. no point having minor and patch version.

@alxiong I don't think I understand. I thought this is just for checking the currently deployed version. So we could deploy a minor upgrade and this function would tell us what version we are on. I thought we did some kind of patch upgrade for the LC once.

tests/reward_claims_e2e.rs
sveitser: `slots-in-an-epoch` is needed to finalize block immediately, the rest isn't.

tests/reward_claims_e2e.rs
sveitser: Yeah some of the code can be simplified now that the rest of the code base have caught up I think.

sveitser: I removed some cruft in 2b15be8deb15e672f553f5d52b2c2e89308960b1

I think the main remaining stuff is from the prover setup which I'm really not familiar with. I think it's fine if this test is verbose. I don't think we need to touch it much. If anything we should probably look into a refactoring after we've done everything else for reward claims.

tests/reward_claims_e2e.rs
sveitser: https://github.com/EspressoSystems/espresso-network/commit/2b15be8deb15e672f553f5d52b2c2e89308960b1

tests/reward_claims_e2e.rs
sveitser: https://github.com/EspressoSystems/espresso-network/commit/2b15be8deb15e672f553f5d52b2c2e89308960b1

contracts/rust/deployer/src/builder.rs
sveitser: https://github.com/EspressoSystems/espresso-network/commit/2b15be8deb15e672f553f5d52b2c2e89308960b1

contracts/src/RewardClaim.sol
sveitser: for the virtual addition: https://github.com/EspressoSystems/espresso-network/commit/2b15be8deb15e672f553f5d52b2c2e89308960b1

contracts/rust/adapter/src/reward.rs
sveitser: Yes but I don't quite see the point of doing this in rust. The rust code only encodes. The decoding is just "pretending to be the contract" because that's much quicker feedback than waiting for tests to fail when we send it to the contract.

contracts/rust/deployer/src/builder.rs
sveitser: I think safe exit but we can also discuss later

contracts/src/RewardClaim.sol
sveitser: Good point about limiting attack surface. A few options to consider:

1. **Per-claim cap**: Limit how much can be claimed in a single transaction
2. **Per-address lifetime cap**: Limit total rewards per address (could be based on their stake)
3. **Global rate limiting**: Limit total tokens minted per time period
4. **Emergency pause**: Circuit breaker that triggers if unusual minting patterns detected

For this prototype, I think option 1 (per-claim cap) might be simplest - we could add a configurable maxClaimAmount that owner can adjust. This limits damage from a single exploit while still allowing legitimate large claims via multiple transactions.

Thoughts on whether we should add this now or defer to a future iteration? It does add complexity but the security benefit seems worthwhile.

contracts/src/EspTokenV2.sol
sveitser: thanks: https://github.com/EspressoSystems/espresso-network/pull/3491/commits/53adee78451ec1c55b43ca1cb0b7d41c0b743a16

contracts/rust/adapter/src/reward.rs
sveitser: not sure, but here it is: 719b50dff2f6c8bd3e89fff083c35e5bcc08b05e

contracts/src/EspTokenV2.sol
sveitser: sure

contracts/src/EspTokenV2.sol
sveitser: 995f3a2a64f54a8754776ad7b292446ba8c3f7fe

contracts/rust/deployer/src/builder.rs
alysiahuggins: okie, started a convo offline

PR #3495, opened 2025-08-08
fix: check merkle proof heights

sveitser:
Since we're using an outdated version of jf I think the proof length is not enforced on the proof API when verifying proofs.

To add a regression test I think we'd need to find a collision which is hard.

cc @mrain @alxiong


alxiong:
LGTM! Leaving to Chengyu for final approval.

sveitser:
Need to add +1 to the expected length, that's why tests are failing.

PR #3497, opened 2025-08-08
Add Stake Table Hash to header

bfish713:
### This PR:

Adds a stake table hash over all stake table events for each epoch to the block header and checks it on every view.

The hashing works by by hashing the serailized values for each validator in the stake table contract, which should be stored in order of update (which must be the same for any sequence of events).  The delegators for each validator are stored in sorted order by Ethereum key (it must be sorted or different validators will get a different hash).  We also add the serialized bls, and state prover keys, in sorted order.

This PR also adds the stake table hash to the LC contract state.

### Key places to review:

- The hash functions implementation for the stake table
- The storage implementation changes (sql and fs)
- `Header::new` and `validate_and_apply_header` changes

---
  - https://app.asana.com/0/0/1210939588593695


data/v4/header-actual.json
```diff
@@ -0,0 +1,50 @@
+{
```
imabdulbasit: This file can be deleted.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -37,6 +37,8 @@ type EpochSender<TYPES> = (
     Sender<Result<EpochMembership<TYPES>>>,
 );
 
+pub type StakeTableHash = [u8; 32];
```
ss-es: can we move this to `election.rs`?

crates/hotshot/types/src/traits/election.rs
```diff
@@ -182,6 +182,12 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
     fn first_epoch(&self) -> Option<TYPES::Epoch> {
         None
     }
+
+    /// Returns the commitment of the stake table for the given epoch,
+    /// Errors if the stake table is not available for the given epoch
+    fn stake_table_hash(&self, _epoch: TYPES::Epoch) -> Option<StakeTableHash> {
+        None
+    }
```
ss-es, line 201: I feel like we should not have a default implementation like this (always returning `None`) here -- in my experience these can cause really annoying to debug silent failures with the test membership implementations

sequencer/src/persistence/fs.rs
```diff
@@ -1631,11 +1636,11 @@ impl MembershipPersistence for Persistence {
                                 )
                             })?;
 
-                        (validatormap, None)
+                        (validatormap, None, StakeTableStateHash::default())
```
ss-es: maybe I'm not understanding this, but using `default()` here seems wrong? shouldn't we return an `Option<>` or just fail the deserialization

types/src/v0/impls/header.rs
```diff
@@ -1105,8 +1140,9 @@ impl BlockHeader<SeqTypes> for Header {
         match self {
             Header::V1(_) | Header::V2(_) | Header::V3(_) => Ok(FixedBytes::from([0u8; 32])),
             Header::V4(header) => {
+                // Add next stake table hash to the auth root, will be 0 except during epoch transition
+                let next_stake_table_hash = self.next_stake_table_hash().unwrap_or([0; 32]);
```
ss-es: I don't know exactly what this is for but this makes me a bit uncomfortable (it seems like it's not just a test thing?)

ss-es (commented):
I think it looks fine to me overall but someone else should also review

I did have a few minor nits/questions

types/src/v0/impls/header.rs
bfish713: This is for adding the state to the light client.  I think this logic is ok, we have some placeholders for any new things we want to store, and the stake table is one of them.  Basically it'll just be a placeholder if the hash isn't there, and the actual hash if it is.

mrain: but light client already has some stake table commitment here: https://github.com/EspressoSystems/espresso-network/blob/3a23e68ecdb106c28e40bb72615bab622006c838/crates/hotshot/types/src/light_client.rs#L285

Maybe I lost somewhere, why are we adding this?

types/src/v0/impls/stake_table.rs
```diff
@@ -222,6 +226,26 @@ impl StakeTableState {
         }
     }
 
+    pub fn hash(&self) -> Result<StakeTableStateHash, bincode::Error> {
+        let mut hasher = Keccak256::new();
```
mrain: I see. If we'd like some contract-friendly commitment to the stake table, maybe use the one in alloy crate like here: https://github.com/EspressoSystems/espresso-network/blob/97e1af5c2274e986d604ad7ea5da94f0e7433e70/crates/hotshot/task-impls/src/helpers.rs#L1494-L1501

crates/hotshot/types/src/traits/election.rs
bfish713: hmm for this one I feel it's ok as only sequencer really needs to implement this one, and so long a membership has the same impl and returns the same hash (or None) for a given stake table.  I guess I just don't see a reason we'd implement this separately for each test membership.

I'm also thinking maybe this should return `Some(StakeTableHash::default)`

types/src/v0/impls/header.rs
bfish713: These are two different values, one is just the hash of the actual stake table in use (after finding the top 100 and ignoring each individual delegator), the other is the summation of all the events on the contract and therefore can be used for catchup.

types/src/v0/impls/stake_table.rs
bfish713: Yes that's a good idea

sequencer/src/persistence/fs.rs
bfish713: This would be the fallback case where the storage was from before and doesn't have the hash (or reward ammount).  I aggree this should probably return None rather than default

types/src/v0/impls/state.rs
```diff
@@ -1003,6 +1013,52 @@ pub async fn get_l1_deposits(
     }
 }
 
+async fn validate_next_stake_table_hash(
+    instance: &NodeState,
+    proposed_header: &Header,
+) -> Result<(), ProposalValidationError> {
+    let Some(epoch_height) = instance.epoch_height else {
+        return Err(ProposalValidationError::NextStakeTableNotFound);
```
imabdulbasit: can we return correct error here i.e epoch height is not present

types/src/v0/impls/state.rs
```diff
@@ -1003,6 +1013,52 @@ pub async fn get_l1_deposits(
...
+    if !is_ge_epoch_root(proposed_header.height(), epoch_height) {
+        return Ok(());
+    }
+    let epoch = EpochNumber::new(epoch_from_block_number(
+        proposed_header.height(),
+        epoch_height,
+    ));
+    let coordinator = instance.coordinator.clone();
+    let Some(first_epoch) = coordinator.membership().read().await.first_epoch() else {
+        return Err(ProposalValidationError::NextStakeTableNotFound);
```
imabdulbasit: same here

data/v4/header.json
```diff
@@ -31,6 +31,7 @@
       "timestamp": "0x456"
     },
     "l1_head": 124,
+    "next_stake_table_hash": null,
```
imabdulbasit: This shouldn't be null for v4 header so that the reference test fails if the stake table hash type changes

types/src/v0/impls/stake_table.rs
```diff
@@ -222,6 +224,26 @@ impl StakeTableState {
         }
     }
 
+    pub fn hash(&self) -> Result<StakeTableStateHash, bincode::Error> {
+        let mut hasher = alloy::primitives::Keccak256::new();
+        for (address, validator) in &self.validators {
+            hasher.update(address);
+            hasher.update(bincode::serialize(&validator)?);
+        }
+        for key in self.used_bls_keys.iter().sorted() {
```
imabdulbasit, line 230: why do we want to also hash the used keys?

types/src/v0/impls/state.rs
```diff
@@ -1003,6 +1013,52 @@ pub async fn get_l1_deposits(
...
+    };
+
+    // Rewards are distributed only if the current epoch is not the first or second epoch
```
imabdulbasit, line 1039: comment needs to be fixed

types/src/v0/impls/header.rs
```diff
@@ -976,6 +988,41 @@ impl BlockHeader<SeqTypes> for Header {
             .await?;
         };
 
+        let mut next_stake_table_hash = None;
+
+        if version >= DrbAndHeaderUpgradeVersion::version() {
+            let epoch_height = instance_state
+                .epoch_height
+                .context("epoch height not in instance state")?;
+            if is_ge_epoch_root(height + 1, epoch_height) {
```
imabdulbasit, line 996: so headers from the epoch root up until the next epoch change will include the stake table hash, or am I misunderstanding?

types/src/v0/impls/stake_table.rs
bfish713: We don't necessarily need to, but it's just for completeness.

types/src/v0/impls/header.rs
bfish713: yeah, the next stake table hash, similar to how we do it with drb result and the lightclient state stuff Chengyu did

types/src/v0/impls/stake_table.rs
```diff
@@ -222,6 +224,26 @@ impl StakeTableState {
...
+            hasher.update(bincode::serialize(&key)?);
```
mrain: @sveitser @alxiong will there be any encoding problem here if we are hashing it in the contract?

sequencer/src/persistence/sql.rs
```diff
@@ -2234,13 +2240,17 @@ impl MembershipPersistence for Persistence {
             .map(|row| {
                 let stake_table_bytes: Vec<u8> = row.get("stake");
                 let reward_bytes: Option<Vec<u8>> = row.get("block_reward");
-                let stake_table = bincode::deserialize(&stake_table_bytes)
+                let (stake_table, _reward, stake_table_hash): (
```
imabdulbasit: I think this wouldn't work for older stake tables.

imabdulbasit: we should insert a new column i.e stake table hash and try to load that.

sequencer/src/persistence/sql.rs
```diff
@@ -2285,11 +2302,12 @@ impl MembershipPersistence for Persistence {
         epoch: EpochNumber,
         stake: ValidatorMap,
         block_reward: Option<RewardAmount>,
+        stake_table_hash: Option<StakeTableStateHash>,
     ) -> anyhow::Result<()> {
         let mut tx = self.db.write().await?;
 
-        let stake_table_bytes =
-            bincode::serialize(&(stake, block_reward)).context("serializing stake table")?;
+        let stake_table_bytes = bincode::serialize(&(stake, block_reward, stake_table_hash))
```
imabdulbasit: I think we should just store stake table bytes into stake column. We were already doing it wrong for SQL because it was serializing (stake table, block_reward) and deserializing it into (stake table)

sequencer/api/migrations/postgres/V1002__add_block_reward_column.sql
```diff
@@ -1,2 +1,3 @@
 ALTER TABLE epoch_drb_and_root
-ADD COLUMN block_reward BYTEA;
\ No newline at end of file
+ADD COLUMN block_reward BYTEA;
```
imabdulbasit:
```suggestion
ADD COLUMN block_reward BYTEA,
```

types/src/v0/impls/stake_table.rs
bfish713: We shouldn't need to do any hashing in the contract, the hash is just stored there so a light client could easily verify the stake table hash

bfish713: The flow would be to get the stake table state from another query node, hash it, then check the has matches what's stored in the LC contract.

sequencer/api/migrations/sqlite/V606__add_block_reward_column.sql
```diff
@@ -1,2 +1,3 @@
 ALTER TABLE epoch_drb_and_root
-ADD COLUMN block_reward BYTEA;
\ No newline at end of file
+ADD COLUMN block_reward BYTEA;
```
imabdulbasit: actually this BYTEA type is also wrong, not sure why the test didn't fail.
SQLite does not have BYTEA type
```suggestion
ADD COLUMN block_reward BLOB,
```

imabdulbasit (approved):

types/src/v0/impls/stake_table.rs
alxiong: i lack context, what is this new hash? a different stake table digest from the stake table commitment? 
do light-client/state-signer need to sign over (certify) this field?

(and yes, glad to hear that we don't need to do bincode serialize and hashing in contract :wink:)

crates/hotshot/types/src/traits/election.rs
```diff
@@ -17,6 +17,7 @@ use crate::{
     traits::signature_key::StakeTableEntryType, PeerConfig,
 };
 
+pub type StakeTableHash = [u8; 32];
```
mrain: nit: recommend to use something like `FixedBytes<32>` or wrapper so that it's prettier when printed out.

types/src/v0/impls/stake_table.rs
mrain: I feel like this is some ST commitment that's easier to compute in the contract. If you are not computing it in the contract, why don't we use `StakeTableState` that's already in the light client contract?

mrain: so as clarified by Brendon, the `StateTableState` already in the LC contract will only commit to the validators with voting power (and also it has a limit on the number of stake table entries), and this hash will touch all network participants. cc @alxiong

types/src/v0/impls/stake_table.rs
sveitser: Sorry bit late to the party. Do I understand correctly that this will commit to the currently internal representation of validators, so making any change to the validator struct or contained types will then require a protocol upgrade? I feel like we could get around this by hashing the L1 events instead.

imabdulbasit: I agree that changing anything in the struct shouldn't affect the hash, but shouldn't the hash be based on the processed events rather than the raw events, since that would mean nodes also agree on how the events are processed? but, maybe that doesn’t matter because  different stake table would stall consensus anyway

types/src/v0/impls/header.rs
```diff
@@ -1119,7 +1171,7 @@ impl BlockHeader<SeqTypes> for Header {
                 // Start with the reward Merkle tree root digest as the base input
                 let digest = header.reward_merkle_tree_root.digest();
                 hasher.update(digest.0);
-                hasher.update(placeholder_1);
+                hasher.update(next_stake_table_hash);
```
sveitser: @bfish713 @imabdulbasit @alxiong do we need this?

The auth root is for data that we need to authenticate on L1 (because we used non keccak hash for the block merkle tree, which is too expensive in the EVM).

Can't a validator complete a block merkle tree proof and authenticate that against the state of the LC contract?

types/src/v0/impls/stake_table.rs
sveitser: We already assume that honest validators compute the same stake table from the same sequence of L1 events, or in other words correctly filter out any bad events.

bfish713: Yeah it kind of does commit to the internal structure. I've been thinking on this because I didn't like this fact either.  Hashing all events would work, or hashing just the necessary information.  I think if we were to hash the information in the validator struct right now in a specific order we also don't really depend on the internal representation.  It would be just a hash of ever `Address` with it's stake amount, commission, stake key and then it's delegators (hashed the same way as validator w/o commission and w/o state_ver_key).  If any of those changed we'd have to do a protocol upgrade anyway.

That said I think just hashing the events might be the most straightforward/best way

types/src/v0/impls/header.rs
alxiong: Agree with Mathis, using auth root for this seem misplaced. We can't efficiently verify against this hash in contract, thus shouldn't put here

types/src/v0/impls/stake_table.rs
```diff
@@ -215,33 +218,46 @@ pub struct StakeTableState {
...
+
+        builder = builder.constant_str("used_schnorr_keys");
         for key in self
             .used_schnorr_keys
             .iter()
             .sorted_by(|a, b| a.to_affine().xy().cmp(&b.to_affine().xy()))
         {
-            hasher.update(bincode::serialize(&key)?);
+            let mut schnorr_key_bytes = vec![];
+            key.serialize_with_mode(&mut schnorr_key_bytes, ark_serialize::Compress::Yes)
```
sveitser, line 241: There is a `to_bytes!` macro shortcut in jellyfish that does this. There is also a slightly shorter `serialize_compressed` in `CanonicalSerialize`.

I think we don't really need compression here and for some types compression doesn't do anything. It annoys me a bit that we always have to choose. @alxiong @mrain is there an easy way to check if a type even supports compression? If compression is supported and we are going to hash the bytes immediately is the overhead of hashing the extra bytes or the overhead of compression larger?

types/src/v0/v0_3/stake_table.rs
```diff
@@ -44,7 +47,38 @@ pub struct Validator<KEY: SignatureKey> {
...
+    fn commit(&self) -> Commitment<Self> {
+        let mut schnorr_key_bytes = vec![];
+        self.state_ver_key.serialize_with_mode(&mut schnorr_key_bytes, ark_serialize::Compress::Yes).unwrap();
+
+        let mut builder = RawCommitmentBuilder::new(&Self::tag())
+            .fixed_size_field("account", &self.account)
+            .var_size_field("stake_table_key", self.stake_table_key.to_bytes().as_slice())
+            .var_size_field("state_ver_key", &schnorr_key_bytes)
+            .fixed_size_field("stake", &to_fixed_bytes(self.stake))
+            .u16(self.commission);
```
sveitser, line 66: nit: I think we usually use a domain separator, even if it's sometimes not strictly required.

types/src/v0/impls/stake_table.rs
bfish713: @mrain mentioned during a meeting that using compressed is best to do if we are not going to do any calculation in the contract, which we are not planning here.  I used teh `to_bytes` macro as well.

PR #3500, opened 2025-08-12
Add timestamp to load generator transaction payload

imabdulbasit:
addresses https://app.asana.com/1/1208976916964769/project/1210110822711063/task/1210964016342895


sveitser (approved):
LGTM. We could also build a TX that is easier to parse, e. g. something with `inner_payload` and `timestamp` field.

imabdulbasit:
> LGTM. We could also build a TX that is easier to parse, e. g. something with `inner_payload` and `timestamp` field.

the problem is we store the transactions in query service, so that requires more effort if we change the serialization

PR #3506, opened 2025-08-13
Update load generator to submit transaction to multiple nodes 

imabdulbasit:
addresses: https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1210964016342893

updates `ESPRESSO_SEQUENCER_URL` in submit-transactions binary  to `ESPRESSO_SEQUENCER_URLS` which takes in list of urls. When sending transaction to sequencer instead of builder, it selects a random node from the list

### DIFF
--- a/sequencer/src/bin/submit-transactions.rs
+++ b/sequencer/src/bin/submit-transactions.rs
@@ -153,9 +153,12 @@ struct Options {
 impl Options {
     fn submit_url(&self, rng: &mut ChaChaRng) -> Url {
         let sequencer_urls = self.urls.clone();
-        self.submit_url
-            .clone()
-            .unwrap_or_else(|| sequencer_urls[rng.gen_range(0..sequencer_urls.len())].clone())
+        self.submit_url.clone().unwrap_or_else(|| {
+            sequencer_urls[rng.gen_range(0..sequencer_urls.len())]
+                .clone()
+                .join("submit")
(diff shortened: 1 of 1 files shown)


sequencer/src/bin/submit-transactions.rs
```diff
@@ -125,9 +125,11 @@ struct Options {
     #[clap(long, env = "ESPRESSO_SUBMIT_TRANSACTIONS_SUBMIT_URL")]
     submit_url: Option<Url>,
 
-    /// URL of the query service.
-    #[clap(env = "ESPRESSO_SEQUENCER_URL")]
-    url: Url,
+    /// list of query service URLs.
+    /// If `ESPRESSO_SUBMIT_TRANSACTIONS_SUBMIT_URL` is not provided,
+    /// transactions will be sent to a random node from this list.
+    #[clap(env = "ESPRESSO_SEQUENCER_URLS", value_delimiter = ',', num_args = 1..,)]
```
sveitser, line 131: This seems kind of difficult to work with. Like if someone sets these URLs and also sets `ESPRESSO_SUBMIT_TRANSACTIONS_SUBMIT_URL` the URLs will be ignored?

Could we just make `ESPRESSO_SUBMIT_TRANSACTIONS_SUBMIT_URL` accept a `Vec<_>` instead of `Opt<_>`?

PR #3511, opened 2025-08-14
feat: consensus changes for LCV3

mrain:
Closes #<ISSUE_NUMBER>

### This PR:
Finishes the consensus for LCV3.


hotshot-state-prover/src/bin/state-prover.rs
```diff
@@ -170,44 +170,45 @@ async fn main() {
...
+        };
+    tracing::info!("Detected contract version v{contract_version}");
 
     // This bind version doesn't represent anything now, but it's required by the service trait
     let bind_version = StaticVersion::<0, 1> {};
 
     if args.daemon {
         // Launching the prover service daemon
-        let result = if is_legacy {
+        let result = if contract_version == 1 {
```
alxiong: let's write in `match case`, it's clearer.

same for the next nested `if else else`

sequencer/src/api.rs
```diff
@@ -5640,8 +5641,11 @@ mod test {
 
             // verify auth root if the consensus version is v4
             if header.version() == DrbAndHeaderUpgradeVersion::VERSION {
-                let auth_root = state_cert_v2.auth_root.expect("auth root is not None");
-                let header_auth_root = header.auth_root().unwrap().unwrap();
+                let auth_root = state_cert_v2.auth_root;
+                let header_auth_root = header.auth_root().unwrap();
+                if auth_root.is_zero() || header_auth_root.is_zero() {
+                    panic!("auth root shouldn't be zero");
```
alxiong, line 5647: just checking, is it okay to panic here? or return err?

alxiong (approved):
Can we also add a test logic (or `assert!`) in test-demo that after `newFinalizedState` the `authRoot` is no longer `bytes32(0)` and matches that of the `mock_ledger` or some states.

I wanted to make sure we didn't accidentally set to `::default()` somewhere we shouldn't.

sequencer/src/api.rs
mrain: Originally it was `Option<[u8; 32]>` and there're "unwrap"s here. cc @imabdulbasit

hotshot-state-prover/src/bin/state-prover.rs
mrain: [51094b0](https://github.com/EspressoSystems/espresso-network/pull/3511/commits/51094b08f973695a1f54f611eb36b9ce1d58460b)

sequencer/src/api.rs
imabdulbasit: this is for V4 header where we do calculate the auth_root field so yeah its fine to panic here if its not zero

types/src/v0/impls/header.rs
```diff
@@ -1101,9 +1101,9 @@ impl BlockHeader<SeqTypes> for Header {
         })
     }
 
-    fn auth_root(&self) -> anyhow::Result<Option<[u8; 32]>> {
+    fn auth_root(&self) -> anyhow::Result<FixedBytes<32>> {
         match self {
-            Header::V1(_) | Header::V2(_) | Header::V3(_) => Ok(None),
+            Header::V1(_) | Header::V2(_) | Header::V3(_) => Ok(FixedBytes::from([0u8; 32])),
```
imabdulbasit, line 1106: Nit: I do think using Option is better if the field is not present for versions <= V3, instead of using zeros

imabdulbasit (approved):

types/src/v0/impls/header.rs
mrain: How about putting the `Option` here but no `Result`? It's hard to imagine a failure case.

mrain: wait it's still weird to make it an option in `LightClientStateUpdateCertificateV2`

imabdulbasit: yeah I guess that is true as well. ok lets keep it as non Option field.

PR #3522, opened 2025-08-19
Unify the event streaming service with existing APIs

rob-maron:
Right now we have two separate endpoints for getting data from a node: 
1. One serves the majority of the data (availability, state, etc)
2. The other is solely the HotShot event stream [of consensus events]

This PR unifies both by moving the HotShot event stream handler into the existing endpoint. This will reduce headaches by not having to have two different endpoints to manage (ports, configuration, etc)

This PR also adds the event stream endpoint to the dev node, manifesting as the event stream from the first node.


crates/builder/src/bin/permissionless-builder.rs
```diff
@@ -20,10 +20,10 @@ struct NonPermissionedBuilderOptions {
     /// The builder will subscribe to this server to receive hotshot events
     #[clap(
         long,
-        env = "ESPRESSO_SEQUENCER_HOTSHOT_EVENT_STREAMING_API_URL",
-        default_value = "http://localhost:22001"
+        env = "ESPRESSO_SEQUENCER_API_URL",
```
imabdulbasit: why do we need another env var? We already pass `ESPRESSO_SEQUENCER_API_PORT` and can use that

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -857,6 +938,16 @@ async fn run_dev_node_server<ApiVer: StaticVersionType + 'static>(
 
     app.register_module("api", api).map_err(io::Error::other)?;
 
+    // Add the hotshot-events API to the dev-node API
+    let hotshot_events_api = hotshot_events_service::events::define_api::<_, _, ApiVer>(
```
imabdulbasit: why not just enable the hotshot events module in the query service?

sequencer/src/bin/espresso-dev-node.rs
```diff
@@ -735,18 +749,48 @@ pub struct ApiState {
     pub wallet: EthereumWallet,
     /// L1 chain id
     pub l1_chain_id: u64,
+    /// The HotShot event streamer for the first node
+    pub event_streamer: Option<Arc<RwLock<EventsStreamer<SeqTypes>>>>,
```
imabdulbasit: I think this whole thing can be removed if we just enable hotshot events module here 
```
 let api_options = options::Options::from(options::Http {
        port: sequencer_api_port,
        max_connections: sequencer_api_max_connections,
    })
    .submit(Default::default())
    .config(Default::default())
    .explorer(Default::default())
    .query_sql(Default::default(), sql)
    .hotshot_events(Default::default())
  ;
```

hotshot-events-service/src/events.rs
```diff
@@ -84,7 +84,7 @@ impl tide_disco::error::Error for Error {
 pub fn define_api<State, Types, Ver>(
     options: &Options,
     api_ver: semver::Version,
-) -> Result<Api<State, Error, Ver>, ApiError>
+) -> anyhow::Result<Api<State, Error, Ver>>
```
imabdulbasit: what is the reason for using `anyhow` here?

crates/builder/src/bin/permissionless-builder.rs
rob-maron: We do have that one, but we need the full URL since we cannot expect them to always run on the same host. I will look into removing the other though, good catch

sequencer/src/bin/espresso-dev-node.rs
rob-maron: Sure, will try this

hotshot-events-service/src/events.rs
rob-maron: It was to unify the error types, since `get_legacy_event_stream` is now fallible

hotshot-events-service/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT"
 
 [dependencies]
 alloy = { workspace = true }
+anyhow = "1"
```
ss-es: can this be `anyhow = { workspace = true }`?

sequencer/src/api/options.rs
```diff
@@ -639,11 +623,7 @@ pub struct State;
 
 /// Options for the Hotshot events streaming API module.
 #[derive(Parser, Clone, Copy, Debug, Default)]
-pub struct HotshotEvents {
-    /// Port that the HTTP Hotshot Event streaming API will use.
-    #[clap(long, env = "ESPRESSO_SEQUENCER_HOTSHOT_EVENT_STREAMING_API_PORT")]
-    pub events_service_port: u16,
-}
+pub struct HotshotEvents;
```
ss-es, line 626: do we still need this struct if it has no fields?

rob-maron: There are a few others we still use with no fields, like `pub struct Explorer;`, `pub struct State;` which is why I left it. I suppose it makes things more extensible if we need to add arguments, but I agree removing it is probably the best course of action

rob-maron: Okay, I see why, it's because the setup function uses the options presence as a way to figure out if the caller wants it enabled or not

crates/builder/src/bin/permissionless-builder.rs
rob-maron: Sorry - where is the `ESPRESSO_SEQUENCER_API_PORT` defined? I see this only in the dev node and the sequencer itself

imabdulbasit: sorry I pasted the wrong env variable.

we use `ESPRESSO_SEQUENCER_URL` in some binaries for query service url, so I was trying to say that can we use here as well instead of creating a new one?

rob-maron: Oh I see what you're saying, yeah will rename it

sequencer/src/bin/espresso-dev-node.rs
rob-maron: Made this change

sequencer/src/bin/espresso-dev-node.rs
rob-maron: Made this change

crates/builder/src/bin/permissionless-builder.rs
```diff
@@ -20,10 +20,10 @@ struct NonPermissionedBuilderOptions {
     /// The builder will subscribe to this server to receive hotshot events
     #[clap(
         long,
-        env = "ESPRESSO_SEQUENCER_HOTSHOT_EVENT_STREAMING_API_URL",
-        default_value = "http://localhost:22001"
+        env = "ESPRESSO_SEQUENCER_URL",
+        default_value = "http://localhost:24000"
```
imabdulbasit: nit: I don't like providing the default value here. I believe its better to explicitly provide it in demo

PR #3534, opened 2025-08-22
Fix get_epoch_drb

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/types/src/epoch_membership.rs
+++ b/crates/hotshot/types/src/epoch_membership.rs
@@ -45,7 +45,7 @@ pub struct EpochMembershipCoordinator<TYPES: NodeType> {
     /// Any in progress attempts at catching up are stored in this map
     /// Any new callers wantin an `EpochMembership` will await on the signal
     /// alerting them the membership is ready.  The first caller for an epoch will
-    /// wait for the actual catchup and allert future callers when it's done
+    /// wait for the actual catchup and alert future callers when it's done
     catchup_map: Arc<Mutex<EpochMap<TYPES>>>,
 
     drb_calculation_map: Arc<Mutex<DrbMap<TYPES>>>,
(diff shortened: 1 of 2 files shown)


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -263,42 +263,49 @@ where
 
         // Iterate through the epochs we need to fetch in reverse, i.e. from the oldest to the newest
         while let Some((current_fetch_epoch, tx)) = fetch_epochs.pop() {
-            if let Err(err) = <TYPES::Membership as Membership<TYPES>>::get_epoch_drb(
+            let root_leaf = match self.fetch_stake_table(current_fetch_epoch).await {
+                Ok(root_leaf) => root_leaf,
```
ss-es, line 267: should we `add_epoch_root` after this?

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -263,42 +263,49 @@ where
...
+                        .await
+                        .add_drb_result(current_fetch_epoch, drb_result);
+                },
+                Err(err) => {
+                    tracing::warn!(
+                        "DRB result for epoch {} missing from membership. Beginning catchup to \
+                         recalculate it. Error: {}",
+                        current_fetch_epoch,
+                        err
+                    );
```
ss-es, line 292: I think we could move this whole "fetch drb or recalculate it" logic to EpochMembershipCoordinator::get_epoch_drb for organization? and it would be nice to have only one place where we perform the raw call to `Membership::get_epoch_drb`

types/src/v0/impls/stake_table.rs
```diff
@@ -2208,11 +2208,11 @@ impl Membership<SeqTypes> for EpochCommittees {
             },
         };
 
-        let stake_table = membership_reader.stake_table(Some(epoch)).clone();
-        let success_threshold = membership_reader.success_threshold(Some(epoch));
+        let stake_table = membership_reader.stake_table(Some(previous_epoch)).clone();
+        let success_threshold = membership_reader.success_threshold(Some(previous_epoch));
```
ss-es, line 2212: this change looks correct to me. just to confirm our understandings agree:

when fetching the drb for epoch (say) 10, we want the root leaf from epoch 9 (which itself is the epoch root for epoch 11). we fetch that as part of a leaf chain of leaves from epoch 9. the root leaf in that chain has `next_drb_result` for epoch 10

ss-es (approved):
this looks correct to me

crates/hotshot/types/src/epoch_membership.rs
bfish713: This is confusing, it's actually contained in `fetch_stake_table`  I don't want to refactor so close to launch but it is being called if this returns `Ok(_)`

types/src/v0/impls/stake_table.rs
bfish713: basically my thinking, but we actually get a transition block from epoch 9, not the root.  For this code part of the code it just matters that it's a leaf from epoch 9 and we agree on that

crates/hotshot/types/src/epoch_membership.rs
bfish713: What if we delete `get_epoch_drb` from the `EpochMembership` (it's not even in the coordinator).  Then only catchup would call this.  I agree that something needs a reworks to organize all the drb work

PR #3536, opened 2025-08-25
update jellyfish merkle tree to 0.4.5-patch1

imabdulbasit:


hotshot-query-service/src/data_source/storage/sql/queries/state.rs
```diff
@@ -854,7 +855,8 @@ mod test {
                 .await
                 .unwrap();
             assert_eq!(proof.elem(), None);
-            assert!(test_tree.non_membership_verify(100, proof).unwrap());
+
+            assert!(UniversalMerkleTree::<usize, Sha3Digest, usize, 8, Sha3Node>::non_membership_verify(test_tree.commitment(), 100, proof).unwrap());
```
sveitser: @alxiong @imabdulbasit just curious, we don't have the previous API anymore? just looks a lot simpler.

alxiong: we do. Just import that `trait UniversalMerkleTree`, i think we should pass compiler without fully qualified syntax.

PR #3538, opened 2025-08-26
AF1.2.1 Improve error handling for transaction jobs

shenkeyao:
Closes https://app.asana.com/1/1208976916964769/project/1209392461754458/task/1211134150841486?focus=true.

### This PR:
* Adds `ErrPermanent` and `ErrEphemeral ` to better represent the errors due to transaction submission or fetch.
* Updates places where the errors are referenced.

### This PR does not:
* Change error handling logic in the repo. Will only affect the error handling in the OP integration repo.

### Key places to review:
* Whether the `ErrPermanent` and `ErrEphemeral` types are assigned correctly.


sdks/go/client/query.go
```diff
@@ -15,14 +15,14 @@ type QueryService interface {
 	// Get the header for block number `height`.
 	FetchHeaderByHeight(ctx context.Context, height uint64) (types.HeaderImpl, error)
 	// Get the raw header for block number `height`.
-	FetchRawHeaderByHeight(ctx context.Context, height uint64) (json.RawMessage, error)
```
ImJeremyHe: It’s a bit odd that only these endpoints return different error types. Your point is that we should identify whether an error is ephemeral. Could we return an extra bool to indicate that?

ImJeremyHe (commented):
Overall, this PR LGTM. Maybe I’m biased by my prior experience with Nitro’s code, where errors have always been returned as the error interface without wrapping them in additional error types.

Therefore, I’m a bit more inclined toward this style of design:
```go
var (
    ErrInvalidInfo = errors.New("invalid input")
    ErrRetryable   = errors.New("retryable error")
)

// Helper functions
func IsInvalidInfo(err error) bool {
    return errors.Is(err, ErrInvalidInfo)
}
func IsRetryable(err error) bool {
    return errors.Is(err, ErrRetryable)
}
```
Users can use these functions to determine the type of an error. And downstream code won’t need to make any intrusive changes either.

I think returning an extra `bool` is also a common approach.
```go
FetchRawHeaderByHeight(ctx context.Context, height uint64) (json.RawMessage, bool, error)
```

All three approaches, including the current PR, are fine with me, and I’d be eager to hear more thoughts on this.

shenkeyao:
@ImJeremyHe thank you for the review and the proposed solutions!

I'd prefer to avoid the first approach because that is similar to what we currently do in the `optimism-espresso-integration` branch. According to the audit report:
> This error handling mechanism exhibits extreme fragility through string-based error
classification, creating a significant maintenance burden and potential for runtime
failures.

I like the second approach of using `bool`. It's very similar to this PR's approach of using `TransactionErrorType`, with the difference that the former has 2 types, whereas the latter has 3. Originally, I found it counterintuitive to have either `true` or `false` to represent the "no error" case, so I added an extra `Success` type for it. However, on second thought, assigning `false` in case of "no error" actually seems reasonable.

How about the following structure combining these approaches?
`FetchRawHeaderByHeight(ctx context.Context, height uint64) (json.RawMessage, TransactionError)`
where `TransactionError` is
```
{
	err     error
	Retryable bool
}
```

shenkeyao:
@ImJeremyHe thanks for the clarification. I'm wondering how Nitro code handles errors like this: https://github.com/EspressoSystems/espresso-network/blob/main/sdks/go/client/client.go#L178.

In this example, it returns `return nil, err` instead of an explicit error, so I'd imagine the caller would have to analyze all possible errors covered by this `err`, which seems complicated. I think returning a `bool` would help simplify this, but maybe there's a better way to handle this. Curious about your thoughts!

ImJeremyHe:
@shenkeyao I agree that returning a bool can be very helpful. However, in my observation, most Go libraries return the error interface. So I wonder if we can make our Go SDK consistent with the wider Go community.

In the example you mentioned, if you consider this err to be ephemeral, regardless of its origin, you could do something like:
```go
var ErrEphemeral = errors.New("retryable")
```

```go
if err != nil {
    // Wrap err with ErrEphemeral. errors.Is can still detect ErrEphemeral.
    return fmt.Errorf("%w: %v", ErrEphemeral, err)
}
```

sdks/go/client/client.go
```diff
@@ -33,41 +33,52 @@ func NewClient(url string) *Client {
...
-	res, err := c.getRawMessage(ctx, "availability/header/%d", blockHeight)
-	if err != nil {
-		return nil, err
+func (c *Client) FetchRawHeaderByHeight(ctx context.Context, blockHeight uint64) (json.RawMessage, TransactionError) {
+	res, txnErr := c.getRawMessage(ctx, "availability/header/%d", blockHeight)
+	if txnErr.err != nil {
+		return nil, txnErr
 	}
-	return res, nil
+	return res, TransactionError{nil, false}
```
Sneh1999: why cant we just `return res, nil`?

sdks/go/client/client.go
```diff
@@ -78,35 +89,35 @@ func (c *Client) FetchExplorerTransactionByHash(ctx context.Context, hash *types
...
+		return types.TransactionQueryData{}, TransactionError{fmt.Errorf("hash is nil"), false}
 	}
 	var res types.TransactionQueryData
-	if err := c.get(ctx, &res, "availability/transaction/hash/%s", hash.String()); err != nil {
-		return types.TransactionQueryData{}, err
+	if txnErr := c.get(ctx, &res, "availability/transaction/hash/%s", hash.String()); txnErr.err != nil {
+		return types.TransactionQueryData{}, txnErr
 	}
-	return res, nil
+	return res, TransactionError{nil, false}
```
Sneh1999: same here

sdks/go/client/client.go
```diff
@@ -144,38 +155,42 @@ func (c *Client) FetchTransactionsInBlock(ctx context.Context, blockHeight uint6
...
 	}
 
 	var hash types.TaggedBase64
 	if err := json.Unmarshal(body, &hash); err != nil {
-		return nil, err
+		return nil, TransactionError{err, false}
 	}
-	return &hash, nil
+
+	return &hash, TransactionError{nil, false}
```
Sneh1999: same here

sdks/go/client/client.go
```diff
@@ -185,28 +200,33 @@ func (c *Client) getRawMessage(ctx context.Context, format string, args ...any)
...
 	// Read the response body into memory before we unmarshal it, rather than passing the io.Reader
 	// to the json decoder, so that we still have the body and can inspect it if unmarshalling
 	// failed.
 	body, err := io.ReadAll(res.Body)
 	if err != nil {
-		return nil, err
+		return nil, TransactionError{err, true}
 	}
-	return body, nil
+	return body, TransactionError{nil, false}
```
Sneh1999: here

sdks/go/client/client.go
```diff
@@ -185,28 +200,33 @@ func (c *Client) getRawMessage(ctx context.Context, format string, args ...any)
...
+	body, txnErr := c.getRawMessage(ctx, format, args...)
+	if txnErr.err != nil {
+		return txnErr
 	}
 	if err := json.Unmarshal(body, out); err != nil {
-		return fmt.Errorf("request failed with body %s and error %v", string(body), err)
+		return TransactionError{fmt.Errorf("request failed with body %s and error %v", string(body), err), false}
 	}
-	return nil
+	return TransactionError{nil, false}
```
Sneh1999: here

shenkeyao:
@Sneh1999 I've simplified the types so there's no longer `TransactionError{nil, false}`!

Sneh1999 (approved):
Looks good, one test seems to be failing though!

shenkeyao:
> Looks good, one test seems to be failing though!

@Sneh1999 It was an error mismatch. Should be fixed now!

PR #3539, opened 2025-08-26
Retry SQL transaction upsert

imabdulbasit:
addresses https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1211136528397777


hotshot-query-service/src/data_source/storage/sql/transaction.rs
```diff
@@ -399,17 +401,38 @@ impl Transaction<Write> {
...
+        let mut retries = 5;
+
+        loop {
+            let query = query_builder.build();
+            let statement = query.sql();
+            match self.execute(query).await {
+                Ok(res) => {
+                    let rows_modified = res.rows_affected() as usize;
+                    if rows_modified as usize != num_rows {
+                        tracing::error!(
```
sveitser: Do you know when this would happen? Currently we log an error but don't return an error. Is it something we should investigate? We could leave some more information in the error message.

sveitser (approved):

hotshot-query-service/src/data_source/storage/sql/transaction.rs
imabdulbasit: I don't think this can happen because we update incase of conflicts, but its good to log it. We were doing it when we had retries before

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3539/commits/eb6342e121a3c6e9d9a614fab7ab369aedb15b91

PR #3540, opened 2025-08-26
test: re-register validator with same Eth account

sveitser:
This test currently fails, but it should pass.


sveitser:
Will be addressed in https://github.com/EspressoSystems/espresso-network/pull/3541

PR #3543, opened 2025-08-26
validator delegatedAmount only decreased when withdrawal occurs

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- updates de-registration and undelegation function logic to only deduct from the validators' delegatedAmount when the delegator withdraws

### This PR does not:

- Check to see if there were any de-registrations in any deployments (sepolia)

### Key places to review:
StakeTable.sol
- claimWithdrawal functions and undelegate function
StakeTable.t.sol

### How to test this PR: 
forge test --match-contract StakeTable

---
  - https://app.asana.com/0/0/1210948765340116

### DIFF
--- a/contracts/test/StakeTable.t.sol
+++ b/contracts/test/StakeTable.t.sol
@@ -1368,7 +1368,7 @@ contract StakeTableUpgradeV2Test is Test {
         assertEq(
             stakeTable.token().balanceOf(delegator),
             remainderDelegation,
-            "delegator should have received the remaininder delegation"
+            "delegator should have received the remainder delegation"
         );
         assertEq(
             stakeTable.token().balanceOf(address(stakeTable)),
@@ -1412,7 +1412,6 @@ contract StakeTableUpgradeV2Test is Test {
     }
 
     function test_WithdrawalAfterValidatorExit() public {
-        // Should test withdrawing after validator exit
         stakeTableRegisterTest.setUp();
         vm.startPrank(stakeTableRegisterTest.admin());
         stakeTableRegisterTest.stakeTable().upgradeToAndCall(address(new StakeTableV2()), "");
@@ -1463,7 +1462,6 @@ contract StakeTableUpgradeV2Test is Test {
         vm.expectRevert(abi.encodeWithSelector(S.ValidatorAlreadyExited.selector));
         stakeTable.undelegate(validator, amountDelegated);
 
-        //validator already exited so call claim withdaral exit
         vm.warp(block.timestamp + stakeTable.exitEscrowPeriod());
         vm.expectEmit(false, false, false, true, address(stakeTable));
         emit S.Withdrawal(delegator, amountDelegated);


contracts/src/StakeTableV2.sol
```diff
@@ -94,32 +101,130 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        }
+
+        uint256 amount = delegations[validator][delegator];
+        if (amount == 0) {
+            revert NothingToWithdraw();
+        }
+
+        // Mark funds as spent
+        delegations[validator][delegator] = 0;
+        // deduct the amount from the validator's delegatedAmount
```
sveitser: I think this comment just explains in other words what the code is doing and can be removed.

We should however add a comment like `// new in V2`.

Please also add the changes to the summary of changes comment on top of the contract.

contracts/src/StakeTableV2.sol
```diff
@@ -94,32 +101,130 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        if (undelegations[validator][delegator].amount != 0) {
+            revert UndelegationAlreadyExists();
+        }
+
+        uint256 balance = delegations[validator][delegator];
+        if (balance < amount) {
+            revert InsufficientBalance(balance);
+        }
+
+        delegations[validator][delegator] -= amount;
```
sveitser: add comment about change w.r.t. to V1

contracts/src/StakeTableV2.sol
```diff
@@ -94,32 +101,130 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    /// @dev and to ensure that the validator's delegatedAmount is not updated until withdrawal
+    /// @dev delegatedAmount represents the no. of tokens that have been delegated to a validator,
+    /// even if it's not participating in consensus
     function deregisterValidator() public virtual override whenNotPaused {
-        super.deregisterValidator();
+        address validator = msg.sender;
+        ensureValidatorActive(validator);
+
+        validators[validator].status = ValidatorStatus.Exited;
+        validatorExits[validator] = block.timestamp + exitEscrowPeriod;
```
sveitser, line 204: add comment about change w.r.t. to V1

contracts/src/StakeTableV2.sol
```diff
@@ -94,32 +101,130 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        }
+
+        uint256 balance = delegations[validator][delegator];
+        if (balance < amount) {
+            revert InsufficientBalance(balance);
+        }
+
+        delegations[validator][delegator] -= amount;
+        undelegations[validator][delegator] =
+            Undelegation({ amount: amount, unlocksAt: block.timestamp + exitEscrowPeriod });
```
sveitser: Should we really remove decrementing `validator.delegatedAmount` here?

I think it would be better to keep that.

There are 2 ways for funds to be undelegated: undelegation, exit. For exit we can't do the accounting in the contract on exit because it would mean iterating over all delegators so we delay it until funds are claimed by the delegator.

With the current change I think we have the problem that: sum of delegated amounts to validator != validator.delegatedAmounts. So it seems to me like we're trading one problem for another, instead of fixing a problem.

cc @0xkato @alysiahuggins

sveitser: We should probably add whatever invariants we want to have to the invariant tests. So that we have some guarantees that they are satisfied and have the invariants captured more explicitly.

sveitser: @alysiahuggins @0xkato here is an invariant test that highlights what I mean. This fails with the current change.

- https://github.com/EspressoSystems/espresso-network/pull/3559

alysiahuggins: thank you @sveitser,i agree. my intentions were to have the validator.delegatedAmount only be adjusted upon withdrawals so that the semantics could be more easily understood but i agree with you and would prefeer that the sum of delegatedAmounts == validator.delegatedAmounts

contracts/src/StakeTableV2.sol
alysiahuggins: the comment was added here https://github.com/EspressoSystems/espresso-network/blob/e5665ac97224526fbbdb19634647b85d3bf53550/contracts/src/StakeTableV2.sol#L216-L219 but do you prefer it next to the line of code?

sveitser: yeah I think if we have to (mostly) copy the function then we should leave comments in the code what we changed exactly where that's feasible (unless it's a total rewrite).

contracts/src/StakeTableV2.sol
alysiahuggins: thank you, updated comments in the following commits:
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/6c87ffa4dacc13f1fb8b30ce5e5b30c49a522432
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/47c72f4f68d90d5ae1fe8635032506522b164f1d
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/8a01cbae324333f8d1c6e39132667a6f6177e67d

contracts/src/StakeTableV2.sol
alysiahuggins: thank you, updated comments in the following commits:
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/6c87ffa4dacc13f1fb8b30ce5e5b30c49a522432
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/47c72f4f68d90d5ae1fe8635032506522b164f1d
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/8a01cbae324333f8d1c6e39132667a6f6177e67d

contracts/src/StakeTableV2.sol
alysiahuggins: thank you, updated comments in the following commits:
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/6c87ffa4dacc13f1fb8b30ce5e5b30c49a522432
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/47c72f4f68d90d5ae1fe8635032506522b164f1d
https://github.com/EspressoSystems/espresso-network/pull/3543/commits/8a01cbae324333f8d1c6e39132667a6f6177e67d

contracts/src/StakeTableV2.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3543/commits/983be3219df0d90d3105b4c86df3711a99011e6a

contracts/test/StakeTable.t.sol
```diff
@@ -1547,6 +1290,455 @@ contract StakeTableUpgradeV2Test is Test {
...
+
+        assertEq(
+            stakeTable.token().balanceOf(delegator),
+            originalDelegateAmount,
+            "the delegator should have now received their full delegation"
+        );
+        assertEq(
+            stakeTable.token().balanceOf(address(stakeTable)),
+            0,
+            "the contract's balance should be zero"
```
sveitser, line 1408: nit: I think these assert strings are a bit noisy. The code should be clear enough.

contracts/test/StakeTable.t.sol
```diff
@@ -1547,6 +1290,455 @@ contract StakeTableUpgradeV2Test is Test {
...
+        vm.warp(block.timestamp + stakeTable.exitEscrowPeriod());
+        vm.expectEmit(false, false, false, true, address(stakeTable));
+        emit S.Withdrawal(delegator, remainderDelegation);
+        stakeTable.claimValidatorExit(validator);
+        vm.stopPrank();
+
+        assertEq(
+            stakeTable.token().balanceOf(delegator),
+            remainderDelegation,
+            "delegator should have received the remaininder delegation"
```
sveitser: nit: typo
```suggestion
            "delegator should have received the remainder delegation"
```

contracts/test/StakeTable.t.sol
```diff
@@ -1547,6 +1290,455 @@ contract StakeTableUpgradeV2Test is Test {
...
+            stakeTable.token().balanceOf(address(stakeTable)),
+            0,
+            "the contract's balance should be zero"
+        );
+        (validatorAmountDelegated,) = stakeTable.validators(validator);
+        assertEq(validatorAmountDelegated, 0, "the validator's delegatedAmount should be zero");
+    }
+
+    function test_WithdrawalAfterValidatorExit() public {
+        // Should test withdrawing after validator exit
```
sveitser:
```suggestion
```

contracts/test/StakeTable.t.sol
```diff
@@ -1547,6 +1290,455 @@ contract StakeTableUpgradeV2Test is Test {
...
+            0,
+            "the undelegated amount should be zero because the user did not undelegate"
+        );
+        assertEq(unlocksAt, 0, "the unlocks at should be zero because the user did not undelegate");
+
+        vm.startPrank(delegator);
+        vm.expectRevert(abi.encodeWithSelector(S.ValidatorAlreadyExited.selector));
+        stakeTable.undelegate(validator, amountDelegated);
+
+        //validator already exited so call claim withdaral exit
```
sveitser:
```suggestion
        //validator already exited so call claim validator exit
```
but I also think the code already explains this well enough

contracts/test/StakeTable.t.sol
```diff
@@ -1299,6 +1020,28 @@ contract StakeTableUpgradeV2Test is Test {
         stakeTableRegisterTest.setUp();
     }
 
+    function registerValidatorOnStakeTableV2(
```
sveitser, line 1023: Nice. I think if we added some similar helper functions for delegation, undelegation etc. the tests would be a bit easier to read but no need to make the change in this PR.

0xkato:
LGTM

contracts/test/StakeTable.t.sol
alysiahuggins: i agree but i added them when i was testing the new logic and found it helpful, the tests are quite long, i can remove these when we spend time to refactor the tests so that they are shorter and thus easier to read

PR #3544, opened 2025-08-26
Garbage collect view sync tasks on decide

ss-es:
Broadcasts an internal event when HotShot reaches a decide, and garbage collects view sync tasks based on that event

### DIFF
--- a/crates/hotshot/task-impls/src/quorum_vote/handlers.rs
+++ b/crates/hotshot/task-impls/src/quorum_vote/handlers.rs
@@ -219,13 +219,18 @@ pub(crate) async fn handle_quorum_proposal_validated<
                 leaf_info.leaf.view_number(),
                 leaf_info.leaf.block_header().block_number(),
             );
-
-            broadcast_event(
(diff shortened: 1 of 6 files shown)


crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -219,6 +219,12 @@ pub(crate) async fn handle_quorum_proposal_validated<
                 leaf_info.leaf.view_number(),
                 leaf_info.leaf.block_header().block_number(),
             );
+
```
bfish713: Rather than do this here, can we do it in the same place we broadcast the external decide event?

ss-es: actually, isn't this already in the right place? this is the log saying we're broadcasting the external event which happens immediately after on L231; the only reason I put the internal broadcast inside the loop is because the internal decide event is for an individual `Leaf` (but I can change this)

bfish713: oh duh, I think we should do it like the other event still and send the whole chain in 1 event

PR #3551, opened 2025-08-26
Revert "Fix quorum proposal struct"

Sneh1999:
Reverts EspressoSystems/espresso-network#3548 because we realized this is for v0 endpoint but v1 supports the older interface


zacshowa (approved):
👍

PR #3553, opened 2025-08-27
fix: nsproof serialization

mrain:
Closes #<ISSUE_NUMBER>

### This PR:

Now json serialization of `NsProof` looks much better.

### DIFF
--- a/sdks/go/client/client.go
+++ b/sdks/go/client/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


sveitser:
@mrain this looks like a breaking change for clients on decaf. Do you know if the integrations team are aware of this change?

cc @Sneh1999 @philippecamacho @ImJeremyHe

mrain:
> Why were the files in the directory `data` so big?

Originally they (the payload) are raw bytes, like `[3, 0, 0, 0, ...]` which have thousands of lines. Now it's base64 encoded.

PR #3555, opened 2025-08-27
Catchup for previous epoch stake table in dynamic reward calculation

imabdulbasit:
---
  - https://app.asana.com/0/0/1211173012986263

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -37,13 +37,14 @@ use hotshot_types::{
         election::{generate_stake_cdf, select_randomized_leader, RandomizedCommittee},
         DrbResult,
     },
+    epoch_membership::EpochMembershipCoordinator,
     stake_table::{HSStakeTable, StakeTableEntry},
     traits::{
         election::Membership,
         node_implementation::{ConsensusTime, NodeType},
         signature_key::StakeTableEntryType,
     },
-    utils::{epoch_from_block_number, transition_block_for_epoch},
+    utils::{epoch_from_block_number, root_block_in_epoch, transition_block_for_epoch},
     PeerConfig,
 };
 use humantime::format_duration;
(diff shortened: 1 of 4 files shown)


types/src/v0/impls/stake_table.rs
```diff
@@ -1557,6 +1559,66 @@ impl EpochCommittees {
...
+        let first_epoch = *self.first_epoch().context("first epoch not initialized")?;
+
+        if *current_epoch <= first_epoch + 1 {
+            bail!("no block reward for first two epochs");
+        }
+
+        let header = committee
+            .header
+            .clone()
+            .context(format!("missing header for epoch={current_epoch:?}"))?;
```
bfish713: I'm worried there is an unrecoverable condition if the header is not here.  Of course it should be, but I am imagining we have a bug where we reload the stake table from storage, but fail to get the reward somehow.  We would not have the header on the committee in this case and there is no recover path.

imabdulbasit: yeah I am also thinking about this. Should we catchup if header is not there?

bfish713: yeah I think we have to

types/src/v0/impls/stake_table.rs
```diff
@@ -1557,6 +1559,100 @@ impl EpochCommittees {
         Ok(block_reward_u256.into())
     }
 
+    pub async fn get_block_reward(
+        epoch: Option<Epoch>,
```
sveitser: @imabdulbasit Don't quite understand why this is optional. In what case do we call this function without epoch?

imabdulbasit: there was no block reward per committee in V3, so if the version is V3 then we just pass None to get fixed block reward.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3555/commits/1231dfcccaad0a8e58611a51720c79b5b3d3135b

bfish713 (approved):

PR #3556, opened 2025-08-27
Move DRB calculation outside of the catchup loop

ss-es:


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -323,6 +287,46 @@ where
...
         }
+
+        let root_leaf = match self.fetch_stake_table(epoch).await {
+            Ok(root_leaf) => root_leaf,
+            Err(err) => {
+                fetch_epochs.push((epoch, epoch_tx));
+                self.catchup_cleanup(epoch, fetch_epochs, err).await;
+                return;
+            },
+        };
```
bfish713: If we're doing this here it's just for getting the root leaf and we only need it if we fail to fetch, so lets move this to the Err block below

bfish713 (approved):

PR #3560, opened 2025-08-28
Add reward account proof endpoint

imabdulbasit:
---
  - https://app.asana.com/0/0/1211150115961747

### DIFF
--- a/sequencer/api/reward.toml
+++ b/sequencer/api/reward.toml
@@ -8,3 +8,9 @@ PATH = ["reward-balance/:height/:address"]
 ":address" = "Literal"
 ":height" = "Integer"
 DOC = "Get balance in reward state at a specific height. Expected parameters are height which is an integer and an Ethereum address in hex format"
+
+[route.get_reward_account_proof]
+PATH = ["proof/:height/:address"]
+":address" = "Literal"
+":height" = "Integer"
+DOC = "Get reward account merkle proof"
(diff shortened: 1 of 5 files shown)


sequencer/src/api.rs
```diff
@@ -5676,4 +5678,154 @@ mod test {
...
+        // wait for 4 epochs
+        let mut events = network.server.event_stream().await;
+        wait_for_epochs(&mut events, EPOCH_HEIGHT, 4).await;
+
+        let url = format!("http://localhost:{api_port}").parse().unwrap();
+        let client: Client<ServerError, StaticVersion<0, 1>> = Client::new(url);
+        network.stop_consensus().await;
+
+        let validated_state = network.server.decided_state().await;
+        // let decided_leaf = network.server.decided_leaf().await;
```
sveitser:
```suggestion
```

sequencer/api/reward.toml
```diff
@@ -8,3 +8,9 @@ PATH = ["reward-balance/:height/:address"]
 ":address" = "Literal"
 ":height" = "Integer"
 DOC = "Get balance in reward state at a specific height. Expected parameters are height which is an integer and an Ethereum address in hex format"
+
+[route.get_reward_account_proof]
```
sveitser, line 12: IIUC currently the client will get a proof for the V1 or the V2 rewards merkle tree but doesn't have an easy way to check what it is. Since we don't need this to work for the V1 tree maybe it would be simpler to only enable the API for V2? 

Wdyt @imabdulbasit ?

imabdulbasit: if we only need v2 proof then we can call v2 endpoint i.e `reward-state-v2/`?

sveitser: @imabdulbasit reminded me that these two versions are served on different API versions so my comment is moot.

PR #3568, opened 2025-08-29
pass genesis version when creating genesis header from node state

imabdulbasit:


data/v2/messages.json
```diff
@@ -46,11 +46,11 @@
               "justify_qc": {
                 "_pd": null,
                 "data": {
-                  "leaf_commit": "COMMIT~tqpBNqqt63akxJWcfI46uDV8c_Sn_aZ777yCj4X_ukDG"
```
imabdulbasit, line 49: This is the genesis leaf commit, which is different now because we use the base version as the genesis version when creating this message, instead of using V1

PR #3569, opened 2025-08-29
Store eQC

ss-es:


sequencer/src/persistence/fs.rs
```diff
@@ -1064,6 +1068,45 @@ impl SequencerPersistence for Persistence {
...
+    }
+
+    async fn load_eqc(
+        &self,
+    ) -> Option<(
+        QuorumCertificate2<SeqTypes>,
+        NextEpochQuorumCertificate2<SeqTypes>,
+    )> {
+        let inner = self.inner.read().await;
+        let path = inner.next_epoch_qc();
```
pls148: should be `inner.eqc()`

ss-es: oops, good catch -- thank you!

sequencer/src/persistence/sql.rs
```diff
@@ -2000,6 +2000,43 @@ impl SequencerPersistence for Persistence {
             .transpose()
     }
 
+    async fn store_eqc(
+        &self,
+        high_qc: QuorumCertificate2<SeqTypes>,
+        next_epoch_high_qc: NextEpochQuorumCertificate2<SeqTypes>,
+    ) -> anyhow::Result<()> {
+        let eqc_bytes =
+            bincode::serialize(&(high_qc, next_epoch_high_qc)).context("serializing eqc")?;
```
imabdulbasit, line 2009: I would prefer if there is separate column for each certificate in the SQL table

PR #3578, opened 2025-09-02
fix first_reward_block calculation

imabdulbasit:


types/src/v0/impls/reward.rs
```diff
@@ -848,22 +848,33 @@ pub async fn distribute_block_reward(
...
+            // is beyond the first rewardable block.
+            if height > first_reward_block {
+                // If v4 upgrade started at block 101, and first_reward_block is 21:
+                // total_distributed = (101 - 21) * fixed_block_reward
+                let blocks = height.checked_sub(first_reward_block).with_context(|| {
+                    format!(
+                        "height ({height}) - first_reward_block ({first_reward_block}) underflowed"
+                    )
+                })?;
+                previously_distributed = U256::from(blocks)
```
sveitser: @imabdulbasit can we separate the computation of block rewards and previously distributed reward? I think that would make it easier to read.

sveitser: The computation of `previously_distributed` does not need to know the block reward. IIUC

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3578/commits/5ace34b2baaf4c06da8b8632d3e244b82fe070db

PR #3585, opened 2025-09-04
Fetch multiple stake table events in one rpc call 

imabdulbasit:
This PR removes separate RPC call for each stake table event type

https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1211245502625859

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -112,7 +112,43 @@ impl TryFrom<StakeTableV2Events> for StakeTableEvent {
             StakeTableV2Events::Undelegated(v) => Ok(StakeTableEvent::Undelegate(v)),
             StakeTableV2Events::ConsensusKeysUpdated(v) => Ok(StakeTableEvent::KeyUpdate(v)),
             StakeTableV2Events::ConsensusKeysUpdatedV2(v) => Ok(StakeTableEvent::KeyUpdateV2(v)),
-            _ => Err(anyhow::anyhow!("Unsupported StakeTableV2Events variant ")),
+            StakeTableV2Events::CommissionUpdated(v) => Ok(StakeTableEvent::CommissionUpdate(v)),
+            StakeTableV2Events::ExitEscrowPeriodUpdated(v) => Err(anyhow::anyhow!(
+                "Unsupported StakeTableV2Events::ExitEscrowPeriodUpdated({v:?})"
+            )),
+            StakeTableV2Events::Initialized(v) => Err(anyhow::anyhow!(
+                "Unsupported StakeTableV2Events::Initialized({v:?})"
+            )),
+            StakeTableV2Events::MaxCommissionIncreaseUpdated(v) => Err(anyhow::anyhow!(
+                "Unsupported StakeTableV2Events::MaxCommissionIncreaseUpdated({v:?})"
+            )),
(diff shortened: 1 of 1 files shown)


types/src/v0/impls/stake_table.rs
```diff
@@ -853,240 +850,112 @@ impl Fetcher {
...
-        let deregistrations = deregistered_events.flatten().collect().await;
-        let delegated = delegated_events.flatten().collect().await;
-        let undelegated = undelegated_events.flatten().collect().await;
-        let keys = keys_update_events.flatten().collect().await;
-        let keys_v2 = keys_update_events_v2.flatten().collect().await;
-
-        StakeTableEvents::from_l1_logs(
+        }
+
+        Ok(StakeTableEvents::from_l1_logs(
```
sveitser: @imabdulbasit Currently I think we break up the events such that each has their own container then merge them again for sorting. I think we only did this because we fetched them individually so I think we can keep them all in the same container now. What do you think?

sequencer/src/persistence.rs
```diff
@@ -1546,6 +1548,8 @@ mod tests {
         for _i in 0..10 {
             // Wait for more than update interval to assert that persistence was updated
             // L1 update interval is 7s in this test
+
+            use espresso_types::sort_stake_table_events;
```
sveitser: we don't usually import in the middle of the code

types/src/v0/impls/stake_table.rs
```diff
@@ -853,240 +782,82 @@ impl Fetcher {
...
-        StakeTableEvents::from_l1_logs(
-            registrations,
-            registrations_v2,
-            deregistrations,
-            delegated,
-            undelegated,
-            keys,
-            keys_v2,
-        )
+        Ok(events)
```
sveitser: We could already sort the events here?

types/src/v0/impls/stake_table.rs
```diff
@@ -102,114 +100,46 @@ impl DisplayLog for Log {
...
+    fn try_from(value: StakeTableV2Events) -> anyhow::Result<Self> {
+        match value {
+            StakeTableV2Events::ValidatorRegistered(v) => Ok(StakeTableEvent::Register(v)),
+            StakeTableV2Events::ValidatorRegisteredV2(v) => Ok(StakeTableEvent::RegisterV2(v)),
+            StakeTableV2Events::ValidatorExit(v) => Ok(StakeTableEvent::Deregister(v)),
+            StakeTableV2Events::Delegated(v) => Ok(StakeTableEvent::Delegate(v)),
+            StakeTableV2Events::Undelegated(v) => Ok(StakeTableEvent::Undelegate(v)),
+            StakeTableV2Events::ConsensusKeysUpdated(v) => Ok(StakeTableEvent::KeyUpdate(v)),
+            StakeTableV2Events::ConsensusKeysUpdatedV2(v) => Ok(StakeTableEvent::KeyUpdateV2(v)),
+            _ => Err(anyhow::anyhow!("Unsupported StakeTableV2Events variant")),
```
sveitser: I think would be good to include the event in the error for easier inspection.

imabdulbasit: StakeTableV2Events doesn't implement debug or display trait :/

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3585/commits/a0083ebd8f09d33f9a50bbd83676aec4555916af

PR #3587, opened 2025-09-04
Add genesis chain config in the genesis toml file

imabdulbasit:


sequencer/src/genesis.rs
```diff
@@ -58,6 +58,7 @@ pub struct Genesis {
     pub drb_upgrade_difficulty: Option<u64>,
     pub epoch_start_block: Option<u64>,
     pub stake_table_capacity: Option<usize>,
+    pub genesis_chain_config: Option<ChainConfig>,
```
ss-es: I'm not sure about this being optional -- is it any more difficult to make it mandatory? genesis_chain_config as a concept always makes sense so why not always require it

ss-es (approved):
looks fine, though question about the `Option`

PR #3590, opened 2025-09-05
fix: handle `EpochRootQC` message from older hotshot version

mrain:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/types/src/message.rs
+++ b/crates/hotshot/types/src/message.rs
@@ -256,8 +256,8 @@ pub enum GeneralConsensusMessage<TYPES: NodeType> {
         NextEpochQuorumCertificate2<TYPES>,
     ),
 
-    /// Message for the next leader containing the epoch root QC
-    EpochRootQc(EpochRootQuorumCertificateV2<TYPES>),
+    /// Message for the next leader containing the epoch root QC from older consensus version.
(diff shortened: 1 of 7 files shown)


crates/hotshot/types/src/message.rs
```diff
@@ -257,7 +257,10 @@ pub enum GeneralConsensusMessage<TYPES: NodeType> {
     ),
 
     /// Message for the next leader containing the epoch root QC
-    EpochRootQc(EpochRootQuorumCertificate<TYPES>),
+    EpochRootQc(EpochRootQuorumCertificateV2<TYPES>),
```
ss-es: can we rename this `EpochRootQcV2` for clarity? we can leave it for later if the type is used in a lot of places

ss-es (approved):
looks good to me

crates/hotshot/types/src/message.rs
```diff
@@ -257,7 +257,10 @@ pub enum GeneralConsensusMessage<TYPES: NodeType> {
...
+
+    /// Message for the next leader containing the epoch root QC from older consensus version.
+    EpochRootQcV1(EpochRootQuorumCertificateV1<TYPES>),
```
ss-es, line 260: this causes serialization issues, the extra variant must be added to the end

ss-es (changes requested):

PR #3591, opened 2025-09-05
feat: updatable validator commissions

sveitser:
- stake table contract: allow updates with rate limits
- migrate commissions from V1 stake table into storage for V2
- handle commission changes in the GCL
- test: stake table state gets invalid commission
- break up some lines to fix cargo fmt
- handle patch upgrades for stake table
- stake table upgrade: dedupe upgrade preparation
- test: unittest commission updates

### What's not done
- Enforcing of limits to commission changes in GCL. I left a comment, not quite clear to me if we want to do that since it doesn't seem relevant for consensus.
- Add a new protocol version. I think we may be able to do this but it's pending some discussion.

Tests:
```
forge test --match-contract StakeTableV2Comm
cargo test -p espresso-types
cargo test -p espresso-contract-deployer
```

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -2329,25 +2329,42 @@ impl Membership<SeqTypes> for EpochCommittees {
         };
 
         // If the epoch committee:
-        // - exists and has a header and block reward, return early.
+        // - exists and has a header stake table hash and block reward, return early.
         // - exists without a reward, reuse validators and update reward.
+        // and fetch from L1 if the stake table hash is missing.
         // - doesn't exist, fetch it from L1.
         let (validators, stake_table_hash) = match epoch_committee {
-            Some(ref committee) => {
-                if committee.block_reward.is_some() && committee.header.is_some() {
-                    tracing::info!(
-                        ?epoch,
-                        "committee already has block reward and header, skipping add_epoch_root"
-                    );
-                    return Ok(());
-                }
+            Some(committee)
+                if committee.block_reward.is_some()
+                    && committee.header.is_some()
+                    && committee.stake_table_hash.is_some() =>
+            {
+                tracing::info!(
+                    ?epoch,
+                    "committee already has block reward, header, and stake table hash; skipping \
+                     add_epoch_root"
+                );
+                return Ok(());
+            },
 
+            Some(committee) => {
                 if let Some(reward) = committee.block_reward {
                     block_reward = Some(reward);
                 }
 
-                (committee.validators.clone(), committee.stake_table_hash)
+                if let Some(hash) = committee.stake_table_hash {
+                    (committee.validators.clone(), Some(hash))
+                } else {
+                    // if stake table hash is missing then recalculate from events
+                    tracing::info!(
+                        "Stake table hash missing for epoch {epoch}. recalculating by fetching \
+                         from l1."
+                    );
+                    let (map, hash) = fetcher.fetch(epoch, &block_header).await?;
+                    (map, Some(hash))
+                }
             },
+
             None => {
                 tracing::info!("Stake table missing for epoch {epoch}. Fetching from L1.");
                 let (map, hash) = fetcher.fetch(epoch, &block_header).await?;


contracts/rust/deployer/src/lib.rs
```diff
@@ -1568,6 +1678,133 @@ mod tests {
...
+            &provider,
+            *stake_table_v2.address(),
+        )
+        .await
+        .unwrap_err();
+        assert!(err.to_string().contains("Expected StakeTable V1"));
+
+        Ok(())
+    }
+
```
sveitser, line 1769: Should add a test that does a deployment with (non-empty) commission migration.

contracts/src/StakeTableV2.sol
```diff
@@ -55,9 +55,47 @@ import { SafeTransferLib } from "solmate/utils/SafeTransferLib.sol";
...
+///
 /// @notice The StakeTableV2 contract ABI is a superset of the original ABI. Consumers of the
 /// contract can use the V2 ABI, even if they would like to maintain backwards compatibility.
 contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeable {
+    // === Types ===
+
+    /// @notice Struct for tracking validator commission and last increase time
+    struct CommissionTracking {
+        uint16 commission;
+        uint256 lastIncreaseTime;
```
alysiahuggins, line 68: a validator may also decrease commission so maybe rename to `lastModified` or `lastChanged`

sveitser: The decreases are not rate limited, and therefore the timestamp isn't tracked and and the amount of the decrease isn't checked.

contracts/rust/deployer/src/lib.rs
sveitser: for test added here: 1bc21809e13a3802dad082e07f476336de99382c

contracts/src/StakeTableV2.sol
alysiahuggins: ok gotcha

contracts/src/StakeTableV2.sol
```diff
@@ -55,9 +55,47 @@ import { SafeTransferLib } from "solmate/utils/SafeTransferLib.sol";
...
+        address validator;
+        uint16 commission;
+    }
+
+    // === Storage ===
+
+    bytes32 public constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
+
+    /// @notice Minimum time interval between commission updates (in seconds)
+    uint256 public minCommissionUpdateInterval;
```
alysiahuggins: should this be `minCommissionUpdateInterval` since it only applies to commission increases as only increases are rate limited as mentioned [here](https://github.com/EspressoSystems/espresso-network/pull/3591/files#r2329362640). 

also maybe this can be uint32?

contracts/src/StakeTableV2.sol
```diff
@@ -290,6 +373,93 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    /// These limits protect stakers from sudden large commission increases,
+    /// particularly by exiting validators.
+    function updateCommission(uint16 newCommission) external virtual whenNotPaused {
+        address validator = msg.sender;
+        ensureValidatorActive(validator);
+        require(newCommission <= 10000, InvalidCommission());
+
+        CommissionTracking storage tracking = commissionTracking[validator];
+        uint16 currentCommission = tracking.commission;
+        uint256 lastIncreaseTime = tracking.lastIncreaseTime;
```
alysiahuggins: you can put this line within the for loop as this is the only place where `lastIncreaseTime` is used

contracts/src/StakeTableV2.sol
```diff
@@ -290,6 +373,93 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        if (newCommission > currentCommission) {
+            // Allow immediate first increase or after interval has passed
+            require(
+                lastIncreaseTime == 0
+                    || block.timestamp >= lastIncreaseTime + minCommissionUpdateInterval,
+                CommissionUpdateTooSoon()
+            );
+
+            uint16 increase = newCommission - currentCommission;
+            require(increase <= maxCommissionIncrease, CommissionIncreaseExceedsMax());
```
alysiahuggins: i'm thinking we can remove the line above and change this to
```suggestion
           if (newCommission > currentCommission + maxCommissionIncrease) revert CommissionIncreaseExceedsMax();
```

contracts/src/StakeTableV2.sol
```diff
@@ -290,6 +373,93 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        // to commission increases.
+        //
+        // 1. Limits only enforced for commission *increase*.
+        // 2. Time of last change only tracked for commission *increase*.
+        if (newCommission > currentCommission) {
+            // Allow immediate first increase or after interval has passed
+            require(
+                lastIncreaseTime == 0
+                    || block.timestamp >= lastIncreaseTime + minCommissionUpdateInterval,
+                CommissionUpdateTooSoon()
```
alysiahuggins: same nit as above about the use of require
```suggestion
            if(
                lastIncreaseTime != 0
                    && block.timestamp < lastIncreaseTime + minCommissionUpdateInterval)
                revert CommissionUpdateTooSoon()
```

contracts/src/StakeTableV2.sol
```diff
@@ -84,6 +122,22 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @notice The exit escrow period is updated
     event ExitEscrowPeriodUpdated(uint64 newExitEscrowPeriod);
 
+    /// @notice A validator updates their commission rate
+    /// @param validator The address of the validator
+    /// @param timestamp The timestamp of the update
+    /// @param newCommission The new commission rate
+    ///
+    /// @dev the timestamp is emitted to simplify processing in the GCL
+    event CommissionUpdated(address indexed validator, uint256 timestamp, uint16 newCommission);
```
alysiahuggins: could be useful to log the previous commission too

contracts/src/StakeTableV2.sol
```diff
@@ -290,6 +373,93 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    function updateCommission(uint16 newCommission) external virtual whenNotPaused {
+        address validator = msg.sender;
+        ensureValidatorActive(validator);
+        require(newCommission <= 10000, InvalidCommission());
+
+        CommissionTracking storage tracking = commissionTracking[validator];
+        uint16 currentCommission = tracking.commission;
+        uint256 lastIncreaseTime = tracking.lastIncreaseTime;
+
+        require(newCommission != currentCommission, CommissionUnchanged());
```
alysiahuggins, line 394: random nit: in our codebase, we normally write it like 
```suggestion
        if(newCommission == currentCommission) revert CommissionUnchanged();
```

contracts/src/StakeTableV2.sol
```diff
@@ -290,6 +373,93 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    }
+
+    /// @notice Initialize validator commissions during V2 migration
+    /// @dev This function is used to retroactively initialize commission storage for validators
+    /// that were registered before the V2 upgrade. On decaf, this will be called with current
+    /// commission values read from L1 events. On mainnet, this will be called with an empty array
+    /// since there are no pre-existing validators.
+    /// @param initialCommissions Array of InitialCommission structs containing validator addresses
+    /// and their commissions
+    function _initializeCommissions(InitialCommission[] calldata initialCommissions) private {
```
alysiahuggins, line 446: before we start the for loop, we should do a check to see if the length of initialCommissions is equal to the number of validators in the contract, and revert otherwise

contracts/test/StakeTableV2Commission.t.sol
```diff
@@ -0,0 +1,209 @@
...
+        vm.stopPrank();
+    }
+
+    function test_DefaultValues() public view {
+        // This is the only test that checks default values - if defaults change, only this test
+        // should fail
+        assertEq(proxy.minCommissionUpdateInterval(), 7 days);
+        assertEq(proxy.maxCommissionIncrease(), 500);
+    }
+}
```
alysiahuggins, line 268: we should add a test for handling an upgrade from v1 to v2 where the initial commission array is not empty

contracts/src/StakeTableV2.sol
sveitser: I'll change the name. Yeah it could be a smaller type but I think just using the same type as block.timestamp is okay.

contracts/src/StakeTableV2.sol
sveitser: Why avoid require?

contracts/src/StakeTableV2.sol
sveitser: I think this is for historical reasons because require did not support custom errors when we started. It supports that now.

contracts/test/StakeTableV2Commission.t.sol
sveitser: There are rust tests for this (which also checks that we are correctly fetching the commissions), but sure can also add a solidity test.

contracts/src/StakeTableV2.sol
alysiahuggins: mostly because we use the `if (!cond) revert Error();` format in the rest of the codebase

contracts/src/StakeTableV2.sol
sveitser: Uhm... we can't get the number of validators from the validators mapping and don't store it otherwise.

sveitser: I think this is okay though, the migration will only be applied on decaf. There are checks that the validators are registered so we can't accidentally pass anything on the fresh mainnet deployment.

sveitser: There is a problem though if the validator has zero commission, then after initialization it's not possible to tell if the tracking is initialized or not.

contracts/src/StakeTableV2.sol
sveitser: partially done in https://github.com/EspressoSystems/espresso-network/pull/3591/commits/6e6f3b2ea1ff644b9cba7836dd863f8fb7910562

contracts/src/StakeTableV2.sol
sveitser: I think using require is fine. It reads better. I think the use of if ... revert just became common when custom errors weren't supported with require.

contracts/src/StakeTableV2.sol
sveitser: https://github.com/EspressoSystems/espresso-network/pull/3591/commits/6e6f3b2ea1ff644b9cba7836dd863f8fb7910562

contracts/src/StakeTableV2.sol
sveitser: Added more tests in https://github.com/EspressoSystems/espresso-network/pull/3591/commits/6e6f3b2ea1ff644b9cba7836dd863f8fb7910562

sveitser: So basically if a validator has zero commission it can be duplicated. I think this is fine as long as our rust code fetches and prepares the commissions correctly which I think it does.

contracts/src/StakeTableV2.sol
sveitser: https://github.com/EspressoSystems/espresso-network/commit/6e6f3b2ea1ff644b9cba7836dd863f8fb7910562

contracts/src/StakeTableV2.sol
sveitser: renamed in https://github.com/EspressoSystems/espresso-network/commit/6e6f3b2ea1ff644b9cba7836dd863f8fb7910562

alysiahuggins:
the contract side of things lgtm

contracts/rust/deployer/src/lib.rs
```diff
@@ -1568,6 +1686,183 @@ mod tests {
...
+    /// env RPC_URL=... cargo test -p espresso-contract-deployer -- --ignored test_fetch_commissions_sepolia
+    #[ignore]
+    #[test_log::test(tokio::test)]
+    async fn test_fetch_commissions_sepolia() -> Result<()> {
+        let rpc_url = std::env::var("RPC_URL")
+            .expect("RPC_URL environment variable not set")
+            .parse()?;
+        let provider = ProviderBuilder::new().on_http(rpc_url);
+
+        let stake_table_address: Address = "0x40304FbE94D5E7D1492Dd90c53a2D63E8506a037".parse()?;
```
imabdulbasit, line 1785: please add a comment that this is decaf contract address

sequencer/src/api.rs
```diff
@@ -5684,6 +5706,130 @@ mod test {
...
+            .unwrap();
+        let target_epoch = current_epoch.u64() + 3;
+        let mut events = network.peers[0].event_stream().await;
+        wait_for_epochs(&mut events, EPOCH_HEIGHT, target_epoch).await;
+
+        // the last epoch with the old commissions
+        let client: Client<ServerError, SequencerApiVersion> =
+            Client::new(format!("http://localhost:{api_port}").parse().unwrap());
+        let validators = client
+            .get::<ValidatorMap>(&format!("node/validators/{}", target_epoch - 1))
```
imabdulbasit, line 5796: Shouldn’t this quer target_epoch - 2? Since a contract update only takes two epochs to take effect, there’s a possibility of overlap

sequencer/src/api.rs
```diff
@@ -5684,6 +5706,130 @@ mod test {
...
+            .await
+            .expect("validators");
+        assert!(!validators.is_empty());
+        for (val, _, new_comm) in commissions.clone() {
+            assert_eq!(validators.get(&val).unwrap().commission, new_comm.to_evm());
+        }
+
+        // check validator 0 received zero rewards in the last epoch where its
+        // commission was zero
+        for block in EPOCH_HEIGHT * (target_epoch - 1) + 1..EPOCH_HEIGHT * target_epoch {
```
imabdulbasit: can we also just assert that other two validators received some rewards?

staking-cli/src/deploy.rs
```diff
@@ -65,11 +65,25 @@ impl TestSystem {
...
+        let mut attempts = 0;
+        let (port, provider) = loop {
+            let port = portpicker::pick_unused_port().unwrap();
+            match ProviderBuilder::new().on_anvil_with_wallet_and_config(|anvil| {
+                anvil.port(port).arg("--accounts").arg("20")
+            }) {
+                Ok(provider) => break (port, provider),
+                Err(e) => {
+                    attempts += 1;
+                    if attempts >= 5 {
```
imabdulbasit, line 81: should there be some delay before retrying like 1 s sleep?

staking-cli/src/registration.rs
```diff
@@ -210,6 +243,44 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_update_commission() -> Result<()> {
```
imabdulbasit, line 247: I think we should also test that a commission update fails if it doesn’t respect the update interval

types/src/v0/impls/stake_table.rs
```diff
@@ -503,6 +511,27 @@ impl StakeTableState {
                 validator.stake_table_key = stake_table_key;
                 validator.state_ver_key = state_ver_key;
             },
+
+            StakeTableEvent::CommissionUpdate(CommissionUpdated {
```
imabdulbasit, line 516: should add comments that we don't expect commission update events for protocol version V3 etc

types/src/v0/impls/stake_table.rs
```diff
@@ -1034,7 +1066,7 @@ impl Fetcher {
             }
         });
         // fetch consensus keys updated v2 events
-        let keys_update_events_v2 = stream::iter(chunks).then(|(from, to)| {
+        let keys_update_events_v2 = stream::iter(chunks.clone()).then(|(from, to)| {
```
imabdulbasit, line 1072: why does it require `clone` now?

types/src/v0/impls/stake_table.rs
```diff
@@ -1500,9 +1564,10 @@ impl EpochCommittees {
     /// We used a fixed block reward for version v3
     /// Version v4 uses the dynamic block reward based
     /// Assumes the stake table contract proxy address does not change
-    /// In the future, if we want to support updates to the stake table contract address via chain config,
-    /// or allow the contract to handle additional block reward calculation parameters (e.g., inflation, block time),
-    /// the `fetch_block_reward` logic can be updated to support per-epoch rewards.
+    /// In the future, if we want to support updates to the stake table contract address
+    /// via chain config, or allow the contract to handle additional block reward calculation
+    /// parameters (e.g., inflation, block time), the `fetch_block_reward` logic can be
```
imabdulbasit: this comment is outdated now.

staking-cli/src/deploy.rs
sveitser: I don't really know what the underlying issue is so I'm not sure waiting has any positive effect. I haven't seen things fail after making this change.

staking-cli/src/registration.rs
sveitser: This is tested in the solidity tests but I'll add it here as well.

types/src/v0/impls/stake_table.rs
sveitser: because we use it again later, but you're right the last use doesn't require clone

sequencer/src/api.rs
sveitser: ah, forgot this

sveitser: aec6dae25fdf2c812d657c05b598a3ff5ba2addb

staking-cli/src/parse.rs
```diff
@@ -126,7 +130,11 @@ mod test {
             ("2", 200),
             ("1.000000", 100),
             ("1.2", 120),
+            ("1.2", 120),
```
imabdulbasit: looks the same as above row ?

sveitser: https://github.com/EspressoSystems/espresso-network/pull/3591/commits/d69c930183497029a2bec7dbc13a4ae2fc9092db

sequencer/src/api.rs
sveitser: @imabdulbasit you mean in case the commission changes happen between the epoch root block and the final block in an epoch?

imabdulbasit: Let’s say we are in epoch 4. We updated the commission in epoch 4. The target epoch = 4 + 3 = 7. Epoch 6 will still have the old commission because it takes 2 full epochs?

imabdulbasit: so I think it should be fine actually

PR #3592, opened 2025-09-08
CI: fix docker build on tag and schedule

sveitser:
Every Monday I think our build broke but it's due to how we don't handle the docker tags correctly in the CI smoke test for scheduled builds and builds on tags.

One problem is that sometimes the docker metadata action will produce multiple tags, so I think it's easier to just re-implement how the tag is derived.

It's described here how the tags are derived by the metadata action: https://github.com/docker/metadata-action?tab=readme-ov-file#basic


sveitser:
Manual run: https://github.com/EspressoSystems/espresso-network/actions/runs/17544162173

alxiong (approved):

PR #3593, opened 2025-09-08
Update alloy version from 0.13 to 1.0

imabdulbasit:
This PR updates the Alloy and Foundry versions. The latest Foundry version adds serde derives to the bindings. The older version did not have that, so we implemented a custom version of `Serialize` and `Deserialize`, which unfortunately differs from what serde generates using the derives. This PR migrates the stake table events in our storage to use the new serialization so that we can rely on the default implementation. This is important because our custom implementation used the tuple form, making it easy to miss field or the ordering.

---
  - https://app.asana.com/0/0/1211291700109579


Cargo.toml
```diff
@@ -89,7 +89,7 @@ edition = "2021"
 [workspace.dependencies]
 # The --alloy-version in the justfile gen-bindings recipe should match the version here.
 # TODO: to update past 0.13 we need https://github.com/foundry-rs/foundry/issues/10321
-alloy = { version = "0.13", default-features = false, features = [
+alloy = { version = "1.0.30", default-features = false, features = [
```
mrain, line 91: nit: remove the line above

mrain:
I'm suggesting to merge this PR with #3601 otherwise it'll be a pain to resolve the build w/o Cargo.lock issue.
WDYT @sveitser @imabdulbasit @alxiong

contracts/rust/deployer/src/lib.rs
```diff
@@ -2184,25 +2152,22 @@ mod tests {
...
             total_supply,
             parse_ether(&initial_supply.to_string()).unwrap()
         );
-        assert_eq!(
-            token.balanceOf(init_recipient).call().await?._0,
-            total_supply,
-        );
-        assert_eq!(token.name().call().await?._0, name);
-        assert_eq!(token.symbol().call().await?._0, symbol);
+        assert_eq!(token.balanceOf(init_recipient).call().await?, total_supply,);
```
```suggestion
        assert_eq!(token.balanceOf(init_recipient).call().await?, total_supply);
```

hotshot-state-prover/src/v3/mock_ledger.rs
```diff
@@ -450,7 +450,9 @@ impl MockLedger {
 
     // return a dummy auth root
     pub fn new_dummy_auth_root(&mut self) -> FixedBytes<32> {
-        FixedBytes::random_with(&mut self.rng)
+        let mut bytes = [0; 32];
```
sveitser, line 453: @imabdulbasit according to the docs `random_with` should still be there, so wonder why this was changed.

sequencer/src/persistence/sql.rs
```diff
@@ -1966,6 +1966,123 @@ impl SequencerPersistence for Persistence {
...
+            // Insert batch into new table
+            let mut query_builder: sqlx::QueryBuilder<Db> = sqlx::QueryBuilder::new(
+                "INSERT INTO stake_table_events (l1_block, log_index, event) ",
+            );
+
+            query_builder.push_values(values, |mut b, (l1_block, log_index, event)| {
+                b.push_bind(l1_block).push_bind(log_index).push_bind(event);
+            });
+            query_builder.push(
+                " ON CONFLICT (l1_block, log_index)
```
sveitser, line 2043: Why do we need this? Or what does this do exactly?

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -373,38 +419,20 @@ impl<'de> Deserialize<'de> for ValidatorRegisteredV2 {
     }
 }
 
-impl Serialize for CommissionUpdated {
```
sveitser, line 376: We're assuming there are no `CommissionUpdated` events in storage right? It's a bit risky because it's currently not known when this change will be deployed. So we do need to make sure we don't deploy `main` from before this change after we have upgraded the stake table contract and hope that no other validator does that either. Notably if a validator does not update in time I think we would corrupt their storage by losing the `CommissionUpdated` events.

sequencer/Cargo.toml
```diff
@@ -58,7 +58,7 @@ derive_more = { workspace = true }
 dotenvy = { workspace = true }
 either = { workspace = true }
 espresso-contract-deployer = { path = "../contracts/rust/deployer" }
-espresso-types = { path = "../types" }
+espresso-types = { path = "../types", features = ["testing"] }
```
sveitser: Why do we need this? The testing feature is enabled for the test build already. It would be best to not enable it for the non-test build.

justfile
```diff
@@ -157,7 +157,7 @@ gen-bindings:
 
     # Generate the alloy bindings
     # TODO: `forge bind --alloy ...` fails if there's an unliked library so we pass pass it an address for the PlonkVerifier contract.
-    forge bind --skip test --skip script --use "0.8.28" --alloy --alloy-version "0.13.0" --contracts ./contracts/src/ \
+    forge bind --skip test --skip script --use "0.8.28" --alloy --alloy-version "1.0.30" --contracts ./contracts/src/ \
```
sveitser: Do we still need to specify the alloy version here? Does it have any effect on the bindings?

sveitser:
@imabdulbasit I looked through all the changes. Looks good. Left some questions.

contracts/rust/adapter/src/sol_types.rs
sveitser: I don't think this is likely to happen but it's still some extra risk and another thing to keep track of.

hotshot-state-prover/src/v3/mock_ledger.rs
mrain: I guess it's about unmatched `serde` version. Encountered this before

imabdulbasit: the `StdRng` we have is coming from some older version `rand` crate in jellyfish. FixedBytes is imported from alloy which expects the newer version `StdRng` here.

sequencer/src/persistence/sql.rs
imabdulbasit: we already have events in the stake table events table. The ON CONFLICT will overwrite those events with the new serialization

contracts/rust/adapter/src/sol_types.rs
imabdulbasit: yeah I don't think we are deploying the commission changes on decaf, so it should be fine IMO

justfile
imabdulbasit: No, I checked and we don't need it

contracts/rust/adapter/src/sol_types.rs
alxiong: I think the `alloy v1.0` will derive serde by default, we don't need those manual derivation?

imabdulbasit:
> @mrain said he merged this into his branch earlier today. Shall we centralized our discussion there and close this PR? what's the plan @imabdulbasit ?

I pushed some small changes so if @mrain can merge this PR again, I can close it

mrain:
I'll merge it again later.

contracts/rust/adapter/src/sol_types.rs
sveitser: We need it for migrating existing storage.

PR #3594, opened 2025-09-08
Version gate quorum proposal

ss-es:


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -1326,12 +1326,24 @@ pub async fn validate_qc_and_next_epoch_qc<TYPES: NodeType, V: Versions>(
 }
 
 /// Validates the light client state update certificate
-pub async fn validate_light_client_state_update_certificate<TYPES: NodeType>(
+pub async fn validate_light_client_state_update_certificate<TYPES: NodeType, V: Versions>(
     state_cert: &LightClientStateUpdateCertificateV2<TYPES>,
     membership_coordinator: &EpochMembershipCoordinator<TYPES>,
+    view_number: TYPES::View,
```
mrain: I think we already have it inside the `state_cert`. The light client state has an associated view number

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -1356,17 +1368,30 @@ pub async fn validate_light_client_state_update_certificate<TYPES: NodeType>(
...
-            ) || !<TYPES::StateSignatureKey as LCV2StateSignatureKey>::verify_state_sig(
-                key,
-                sig_v2,
-                &state_cert.light_client_state,
-                &state_cert.next_stake_table_state,
-            ) {
-                bail!("Invalid light client state update certificate signature");
+            #[allow(clippy::collapsible_else_if)]
+            // If the auth_root is `Default`, the state_cert was cast from an older version of the struct. We only perform the second signature check.
+            if state_cert.auth_root == <FixedBytes<32> as Default>::default() {
```
mrain: I think it's safer to replace this condition with `!upgrade_lock.proposal2_version(view_number).await`

crates/hotshot/task-impls/src/helpers.rs
ss-es: updated

crates/hotshot/task-impls/src/helpers.rs
ss-es: I think this has to match whatever we do when accumulating votes, because we need to similarly gate that

crates/hotshot/types/src/vote.rs
```diff
@@ -291,11 +291,14 @@ impl<TYPES: NodeType> LightClientStateUpdateVoteAccumulator<TYPES> {
             error!("Invalid light client state update vote {vote:?}");
             return None;
         }
-        if !<TYPES::StateSignatureKey as LCV3StateSignatureKey>::verify_state_sig(
-            &state_ver_key,
-            &vote.signature,
-            vote.signed_state_digest,
-        ) {
+        // only verify the new state signature if the auth_root is not default, indicating a vote that was cast from an older vote
+        if vote.auth_root != <FixedBytes<32> as Default>::default()
```
mrain: we should add `auth_root` to the keys of `vote_outcomes`. Otherwise malicious replica will do sth bad here.

crates/hotshot/task-impls/src/helpers.rs
ss-es: updated to use the upgrade lock rather than the auth_root everywhere

PR #3595, opened 2025-09-09
recalculate stake table hash if missing in add_epoch_root()

imabdulbasit:


ss-es (approved):
looks good to me as a quick fix

PR #3596, opened 2025-09-10
View sync byzantine tests

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Adds two byzantine tests:
* [view_sync_next_epoch](https://www.notion.so/espressosys/View-sync-to-new-epoch-21c2431b68e980d9b163e6d743bfc74b)
* [view_sync_old_epoch](https://www.notion.so/espressosys/View-sync-to-old-epoch-21b2431b68e980f3b7fac3e996c11564)

### Key places to review:

---
  - https://app.asana.com/0/0/1210134443040116

### DIFF
--- a/crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
+++ b/crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
@@ -32,7 +32,7 @@ use hotshot_types::{
     },
     traits::{
         election::Membership,
-        network::{ConnectedNetwork, ViewMessage},
+        network::ConnectedNetwork,
         node_implementation::{ConsensusTime, NodeImplementation, NodeType, Versions},
     },
     vote::HasViewNumber,
@@ -723,8 +723,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> EventTransforme
         private_key: &<TYPES::SignatureKey as SignatureKey>::PrivateKey,
         upgrade_lock: &UpgradeLock<TYPES, V>,
         _consensus: OuterConsensus<TYPES>,
-        membership_coordinator: EpochMembershipCoordinator<TYPES>,
-        network: Arc<I::Network>,
+        _membership_coordinator: EpochMembershipCoordinator<TYPES>,
+        _network: Arc<I::Network>,
     ) -> Vec<HotShotEvent<TYPES>> {
         match event {
             HotShotEvent::QuorumProposalSend(proposal, _) => {


crates/hotshot/testing/tests/tests_3/byzantine_tests.rs
```diff
@@ -281,3 +283,81 @@ cross_tests!(
         metadata
     },
 );
+
```
ss-es, line 286: can you add a comment describing what these tests are meant to do?

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -701,3 +705,131 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> EventTransforme
...
+                if self.first_dishonest_view_number > vote.view_number().u64() {
+                    return vec![event.clone()];
+                }
+                let view_number = vote.data.round;
+                let message_kind = if upgrade_lock.epochs_enabled(view_number).await {
+                    let Ok(vote) = ViewSyncPreCommitVote2::<TYPES>::create_signed_vote(
+                        ViewSyncPreCommitData2 {
+                            relay: 0,
+                            round: view_number,
+                            epoch: vote.data.epoch.map(self.epoch_modifier),
```
ss-es, line 758: is this basically just creating a vote that view syncs into the "wrong" epoch? (depending on the configuration)

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -701,3 +705,131 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> EventTransforme
...
+                        },
+                        view_number,
+                        public_key,
+                        private_key,
+                        upgrade_lock,
+                    )
+                    .await
+                    else {
+                        tracing::error!("Failed to sign pre commit vote!");
```
ss-es: this could probably just be a panic right?

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -701,3 +705,131 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> EventTransforme
...
+                    .await;
+                match transmit_result {
+                    Ok(()) => tracing::info!(
+                        "Sent ViewSyncPreCommitVoteSend for view {} and epoch {:?} to the leader",
+                        view_number,
+                        message.kind.epoch(),
+                    ),
+                    Err(e) => panic!("Failed to send message task: {e:?}"),
+                }
+                vec![]
```
ss-es: I'm not sure about this part, maybe I forgot how the byzantine stuff is supposed to work. shouldn't we just return the `ViewSyncPreCommitVoteSend` event? the normal network task should then serialize/transmit this as normal

ss-es (commented):
overall looks good, but some comments/questions before approving

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
lukaszrzasik: yes, correct.

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
lukaszrzasik: You're right. I used the same pattern here that I used in a different test but there I actually had to do it like that because instead of broadcasting a message I had to send it to one particular node.

crates/hotshot/testing/src/byzantine/byzantine_behaviour.rs
lukaszrzasik: yeah, it can, I'll change it.

crates/hotshot/testing/tests/tests_3/byzantine_tests.rs
```diff
@@ -281,3 +283,83 @@ cross_tests!(
...
+            overall_safety_properties: OverallSafetyPropertiesDescription {
+                num_successful_views: 30,
+                ..OverallSafetyPropertiesDescription::default()
+            },
+            view_sync_properties: ViewSyncTaskDescription::Threshold(0, 13),
+            behaviour,
+            ..TestDescription::default()
+        }.set_num_nodes(10, 10);
+
+        metadata.overall_safety_properties.possible_view_failures = (0..100).collect();
```
ss-es, line 321: so every view might fail but we want 30 successful views? this seems a little suspicious but there's nothing strictly wrong with it

lukaszrzasik: yeah, the idea is that the byzantine nodes can cause some views to fail, there's no way around it but they're not able to stall the network progression.

PR #3601, opened 2025-09-15
chore: update jellyfish & arkworks deps

mrain:
Closes #<ISSUE_NUMBER>

### This PR:

- Upgrade arkworks dependencies to 0.5
- Upgrade alloy dependency to 1.0
- Upgrade jellyfish dependency that comes with a circuit fix

### Key places to review:

WARN: the light client contract V1/V2 in this PR are different from the ones deployed on Mainnet/Decaf!


.cargo/audit.toml
```diff
@@ -13,6 +13,6 @@ ignore = [
     # To upgrade async-tungstenite we need a new release of tide-websockets, which seems unlikely
     # https://github.com/http-rs/tide-websockets
     "RUSTSEC-2025-0009",
-    # Requires jf to upgrade ark from 0.4 to 0.5, so we can use tracing-subscriber 0.3 exclusively
+    # Still here because of `ark-r1cs-std` crate.
```
alxiong: curious, can we take the same dummy crate approach as: https://github.com/EspressoSystems/jellyfish/commit/3ce504ec2b75b23ba41124521b3baaeb4050c03e

Cargo.toml
```diff
@@ -88,8 +88,7 @@ edition = "2021"
 
 [workspace.dependencies]
 # The --alloy-version in the justfile gen-bindings recipe should match the version here.
-# TODO: to update past 0.13 we need https://github.com/foundry-rs/foundry/issues/10321
-alloy = { version = "0.13", default-features = false, features = [
+alloy = { version = "1.0", default-features = false, features = [
```
alxiong: i'd simplify to `alloy = { version = "1.0", features = ["provider-ws", "json-rpc", "signer-mnemonic", "node-bindings", "getrandom"] }` 

alloy 1.0 has very clear [feature list](https://docs.rs/crate/alloy/latest/features)

also we can remove the comment above, since bindings will be the same, and we didn't pin to `--alloy-version` anymore.

Cargo.toml
```diff
@@ -198,32 +196,33 @@ hotshot-utils = { path = "crates/hotshot/utils" }
 vid = { path = "vid", features = ["parallel", "keccak256"] }
 
 bytesize = "1.3"
-committable = "0.2"
+# committable = "0.2"
```
alxiong, line 177:
```suggestion
```

utils/src/lib.rs
```diff
@@ -84,12 +85,13 @@ macro_rules! impl_to_fixed_bytes {
 /// - `wait_for_transaction_to_be_mined` is removed thanks to alloy's better builtin PendingTransaction await
 /// - DON'T use this if you want parse the exact revert reason/type, since this func will only give err msg like: "custom error 0x23b0db14",
 ///   instead, follow <https://docs.rs/alloy/0.12.5/alloy/contract/enum.Error.html#method.as_decoded_interface_error> to pattern-match err type
-pub async fn contract_send<T, P, C>(
-    call: &SolCallBuilder<T, P, C>,
+pub async fn contract_send<P, C, N>(
+    call: &SolCallBuilder<P, C, N>,
```
alxiong: we can simply drop this generic parameter `N` iiuc. because it has default to `N=Ethereum`, which is always what we need.

Cargo.toml
alxiong: and `deployer/Cargo.toml` we should use: `alloy = { workspace = true, features = ["signer-ledger"]}` since we don't use Ledger signer elsewhere.

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -16,42 +16,46 @@
 
 use alloy::sol;
 
+use crate::bindings::stake_table_v2::{
+    EdOnBN254,
+    BN254::{self, BaseField},
+};
```
alxiong, line 22: This addition is weird. revisit later. 
(EdOnBN254 should be imported already somewhere iirc, so are other types, plus importing them from `stake_table_v2` also seems strange)

contracts/rust/adapter/src/sol_types.rs
```diff
@@ -312,289 +313,375 @@ impl PartialEq for ConsensusKeysUpdatedV2 {
...
+```*/
+#[allow(
+    non_camel_case_types,
+    non_snake_case,
+    clippy::pub_underscore_fields,
+    clippy::style
+)]
+#[derive(Clone, Serialize, Deserialize)]
+pub struct ValidatorRegisteredLegacy {
+    #[allow(missing_docs)]
```
alxiong, line 329: rather than manually replicating these, can we use [`sol!` to generate](https://docs.rs/alloy/latest/alloy/sol_types/macro.sol.html) instead? 

we can define like:

```rust
sol! {
  struct ValidatorRegsiteredLegacy {}
}
```

all the serde should also be auto-derived, if not then use `#[sol(all_derives)]` on the struct

hotshot-state-prover/src/v1/snark.rs
```diff
@@ -159,18 +158,12 @@ mod tests {
...
-        // TODO: parallelization
-        let g_table = FixedBase::get_window_table(scalar_bits, window_size, g);
-        let powers_of_g = FixedBase::msm::<<Bn254 as Pairing>::G1>(
-            scalar_bits,
-            window_size,
-            &g_table,
-            &powers_of_beta,
-        );
+        let powers_of_g = g
+            .batch_mul(&powers_of_beta)
```
alxiong, line 163: btw, recently i learned that `batch_mul()` internally does two things: 1) build a lookup table, 2) does batched mul
https://github.com/EspressoSystems/timeboost/pull/456#issuecomment-3209054951

so it makes little sense to use this api directly if we only batch once. because each time will incur a "build preprocessing table" high cost. 

a simple parallelization of `.mul()` is actually faster.
(again, feel free to fix this later)

contracts/src/libraries/LightClientStateUpdateVK.sol
alxiong, line 1: I understand that we want to update V1 and V2 verifying key, the slightly scary thing is our test is always targeting the latest LightClient only, so contract tests are not covering it anymore)

are you confident that unit test in `hotshot-state-prover/v1` is using the old circuit (but ark-upgraded and circuit-fixed proving) against this VK? 
(if so, then my worry is addressed)

@mrain

hotshot-state-prover/src/v3/mock_ledger.rs
```diff
@@ -450,7 +450,9 @@ impl MockLedger {
 
     // return a dummy auth root
     pub fn new_dummy_auth_root(&mut self) -> FixedBytes<32> {
-        FixedBytes::random_with(&mut self.rng)
+        let mut bytes = [0; 32];
+        self.rng.fill_bytes(&mut bytes);
+        bytes.into()
```
alxiong, line 455: this is because you drop `getrandom` feature on alloy? should add back.

alxiong (commented):
👍 🚀 ❤️ amazing!

Thanks you Chengyu and Abdul!! I'm very grateful for this amount of tedious chore being done by both you!

Salute!

Cargo.toml
mrain: @jbearer @Ancient123 could we get this committable@0.2.5 to crates.io?

.cargo/audit.toml
mrain: how do you handle the deps from ark-srs
```
tracing-subscriber 0.2.25
└── ark-relations 0.5.1
    ├── ark-snark 0.5.1
    │   └── ark-crypto-primitives 0.5.0
    │       └── ark-poly-commit 0.5.0
    │           └── ark-srs 0.3.3
```

Cargo.toml
mrain: I'm getting the following error with ur suggestion:
```
error[E0308]: mismatched types
   --> hotshot-query-service/src/data_source/storage/sql/queries/node.rs:358:13
    |
358 |             num_transactions,
    |             ^^^^^^^^^^^^^^^^ expected `HashMap<Option<...>, usize, ...>`, found `HashMap<Option<_>, usize>`
    |
    = note: `std::hash::RandomState` and `RandomState` have similar names, but are actually distinct types
note: `std::hash::RandomState` is defined in crate `std`
   --> /Users/chengyu/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/hash/random.rs:36:1
    |
36  | pub struct RandomState {
    | ^^^^^^^^^^^^^^^^^^^^^^
note: `RandomState` is defined in crate `foldhash`
   --> /Users/chengyu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/foldhash-0.1.5/src/fast.rs:153:1
    |
153 | pub struct RandomState {
    | ^^^^^^^^^^^^^^^^^^^^^^
    = note: the full name for the type has been written to '/Users/chengyu/espressosys/espresso-network/target/debug/deps/hotshot_query_service-5f33ea8368db6ebb.long-type-5495656606199224291.txt'
    = note: consider using `--verbose` to print the full type name to the console


```

contracts/rust/adapter/src/sol_types.rs
mrain: dunno how to apply this, directly copy your code above doesn't work

contracts/src/libraries/LightClientStateUpdateVK.sol
mrain: yes we didn't touch the circuit impl here. So the differences are only ark upgrade and circuit fix.

hotshot-state-prover/src/v3/mock_ledger.rs
mrain: no, it's because of the `rand` version mismatch.
https://github.com/EspressoSystems/espresso-network/pull/3593#discussion_r2352418557

utils/src/lib.rs
mrain: removed in [baf9b3a](https://github.com/EspressoSystems/espresso-network/pull/3601/commits/baf9b3af6a6e3fe2c7109867b6da382a34dab90a)

contracts/rust/adapter/src/sol_types.rs
sveitser: Would these derives be compatible with our manual derives? The reason for keeping the manual derives is because we need to migrate existing storage to the new auto derived serialization format.

alxiong: > Would these derives be compatible with our manual derives?

i'm not sure, but don't see why it wasn't. our manual serialization wasn't unconventional.

Cargo.toml
alxiong: Screenshot 2025-09-17 at 10 27 08 PM


I think this solves it

mrain: ty! [0c7a4ac](https://github.com/EspressoSystems/espresso-network/pull/3601/commits/0c7a4ac876a6dd857197a228565bd7c68a409eb1)

.cargo/audit.toml
mrain: I patched ark-relations@0.5.1 here: [c251415](https://github.com/EspressoSystems/espresso-network/pull/3601/commits/c2514153a2d2c952b40a82689d1c0812b282a9c8)
solved

mrain: nvm reverted. causing compilation issue

sveitser:
I didn't get to this today but we should quickly check how the contracts bytecode changed. We didn't change the solc version. So either it's the bytecode hash (dont remember if we use that from the top of my head) or something else changed via foundry defaults changes, like optimizer runs. I will do it later today or tomorrow

sveitser:
> WARN: the light client contract V1/V2 in this PR are different from the ones deployed on Mainnet/Decaf!

@alxiong @mrain 

So I believe this is mainly due to the bn254 update.

alxiong:
> 
>    WARN: the light client contract V1/V2 in this PR are different from the ones deployed on Mainnet/Decaf!
> 
> @alxiong @mrain
>
> So I believe this is mainly due to the bn254 update.

no, what we meant is we know the circuit is different because we fix the circuit bug and upgraded to ark5.0, so the verification key would be diff. It's not due to BN254 update. 

>  they don't work if we upgrade the dependency also for the older version of the contract in the repo. Because the tests the only compare what's in the repo. We currently don't run the fork tests on CI to check against deployed contracts either I think,

I was worrying about this, https://github.com/EspressoSystems/espresso-network/pull/3601#discussion_r2355044410
but our state-prover unit tests verify that the new VK lib will work. 

> using soldeer as a package manager we could more easily depend on multiple versions

yes, we should adopt that.


---

on the topic of V1 and V2 being different from what's being deployed:

is it okay to assume that we will be upgrading all LC to V3 all on testnet and mainnet soon, so it's fine if we change them in this repo?  as long as we know that we can upgrade existing deployments to V3. 
@sveitser @mrain

contracts/rust/adapter/src/sol_types.rs
sveitser: I think @imabdulbasit mentioned they are not the same, hence the migration of storage.

sveitser:
> no, what we meant is we know the circuit is different because we fix the circuit bug and upgraded to ark5.0, so the verification key would be diff. It's not due to BN254 update.

I see. I understood it as the LC bytecode changed (which is independent of the plonk verifier library, since that library is external). And it did change and I think it's due to the bn254 update. In any event, I think it's fine.

> is it okay to assume that we will be upgrading all LC to V3 all on testnet and mainnet soon, so it's fine if we change them in this repo? as long as we know that we can upgrade existing deployments to V3.

I think it's fine and yes, the plan is to upgrade everything soon.

sveitser:
@alxiong are there remaining comments from you that need to be addressed?

contracts/rust/adapter/src/sol_types.rs
imabdulbasit: image

alxiong (approved):

contracts/rust/adapter/src/sol_types.rs
imabdulbasit: we were serializing the struct in tuple form.
example: 
manual serialization 
`Object {"Delegate": Array [String("0x9f6e61d9307d924c7c8af4c2ef4e98d5f3eeb8d1"), String("0x71a0bd27625cab15c1f7a04b28c97bb800a84e27"), String("0x15af1d78b58c400000")]}`

serde derive:
`Object {"Delegate": Object {"amount": String("0x15af1d78b58c400000"), "delegator": String("0x9f6e61d9307d924c7c8af4c2ef4e98d5f3eeb8d1"), "validator": String("0x71a0bd27625cab15c1f7a04b28c97bb800a84e27")}}`

PR #3604, opened 2025-09-15
Implement transaction streaming in Go SDK

QuentinI:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/sdks/go/client/client.go
+++ b/sdks/go/client/client.go
@@ -178,6 +178,7 @@ func (c *Client) SubmitTransaction(ctx context.Context, tx types.Transaction) (*
 	return &hash, nil
 }
 
+// Stream of JSON-encoded objects over a WebSocket connection
 type WsStream[S any] struct {
 	conn *websocket.Conn
 }
@@ -212,6 +213,7 @@ func (s *WsStream[S]) Close() error {
 	return s.conn.Close(websocket.StatusNormalClosure, "")
 }
 
+// Open a `Stream` of Espresso transactions starting from a specific block height.
 func (c *Client) StreamTransactions(ctx context.Context, height uint64) (Stream[types.TransactionQueryData], error) {
 	opts := &websocket.DialOptions{}
 	opts.HTTPClient = c.client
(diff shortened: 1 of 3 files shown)


sdks/go/client/client.go
```diff
@@ -177,6 +181,51 @@ func (c *Client) SubmitTransaction(ctx context.Context, tx types.Transaction) (*
...
+		return nil, fmt.Errorf("%w: %v", ErrPermanent, err)
+	}
+	return &data, nil
+}
+
+func (s *WsStream[S]) Close() error {
+	return s.conn.Close(websocket.StatusNormalClosure, "")
+}
+
+func (c *Client) StreamTransactions(ctx context.Context, height uint64) (Stream[types.TransactionQueryData], error) {
```
Ayiga, line 217: Should we also support the transaction stream with the [filtered namespace](https://github.com/EspressoSystems/espresso-network/blob/c9d0aa28463329a3c2249d0b08534fb680a45488/hotshot-query-service/src/availability.rs#L641), for convenience?

sdks/go/client/stream.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+var _ Stream[any] = (*StreamWithTimeout[any])(nil)
+
+type StreamWithTimeout[T any] struct {
```
Ayiga: For the `StreamWithTimeout`, I see that we're utilize the `timeout` parameter for calls to `Close` and `NextRaw`.  The only real use I see for this is the `timeout` call for the `Close`.

If we're wanting the User to be able to control when these timeout, then the user *should* already have the ability to do this with the `ctx` parameter.  Additionally, utilizing context in this way would give the consuming  user more control over what he/she wants to timeout or not.

QuentinI: > If we're wanting the User to be able to control when these timeout, then the user should already have the ability to do this with the ctx parameter

Yeah, that's fair. I'm still unused to Go and straight up missed this
I'd say we can outright remove the StreamWithTimeout in this case.

Ayiga (approved):
Could use some comments on public methods and structs, but looks good overall

PR #3606, opened 2025-09-16
Add endpoint to fetch all the validators

imabdulbasit:
This PR adds persistence methods to store and load all validators, including inactive ones, and adds a new `all-validators `API endpoint to fetch them with pagination.

---
  - https://app.asana.com/0/0/1211332042406112

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -2231,6 +2231,7 @@ mod test {
     };
     use jf_merkle_tree_compat::prelude::{MerkleProof, Sha3Node};
     use portpicker::pick_unused_port;
+    use pretty_assertions::assert_matches;
     use rand::seq::SliceRandom;
     use rstest::rstest;
     use staking_cli::{
@@ -2242,7 +2243,7 @@ mod test {
         catchup_test_helper, state_signature_test_helper, status_test_helper, submit_test_helper,
         TestNetwork, TestNetworkConfigBuilder,
     };
-    use tide_disco::{app::AppHealth, error::ServerError, healthcheck::HealthStatus};
+    use tide_disco::{app::AppHealth, error::ServerError, healthcheck::HealthStatus, StatusCode};
     use tokio::time::sleep;
     use vbs::version::{StaticVersion, StaticVersionType};
 
@@ -2763,7 +2764,8 @@ mod test {
     async fn test_chain_config_from_instance() {
         // This test uses a ValidatedState which only has the default chain config commitment.
         // The NodeState has the full chain config.
-        // Both chain config commitments will match, so the ValidatedState should have the full chain config after a non-genesis block is decided.
+        // Both chain config commitments will match, so the ValidatedState should have the
+        // full chain config after a non-genesis block is decided.
 
         let port = pick_unused_port().expect("No ports free");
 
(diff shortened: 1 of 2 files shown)


sequencer/src/api.rs
```diff
@@ -219,6 +219,15 @@ impl<N: ConnectedNetwork<PubKey>, D: Sync, V: Versions, P: SequencerPersistence>
...
         self.as_ref().previous_proposal_participation().await
     }
+
+    async fn get_all_validators(
+        &self,
+        epoch: <SeqTypes as NodeType>::Epoch,
+        offset: u64,
+        limit: u64,
+    ) -> anyhow::Result<Vec<Validator<PubKey>>> {
+        self.as_ref().get_all_validators(epoch, offset, limit).await
```
Ayiga, line 232: Do we want a potential cap on `limit`, so the user isn't able to request arbitrarily many entries at once?

sequencer/src/api/endpoints.rs
```diff
@@ -625,6 +625,33 @@ where
...
+            let epoch = req.integer_param::<_, u64>("epoch_number").map_err(|_| {
+                hotshot_query_service::node::Error::Custom {
+                    message: "Epoch number is required".to_string(),
+                    status: StatusCode::BAD_REQUEST,
+                }
+            })?;
+
+            let offset = req.integer_param::<_, u64>("offset")?;
+
+            let limit = req.integer_param::<_, u64>("limit")?;
```
Ayiga, line 735: I suppose a cap on `limit` should **really** be realized here if anywhere.

sequencer/src/api/endpoints.rs
```diff
@@ -625,6 +625,33 @@ where
...
+            state
+                .read(|state| {
+                    state
+                        .get_all_validators(EpochNumber::new(epoch), offset, limit)
+                        .boxed()
+                })
+                .await
+                .map_err(|err| hotshot_query_service::node::Error::Custom {
+                    message: format!("failed to get all validators : err: {err}"),
+                    status: StatusCode::NOT_FOUND,
```
Ayiga: Surely any error encountered here isn't due to things not being found, but rather an `SQL` error, a database error, a connection error, or the tables themselves not existing.

In all of these cases, I would think this is actually a `server` error.

types/src/v0/impls/instance_state.rs
```diff
@@ -121,6 +121,23 @@ impl MembershipPersistence for NoStorage {
     )> {
         bail!("unimplemented")
     }
+
+    async fn store_all_validators(
+        &self,
+        _epoch: EpochNumber,
+        _all_validators: ValidatorMap,
+    ) -> anyhow::Result<()> {
+        Ok(())
```
Ayiga, line 130: Everything else is `bail!("unimplemented")` but not this?

sveitser: @Ayiga I see storing is a no-op on `NoStorage` and loading is an error. Seems reasonable to me.

sequencer/src/api/endpoints.rs
sveitser: 814acc09e8fe1619a37c65b1591c4c1eb015e1fd

sequencer/src/api/endpoints.rs
sveitser: 814acc09e8fe1619a37c65b1591c4c1eb015e1fd

sveitser:
@Ayiga I fixed conflicts, reverted IMO accidental submodule downgrade, fixed 2 of your comments and answered the remaining comment. Could you take another look and approve if you're happy with it?

Ayiga (approved):
LGTM

PR #3610, opened 2025-09-17
add IRewardClaim interface

sveitser:
This is taken from PR

https://github.com/EspressoSystems/espresso-network/pull/3491

Would like to review and merge it into main first before we finish the contract so that it's easy to share with external parties.


contracts/src/interfaces/IRewardClaim.sol
```diff
@@ -0,0 +1,38 @@
...
+pragma solidity ^0.8.28;
+
+interface IRewardClaim {
+    /// @notice Proof for rewards merkle tree. Obtained from Espresso query service API.
+    struct LifetimeRewardsProof {
+        bytes32[] siblings;
+    }
+
+    /// @notice User claimed rewards
+    event RewardClaimed(address indexed user, uint256 amount);
```
sveitser:
```suggestion
    event RewardsClaimed(address indexed user, uint256 amount);
```

Probably a bit more consistent to use the plural here too.

sveitser: https://github.com/EspressoSystems/espresso-network/pull/3610/commits/39ee5deb0a474f183b0d1678eea9af6ff9e10243

contracts/src/interfaces/IRewardClaim.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @param lifetimeRewards Total earned lifetime rewards for the user
+    /// @param proof Merkle proof attesting to lifetime rewards for the user
+    /// @param authRootInputs The authRootInputs must all be zero at the moment,
+    ///        this may change in the future with Espresso protocol upgrades.
+    ///
+    /// @notice Obtain rewards and proof from the Espresso query service API.
+    function claimRewards(
+        uint256 lifetimeRewards,
+        LifetimeRewardsProof calldata proof,
+        bytes32[7] calldata authRootInputs
```
alxiong, line 33: I wonder if want to hardcode the arbitrarily chosen 7 placeholder in the interface. I do agree that implementation would require upgrade if we expand to more auth roots, but I think we should minimize interface changes. 

To wit, I'm proposing we change this to more general bytes

sveitser: @alxiong If we added something to the auth root the claim application would only to change the values of these other `authRootInputs` but we don't expect to change the number of inputs in the foreseeable future. We have already "future prooved" by adding 7 empty fields.

If we want to make it easy for the claim application we could for example return these other auth root inputs from the query service API when we fetch a reward proof. Then the UIs don't need to change anything if we use more auth root fields.

cc @imabdulbasit

imabdulbasit: we don't track the auth_root inputs currently, but should be doable. We can store them in storage and return through an endpoint

sveitser: I mean for the time being we can just return the array with 7 zeros. It's just nice for the UI if we get this flow correct then if we add nonzero value later the UI doesn't need to care about this.

alxiong: >but we don't expect to change the number of inputs in the foreseeable future

okay. sounds good. I was being more conservative, suggesting that `bytes` is the longest-term future-proof data type. but if 7 fields is good enough for long enough, then yes. let's keep it.

alxiong (approved):

PR #3616, opened 2025-09-19
rust 1.89 -> 1.90

sveitser:
fix clippy

- use `is_multiple_of`
- remove some dead code
- update flake.lock, leave that to https://github.com/EspressoSystems/espresso-network/pull/3601

### DIFF
--- a/crates/hotshot/testing/tests/tests_3/memory_network.rs
+++ b/crates/hotshot/testing/tests/tests_3/memory_network.rs
@@ -11,16 +11,13 @@ use hotshot::{
     traits::{
         election::static_committee::StaticCommittee,
         implementations::{MasterMap, MemoryNetwork},
-        NodeImplementation,
     },
     types::SignatureKey,
 };
 use hotshot_example_types::{
-    
     block_types::{TestBlockHeader, TestBlockPayload, TestTransaction},
     node_types::TestVersions,
     state_types::{TestInstanceState, TestValidatedState},
-    storage_types::TestStorage,
 };
 use hotshot_types::{
     data::{EpochNumber, ViewNumber},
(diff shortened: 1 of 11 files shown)


crates/hotshot/testing/tests/tests_3/memory_network.rs
```diff
@@ -229,7 +219,8 @@ async fn memory_network_broadcast_queue() {
             .recv_message()
             .await
             .expect("Failed to receive message");
-        let (deserialized_message, _version) = upgrade_lock.deserialize(&recv_message).await.unwrap();
+        let (deserialized_message, _version) =
```
sveitser, line 222: not sure why these are formatted differently now, probably some fix in rustfmt

vid/src/avid_m/config.rs
```diff
@@ -29,52 +25,6 @@ pub trait AvidMConfig {
     fn raw_share_digest(raw_shares: &[Self::BaseField]) -> VidResult<Self::Digest>;
 }
```
sveitser, line 27: @mrain ping to check if deleting this code is fine or we should keep it. It seems to be unused.

vid/src/avid_m/config.rs
```diff
@@ -29,52 +25,6 @@ pub trait AvidMConfig {
...
 
-/// Configuration of Poseidon2 based AVID-M scheme
```
sveitser, line 33: @mrain ok to delete this? Seems it's unused.

flake.nix
```diff
@@ -111,6 +103,10 @@
         })
       ];
       pkgs = import nixpkgs { inherit system overlays; };
+      myShell = pkgs.mkShellNoCC.override (pkgs.lib.optionalAttrs pkgs.stdenv.isLinux {
```
sveitser, line 106: This uses the `NoCC` shell also on darwin, not just on Linux.

PR #3621, opened 2025-09-23
Return all reward claim inputs from reward API

sveitser:
Note, this PR is into [ab/reward-proof-endpoint](https://github.com/EspressoSystems/espresso-network/tree/ab/reward-proof-endpoint), fixup for PR:
- https://github.com/EspressoSystems/espresso-network/pull/3611

### DIFF
--- a/contracts/rust/adapter/src/reward.rs
+++ b/contracts/rust/adapter/src/reward.rs
@@ -2,8 +2,8 @@ use alloy::{
     primitives::{Bytes, B256, U256},
     sol_types::SolValue,
 };
-use derive_more::From;
-use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+use derive_more::{From, Into};
+use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Eq, PartialEq, From)]
 pub struct RewardProofSiblings([B256; 160]);
@@ -24,40 +24,43 @@ impl RewardAuthData {
             auth_root_inputs,
         }
     }
+}
 
-    pub fn from_bytes(bytes: impl AsRef<[u8]>) -> alloy::sol_types::Result<Self> {
-        let (siblings, auth_root_inputs): ([B256; 160], [B256; 7]) =
-            SolValue::abi_decode(bytes.as_ref())?;
-        Ok(RewardAuthData {
+impl From<([B256; 160], [B256; 7])> for RewardAuthData {
+    fn from((siblings, auth_root_inputs): ([B256; 160], [B256; 7])) -> Self {
+        RewardAuthData {
             siblings: siblings.into(),
             auth_root_inputs: auth_root_inputs.into(),
-        })
-    }
-
-    pub fn to_bytes(&self) -> Bytes {
-        (self.siblings.0, self.auth_root_inputs.0)
-            .abi_encode()
-            .into()
+        }
     }
 }
 
-impl Serialize for RewardAuthData {
-    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        alloy::hex::serde::serialize(self.to_bytes(), serializer)
+impl TryFrom<RewardAuthDataEncoded> for RewardAuthData {
+    type Error = alloy::sol_types::Error;
+
+    fn try_from(value: RewardAuthDataEncoded) -> Result<Self, Self::Error> {
+        let decoded: ([B256; 160], [B256; 7]) = SolValue::abi_decode(&value.0)?;
+        Ok(decoded.into())
     }
 }
 
-impl<'de> Deserialize<'de> for RewardAuthData {
-    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        let bytes: Bytes = alloy::hex::serde::deserialize(deserializer)?;
-        RewardAuthData::from_bytes(bytes).map_err(D::Error::custom)
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, From, Into)]
+pub struct RewardAuthDataEncoded(Bytes);
+
+impl From<RewardAuthData> for RewardAuthDataEncoded {
+    fn from(value: RewardAuthData) -> Self {
+        Self(
+            (value.siblings.0, value.auth_root_inputs.0)
+                .abi_encode()
+                .into(),
+        )
     }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct RewardClaimInput {
     pub lifetime_rewards: U256,
-    pub auth_data: RewardAuthData,
+    pub auth_data: RewardAuthDataEncoded,
 }
 
 #[cfg(test)]
@@ -71,7 +74,7 @@ mod tests {
         let auth_root_inputs = RewardAuthRootInputs([B256::random(); 7]);
         RewardClaimInput {
             lifetime_rewards: B256::random().into(),
-            auth_data: RewardAuthData::new(siblings, auth_root_inputs),
+            auth_data: RewardAuthData::new(siblings, auth_root_inputs).into(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


data/insta_snapshots/espresso_types__reference_tests__reward_claim_input_v2.snap
```diff
@@ -0,0 +1,175 @@
+---
```
sveitser, line 1: @Ayiga please check this file to see if the format is nice for the UI. I think it should be. The API will of course return this as Json.

sequencer/src/api.rs
```diff
@@ -5910,7 +5912,7 @@ mod test {
             .build();
 
         let api_port = pick_unused_port().expect("No ports free for query service");
-        tracing::info!("API PORT = {api_port}");
+        println!("API PORT = {api_port}");
```
alysiahuggins, line 5915: curious about the switch to println here

contracts/src/libraries/RewardMerkleTreeVerifier.sol
```diff
@@ -17,10 +19,6 @@ library RewardMerkleTreeVerifier {
 
     uint256 public constant TREE_DEPTH = 160;
 
-    struct AccruedRewardsProof {
-        bytes32[] siblings;
-    }
-
     function _hashLeaf(uint256 value) internal pure returns (bytes32) {
         bytes32 firstHash = keccak256(abi.encodePacked(value));
```
alysiahuggins, line 23: we're good on using encodePacked here?

sveitser: You mean instead of assembly? Basically we encode the raw 32 bytes in rust so in order to match we need to use encodePacked (or use assembly, to do the same).

sveitser: Since it is only called once (as opposed to 160 times for the internal nodes) I didn't bother with assembly optimization, happy to change it though.

sequencer/src/api.rs
sveitser: I think for tests it's slightly preferrable to use `println` so that doing something like `env RUST_LOG=error cargo test ...` only shows minimal logs plus the test output without having to write a complex filtering expression for rust log.

data/insta_snapshots/espresso_types__reference_tests__reward_claim_input_v2.snap
sveitser: @Ayiga I updated the serialization again. I think this is the best we can do and I think the values can be passed directly to wagmi:

```ts
import { writeContract } from 'wagmi';

const inputs = // call espresso query service "reward-state-v2/reward-claim-input/{height}/{address}"

await writeContract({
  abi: contractAbi,
  address: stakeTableContractAddress,
  functionName: 'claimRewards',
  args: [inputs.lifetimeRewards, inputs.authData],
});
```

Ayiga: Alright, so this means that we're also specifying up to the the `lifetimeRewards` that we've claimed to this point (as a running total)?  Also, this *should* be the `rewardStateContractAddress` and not the `stakeTableContactAddress`, right?  Or are we just going to be adding the `RewardClaim` interface to the `StakeTable`?

For instance if I have `10` units of reward to claim, and I wanted to claim `5` of them initially I could call:

```typescript
await writeContract({
  abi: contractABI,
  address: rewardClaimContractAddress,
  functionName: 'claimRewards',
  args: ['0x5', inputs.authData],
});
```

But then if I wanted to claim the other `5` from that point, it would require me calling with `5` added to what I've previously `claimed`.  Is that correct?

Example:
```typescript
await writeContract({
  abi: contractABI,
  address: rewardClaimContractAddress,
  functionName: 'claimRewards',
  args: ['0xa', inputs.authData],
});
```

contracts/rust/adapter/src/reward.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    pub fn to_bytes(&self) -> Bytes {
+        (self.siblings.0, self.auth_root_inputs.0)
+            .abi_encode()
+            .into()
+    }
+}
+
+impl Serialize for RewardAuthData {
```
alxiong: can't we auto derive this? constituting fields should also be able to auto-derive serde.

contracts/src/interfaces/IRewardClaim.sol
```diff
@@ -21,17 +16,11 @@ interface IRewardClaim {
...
+    /// @param authData inputs required for authentication of lifetime rewards amount.
     ///
-    /// @notice Obtain rewards and proof from the Espresso query service API.
-    function claimRewards(
-        uint256 lifetimeRewards,
-        LifetimeRewardsProof calldata proof,
-        bytes32[7] calldata authRootInputs
-    ) external;
+    /// @notice Obtain authData from the Espresso query service API.
+    function claimRewards(uint256 lifetimeRewards, bytes calldata authData) external;
```
alxiong, line 23: wait, i thought we said might as well use `memory` since calldata-to-memory copy is inevitable.

justfile
```diff
@@ -163,6 +163,9 @@ gen-bindings:
       --libraries contracts/src/libraries/PlonkVerifierV2.sol:PlonkVerifierV2:0xffffffffffffffffffffffffffffffffffffffff \
       --libraries contracts/src/libraries/PlonkVerifierV3.sol:PlonkVerifierV3:0xffffffffffffffffffffffffffffffffffffffff
 
+    # HACK: add serde support for fixed byte arrays in the generated bindings
+    sed -i '/pub proof: \[alloy::sol_types::private::FixedBytes<32>; 160usize\],/i \        #[serde(with = "serde_arrays")]' contracts/rust/adapter/src/bindings/*.rs
```
alxiong, line 167: without this, serde won't be there? this feels like an upstream bug.

contracts/rust/adapter/src/reward.rs
```diff
@@ -0,0 +1,102 @@
...
+    sol_types::SolValue,
+};
+use derive_more::From;
+use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Clone, Debug, Eq, PartialEq, From)]
+pub struct RewardProofSiblings([B256; 160]);
+
+#[derive(Clone, Debug, Eq, PartialEq, From, Default)]
+pub struct RewardAuthRootInputs([B256; 7]);
```
alxiong, line 12: it's kinda of unclear how to construct this, do we want to add methods like `fn from_esp_reward_tree()-> Self` ?

also add doc about what are these 7 inputs.

alxiong (commented):
Looks good to me!

data/insta_snapshots/espresso_types__reference_tests__reward_claim_input_v2.snap
sveitser: Yup sorry, wrong contract. I copied this from the explorer code to see how we send transactions.

A claim will claim everything that's available. The total lifetime rewards must be used, otherwise the value won't authenticate.

The reward claim contract will mint the total lifetime rewards minus what has previously been claimed.

contracts/src/interfaces/IRewardClaim.sol
sveitser: memory to memory copy is worse than calldata to memory

contracts/rust/adapter/src/reward.rs
sveitser: I don't know if/how to derive it to get abi encoded hex bytes.

justfile
sveitser: It's a limitation of serde. See https://github.com/foundry-rs/foundry/issues/10911

contracts/rust/adapter/src/reward.rs
sveitser: Yeah we don't have or need the any construction of non-zero values yet.

contracts/src/interfaces/IRewardClaim.sol
sveitser: Also I think this part of the interface is not so important here. We will of course extensively test the gas usage for the real reward claim contract. The UIs that consume the interface actually don't need to worry about if this is calldata or memory. It would only matter for other contracts calling this contract.

contracts/rust/adapter/src/reward.rs
alxiong: via [`SolType` trait](https://docs.rs/alloy/latest/alloy/dyn_abi/trait.SolType.html), and use `.abi_encode().encode_hex()` where `alloy::hex::ToHexExt` will bring `.encode_hex()` to the scope. 

Can you try this? 
(if it's too troublesome, we can also just keep this serde)

alxiong (approved):

PR #3622, opened 2025-09-24
Fix order of `add_epoch_root` and `add_drb_result`

ss-es:
Fixes an issue where `add_drb_result` could sometimes be called before `add_epoch_root` due to executor timing, especially with a low DRB difficulty.

Found thanks to #3490, but I'd like to merge this fix separately.


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -287,7 +287,9 @@ async fn decide_epoch_root<TYPES: NodeType, I: NodeImplementation<TYPES>>(
 
         tokio::spawn(async move {
```
ss-es, line 288: Note that the entire block is already in a `tokio::spawn`, and the minor advantage from starting the DRB calculation in parallel with adding the epoch root is definitely not worth the risk of `add_drb_result` happening out of order

PR #3623, opened 2025-09-25
Remove workspace-hack.

twittner:
https://github.com/EspressoSystems/HotShot/pull/4066 added a "workspace hack" to always compile with all features. Unfortunately this inflates the number of dependencies for clients a lot.

A crate whose sole dependency is `hotshot-types = { git = "https://github.com/EspressoSystems/espresso-network" }` for example compiles a total of 1094 crates, whereas on this PR branch only 518 crates are compiled. This PR proposes to remove `workspace-hack`.

In addition, to ensure workspace crates do not implicitly depend on feature unification, a justfile recipe is added that invokes `cargo check` on each individual crate and is executed as part of CI.


crates/hotshot/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -283,7 +282,7 @@ impl<K: SignatureKey + 'static> PushCdnNetwork<K> {
 }
 
 #[cfg(feature = "hotshot-testing")]
-impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
+impl<TYPES: crate::NodeType> TestableNetworkingImplementation<TYPES>
```
ss-es: can we revert the changes in this file? stylistically I think this is inconsistent with the rest of the crates, we never write `crate::NodeType` anywhere

twittner: Will do. The import needs to be feature-gated though to avoid a warning.

twittner: Done in https://github.com/EspressoSystems/espresso-network/pull/3623/commits/a18eebc15471d9d51a0d8db2e759be72fc274c0c.

.github/workflows/lint.yml
```diff
@@ -57,3 +57,12 @@ jobs:
         - name: Run clippy
           run: |
             cargo clippy --workspace --features "embedded-db testing" --all-targets --keep-going -- -D warnings
+
+  check-individually:
```
sveitser: I think the `check-features` recipe already includes this check. Not at computer now but can check the CI logs later to make sure.

It's used in this CI job:

https://github.com/EspressoSystems/espresso-network/blob/main/.github%2Fworkflows%2Fcargo-features.yml

twittner: That's nice. I did not see this and `cargo-hack` is certainly much more exhaustive.

twittner: I have removed the justfile recipe and the CI job.

sveitser (approved):
LGTM. Thanks!

sveitser:
Ok. It's failing on `main` too now.

sveitser:
Might have found it

```
3446:[demo-l1-network	] demo-l1-network-1  | ERROR[09-29|09:51:58.501] Low disk space. Gracefully shutting down Geth to prevent database corruption. available=365.08MiB path=/db/geth
```

of course totally unrelated to this PR

PR #3624, opened 2025-09-26
added totalStake and totalValidatorStake along with unit tests

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
* adds total stake tracking to StakeTableV2 with two new fields: `totalStake` and `totalValidatorStake`
* updates all delegation/withdrawal functions to maintain accurate stake totals
* adds unit test coverage for the new tracking variables


contracts/lib/bn254
sveitser, line 1: This looks like an accidental submodule downgrade. Probably missing a `git submodule update` after pulling main.

contracts/src/StakeTableV2.sol
```diff
@@ -84,6 +89,12 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @notice Maximum commission increase allowed per increase (in basis points)
     uint16 public maxCommissionIncrease;
 
+    /// @notice Total stake in active (not marked for exit) validators in the contract
+    uint256 public totalValidatorStake;
```
sveitser: I'm not sure about the naming because I couldn't tell just from the names what the variables track. 

I feel like it would be best to have pendingWithdrawals, pendingExits, activeStake and totalTracked but this would be 4 storage edits for each delegation which is quite expensive.

alysiahuggins: the `delegate` gas cost is up 46% so i'm looking at other gas saving measures

sveitser: Did you make sure this isn't the first delegation? The first one will create 4 new storage entries, subsequent ones only edit the 4 entries.

sveitser: But it also sounds about right/expected. A delegation doesn't do much other than moving the token (changing balance of delegator and stake table) and increasing the balance of the delegator for the validator in the stake table.

Overall it feels not ideal to have a user pay so much for something that isn't useful for them. However at current gas prices it's probably fine.

alysiahuggins: that's fair at first delegation, in this case, i used the average fees to compare. 
If we only store totalValidatorStake which i'm thinking to rename to `totalEffectiveStake` or `totalConsensusStake`, the (average) gas increase for a delegation is 25%.

I've made the change to only track `totalValidatorStake` here, if you agree, i'll merge it into this pr https://github.com/EspressoSystems/espresso-network/pull/3630

sveitser:
@alysiahuggins this needs to be synced with main

contracts/src/StakeTableV2.sol
alysiahuggins: pr https://github.com/EspressoSystems/espresso-network/pull/3630 has been merged into this branch and we are only tracking "activeStake"

contracts/src/StakeTableV2.sol
```diff
@@ -263,7 +276,23 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @param validator The validator to withdraw from
     /// @dev This function is overridden to add pausable functionality
     function claimWithdrawal(address validator) public virtual override whenNotPaused {
-        super.claimWithdrawal(validator);
+        address delegator = msg.sender;
```
sveitser: @alysiahuggins I don't think there's now any change to this function.

contracts/rust/deployer/src/lib.rs
```diff
@@ -959,6 +959,60 @@ pub async fn fetch_commissions_for_stake_table_storage_migration(
...
+        start_block
+    );
+
+    // Query Delegated events (V1) to get initial active stake values
+    let delegated_events = stake_table
+        .Delegated_filter()
+        .from_block(start_block)
+        .query()
+        .await
+        .context("Failed to query Delegated events from V1 contract")?;
```
sveitser: Revert if there are any undelegated events or validator exits?

contracts/rust/deployer/src/lib.rs
```diff
@@ -1741,6 +1813,128 @@ mod tests {
...
+        let token = EspToken::new(token_addr, &provider);
+
+        for (delegator_addr, amount) in test_delegators.iter() {
+            let _receipt = token
+                .transfer(*delegator_addr, *amount)
+                .from(Address::from(*owner))
+                .send()
+                .await?
+                .get_receipt()
+                .await?;
```
sveitser, line 1851: nit: why not also check status here?

contracts/rust/deployer/src/lib.rs
```diff
@@ -1741,6 +1813,128 @@ mod tests {
...
+        assert_eq!(
+            fetched_active_stake,
+            test_delegators
+                .iter()
+                .map(|(_, amount)| *amount)
+                .sum::<U256>(),
+        );
+
+        assert!(
+            fetched_active_stake <= token.balanceOf(stake_table_proxy_addr).call().await?,
```
sveitser: Equality check instead? We did not undelegate or exit so the amounts should match exactly.

contracts/test/StakeTableV2PropTestBase.sol
```diff
@@ -557,6 +567,10 @@ contract StakeTableV2PropTestBase is FunctionCallTracking {
         return testState;
     }
 
+    function getactiveStake() external view returns (uint256) {
```
sveitser:
```suggestion
    function getActiveStake() external view returns (uint256) {
```

contracts/test/StakeTableV2PropTestBase.sol
```diff
@@ -417,6 +422,7 @@ contract StakeTableV2PropTestBase is FunctionCallTracking {
 
     function trackDelegate(address actorAddr, address val, uint256 amount) internal {
         testState.totalDelegated += amount;
+        testState.activeStake += amount; // Add this
```
sveitser:
```suggestion
        testState.activeStake += amount;
```

contracts/test/StakeTableV2PropTestBase.sol
```diff
@@ -757,4 +775,12 @@ contract StakeTableV2PropTestBase is FunctionCallTracking {
             );
         }
     }
+
+    /// @dev Assert that contract's token balance and activeStake match our tracking
```
sveitser: Not sure about this comment here. It's comparing the contracts internal tracking to the test's tracking.

sveitser (approved):
LGTM. @alysiahuggins I cleaned up some stuff that came from my PR into this one and resolved conflicts with main.

PR #3626, opened 2025-09-26
Fix header reward check

ss-es:
Fixes the version check so that it also works for a 0.2 -> 0.4 upgrade


ss-es:
It is probably correct as implemented -- there are no previous rewards if the last block header version is 0.2

PR #3630, opened 2025-09-29
only track totalValidatorStake

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
remove tracking of totalStake and just tracks totalValidatorStake 

### This PR does not:
(TODO, remove totalStake from tests etc)


contracts/src/StakeTableV2.sol
```diff
@@ -92,9 +92,6 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @notice Total stake in active (not marked for exit) validators in the contract
     uint256 public totalValidatorStake;
```
sveitser: Should we just call it `activeStake` then?

alysiahuggins: agreed https://github.com/EspressoSystems/espresso-network/pull/3630/commits/970a4c68fc2433c24937ce1e6eaaeefdcba724ca

PR #3634, opened 2025-10-01
load da_committee from genesis file

pls148:
### This PR:

* Adds support for genesis -> da_committees
* Parsing of da_committees
* EpochCommittees now holds da_committees as data instead of known_da_nodes

### This PR does not:

* Actually do anything with the new data

### Key places to review:

Example da_committees in data/genesis/X.toml: 

```
[da_committees]
0.3 = {
    0 = [
        { stake_table_key = "BLS_VER_KEY~bQszS-QKYvUij2g20VqS8asttGSb95NrTu2PUj0uMh1CBUxNy1FqyPDjZqB29M7ZbjWqj79QkEOWkpga84AmDYUeTuWmy-0P1AdKHD3ehc-dKvei78BDj5USwXPJiDUlCxvYs_9rWYhagaq-5_LXENr78xel17spftNd5MA1Mw5U", state_ver_key = "SCHNORR_VER_KEY~lJqDaVZyM0hWP2Br52IX5FeE-dCAIC-dPX7bL5-qUx-vjbunwe-ENOeZxj6FuOyvDCFzoGeP7yZ0fM995qF-CRE", stake = 1 }
    ]
}
```

### DIFF
--- a/sequencer/src/genesis.rs
+++ b/sequencer/src/genesis.rs
@@ -9,7 +9,7 @@ use espresso_types::{
     v0_3::ChainConfig, FeeAccount, FeeAmount, GenesisHeader, L1BlockInfo, L1Client, SeqTypes,
     Timestamp, Upgrade,
 };
-use hotshot_types::{traits::node_implementation::NodeType, version_ser, VersionedDaCommittee};
+use hotshot_types::{version_ser, VersionedDaCommittee};
 use serde::{Deserialize, Serialize};
 use vbs::version::Version;
 


crates/hotshot/types/src/lib.rs
```diff
@@ -266,4 +270,26 @@ impl<TYPES: NodeType> HotShotConfig<TYPES> {
...
+    pub fn build_da_committees(&self) -> BTreeMap<TYPES::Epoch, Vec<PeerConfig<TYPES>>> {
+        if self.da_committees.is_empty() {
+            tracing::warn!(
+                "da_committees is not set, falling back to known_da_nodes, which is deprecated."
+            );
+
+            [(TYPES::Epoch::new(0), self.known_da_nodes.clone())].into()
+        } else {
+            if !self.known_da_nodes.is_empty() {
+                tracing::warn!(
```
ss-es: I think this could be an error

or arguably even a panic -- I think it might be preferable to call this a misconfiguration and immediately panic vs. try to resolve the ambiguity

types/src/v0/impls/stake_table.rs
```diff
@@ -1689,11 +1689,18 @@ impl EpochCommittees {
...
         fixed_block_reward: Option<RewardAmount>,
         fetcher: Fetcher,
         epoch_height: u64,
     ) -> Self {
+        // TODO: Store da_committees in EpochCommittees. Currently we just pull da_members back out, but we
+        // will need to figure out how to integrate da_committees into EpochCommittee and the like.
+        let da_members = da_committees
+            .get(&Epoch::new(0))
+            .cloned()
+            .unwrap_or_default(); // For testing reasons, we want to support being given an empty map
```
ss-es: `or_default` means return an empty vec here right? I think it would be better to panic, since an empty DA committee is always a (fatal) error

sequencer/src/genesis.rs
```diff
@@ -67,6 +110,8 @@ pub struct Genesis {
     #[serde(rename = "upgrade", with = "upgrade_ser")]
     #[serde(default)]
     pub upgrades: BTreeMap<Version, Upgrade>,
+    #[serde(default)]
+    pub da_committees: Option<BTreeMap<TomlKeyU64, Vec<PeerConfigData>>>,
```
ss-es: A `BTreeMap` can be null right? do we need to make this an `Option`?

ss-es: also is there a reason it needs to be in both `HotShotConfig` and `Genesis`?

do we use it out of `HotShotConfig` anywhere? (I think it's fine to leave it in both even if we only use it out of genesis, just wondering)

types/src/v0/config.rs
```diff
@@ -55,6 +56,8 @@ pub struct PublicHotShotConfig {
     num_nodes_with_stake: NonZeroUsize,
     known_nodes_with_stake: Vec<PeerConfig<SeqTypes>>,
     known_da_nodes: Vec<PeerConfig<SeqTypes>>,
+    #[serde(default)]
+    da_committees: BTreeMap<Version, BTreeMap<u64, Vec<PeerConfig<SeqTypes>>>>,
```
ss-es: I think the inner `BTreeMap` is unnecessary -- this should just be a `BTreeMap<Version, (u64, PeerConfig<SeqTypes>)>`

maybe @bfish713 can comment, but I think it's unlikely that we'd schedule an upgrade with multiple DA committees that take effect at different points in the future? and it definitely complicates the logic

sequencer/src/genesis.rs
```diff
@@ -44,6 +45,21 @@ pub enum L1Finalized {
...
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct VersionedDaCommittee {
+    #[serde(with = "version_ser")]
+    pub start_version: Version,
+    pub start_epoch: u64,
+    pub committee: Vec<PeerConfigData>,
+}
+
```
ss-es: I feel like this should be imported from `hotshot-types` instead of re-defined

sequencer/src/lib.rs
```diff
@@ -390,6 +391,28 @@ where
...
+                        stake_table_entry: StakeTableEntry {
+                            stake_key: pcd.stake_table_key,
+                            stake_amount: U256::from(pcd.stake),
+                        },
+                        state_ver_key: pcd.state_ver_key.clone(),
+                    })
+                    .collect(),
+            })
+            .collect();
+    }
```
ss-es, line 396: I'm not sure I see the reason for any of this -- can't we just use the same type? can't this entire block just be `network_config.config.da_committees = da_committees`

crates/hotshot/types/src/lib.rs
```diff
@@ -266,4 +277,11 @@ impl<TYPES: NodeType> HotShotConfig<TYPES> {
     pub fn hotshot_stake_table(&self) -> HSStakeTable<TYPES> {
         self.known_nodes_with_stake.clone().into()
     }
+
+    pub fn build_da_committees(&self) -> Vec<PeerConfig<TYPES>> {
+        // TODO: THIS IS A TEMPORARY FIX WITH THE WRONG RETURN TYPE.
+        // It's done so that we can start using this function and have the existing behavior
+        // (use known_da_nodes) while we transition to using da_committees.
+        self.known_da_nodes.clone()
+    }
```
ss-es, line 313: I don't think this function is necessary. We can keep the base `da_stake_table` (or whatever `Membership` currently stores) separate from the `epoch_da_committees` that apply at specific epochs, as we do for the full quorum stake table anyway

sequencer/src/lib.rs
```diff
@@ -529,7 +552,7 @@ where
     // Create the HotShot membership
     let mut membership = EpochCommittees::new_stake(
         network_config.config.known_nodes_with_stake.clone(),
-        network_config.config.known_da_nodes.clone(),
+        network_config.config.build_da_committees(),
```
ss-es: I think this can (should) remain unchanged

sequencer/src/genesis.rs
```diff
@@ -44,6 +45,13 @@ pub enum L1Finalized {
     Timestamp { timestamp: Timestamp },
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct PeerConfigData {
+    pub stake_table_key: <SeqTypes as NodeType>::SignatureKey,
+    pub state_ver_key: <SeqTypes as NodeType>::StateSignatureKey,
+    pub stake: u64,
+}
+
```
ss-es: is this struct now unused? if so we should delete it

PR #3635, opened 2025-10-02
Build Blocks on Proposal Received if it's Before View Change

bfish713:
### This PR:

This changes the transactions task to start getting a block as soon as we get the prior proposal.  Ususally we won't update the view until we vote on the prior proposal, or form a QC.  This could be 500ms or more after receiving the proposal so it's just wasted time.   We have all the information we need to get a block from the builder as soon as the prior proposal is received

### DIFF
--- a/crates/hotshot/testing/tests/test_da_committees.rs
+++ b/crates/hotshot/testing/tests/test_da_committees.rs
@@ -37,7 +37,7 @@ cross_tests!(
             ..TestDescription::default()
         };
 
-        let node_configs = vec![
+        let node_configs = [
             ValidatorConfig::generated_from_seed_indexed(
                 [0u8; 32],
                 0,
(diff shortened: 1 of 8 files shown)


crates/hotshot/task-impls/src/transactions.rs
```diff
@@ -366,7 +360,50 @@ impl<TYPES: NodeType, V: Versions> TransactionTaskState<TYPES, V> {
...
+                }
+
+                let vid = proposal.data.block_header().payload_commitment();
+                let block_height = proposal.data.block_header().block_number();
+                if is_epoch_transition(block_height, self.epoch_height) {
+                    return Ok(());
+                }
+                if is_last_block(block_height, self.epoch_height) {
+                    return Ok(());
+                }
```
lukaszrzasik: nit: I think this is unnecessary as `is_last_block` is already covered by `is_epoch_transition`.

PR #3636, opened 2025-10-02
Add second QC to decide event

jbearer:
This is necessary to convince an untrusting listener (such as the new light client) that the leaf chain in the event is in fact decided

https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510079577349?focus=true


crates/hotshot/testing/src/consistency_task.rs
```diff
@@ -399,6 +406,23 @@ impl<TYPES: NodeType<BlockHeader = TestBlockHeader>, V: Versions> TestTaskState
...
+                },
+                None => {
+                    // Only the genesis decide is special in that it doesn't have a 2-chain of QCs
+                    // extending the last decided leaf.
+                    ensure!(
+                        view_number == TYPES::View::genesis(),
+                        "expected 2nd deciding QC for non-genesis decide"
+                    );
+                },
+            }
```
jbearer: Is there a better place in the test framework to hook in this new consistency check?

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -339,8 +339,8 @@ pub struct LeafChainTraversalOutcome<TYPES: NodeType> {
     /// The new decided view obtained from a 3 chain starting from the proposal's parent.
     pub new_decided_view_number: Option<TYPES::View>,
 
-    /// The qc for the decided chain.
-    pub new_decide_qc: Option<QuorumCertificate2<TYPES>>,
+    /// The 2-chain of qcs causing the chain to be decided.
+    pub new_decide_qcs: Option<[QuorumCertificate2<TYPES>; 2]>,
```
ss-es: nit: could we make this a tuple or a struct? or just separate fields for the different qc's

crates/hotshot/testing/src/consistency_task.rs
ss-es: I think this is fine -- it's a bit weird because the consistency task really validates the aggregated leaf chain, but I don't see the harm in doing this

crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -563,6 +566,7 @@ pub async fn decide_from_proposal<TYPES: NodeType, I: NodeImplementation<TYPES>>
 
     let mut last_view_number_visited = view_number;
     let mut current_chain_length = 0usize;
+    let mut prev_qc = proposal.justify_qc().clone();
```
bfish713: This function is using the old decide rule from original Hotshot so it requires and additional QC.  It might be best to leave this as is, or increase the number of QCs in the Decide event

crates/hotshot/types/src/event.rs
```diff
@@ -202,6 +202,11 @@ pub enum EventType<TYPES: NodeType> {
         /// Note that the QC for each additional leaf in the chain can be obtained from the leaf
         /// before it using
         qc: Arc<QuorumCertificate2<TYPES>>,
```
ss-es: I wonder if we could rename this e.g. `committing_qc` and `finalizing_qc` for clarity (here and in `LeafChainTraversalOutcome`)

jbearer: I renamed `qc2` to `deciding_qc` but kept `qc` for backwards compatibility. Do you think it's alright to change that one too?

jbearer: Done in [d8f1f7a](https://github.com/EspressoSystems/espresso-network/pull/3636/commits/d8f1f7aedc3658754a7e1507bbac564eb7105cdb)

PR #3638, opened 2025-10-03
Call `add_da_committee` on upgrade decide

ss-es:
Adds a new `add_da_committee` method to `Membership`, and calls it when an upgrade is decided, and on consensus startup.

The method will be implemented in a separate PR.


crates/hotshot/hotshot/src/lib.rs
```diff
@@ -282,13 +282,29 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
...
+        };
+
         debug!("Setting DRB difficulty selector in membership");
         let drb_difficulty_selector = drb_difficulty_selector(upgrade_lock.clone(), &config);
 
         membership_coordinator
             .set_drb_difficulty_selector(drb_difficulty_selector)
             .await;
 
+        for da_committee in &config.da_committees {
```
pls148, line 298: should we also add logic to insert the known_da_nodes for cases where we aren't yet at a version that supports dynamic committees / there is no da_committee with version at our current version?

ss-es: I don't think so -- the logic in the sequencer's `Membership` implementation should select `known_da_nodes` when we have nothing in `da_committees`

pls148: ah i missed the unwrap_or there, looks good then

PR #3639, opened 2025-10-03
feat: concurrent txns, better output

sveitser:
- allow broadcasting txns without awaiting
- output events
- check for events in tests
- add test to inspect output manually
- output to stdout/stderr consistently
- always use tracing if RUST_LOG_FORMAT=json

### DIFF
--- a/staking-cli/src/demo.rs
+++ b/staking-cli/src/demo.rs
@@ -7,12 +7,13 @@ use alloy::{
         Address, U256,
     },
     providers::{Provider, ProviderBuilder, WalletProvider},
-    rpc::types::TransactionRequest,
+    rpc::types::{TransactionReceipt, TransactionRequest},
     signers::local::PrivateKeySigner,
 };
 use anyhow::Result;
 use clap::ValueEnum;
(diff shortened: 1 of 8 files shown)


staking-cli/src/demo.rs
```diff
@@ -1,36 +1,72 @@
...
+pub enum CreateTransactionsError {
+    #[error(
+        "insufficient ESP balance: have {have} ESP, need {need} ESP to fund {delegators} \
+         delegators"
+    )]
+    InsufficientEsp {
+        have: String,
+        need: String,
+        delegators: usize,
+    },
```
rob-maron, line 50: Love these

staking-cli/tests/cli.rs
```diff
@@ -129,7 +129,8 @@ async fn test_cli_register_validator(
                 .arg("--commission")
                 .arg("12.34")
                 .assert()
-                .success();
+                .success()
+                .stdout(str::contains("ValidatorRegistered"));
```
rob-maron, line 133: Nice, I like these tests

PR #3647, opened 2025-10-07
only upgrade to v3 if atleast v2

alysiahuggins:
Closes #<ISSUE_NUMBER>
light cilent upgrade to v3 checks first that the light client contract is atleast v2

### This PR:
- updates rust contract deployment code for light client upgrade to v3 
-

### DIFF
--- a/contracts/rust/deployer/src/lib.rs
+++ b/contracts/rust/deployer/src/lib.rs
@@ -615,6 +615,15 @@ pub async fn upgrade_light_client_v3(
         Some(proxy_addr) => {
             let proxy = LightClient::new(proxy_addr, &provider);
 
+            //check proxy version, if not atleast V2, can't upgrade to V3
+            let version = proxy.getVersion().call().await?;
+            if version.majorVersion < 2 {
+                anyhow::bail!(
+                    "LightClientProxy is V{}, can't upgrade to V3",
+                    version.majorVersion
+                );
+            }
+
             // first deploy PlonkVerifierV3.sol
             let pv3_addr = contracts
                 .deploy(
@@ -672,7 +681,15 @@ pub async fn upgrade_light_client_v3(
 
             // prepare init calldata
             let lcv3 = LightClientV3::new(lcv3_addr, &provider);
-            let init_data = lcv3.initializeV3().calldata().to_owned();
+
+            // set init data unless already initialized
+            let init_data =
+                if already_initialized(&provider, proxy_addr, Contract::LightClientV3, 3).await? {
+                    "0x".into()
+                } else {
+                    lcv3.initializeV3().calldata().to_owned()
+                };
+
             // invoke upgrade on proxy
             let receipt = proxy
                 .upgradeToAndCall(lcv3_addr, init_data)
--- a/contracts/rust/adapter/src/bindings/erc1967_proxy.rs
+++ b/contracts/rust/adapter/src/bindings/erc1967_proxy.rs
@@ -116,12 +116,11 @@ pub mod ERC1967Proxy {
     pub static DEPLOYED_BYTECODE: alloy_sol_types::private::Bytes = alloy_sol_types::private::Bytes::from_static(
         b"`\x80`@R`\n`\x0CV[\0[`\x18`\x14`\x1AV[`PV[V[_`K\x7F6\x08\x94\xA1;\xA1\xA3!\x06g\xC8(I-\xB9\x8D\xCA> v\xCC75\xA9 \xA3\xCAP]8+\xBCT`\x01`\x01`\xA0\x1B\x03\x16\x90V[\x90P\x90V[6__7__6_\x84Z\xF4=__>\x80\x80\x15`iW=_\xF3[=_\xFD\xFE\xA1dsolcC\0\x08\x1C\0\n",
     );
-    #[derive(serde::Serialize, serde::Deserialize)]
-    #[derive(Default, Debug, PartialEq, Eq, Hash)]
+    #[derive(serde::Serialize, serde::Deserialize, Default, Debug, PartialEq, Eq, Hash)]
     /**Custom error with signature `AddressEmptyCode(address)` and selector `0x9996b315`.
-```solidity
-error AddressEmptyCode(address target);
-```*/
+    ```solidity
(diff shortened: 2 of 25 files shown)


contracts/rust/deployer/src/lib.rs
```diff
@@ -548,12 +548,16 @@ pub async fn upgrade_light_client_v2(
...
 
-            // prepare init calldata
+            // prepare init calldata (checks if already initialized)
             let lcv2 = LightClientV2::new(lcv2_addr, &provider);
-            let init_data = lcv2
-                .initializeV2(blocks_per_epoch, epoch_start_block)
-                .calldata()
-                .to_owned();
+            let init_data =
+                if already_initialized(&provider, proxy_addr, Contract::LightClientV2, 2).await? {
```
alxiong: btw, i take another look at the `fn already_initialized()`, we can drop the 3rd input argument. since we know all inherited contract should have `.getVersion()` func, we can simply `LightClient::new()` over the address provided, it's the same abi encoding when sending calls to the proxy.

also I would rename to `expect_light_client_contract_version(provider, addr, major, minor, patch)`

contracts/rust/deployer/src/lib.rs
```diff
@@ -548,12 +548,16 @@ pub async fn upgrade_light_client_v2(
...
+                    vec![].into()
+                } else {
+                    lcv2.initializeV2(blocks_per_epoch, epoch_start_block)
+                        .calldata()
+                        .to_owned()
+                };
```
alxiong: shouldn't we ignore and early return if the proxy is already on V2? why still call initialize on empty initdata?

contracts/rust/deployer/src/lib.rs
```diff
@@ -672,7 +685,15 @@ pub async fn upgrade_light_client_v3(
 
             // prepare init calldata
             let lcv3 = LightClientV3::new(lcv3_addr, &provider);
-            let init_data = lcv3.initializeV3().calldata().to_owned();
+
+            // set init data unless already initialized
+            let init_data =
+                if already_initialized(&provider, proxy_addr, Contract::LightClientV3, 3).await? {
+                    vec![].into()
+                } else {
```
alxiong: same here, should early return, right?

contracts/rust/deployer/src/lib.rs
```diff
@@ -615,6 +619,15 @@ pub async fn upgrade_light_client_v3(
...
             let proxy = LightClient::new(proxy_addr, &provider);
 
+            //check proxy version, if not atleast V2, can't upgrade to V3
+            let version = proxy.getVersion().call().await?;
+            if version.majorVersion < 2 {
+                anyhow::bail!(
+                    "LightClientProxy is V{}, can't upgrade to V3",
+                    version.majorVersion
+                );
+            }
```
alxiong, line 633: instead of `<2`, i'd completely compare against the expected version: `if version != Version { majorVersion: 2, minorVersion: 0, patchVersion: 0}` then bail.

alysiahuggins: in a situation where it's a hotfix and we don't want to increase versions, we would be fine with doing v3 -> v3. We've done that before for a v2 contract

contracts/rust/deployer/src/lib.rs
alysiahuggins: init_data also includes the method name, e.g. `initializeV2` so if the init data is empty no initialized method is called (no method at all is called. it just does the upgrade).

contracts/rust/deployer/src/lib.rs
alysiahuggins: same comment as above, if the init data is empty , no initialize method is called

contracts/rust/deployer/src/lib.rs
alysiahuggins: can you explain that rename a bit more. the aim of the method is to check whether the contract is already initialized for that version, let's say, v2, in the scenarios where we're redeploying a version, let's say we already deployed v2, initializeV2 would have been called. so if you try to deploy v2 again (there might be a valid reason to), then the initializeV2 method should not be called.

alxiong: > the aim of the method is to check whether the contract is already initialized for that version

i don't think it checks whether it's initialized (ie. the `initializedV2()` has been called), it only checks if the proxy has pointed to the right impl contract with the right versions. 
so i think it's a misnormer to call it `check_initialized()`

contracts/rust/deployer/src/lib.rs
alysiahuggins: i'll add comments like this in the code - explaining overriding upgrades

contracts/rust/deployer/src/lib.rs
alysiahuggins: will add a comment to explain this

alxiong: pending clarification from the confirmation team, whether we can always bump up the version.

contracts/rust/deployer/src/lib.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3647/commits/3fa0975d61ad0f1db42d3c2025014ef84c7c50ba

alysiahuggins: also we discussed in a call that there will be no name change, however, i added comments

contracts/rust/deployer/src/lib.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3647/commits/3fa0975d61ad0f1db42d3c2025014ef84c7c50ba

contracts/rust/deployer/src/lib.rs
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3647/commits/3fa0975d61ad0f1db42d3c2025014ef84c7c50ba

contracts/rust/deployer/src/lib.rs
alysiahuggins: from the chat on slack they use the major version only at the moment

PR #3651, opened 2025-10-09
feat: RewardClaim daily limits, pausability

sveitser:
Depends on #3491

- feat: daily limits, pausability
- feat: robuster initialization
- test: reward claim upgrades and roles
- test: reward claim initialization
- test: reward claim admin functions
- test: re-entrancy
- doc: note about daily limits

---
  - https://app.asana.com/0/0/1211505964134163

### DIFF
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -1050,7 +1050,7 @@ async fn validate_next_stake_table_hash(
     let next_stake_table_hash = epoch_membership
         .stake_table_hash()
         .await
-        .ok_or(ProposalValidationError::NextStakeTableNotFound)?;
+        .ok_or(ProposalValidationError::NextStakeTableHashNotFound)?;
     let Some(proposed_next_stake_table_hash) = proposed_header.next_stake_table_hash() else {
         return Err(ProposalValidationError::NextStakeTableHashNotFound);
     };


contracts/src/RewardClaim.sol
```diff
@@ -1,72 +1,206 @@
...
+        // NOTE: __Ownable_init checks _owner != address(0)
+        require(_lightClient != address(0), ZeroLightClientAddress());
+        require(_pauser != address(0), ZeroPauserAddress());
+        require(_espToken != address(0), ZeroTokenAddress());
+
+        // NOTE: external call
+        uint256 totalSupply = EspTokenV2(_espToken).totalSupply();
+        require(totalSupply > 0, ZeroTotalSupply());
+
+        uint256 _dailyLimit = totalSupply / 100;
```
alysiahuggins, line 106: is this a placeholder? if not, 1% feels too high, perhaps 0.001%?

contracts/src/RewardClaim.sol
```diff
@@ -1,72 +1,206 @@
...
 
-    function claimRewards(uint256 lifetimeRewards, bytes memory authData) external {
+    function unpause() external onlyRole(PAUSER_ROLE) {
+        _unpause();
+    }
+
+    /// @notice Updates the daily limit
+    /// @param newLimit New daily limit in wei, must be greater than zero
+    function setDailyLimit(uint256 newLimit) external onlyOwner {
+        require(newLimit > 0, ZeroDailyLimit());
```
alysiahuggins, line 136: should we provide an upper bound? especially since this method is not pausable

contracts/src/RewardClaim.sol
```diff
@@ -1,72 +1,206 @@
...
+    /// reason about. DO NOT REMOVE: intentionally kept for defense-in-depth and
+    /// code clarity. See RewardClaimReentrancy.t.sol for regression test.
+    function claimRewards(uint256 lifetimeRewards, bytes memory authData)
+        external
+        whenNotPaused
+        nonReentrant
+    {
         require(lifetimeRewards != 0, InvalidRewardAmount());
-        require(claimedRewards[msg.sender] < lifetimeRewards, AlreadyClaimed());
+        address claimer = msg.sender;
```
alysiahuggins, line 155: starting the discussion on whether we want a platform to claim on behalf of a user

contracts/src/RewardClaim.sol
sveitser: No this is what has been agreed upon as a good compromise. If we set it too low we need to provide escape hatches which will be much more work.

contracts/src/RewardClaim.sol
sveitser: I don't think that's necessary since this is a fallback security mechanism and it requires an exploit to circumvent the merkle tree authentication.

contracts/src/RewardClaim.sol
sveitser: I think maybe chat thread is more suitable for this at this stage? wdyt?

alysiahuggins: yes

contracts/src/RewardClaim.sol
alysiahuggins: Even though it requires an exploit to bypass merkle tree authentication, the daily limit serves as the last line of defense. If the primary security fails, it would be good if the daily limit is limited to minimize the damage

contracts/src/RewardClaim.sol
alysiahuggins: ok got it

contracts/src/RewardClaim.sol
sveitser: I just can't think of a scenario where this would be set to a too high number unwillingly. It's also behind timelock. I think if the multisig is compromised having a limit here won't help at all.

It might be reasonably to set the upper limit at 1% so that it would require a smart contract upgrade to go higher. Only allowing lowering is also not great because then if it's set too low it requires a contract upgrade to increase it again. I think 1% is a really good tradeoff in this particular case. The chance of this ever being needed is very low and the attacker should be able to mint 2% in the worst case.

alysiahuggins: you're right that if the multisig is compromised, then we have other things to worry about. Cases where the multisig is legitimate but an unusually high value is set would probably be a fat finger error. but i'll leave this to your judgement then

alxiong: it's okay, admin can immediately rectify the fat-finger value.

contracts/src/RewardClaim.sol
```diff
@@ -1,72 +1,206 @@
...
+        uint256 today = block.timestamp / 1 days;
+        if (today != _currentDay) {
+            _currentDay = today;
+            _claimedToday = 0;
+        }
+        uint256 newTotal = _claimedToday + amount;
+        if (newTotal > dailyLimit) {
+            revert DailyLimitExceeded();
+        }
+        _claimedToday = newTotal;
```
alxiong: nit: no need the extra tmp variable `newTotal`, revert will undo changes to `_claimedToday` anyway.

justfile
```diff
@@ -150,7 +150,9 @@ build-docker-images:
     scripts/build-docker-images-native
 
 # generate rust bindings for contracts
-REGEXP := "^LightClient(V\\d+)?$|^LightClientArbitrum(V\\d+)?$|^FeeContract$|PlonkVerifier(V\\d+)?$|^ERC1967Proxy$|^LightClient(V\\d+)?Mock$|^StakeTable$|^StakeTableV2$|^EspToken$|^EspTokenV2$|^OpsTimelock$|^SafeExitTimelock$|^OwnableUpgradeable$|RewardClaim$"
+VERSIONED := "LightClient(Arbitrum)?(V\\d+)?(Mock)?|PlonkVerifier(V\\d+)?|StakeTable(V\\d+)?|EspToken(V\\d+)?|RewardClaim(V\\d+)?"
+EXACT := "FeeContract|ERC1967Proxy|OpsTimelock|SafeExitTimelock|OwnableUpgradeable|IRewardClaim|IPlonkVerifier"
```
alxiong, line 157: 👍 I like it.

contracts/test/RewardClaimPause.t.sol
```diff
@@ -0,0 +1,62 @@
...
+    function test_Pause() public {
+        vm.prank(pauser);
+        rewardClaim.pause();
+
+        vm.prank(claimer);
+        vm.expectRevert(abi.encodeWithSignature("EnforcedPause()"));
+        rewardClaim.claimRewards(1, "");
+    }
+
+    function test_Pause_Upause() public {
```
alxiong:
```suggestion
    function test_Pause_Unpause() public {
```

contracts/src/RewardClaim.sol
sveitser: So in order for the error to make it into the contract we would need

1. Wrong transaction proposed on chain
2. Nobody notices for 2 weeks while proposed TX is on chain (timelock)
3. A quorum of signers signs anyway
4. The transaction is finally executed anyway

But yeah I do think it's good to abort early if something is almost definitely wrong. Like if it's greater than 5% of total supply or something. Not quite sure what that number should be though.

I chose Wei units instead of ESP to avoid accidentally setting it 10^18 too high. But it's different from the token deployment which is also a bit confusing.

contracts/src/RewardClaim.sol
sveitser: I made this change but I think it's kind of unintuitive. Also not sure about gas usage since there may be an extra storage read unless that's optimized away. 🤔

PR #3653, opened 2025-10-09
test: extensive tests for reward claims

sveitser:
Adds what I believe to be good test coverage for reward claim proof verification, daily limits, and claiming.

Almost all changes are adding tests.

There is a small change regarding a max daily limit, follow up from #3651 

---
  - https://app.asana.com/0/0/1211766439562113


contracts/rust/diff-test/src/main.rs
```diff
@@ -583,5 +595,159 @@ fn main() {
...
+            let res = (auth_root_u256, fixtures);
+            println!("{}", res.abi_encode_params().encode_hex());
+        },
+        Action::GenRewardFixtureWithAccountAndAmount => {
+            if cli.args.len() < 2 || cli.args.len() > 3 {
+                panic!("Should provide arg1=account, arg2=amount, arg3(optional)=seed");
+            }
+
+            let account_bytes = hex::decode(&cli.args[0]).unwrap();
+            let account = Address::from_slice(&account_bytes);
```
alxiong: why pass bytes, just directly pass address so that we can directly `.parse::<Address>()`

contracts/test/RewardClaimTestBase.sol
```diff
@@ -0,0 +1,127 @@
...
+
+import "forge-std/Test.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import "../src/RewardClaim.sol";
+import "../src/EspTokenV2.sol";
+import "./mocks/LightClientV3Mock.sol";
+import "../src/interfaces/IRewardClaim.sol";
+
+// Conventions:
+// - Always use validateTestCase to validate succesfull claims before fuzzing without changing
```
alxiong:
```suggestion
// - Always use `validateTestCase()` to validate successful claims before fuzzing without changing the
```

contracts/test/RewardClaimTestBase.sol
```diff
@@ -0,0 +1,127 @@
...
+        cmds[2] = vm.toString(account);
+        cmds[3] = vm.toString(amount);
+        bytes memory result = vm.ffi(cmds);
+        (authRoot, fixtures) = abi.decode(result, (uint256, RewardClaimTestCase[]));
+    }
+
+    function validateTestCase(RewardClaimTestCase memory testCase, uint256 authRoot) internal {
+        RewardClaimTestHelper helper = new RewardClaimTestHelper();
+        helper.setUp();
+        helper.doValidate(testCase, authRoot);
```
alxiong: this relationship between the abstract contract and its realizer feels strange. can you explain why we need the Helper?

contracts/src/RewardClaim.sol
```diff
@@ -131,22 +140,28 @@ contract RewardClaim is
...
     /// @notice Updates the daily limit
-    /// @param newLimit New daily limit in wei, must be greater than zero
-    function setDailyLimit(uint256 newLimit) external onlyOwner {
+    /// @param newLimit New daily limit in wei, must be greater than zero and less than max
+    /// @dev nonReentrant protects against reentrancy during the external call to
+    /// espToken.totalSupply(). While unlikely to be exploited, this provides defense-in-depth
+    /// security for critical security parameters.
+    function setDailyLimit(uint256 newLimit) external onlyOwner nonReentrant {
         require(newLimit > 0, ZeroDailyLimit());
+        uint256 maxLimit = (espToken.totalSupply() * MAX_DAILY_LIMIT_PERCENTAGE) / 100e18;
```
sveitser: @alysiahuggins @alxiong are you happy with this check?

contracts/src/RewardClaim.sol
```diff
@@ -131,22 +140,28 @@ contract RewardClaim is
...
     /// @param authData Authentication data from Espresso query service
+    ///
     /// @dev nonReentrant is not strictly necessary (claimedRewards updated before
     /// external call, and re-entrancy would change msg.sender making proof
     /// verification fail), but makes the security properties much simpler to
     /// reason about. DO NOT REMOVE: intentionally kept for defense-in-depth and
-    /// code clarity. See RewardClaimReentrancy.t.sol for regression test.
-    function claimRewards(uint256 lifetimeRewards, bytes memory authData)
+    /// code clarity. See RewardClaim.Reentrancy.Unit.t.sol for regression test.
+    function claimRewards(uint256 lifetimeRewards, bytes calldata authData)
```
sveitser, line 197: This saves a tiny bit of gas in my test, I'm not 100% sure it's significant though.

contracts/src/RewardClaim.sol
```diff
@@ -131,22 +143,29 @@ contract RewardClaim is
...
 
     /// @notice Updates the daily limit
-    /// @param newLimit New daily limit in wei, must be greater than zero
-    function setDailyLimit(uint256 newLimit) external onlyOwner {
+    /// @param newLimit New daily limit in wei, must be greater than zero and less than max
+    /// @dev nonReentrant protects against reentrancy during the external call to
+    /// espToken.totalSupply(). While unlikely to be exploited, this provides defense-in-depth
+    /// security for critical security parameters.
+    function setDailyLimit(uint256 newLimit) external onlyOwner nonReentrant {
+        require(newLimit != dailyLimit, NoChangeRequired());
```
sveitser: Just noticed this check was missing.

contracts/src/RewardClaim.sol
sveitser: I don't really like the external call but absolute limits also feel clumsy because then we _eventually_ have to update the contract just to have a reasonable limit. Not sure we will still be around to see that though.

contracts/src/RewardClaim.sol
alxiong: technically not a bug anyway haha.

contracts/src/RewardClaim.sol
alxiong: can we replace `100e18` with `ERC.decimals()`? 
(to me, external calls are fine)

I would express percentage via `5 * (decimals() - 2)` instead of hardcode the `e18`

alysiahuggins (approved):

PR #3661, opened 2025-10-14
feat: claim rewards with staking-cli in the demos

sveitser:
- Add claim functionality to staking-cli
- Break up contract deployment in demo in 3 steps to satisfy requirements of deployment order.
- Add claimer service to demos

---
  - https://app.asana.com/0/0/1211766439562114

### DIFF
--- a/staking-cli/src/main.rs
+++ b/staking-cli/src/main.rs
@@ -425,7 +425,7 @@ pub async fn main() -> Result<()> {
         },
         Commands::UpdateMetadataUri { metadata_uri } => {
             tracing::info!("Updating validator {account} metadata URI");
-            update_metadata_uri(&provider, stake_table_addr, metadata_uri).await?
+            update_metadata_uri(&provider, stake_table_addr, metadata_uri).await
         },
         Commands::Approve { amount } => {
             approve(&provider, token_addr, stake_table_addr, amount).await


docker-compose.yaml
```diff
@@ -71,6 +71,61 @@ services:
       deploy-sequencer-contracts:
         condition: service_completed_successfully
 
+  deploy-pos-contracts-upgrades:
+    image: ghcr.io/espressosystems/espresso-sequencer/deploy:${DOCKER_TAG:-main}
+    command: deploy --deploy-reward-claim --upgrade-esp-token-v2 --upgrade-stake-table-v2
```
alysiahuggins, line 112: does the reward claim need lcv3?

staking-cli/README.md
```diff
@@ -73,6 +74,8 @@ Commands:
     undelegate             Initiate a withdrawal of delegated funds from a validator
     claim-withdrawal       Claim withdrawal after an undelegation
     claim-validator-exit   Claim withdrawal after validator exit
+    claim-rewards          Claim staking rewards
+    unclaimed-rewards      Check unclaimed staking rewards
```
alysiahuggins, line 80: nit: somehow `claim-rewards` and `unclaimed-rewards` might be mistaken, maybe `rewards-balance`?

docker-compose.yaml
sveitser: Yes

sveitser: We could also move the LC V3 update to this job I guess.

scripts/claim-rewards-loop
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SLEEP_INTERVAL="${1:-10}"
+
+echo "Starting reward claiming loop (checking every ${SLEEP_INTERVAL}s)..."
+
+while true; do
+  unclaimed=$(staking-cli unclaimed-rewards)
```
alysiahuggins, line 9: nit: add error handling, my assumption is that the staking-cli is used by validators when making this recommendation.

sveitser: So this script is just for testing the demo. I was considering adding a command to the staking CLI to watch for claimable rewards and claim them but it's not really something I think will be used outside the demo so having this bash script seemed simpler. The staking-cli has some error handling in that it tries to show a helpful message when something goes wrong.

When claiming there is a potential race condition in case another command has already claimed and I wasn't quite sure what to do in this case. If a claim command fails because someone else claimed it should sort of exit with a failure code because running to the command didn't do what it was supposed to do. On the other hand the actual claim did happen so the what the command set out to do was accomplished anyway.

alysiahuggins: hmm for that race condition perhaps if the error message also included the current rewards balance the user would see that even though it failed, the balance is zero, which helps them debug why it might have failed.

sveitser: https://github.com/EspressoSystems/espresso-network/issues/3775

PR #3663, opened 2025-10-14
feat: transaction inclusion proof

### LINKED ISSUE #3532: Test transaction proof verification
Here
https://github.com/EspressoSystems/espresso-network/blob/cl/txproof/hotshot-query-service/src/availability.rs#L1074-L1085

we don't have an API to verify the inclusion proof.

mrain:
Closes #3532 

### This PR:
Introduces the transaction inclusion proof trait in `hotshot-query-service` and enriches test in availability test.

### Key places to review:

---
  - https://app.asana.com/0/0/1211180342156956

### DIFF
--- a/types/src/v0/impls/block/test.rs
+++ b/types/src/v0/impls/block/test.rs
@@ -70,9 +70,7 @@ async fn basic_correctness() {
                 assert_eq!(tx, tx2);
                 tx_proof
             };
-            assert!(tx_proof2
-                .verify(block.ns_table(), &tx, &vid_commit, &vid_common)
-                .unwrap());
+            assert!(tx_proof2.verify(block.ns_table(), &tx, &vid_commit, &vid_common));
         }
         assert!(
             all_txs.is_empty(),


hotshot-query-service/src/availability/query_data.rs
```diff
@@ -89,6 +88,26 @@ where
...
+{
+    /// Verify the inclusion proof against a payload commitment.
+    /// Returns `None` on error.
+    fn verify(
+        &self,
+        metadata: &Metadata<Types>,
+        tx: &Transaction<Types>,
+        payload_commitment: &VidCommitment,
+        common: &VidCommon,
+    ) -> Option<bool>;
```
jbearer: I think it's not clear what the difference is between `None` vs `Some(false)`

hotshot-query-service/src/availability/query_data.rs
```diff
@@ -89,6 +88,26 @@ where
     pub position: u32,
 }
 
+/// The proof system and the statement which is proved will vary by application, with different
+/// applications proving stronger or weaker statements depending on the trust assumptions at
+/// play. Some may prove a very strong statement (for example, a shared sequencer proving that
+/// the transaction belongs not only to the block but to a section of the block dedicated to a
+/// specific rollup), otherwise may prove something substantially weaker (for example, a trusted
+/// query service may use `()` for the proof).
+pub trait TransactionInclusionProof<Types: NodeType>:
```
alxiong: nit: `trait ProvableInclusion`, maybe clearer distinction from the `struct InclusionProof`

hotshot-query-service/src/availability/query_data.rs
mrain: It's for some historical reason that I don't really understand either, https://github.com/EspressoSystems/espresso-network/blob/b1c62f2acec6ece063476ba14bed448a2a13ee82/types/src/v0/impls/block/namespace_payload/tx_proof/advz.rs#L122

Should we change them all to `bool` then?

hotshot-query-service/src/availability/query_data.rs
mrain: probably better with `VerifiableInclusion`, WDYT

mrain: [75373fc](https://github.com/EspressoSystems/espresso-network/pull/3663/commits/75373fc2421b4f303d64659a973d57dfb32d2dec)

hotshot-query-service/src/availability/query_data.rs
jbearer: I don't think it's a big deal but I think either plain `bool` or like `Result<()>` would be better

mrain: I changed them all to `bool` [5ec354f](https://github.com/EspressoSystems/espresso-network/pull/3663/commits/5ec354f0f08e32d3bcd3c5a308dd4d9bb5468091)

PR #3676, opened 2025-10-17
test DA committees upgrade

pls148:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,2 @@
 [net]
 git-fetch-with-cli = true
-
-[build]
-jobs = 8
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
 [net]
 git-fetch-with-cli = true
+
+[build]
+jobs = 8
--- a/crates/hotshot/hotshot/src/traits/election/two_static_committees.rs
+++ b/crates/hotshot/hotshot/src/traits/election/two_static_committees.rs
@@ -435,4 +435,7 @@ impl<TYPES: NodeType> Membership<TYPES> for TwoStaticCommittees<TYPES> {
     fn first_epoch(&self) -> Option<TYPES::Epoch> {
         self.first_epoch
     }
+
+    /// add_da_committee is a no-op here as dynamic DA committees don't make sense here.
+    fn add_da_committee(&mut self, _first_epoch: u64, _da_committee: Vec<PeerConfig<TYPES>>) {}
 }
--- a/crates/hotshot/testing/tests/test_da_committees.rs
+++ b/crates/hotshot/testing/tests/test_da_committees.rs
@@ -24,8 +24,6 @@ cross_tests!(
     TestName: test_da_committees,
     Impls: [MemoryImpl],
     Types: [TestTypes, TestTypesRandomizedLeader],
-    // TODO: we need some test infrastructure + Membership trait fixes to get this to work with:
-    // Types: [TestTypes, TestTypesRandomizedLeader, TestTwoStakeTablesTypes],
     Versions: [DaCommitteeTestVersions],
     Ignore: false,
     Metadata: {
--- a/crates/hotshot/testing/tests/test_da_committees.rs
+++ b/crates/hotshot/testing/tests/test_da_committees.rs
@@ -1,0 +1,96 @@
+// Copyright (c) 2021-2024 Espresso Systems (espressosys.com)
+// This file is part of the HotShot repository.
+
+// You should have received a copy of the MIT License
+// along with the HotShot repository. If not, see <https://mit-license.org/>.
+
+use std::time::Duration;
+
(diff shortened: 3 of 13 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,5 @@
 [net]
-git-fetch-with-cli = true
\ No newline at end of file
+git-fetch-with-cli = true
+
+[build]
+jobs = 8
```
ss-es: is this change intentional? do we need it?

crates/hotshot/hotshot/src/traits/election/two_static_committees.rs
```diff
@@ -0,0 +1,441 @@
...
+        self.add_drb_result(epoch, initial_drb_result);
+        self.add_drb_result(epoch + 1, initial_drb_result);
+    }
+
+    fn first_epoch(&self) -> Option<TYPES::Epoch> {
+        self.first_epoch
+    }
+
+    /// add_da_committee is a no-op here as dynamic DA committees don't make sense here.
+    fn add_da_committee(&mut self, _first_epoch: u64, _da_committee: Vec<PeerConfig<TYPES>>) {}
```
ss-es, line 440: unless I'm missing something, I think they still make sense -- I would implement as in the other types

it's perfectly reasonable to add e.g. non-quorum nodes to the DA committee (and we might actually end up in this situation on mainnet)

crates/hotshot/types/src/lib.rs
```diff
@@ -24,6 +24,7 @@ use crate::utils::bincode_opts;
 pub mod bundle;
 pub mod consensus;
 pub mod constants;
+pub mod da_committee;
```
ss-es, line 27: is this a re-export?

crates/hotshot/testing/tests/test_da_committees.rs
```diff
@@ -0,0 +1,96 @@
...
+};
+use hotshot_types::ValidatorConfig;
+use vbs::version::Version;
+
+cross_tests!(
+    TestName: test_da_committees,
+    Impls: [MemoryImpl],
+    Types: [TestTypes, TestTypesRandomizedLeader],
+    // TODO: we need some test infrastructure + Membership trait fixes to get this to work with:
+    // Types: [TestTypes, TestTypesRandomizedLeader, TestTwoStakeTablesTypes],
```
ss-es: does it not work eith the two_static_committees implementation?

crates/hotshot/testing/tests/test_da_committees.rs
```diff
@@ -0,0 +1,96 @@
...
+            ),
+            ValidatorConfig::generated_from_seed_indexed(
+                [0u8; 32],
+                3,
+                U256::from(1),
+                true,
+            ),
+        ];
+
+        metadata.test_config.epoch_height = 50;
```
ss-es, line 68: I think in the test we can lower the epoch height to e.g. 10 (the default), unless there's a reason we don't want to?

ss-es: oh right, the upgrade -- nevermind

ss-es (commented):
just some minor comments -- need to think about the test in the meantime, but on the whole the pr looks really good to me

crates/hotshot/hotshot/src/traits/election/two_static_committees.rs
pls148: do we want to also do the even/odd thing for the da committees? those two goals seem to be contradictory

.cargo/config.toml
pls148: whoops, removed

crates/hotshot/testing/tests/test_da_committees.rs
pls148: missed removing comment when i copied the test over

crates/hotshot/hotshot/src/traits/election/two_static_committees.rs
ss-es: no, the DA committee should be independent of the quorum committee -- if we add a node to `da_committees` it shouldn't matter whether it's in the quorum or not (and in fact, we should be able to function normally even if none of the DA nodes are in the quorum committee)

ss-es: oh, I see the confusion -- we currently alternate DA committees

so it's weird to have an already alternating DA committee then overriding it. ok, I think it's fine to leave it out -- but maybe we can panic here with an informative message ("dynamic DA committees are not functional in this stake table implementation"), so that if we ever invoke this accidentally we won't be left wondering why the DA committee isn't updating

ss-es (approved):

ss-es (changes requested):
wait, actually -- something weird happened with the membership implementation. they should be under `example-types`, maybe this branch was based on an old commit on main?

PR #3680, opened 2025-10-20
Implement light client leaf endpoint and validation

jbearer:
* [Add endpoint to serve light client leaves with finality proof](https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510082563913?focus=true)
* [Validate leaf finality in light client](https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510082563937?focus=true)

### This PR:

* Adds a light client endpoint to serve leaves with finality proofs
* Adds code in the light client crate to validate these proofs (given a stake table)
* Exposes next epoch justify QCs in decide events, and stores them in consensus ephemeral storage (in the same way that #3636 and #3665 did for the deciding QCs). These are also needed to verify a decide event in the case where one of the QCs is part of an epoch transition.
* Deals with upgrades/versioning in the light client validation logic, especially regarding epoch enabling upgrade

### This PR does not:

* Implement stake table tracking/fetching. The leaf validation currently assumes a given stake table.
* Implement the leaf fetching part of the light client. This only implements the server side, and the client side validation which will be used by the fetching logic.

#### Soundness (@bfish713 and @ss-es )
* `CertificatePair::verify_next_epoch_qc` in `simple_certificate.rs`
* Light client leaf finality check: `LeafProof::verify` in `light_client/src/consensus/leaf.rs` (tests in that file provide a number of examples of good and bad proofs)
* QC chain verification: `Quorum::verify_qc_chain_and_get_version` in `light_client/src/consensus/quorum.rs` (tests in that file provide a number of examples of good and bad chains)
* Individual QC verification: `StakeTableQuorum::verify_static` in `light_client/src/consensus/quorum.rs`

#### Server side (@imabdulbasit )
* Exposing next epoch QCs from node storage (changes in `hotshot-query-service`)
* Persisting next epoch QCs in consensus storage until decide event (`sequencer/src/persistence/{sql,fs}.rs`)
  - New migrations in `sequencer/api/migrations`
* Leaf proof construction: `LeafProof::push` in `light_client/src/consensus/leaf.rs` and `get_leaf_proof` in `sequencer/src/api/light_client.rs`

### How to test this PR

Read new unit tests in `light_client/src/consensus/{leaf,quorum}.rs` and `sequencer/src/api/light_client.rs`

### DIFF
--- a/crates/hotshot/types/src/simple_certificate.rs
+++ b/crates/hotshot/types/src/simple_certificate.rs
@@ -987,7 +987,7 @@ impl<TYPES: NodeType> CertificatePair<TYPES> {
             self.qc.data.epoch
         ))?;
         if !is_epoch_transition(block_number, epoch_height) {
-            tracing::trace!(
+            tracing::debug!(
                 block_number,
                 epoch_height,
                 "QC is not in an epoch transition"


crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -901,3 +902,114 @@ impl<TYPES: NodeType> From<EpochRootQuorumCertificateV2<TYPES>>
...
+    /// Verify that the next epoch QC is present and consistent if required.
+    ///
+    /// This checks that if required, the next epoch QC is present and is consistent with the
+    /// primary QC. It does not check the signature on either QC, only that the data being signed
+    /// over is consistent between the two.
+    ///
+    /// Returns the next epoch QC if it is present and invariants are satisfied. Returns an error if
+    /// a required next epoch QC is missing or if it is inconsistent with the primary QC. Returns
+    /// [`None`] if a next epoch QC is not required for this certificate.
+    pub fn verify_next_epoch_qc(
```
jbearer, line 981: Consensus: assuming this function looks good, WDYT about restructuring the existing `validate_qc_and_next_epoch_qc` function to call this one?

The reason I didn't just use that function completely in the light client is:
* I wanted to split apart the invariant checking (e.g. consistency between QC and next epoch QC) from the low level threshold signature checks, which improves testability
* The light client handles versions differently, looking at the version in the parent leaf + upgrade certificate, rather than using an upgrade lock, since it might need to jump around in history and potentially handle more than one upgrade in a single instance

But if we were to replace the invariant checking in  `validate_qc_and_next_epoch_qc` with a call to this function, and just have `validate_qc_and_next_epoch_qc` handle the versioning and the `is_valid_cert` calls, I think we would be pretty immune from logic changes in core consensus not being propagated to the light client, which is something I'm worried about in the current structure.

light-client/src/consensus/leaf.rs
```diff
@@ -0,0 +1,359 @@
...
+
+            if opt_qc.is_none() {
+                // Get the QC signing `leaf` from the justify QC of the subsequent leaf.
+                opt_qc = Some(next.justify_qc().clone());
+            }
+        }
+
+        // Check that the final leaf is actually finalized and get the QC which signs it.
+        let final_qc = match &self.proof {
+            FinalityProof::Assumption => {
```
Sneh1999, line 92: What is the use case of this one?

light-client/src/consensus/leaf.rs
```diff
@@ -0,0 +1,359 @@
...
+        // HotStuff2 QC chain for the leaf before.
+        if len >= 2 && self.leaves[len - 2].block_header().version() >= EpochVersion::version() {
+            let committing_qc = Certificate::for_parent(&self.leaves[len - 1]);
+            let deciding_qc = Certificate::for_parent(new_leaf.leaf());
+            if committing_qc.view_number() == self.leaves[len - 2].view_number()
+                && deciding_qc.view_number() == committing_qc.view_number() + 1
+            {
+                // Sanity check: if we have a chain of QCs from consecutive views, they must refer
+                // to consecutive leaves.
+                debug_assert!(committing_qc.leaf_commit() == self.leaves[len - 2].commit());
```
Sneh1999, line 163: why are these debug assert?

sequencer/api/light-client.toml
```diff
@@ -0,0 +1,43 @@
+[route.leaf]
+PATH = ["/leaf/:number", "/leaf/:number/:finalized"]
```
Sneh1999, line 2: I have a question about this, for integrations when we call this endpoint will it be like if I get a given hotshot block from query node, I will call this endpoint with that hotshot block which will give us this proof?

Sneh1999: If my understanding is right then we will also have to write some code similar to the below to verify this chain of leafs?

```
pub async fn verify(
        &self,
        quorum: &impl Quorum,
        finalized: Option<&Leaf2>,
    ) -> Result<LeafQueryData<SeqTypes>> {
```

Sneh1999 (commented):
just some comments for my own understanding!

light-client/src/consensus/leaf.rs
jbearer: If the light client wants leaf `n` and it already knows about leaf `n + 1`, we don't need to bother with the QC 2-chain/threshold signature check. We can just verify a hash chain from the requested leaf up to the already known leaf. So this is just a way of saying "QC chain omitted because the client told me they didn't need it"

light-client/src/consensus/leaf.rs
jbearer: So we can compile them out in a release build if we want. Because recomputing the leaf commit is not a trivial amount of compute and the check is redundant anyways

sequencer/api/light-client.toml
jbearer: This will all be abstracted. The integrations will just call a light client function that has an interface identical to the current query service (so like `/availability/block/:height/namespace/:id` endpoint). But as a function instead of an HTTP call. Internally the light client will use this new query service endpoint as needed to both fetch and verify data.

> If my understanding is right then we will also have to write some code similar to the below to verify this chain of leafs?

Not sure what you mean, this function already exists as of this PR?

Sneh1999: > This will all be abstracted. The integrations will just call a light client function that has an interface identical to the current query service (so like /availability/block/:height/namespace/:id endpoint). But as a function instead of an HTTP call. Internally the light client will use this new query service endpoint as needed to both fetch and verify data.

Yes I think I understand that we will abstract this as a part of a function which will further call this endpoint to get the proof and then verify the proof. I was just wondering because integrations code is currently written in golang, we might have to write some FFI between rust and golang? Or replicate the verify logic of rust in our golang SDK? 
Is my understanding correct here?

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -901,3 +902,114 @@ impl<TYPES: NodeType> From<EpochRootQuorumCertificateV2<TYPES>>
...
+            "next epoch QC is missing from epoch transition cert at block number {block_number}"
+        ))?;
+
+        // The signature from the next epoch must be over the same data.
+        ensure!(self.qc.view_number == next_epoch_qc.view_number);
+        ensure!(self.qc.data == *next_epoch_qc.data);
+
+        Ok(Some(next_epoch_qc))
+    }
+}
```
ss-es, line 1008: I don't think there's anything really wrong with this function but I'm not sure about the purpose. Are we guarding against our own code passing us an inconsistent pair?

I guess from my limited understanding it seems like either the light client trusts the `CertificatePair` or it doesn't. If it trusts the `CertificatePair`, then this is unnecessary. If it doesn't trust the `CertificatePair`, I don't see why we would ever call this separately from e.g. `validate_qc_and_next_epoch_qc` as you said

ss-es: if your suggestion was using `validate_qc_and_next_epoch_qc` in the light client and then calling this function in that, I think that makes a lot of sense to me and I don't see any reason not to

light-client/src/consensus/leaf.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// A chain of leaves from the requested leaf to a provably finalized leaf.
+    ///
+    /// The chain is in chronological order, so `leaves[0]` is the requested leaf and
+    /// `leaves.last()` is a leaf which is known or can be proven to be finalized. The chain is
+    /// joined by `parent_commitment`, so it can be validated by recomputing the commitment of each
+    /// leaf and comparing to the parent commitment of the next.
+    leaves: Vec<Leaf2>,
+
+    /// Some extra data proving finality for the last leaf in `leaves`.
+    proof: FinalityProof,
```
ss-es, line 59: this structure makes me a bit nervous. `FinalityProof` has information that is not necessarily the same as the information in `leaves`, and you have to perform an extra check to compare e.g. the certs in the leaves against the certs in a `FinalityProof`

light-client/src/consensus/leaf.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+impl LeafProof {
+    /// Verify the proof.
+    ///
+    /// If successful, returns the leaf which is proven finalized.
+    pub async fn verify(
+        &self,
+        quorum: &impl Quorum,
+        finalized: Option<&Leaf2>,
```
ss-es, line 69: maybe I'm misunderstanding how this is meant to be used, but I feel like `FinalityProof::Hotstuff` and `FinalityProof::Hotstuff2` should have no fields, whereas `FinalityProof::Assumption` should have a single `finalized_leaf` field?

for hotstuff/hotstuff2, you'd then extract a corresponding number of leaves from the leaf chain and check those qc's

sequencer/api/light-client.toml
jbearer: Oh yeah I was thinking we would just do an FFI wrapper, like we did with the crypto functions

crates/hotshot/types/src/simple_certificate.rs
jbearer: I didn't want to use `validate_qc_and_next_epoch_qc` directly in the light client because it does two things:
* basic consistency check of the two QCs
* threshold signature check

I wanted to separate these out so that I can mock the signature check with the `Quorum` trait in the light client code, which greatly simplifies testing. So, this function is basically pulling the consistency check out of `validate_qc_and_next_epoch_qc` as a self-contained function.

My suggestion is to call this function in `validate_qc_and_next_epoch_qc` so that both `validate_qc_and_next_epoch_qc` and the light client are checking consistency of these two certificates in the same way (which should be equivalent to how it was previously being checked). But the light client would still not call `validate_qc_and_next_epoch_qc` itself.

light-client/src/consensus/leaf.rs
jbearer: But the `verify` function does that check, and these fields are private. I don't really see the problem with having invariants between different fields of a struct which is meant to serve as one opaque proof type. It is up to clients not to forget to verify the proof, like any proof type

light-client/src/consensus/leaf.rs
jbearer: In the current design, the extra 2 or 3 QCs are not in the leaf chain. You could also do your suggestion and have the leaf chain be 2 or 3 longer, but there are reasons for not doing that:
* the last 2 or 3 leaves would not be used, only their justify QCs, so this is not optimal
* there is a potential confusion with returning potentially unfinalized data in the last 2 or 3 leaves just so the client can pull out their justify QCs
* that doesn't work if the QC 2-chain comes from the most recent decide event's `committing_qc` and `deciding_qc`, in which case we only have access to the QCs, not the corresponding leaves necessarily

As for `Assumption`, the reason why it does not contain a `finalized_leaf` field is because the finalized leaf is meant to be supplied by the client as an input to verification. We will only construct a proof with an `Assumption` variant when the client has specified in their request that they already have some leaf that they know to be finalized, so it's redundant to also include that leaf in the proof.

sequencer/src/api.rs
```diff
@@ -6176,4 +6200,132 @@ mod test {
...
+            if leaf.header().version() < EpochVersion::version() {
+                tracing::info!(version = %leaf.header().version(), height = leaf.header().height(), view = ?leaf.leaf().view_number(), "waiting for epoch upgrade");
+                continue;
+            }
+            break (leaf.height(), leaf.leaf().epoch(EPOCH_HEIGHT));
+        };
+        tracing::info!(upgrade_height, ?first_epoch, "epochs enabled");
+
+        // Wait for an epoch change.
+        let next_epoch_height = loop {
```
imabdulbasit: I don't think it would make a difference but could you please use the epoch 2->3 transition blocks because we use the stake table from contract from epoch 3 onwards

sequencer/src/api/light_client.rs
```diff
@@ -0,0 +1,327 @@
...
+    requested: usize,
+    finalized: Option<usize>,
+    fetch_timeout: Duration,
+) -> Result<LeafProof, Error>
+where
+    State: AvailabilityDataSource<SeqTypes> + VersionedDataSource,
+    for<'a> State::ReadOnly<'a>: NodeStorage<SeqTypes>,
+{
+    let (endpoint, qc_chain) = match finalized {
+        Some(finalized) => {
```
imabdulbasit, line 26: maybe add a check that finalized is > than requested block number?

sequencer/src/api.rs
jbearer: That's gonna make the test really slow lol but yeah I can try that

jbearer: [ba7c301](https://github.com/EspressoSystems/espresso-network/pull/3680/commits/ba7c3015f7909551e7bed2d8c1520d62f5ea04a1)

sequencer/src/api/light_client.rs
jbearer: [e44d300](https://github.com/EspressoSystems/espresso-network/pull/3680/commits/e44d3001fcd15ff46717bf3709bda93f38f0db8e)

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -901,3 +902,113 @@ impl<TYPES: NodeType> From<EpochRootQuorumCertificateV2<TYPES>>
...
+    pub fn verify_next_epoch_qc(
+        &self,
+        epoch_height: u64,
+    ) -> Result<Option<&NextEpochQuorumCertificate2<TYPES>>> {
+        let block_number = self.qc.data.block_number.context(warn!(
+            "QC for epoch {:?} has no block number",
+            self.qc.data.epoch
+        ))?;
+        if !is_epoch_transition(block_number, epoch_height) {
+            tracing::trace!(
```
ss-es: I think this might be better as `debug` -- I feel like we put very little at the `trace` level, and making it `trace` probably means nobody will ever see it

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -901,3 +902,113 @@ impl<TYPES: NodeType> From<EpochRootQuorumCertificateV2<TYPES>>
...
+
+        let next_epoch_qc = self.next_epoch_qc.as_ref().context(warn!(
+            "Received High QC for the transition block {block_number} but not the next epoch QC"
+        ))?;
+
+        // The signature from the next epoch must be over the same data.
+        ensure!(self.qc.view_number == next_epoch_qc.view_number);
+        ensure!(self.qc.data == *next_epoch_qc.data);
+
+        Ok(Some(next_epoch_qc))
```
ss-es, line 1006: looks good to me

ss-es (approved):

crates/hotshot/types/src/simple_certificate.rs
bfish713: Splitting this logic like you said makes sense to me

PR #3681, opened 2025-10-20
Broadcast view change early from proposal

ss-es:
This PR fixes an issue where consensus stalled due to an early exit in `handle_quorum_proposal_recv`. We now broadcast a view change immediately in that function based on any valid certificates we receive

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -1179,7 +1179,9 @@ pub(crate) async fn validate_proposal_view_and_certs<
                      preceding view"
                 );
                 let timeout_cert_epoch = timeout_cert.data().epoch();
-                membership = membership.get_new_epoch(timeout_cert_epoch).await?;
+                membership = membership
+                    .get_new_epoch_stake_table(timeout_cert_epoch)
+                    .await?;
 
                 let membership_stake_table = membership.stake_table().await;
                 let membership_success_threshold = membership.success_threshold().await;
@@ -1204,7 +1206,9 @@ pub(crate) async fn validate_proposal_view_and_certs<
                 );
 
                 let view_sync_cert_epoch = view_sync_cert.data().epoch();
-                membership = membership.get_new_epoch(view_sync_cert_epoch).await?;
+                membership = membership
+                    .get_new_epoch_stake_table(view_sync_cert_epoch)
+                    .await?;
 
                 let membership_stake_table = membership.stake_table().await;
                 let membership_success_threshold = membership.success_threshold().await;
--- a/crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
+++ b/crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
@@ -364,6 +364,22 @@ async fn view_change_from_proposal<TYPES: NodeType, I: NodeImplementation<TYPES>
         },
     }
 
+    // if we are using HS2 we update our locked view for any QC from a leader greater than our current lock
+    {
+        let mut consensus_writer = validation_info.consensus.write().await;
+        let liveness_check =
+            proposal.data.justify_qc().view_number() > consensus_writer.locked_view();
+        if liveness_check
+            && validation_info
+                .upgrade_lock
+                .version(proposal.data.view_number())
+                .await
+                .is_ok_and(|v| v >= V::Epochs::VERSION)
+        {
+            consensus_writer.update_locked_view(proposal.data.justify_qc().view_number());
+        }
+    }
+
     if (network_epoch > task_state.cur_epoch && network_view <= task_state.cur_view)
         || (network_epoch < task_state.cur_epoch && network_view > task_state.cur_view)
     {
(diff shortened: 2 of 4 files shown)


crates/hotshot/task-impls/src/helpers.rs
```diff
@@ -1179,7 +1179,9 @@ pub(crate) async fn validate_proposal_view_and_certs<
                      preceding view"
                 );
                 let timeout_cert_epoch = timeout_cert.data().epoch();
-                membership = membership.get_new_epoch(timeout_cert_epoch).await?;
+                membership = membership
+                    .get_new_epoch_stake_table(timeout_cert_epoch)
```
ss-es, line 1183: note that we do not actually need the full randomized stake table here (since we never query the leader)

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -592,6 +592,10 @@ impl<TYPES: NodeType> EpochMembership<TYPES> {
         self.coordinator.membership_for_epoch(epoch).await
     }
 
+    pub async fn get_new_epoch_stake_table(&self, epoch: Option<TYPES::Epoch>) -> Result<Self> {
```
ss-es, line 595: the naming is not very good, but I don't want to make a bigger change like that here. I feel like it would be worthwhile for us to go through this file and rename a lot of these methods to something more obvious

crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -242,6 +243,157 @@ async fn validate_block_height<TYPES: NodeType>(
...
+    let membership = validation_info.membership.clone();
+
+    let mut network_view = TYPES::View::new(0);
+    let mut network_epoch = None;
+
+    if let Some(vce) = proposal.data.view_change_evidence().clone() {
+        match vce {
+            ViewChangeEvidence2::Timeout(cert) => {
+                match membership.get_new_epoch_stake_table(cert.epoch()).await {
+                    Ok(membership) => {
```
bfish713, line 261: nit, can we use the `let Some(next_membership) = membership.get_new_epoch...() {..{ else return }, or return a `Result` so we can just use '?'

crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -242,6 +243,157 @@ async fn validate_block_height<TYPES: NodeType>(
...
+                            .is_valid_cert(
+                                &StakeTableEntries::<TYPES>::from(membership_stake_table.clone()).0,
+                                membership_success_threshold,
+                                &validation_info.upgrade_lock,
+                            )
+                            .await
+                            .is_ok()
+                        {
+                            network_epoch = max(network_epoch, cert.epoch());
+                            network_view = max(network_view, cert.view_number() + 1);
```
bfish713, line 275: Why do we need max?  These two values are set to 0 and None, and we aren't looping

crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -242,6 +243,157 @@ async fn validate_block_height<TYPES: NodeType>(
...
+                }
+            },
+        }
+    }
+
+    let justify_qc = proposal.data.justify_qc().clone();
+
+    match membership
+        .get_new_epoch_stake_table(justify_qc.epoch())
+        .await
```
bfish713, line 330: I think we are repeating this call and the next few lines of code for each cert type.  Can we just get the membership at the start.  I think there is no situation where the epoch is different for view change evidence and QC except for when the QC is for the last block (if there is it's a violation of the protocol).

crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -242,6 +243,157 @@ async fn validate_block_height<TYPES: NodeType>(
     Ok(())
 }
 
+async fn view_change_from_proposal<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>(
```
bfish713, line 246: Maybe this is minor but IMO the logic should be structured as take the view from the VCE if it exists, else take it from the justify QC.  There is no case where VCE view is <= justify QC view.  I think this plus changing the syntax will make this function more concise and easy to read

bfish713 (commented):
Logic looks correct to me, just hoping the `view_change_from_proposal` code can get simplified some

crates/hotshot/task-impls/src/quorum_proposal_recv/handlers.rs
ss-es: I'm not sure about that. I thought part of the point of this is that we want to view change even based on proposals we could still reject, so long as we receive a valid cert (justify qc/view change evidence) right?

A node may maliciously send us a proposal with view change evidence and justify qc in different epochs. They may both be valid or invalid, and we don't really know which to use until we try. And which epoch should we use at that point?

PR #3682, opened 2025-10-22
Implement Fetching for light client state certificate

imabdulbasit:
addresses https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1211700203604484

This PR moves the state certificate storage and API endpoints from the query service to the sequencer. This change is needed because, during downtime, the query node may miss some state certificates. Therefore, it needs to fetch the state certificate from its peers. However, we cannot blindly trust that a fetched certificate is valid. The sequencer has access to all stake tables, and if any are missing, it can perform a catchup. This makes it more logical for the state certificate to reside in the sequencer, as it can validate the certificate easily using the stake table

The SQL migrations for the state certificate remain in the  query service, since we don't want to lose existing certificates. While we could implement a migration and move those SQL files to sequencer migrations but both the sequencer and query service use the same SQL database, so it makes no difference

This PR also adds a test that restarts a node with fresh storage and verifies that it can successfully fetch the state certificate.

**Validation**
State certificate validation depends on the protocol version that the light client state certificate belongs to. If it belongs to  V3, we need to verify the LightClientV2 signatures in the state certificate. If it belongs to protocol version ≥ V4, we check the LCV3 signature instead. The certificate version can be inferred by checking the auth_root field. This field did not exist in V1 state certificate so auth_root is set to the default value in LightClientStateCertificateV2.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -13,7 +13,7 @@ test(test_process_client_handling_stream_subscribe_node_identity) |
 test(test_process_client_handling_stream_subscribe_voters)
 """
 slow-timeout = { period = "2s", terminate-after = 1 }
-retries  = 10
+retries = 10
 
 [[profile.default.overrides]]
 filter = 'test(test_hotshot_event_streaming_epoch_progression)'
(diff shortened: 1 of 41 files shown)


sequencer/src/state_cert.rs
```diff
@@ -0,0 +1,119 @@
...
+) -> anyhow::Result<()> {
+    let signed_state_digest = derive_signed_state_digest(
+        &cert.light_client_state,
+        &cert.next_stake_table_state,
+        &cert.auth_root,
+    );
+
+    // If auth_root is the default value (all zeros), we're on consensus version V3, so verify LCV2 signatures
+    // Otherwise, verify LCV3 signatures
+    let use_lcv2 = cert.auth_root == FixedBytes::<32>::default();
```
mrain: We had a little mis communication here, for consensus V4 we should verify both LCV3 signature and LCV2 signature.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3682/commits/81a2fff58b36ba275fafe3c0d9e9fdca1482fea5

PR #3683, opened 2025-10-22
Feature gate the stats task

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/hotshot/src/tasks/mod.rs
@@ -20,14 +20,13 @@ use futures::{
 use hotshot_task::task::Task;
 #[cfg(feature = "rewind")]
 use hotshot_task_impls::rewind::RewindTaskState;
-#[cfg(feature = "stats")]
-use hotshot_task_impls::stats::StatsTaskState;
 use hotshot_task_impls::{
     da::DaTaskState,
     events::HotShotEvent,
     network::{NetworkEventTaskState, NetworkMessageTaskState},
     request::NetworkRequestState,
     response::{run_response_task, NetworkResponseState},
+    stats::StatsTaskState,
     transactions::TransactionTaskState,
     upgrade::UpgradeTaskState,
     vid::VidTaskState,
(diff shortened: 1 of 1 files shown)


crates/hotshot/task-impls/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "MIT"
 [features]
 example-upgrade = []
 rewind = []
+stats = []
```
ss-es, line 12: this doesn't do anything right?

crates/hotshot/hotshot/Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "MIT"
 default = ["docs", "doc-images"]
 example-upgrade = ["hotshot-task-impls/example-upgrade"]
 rewind = ["hotshot-task-impls/rewind"]
+stats = ["hotshot-task-impls/stats"]
```
ss-es, line 14: this should probably be `stats = []`?

crates/hotshot/hotshot/src/tasks/mod.rs
```diff
@@ -268,6 +269,7 @@ pub async fn add_consensus_tasks<TYPES: NodeType, I: NodeImplementation<TYPES>,
         handle.add_task(QuorumVoteTaskState::<TYPES, I, V>::create_from(handle).await);
         handle.add_task(QuorumProposalRecvTaskState::<TYPES, I, V>::create_from(handle).await);
         handle.add_task(ConsensusTaskState::<TYPES, I, V>::create_from(handle).await);
+        #[cfg(feature = "stats")]
```
ss-es: nit: can we do `if cfg!(feature = "stats") {}`? it doesn't really matter here, but generally I think it's better to do that since the compiler type-checks the gated code even when the feature is disabled

crates/hotshot/hotshot/Cargo.toml
bfish713: I'm just following the examples of the other features we only use in hotshot

crates/hotshot/task-impls/Cargo.toml
bfish713: I think it just adds the flag to this crate, just following the pattern for other features

ss-es: oh weird

unless I'm missing something I think maybe none of these should be in the `task-impls` cargo file because we don't feature-gate anything in this crate as far as I can tell (the gating only happens in the `hotshot` crate)

PR #3684, opened 2025-10-22
New namespace endpoints

jbearer:
* [Add ranged namespace queries](https://app.asana.com/1/1208976916964769/project/1209380117384039/task/1211709312750778?focus=true)
* [Add namespace subscription streams](https://app.asana.com/1/1208976916964769/project/1209380117384039/task/1210163044689718?focus=true)
* Allow namespace queries to use the alternative block IDs (hash and payload hash) for consistency with other availability endopints

### This PR:
* Provides the new features outlined above
* Moves sequencer availability extension implementations to their own module to declutter the `endpoints.rs` module, which has other concerns
* Simplifies the handling of multiple versions in the namespace proof endpoints, reducing a lot of casework and duplicated code (see first commit)
* Reduces scope of read locks in the implementation of namespace queries, in particular not holding a lock on the state while fetching VID shares

### Testing
* Adds tests for the new endpoints to `test_namespace_query`
* Adds new tests `test_namespace_query_compat_*` to ensure the legacy endpoint versions are covered

### DIFF
--- a/sequencer/src/api.rs
+++ b/sequencer/src/api.rs
@@ -6339,7 +6339,7 @@ mod test {
             let ns_proof: NamespaceProofQueryData = client
                 .get(&format!(
                     "{api_ver}/availability/block/{}/namespace/{ns}",
-                    block + 1
+                    block - 1
                 ))
                 .send()
                 .await
@@ -6412,7 +6412,7 @@ mod test {
         let ns_proof: ADVZNamespaceProofQueryData = client
             .get(&format!(
                 "v0/availability/block/{}/namespace/{ns}",
-                block + 1
+                block - 1
             ))
             .send()
             .await


types/src/v0/nsproof.rs
```diff
@@ -27,8 +27,8 @@ pub enum NsProof {
     V0(ADVZNsProof),
     /// V1 proof for AvidM, contains only correct encoding proof
     V1(AvidMNsProof),
-    /// V1_1 proof for AvidM, contains both correct and incorrect encoding proofs
-    V1_1(AvidMNsProofV1),
+    /// Incorrect encoding proof for AvidM (only supported after API version 1.1)
```
mrain, line 30: I slightly prefer my previous approach, where users don't have to care about whether the payload is correctly encoded. If we change it in this way, users may need some explicit logic about encoding of a payload in their applications

jbearer: Why? They would still just call `proof.verify()`? This is just moving the enum of `{correct encoding, incorrect encoding}` that was already there one level higher up

mrain: I see, users could handle these in `NsProof` level. no problem with that

sequencer/api/availability.toml
```diff
@@ -1,8 +1,44 @@
 [route.getnamespaceproof]
-PATH = ["block/:height/namespace/:namespace"]
+PATH = [
+    "block/:height/namespace/:namespace",
+    "block/hash/:hash/namespace/:namespace",
+    "block/payload-hash/:payload-hash/namespace/:namespace",
```
Sneh1999, line 5: what will be the significance of this endpoint?

sequencer/api/availability.toml
```diff
@@ -1,8 +1,44 @@
...
+
+[route.stream_namespace_proofs]
+PATH = ["stream/blocks/:height/namespace/:namespace"]
+METHOD = "SOCKET"
+":height" = "Integer"
+":namespace" = "Integer"
+DOC = """
+Subscribe to the namespace data from each block, starting at `:height`.
+
+Opens a WebSockets connection and sends a stream of the same data type returned by
```
Sneh1999, line 39: I have one concern in my mind, lets say the websocket stream is down for sometime - would this make us end up in a situation where we will not know if we missed some transactions for a given namespace?

sequencer/api/availability.toml
```diff
@@ -1,8 +1,44 @@
...
+
+[route.getnamespaceproof_range]
+PATH = ["block/:from/:until/namespace/:namespace"]
+":from" = "Integer"
+":until" = "Integer"
+":namespace" = "Integer"
+DOC = """
+Get the transactions in the specified namespace from each block in a range.
+
+Returns a list of the same namespace proof type returned by `block/:height/namespace/namespace`.
```
Sneh1999, line 25: I wonder if we should make it optional to get the proof or not because in our first implementation we are just relying on multi query node and it might be possible that codebase doesnt want proofs. Retrieving the proofs will slow down this endpoint I believe

sequencer/api/availability.toml
jbearer: It's just for consistency with the other query endpoints. I don't think it will be particularly useful for the integrations

sequencer/api/availability.toml
jbearer: This endpoint will always yield a namespace proof for every consecutive block (some of the proofs may be for an empty namespace). If the stream goes down the client will need to reconnect by supplying the block number where they left off. If the stream ever skips a block number, you can just close it and reconnect, or failover, that would basically be a buggy or malicious server

sequencer/api/availability.toml
jbearer: Aren't we already getting the proof though? So this should be strictly faster than what we're currently doing, which is fetching the namespace w/ proof one by one. And ultimately we will need to get the proof when we switch to single query node so it kind of just needs to be fast enough lol.

Also, I thought we were using the multi query node just to get the finalized header, but then using a single query node to get the namespace and checking the proof relative to that header?

sequencer/api/availability.toml
Sneh1999: okay I see, makes sense!

sequencer/api/availability.toml
Sneh1999: > Also, I thought we were using the multi query node just to get the finalized header, but then using a single query node to get the namespace and checking the proof relative to that header?

Now that I read the comment, I think we should have done that but thats not how it works. We just do majority on every data we fetch,  we basically check majority agrees on everything 

> Aren't we already getting the proof though? So this should be strictly faster than what we're currently doing, which is fetching the namespace w/ proof one by one. And ultimately we will need to get the proof when we switch to single query node so it kind of just needs to be fast enough lol.

We had shifted to using the explorer [endpoint](https://github.com/EspressoSystems/espresso-network/blob/main/sdks/go/client/client.go#L86) for Molten, I thought the reason for using that explorer endpoint was because it doesnt give us back proofs and so is faster

PR #3685, opened 2025-10-22
test DA committees upgrade

pls148:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/testing/tests/test_da_committees.rs
+++ b/crates/hotshot/testing/tests/test_da_committees.rs
@@ -21,7 +21,7 @@ use hotshot_types::ValidatorConfig;
 use vbs::version::Version;
 
 cross_tests!(
-    TestName: test_da_committees,
+    TestName: test_da_committees_downhalf,
     Impls: [MemoryImpl],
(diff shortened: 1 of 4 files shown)


crates/hotshot/testing/tests/test_da_committees.rs
```diff
@@ -0,0 +1,94 @@
...
+                3,
+                U256::from(1),
+                true,
+            ),
+        ];
+
+        metadata.test_config.epoch_height = 50;
+        metadata.test_config.da_committees.push(hotshot_types::VersionedDaCommittee {
+            start_version: Version{major: 0, minor: 4},
+            start_epoch: 4,
```
ss-es: I think to speed things up/get better coverage, could we:

- change `start_epoch` to 2
- have 3 tests, one that e.g. cuts the existing da committee by 50%, another that expands it by 50% and a third that swaps out 50% of the existing da committee with new nodes. and then require e.g. 200 successful views?

ss-es: I think this would be a lot more useful than the double da committee change in a single test

PR #3687, opened 2025-10-24
remove mutable reference from the select_active_validator_set, add checks in apply_event()

imabdulbasit:
Removes mutability select_active_validator_set, add checks in apply_event to prevent inconsistent state, and add a test

---
  - https://app.asana.com/0/0/1211736865530853


types/src/v0/impls/stake_table.rs
```diff
@@ -240,7 +240,8 @@ impl StakeTableState {
         self.validators
     }
 
-    pub fn apply_event(&mut self, event: StakeTableEvent) -> ApplyEventResult<()> {
+    pub fn apply_event(&self, event: StakeTableEvent) -> ApplyEventResult<Self> {
+        let mut new_state = self.clone();
```
imabdulbasit: I don't know if I can remove this clone because we apply the events in the loop

jbearer: Hmm I worry that cloning the state for every single event could really hurt performance, if the state is large. I actually think it's fine for this function to be mutable in the first place, as long as it takes the state from one consistent state to the next. The change you've made to do all the validation first (e.g. `contains_key` before `insert`) seems sufficient to ensure that the state is not modified if we fail, making the clone here unnececessary

sveitser: I agree that cloning each time here will likely end up being an issue on mainnet with enough delegators.

Let's add a comment on top that the function must not modify self if the event is invalid and add some tests to check that it behaves correctly plus a simple regression test that hits the issue we're fixing.

Also do we need this method to be `pub`? It feels like it should be possible for it to be private.

types/src/v0/impls/stake_table.rs
```diff
@@ -410,28 +418,34 @@ impl StakeTableState {
...
-                let validator = self
-                    .validators
-                    .get_mut(&account)
-                    .ok_or(StakeTableError::ValidatorNotFound(account))?;
-
                 let stake_table_key: BLSPubKey = blsVK.into();
                 let state_ver_key: SchnorrPubKey = schnorrVK.into();
 
-                if !self.used_bls_keys.insert(stake_table_key) {
+                if !self.validators.contains_key(&account) {
```
sveitser, line 467: @imabdulbasit I think on line 400 above we may abort but already have modified `val.stake`.

```rust
                let val = self
                    .validators
                    .get_mut(&validator)
                    .ok_or(StakeTableError::ValidatorNotFound(validator))?;

                val.stake = val
                    .stake
                    .checked_sub(amount)
                    .ok_or(StakeTableError::InsufficientStake)?;

                let delegator_stake = val
                    .delegators
                    .get_mut(&delegator)
                    .ok_or(StakeTableError::DelegatorNotFound(delegator))?;

                *delegator_stake = delegator_stake
                    .checked_sub(amount)
                    .ok_or(StakeTableError::InsufficientStake)?;
```

imabdulbasit: thanks for catching that

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/3687/commits/6f73ab0e356eed8f653979d33c99c8b3a15ea5d7

sveitser:
I don't know if it's better but now we have to scrutinize every match arm. We could instead have 2 functions. The first one for validation and the second one for application that is infallible, for example: `validate` and `apply_unchecked`.

types/src/v0/impls/stake_table.rs
```diff
@@ -410,28 +426,34 @@ impl StakeTableState {
...
                     )));
                 }
 
+                self.used_bls_keys.insert(stake_table_key);
+                self.used_schnorr_keys.insert(state_ver_key.clone());
+
+                let validator = self
+                    .validators
+                    .get_mut(&account)
+                    .ok_or(StakeTableError::ValidatorNotFound(account))?;
```
sveitser: @imabdulbasit this should be a panic now. It can't really happen because we check earlier but we need to make it clear we won't return an error here.

types/src/v0/impls/stake_table.rs
```diff
@@ -450,28 +472,34 @@ impl StakeTableState {
...
                     ));
                 }
 
+                self.used_bls_keys.insert(stake_table_key);
+                self.used_schnorr_keys.insert(state_ver_key.clone());
+
+                let validator = self
+                    .validators
+                    .get_mut(&account)
+                    .ok_or(StakeTableError::ValidatorNotFound(account))?;
```
sveitser: same here

types/src/v0/impls/stake_table.rs
sveitser: Or there's probably a cleaner way to do it like taking out the reference earlier.

types/src/v0/impls/stake_table.rs
```diff
@@ -528,42 +578,51 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
+                tracing::info!("Validator {address:?} does not have any delegator");
+                return false;
+            }
 
-        if validator.stake.is_zero() {
-            tracing::info!("Validator {address:?} does not have any stake");
-            return false;
-        }
+            if validator.stake.is_zero() {
+                tracing::info!("Validator {address:?} does not have any stake");
```
sveitser, line 611: Should this be a warning or error instead? I think it would always be a bug because we remove the validator if the stake is zero.

types/src/v0/impls/stake_table.rs
```diff
@@ -528,42 +578,51 @@ pub fn validators_from_l1_events<I: Iterator<Item = StakeTableEvent>>(
...
 
-    let maximum_stake = validators.values().map(|v| v.stake).max().ok_or_else(|| {
-        tracing::error!("Could not compute maximum stake from filtered validators");
-        StakeTableError::MissingMaximumStake
-    })?;
+    let maximum_stake = valid_validators
+        .values()
+        .map(|v| v.stake)
+        .max()
+        .ok_or_else(|| {
```
sveitser: this should be impossible with the non-empty check above

types/src/v0/impls/stake_table.rs
imabdulbasit: I always thought we can register a validator with no delegators? Is that not right?

types/src/v0/impls/stake_table.rs
```diff
@@ -384,22 +407,41 @@ impl StakeTableState {
...
-                    .checked_sub(amount)
-                    .ok_or(StakeTableError::InsufficientStake)?;
+                if *delegator_stake < amount {
+                    tracing::warn!("delegator_stake={delegator_stake}, amount={amount}");
+                    return Err(StakeTableError::InsufficientStake);
+                }
+
+                let new_delegator_stake =
+                    delegator_stake.checked_sub(amount).unwrap_or_else(|| {
+                        panic!(
```
sveitser: shouldn't these just unwrap since we check just before?

types/src/v0/impls/stake_table.rs
imabdulbasit: do you mean the delegators check takes care of it above?

types/src/v0/impls/stake_table.rs
imabdulbasit: tbh i dont think it matters

sveitser: It's just easier to read IMO. The way it's currently written I first expected it to be something that we might expect to occur in some cases.

types/src/v0/impls/stake_table.rs
sveitser: Ah yes we can do that.

sveitser: But we remove the validator if it has zero stake. Actually I need to review this more carefully.

imabdulbasit: No we dont remove it in apply_event()

sveitser: ah yes, only the delegators are removed

PR #3691, opened 2025-10-27
feat(VID): new AvidM implementation on GF2 field

mrain:
Closes #<ISSUE_NUMBER>

### This PR:
Implement the super fast AvidM scheme over GF2 field, with `reed-solomon-simd` backend.

Some benchmark number:

For 5MB payload, total stake weight 100, recovery threshold 34

| Field  | Dispersal time | Recovery time |
| ------------- | ------------- | ------------ |
| Bn254  | 26.736 ms  | 1.609 s (quadratic naive impl) | 
| GF2  | 26.852 ms | 13.693 ms |

### DIFF
--- a/vid/src/avidm_gf2.rs
+++ b/vid/src/avidm_gf2.rs
@@ -3,10 +3,12 @@
 use std::{ops::Range, vec};
 
 use anyhow::anyhow;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use jf_merkle_tree::{hasher::HasherNode, MerkleTreeScheme};
 use jf_utils::canonical;
 use serde::{Deserialize, Serialize};
 use sha2::Digest;
+use tagged_base64::tagged;
 
 use crate::{VidError, VidResult, VidScheme};
 
@@ -72,7 +74,20 @@ impl AvidmGf2Share {
 }
 
 /// VID Commitment type
-#[derive(Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Hash,
+    CanonicalSerialize,
+    CanonicalDeserialize,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+)]
+#[tagged("AvidmGf2Commit")]
+#[repr(C)]
 pub struct AvidmGf2Commit {
     /// VID commitment is the Merkle tree root
     pub commit: MerkleCommit,


vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+
+use crate::{VidError, VidResult, VidScheme};
+
+/// Namespaced AvidMGF2 scheme
+pub mod namespaced;
+/// Namespace proofs for AvidMGF2 scheme
+pub mod proofs;
+
+/// Merkle tree scheme used in the VID
+pub type MerkleTree =
```
alxiong: nit: do we need to declare pub type here?

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+    jf_merkle_tree::hasher::HasherMerkleTree<sha3::Keccak256, HasherNode<sha3::Keccak256>>;
+type MerkleProof = <MerkleTree as MerkleTreeScheme>::MembershipProof;
+type MerkleCommit = <MerkleTree as MerkleTreeScheme>::Commitment;
+
+/// Dummy struct for AVID-M scheme over GF2
+pub struct AvidMGF2Scheme;
+
+/// VID Parameters
+#[derive(Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AvidMGF2Param {
```
alxiong: should we also allow configurable RS rate? (but default to 2 or 3)

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+            recovery_threshold,
+        })
+    }
+}
+
+/// Share type to be distributed among the parties.
+#[derive(Clone, Debug, Hash, Serialize, Deserialize, Eq, PartialEq)]
+pub struct RawAvidMGF2Share {
+    /// Range of this share in the encoded payload.
+    range: Range<usize>,
```
alxiong: 1. "Share type to be distributed among the parties." is misleading? i think people should distribute `AvidmGf2Share`? 
2. should `range` be a metadata on the `Share` instead of the `RawShare` here?

vid/src/avidm_gf2.rs
alxiong, line 1: nit-pick on naming, since we rename to `avidm` the direct camelcase translation should be `Avidm` (editor tools will auto translate and smart-map `avid_m` to `AvidM` and `avidm` to `Avidm`)

also I find `AvidmGf2` to be friendlier to eyes.

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+        let original_count = param.recovery_threshold;
+        let recovery_count = param.total_weights - param.recovery_threshold;
+        // Appending an 1u8 to the end of the payload as an implicit commit to the payload byte length.
+        let mut shard_bytes = (payload.len() + 1).div_ceil(original_count);
+        if shard_bytes % 2 == 1 {
+            shard_bytes += 1;
+        }
+        let mut payload = payload.to_vec();
+        payload.push(1u8);
+        payload.resize(shard_bytes * original_count, 0u8);
```
alxiong: define a `bit_padding()` (as defined in [wiki](https://en.wikipedia.org/wiki/Padding_(cryptography)#Bit_padding)) helper func instead.

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+        let original = payload
+            .chunks(shard_bytes)
+            .map(|chunk| chunk.to_owned())
+            .collect::<Vec<_>>();
+        let recovery = reed_solomon_simd::encode(original_count, recovery_count, &original)?;
+
+        let shares = [original, recovery].concat();
+        let share_digests: Vec<_> = shares
+            .iter()
+            .map(|share| HasherNode::from(sha3::Keccak256::digest(share)))
```
alxiong, line 141: use `map().chain()` to avoid creating a intermediate `shares` which allocate more mem?

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+    ) -> VidResult<crate::VerificationResult> {
+        if share.range.end > param.total_weights
+            || share.range.len() != share.payload.len()
+            || share.range.len() != share.mt_proofs.len()
+        {
+            return Err(VidError::InvalidShare);
+        }
+        for (i, index) in share.range.clone().enumerate() {
+            let payload_digest = HasherNode::from(sha3::Keccak256::digest(&share.payload[i]));
+            if MerkleTree::verify(
```
alxiong: can we add TODO to switch to `MerkleMultiProofs`, this feel like a low-handing fruits considering the consecutive leaves ranges.

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+        let mut original_shares = vec![None; original_count];
+        for share in shares {
+            let share = &share.content;
+            if share.range.len() != share.payload.len()
+                || share.payload.iter().any(|p| p.len() != shard_bytes)
+            {
+                return Err(VidError::InvalidShare);
+            }
+            for (i, index) in share.range.clone().enumerate() {
+                if index < original_count {
```
alxiong, line 245: nit: can we turn this as `AvidmGf2Share.is_recovery_share()`?

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+        }
+        let result = decoder.decode()?;
+        original_shares
+            .iter_mut()
+            .enumerate()
+            .for_each(|(i, share)| {
+                if share.is_none() {
+                    *share = result.restored_original(i);
+                }
+            });
```
alxiong, line 261: i'm confused, why not ditch the partial `original_shares`, just directly use `let recover = result.restored_original`?

vid/src/avidm_gf2.rs
```diff
@@ -0,0 +1,337 @@
...
+pub mod tests {
+    use rand::{seq::SliceRandom, RngCore};
+
+    use super::AvidMGF2Scheme;
+    use crate::VidScheme;
+
+    #[test]
+    fn round_trip() {
+        // play with these items
+        let params_list = [4, 9, 16];
```
alxiong: unclear naming.

vid/src/avidm_gf2/proofs.rs
alxiong, line 1: impl in this file feels like boilerplate that shares identical logic with `avidm/proofs.rs`? is it worthwhile to macro it in your opinion? (ask AI to do it)

vid/src/avidm_gf2/namespaced.rs
alxiong, line 1: similarly here. maybe we can have generic blanket impl proofs.rs and namespaced.rs over the VidScheme impl?

alxiong (commented):
good job! nice init benchmark numbers!

vid/src/avidm_gf2.rs
mrain: rate is configured by the `recovery_threshold` and `total_weights`.

vid/src/avidm_gf2.rs
mrain: Now I'm thinking about removing the `index` in `AvidMGF2Share`, not useful anywhere.

vid/src/avidm_gf2.rs
mrain: this `shares` will be returned later.

vid/src/avidm_gf2.rs
mrain: underlying library problem, there's a `restored_original_iter()` but it'll return nothing if we already have all the original shards.

vid/src/avidm_gf2.rs
mrain: `AvidMGF2Share` struct doesn't have the metadata to determine it. I think it's fine b/c this logic is not exposed to the user.

vid/src/avidm_gf2.rs
mrain: [e1f577a](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/e1f577ab85e5e3b49304c3eadc8d14210f2a8704)

vid/src/avidm_gf2.rs
mrain: make it `pub(crate) in [e1f577a](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/e1f577ab85e5e3b49304c3eadc8d14210f2a8704)

vid/src/avidm_gf2.rs
mrain: changed in [e1f577a](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/e1f577ab85e5e3b49304c3eadc8d14210f2a8704)

vid/src/avidm_gf2/namespaced.rs
mrain: maybe not here b/c I just nuked the `index` field, and don't want to touch the deployed code

vid/src/avidm_gf2.rs
mrain: added in [e1f577a](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/e1f577ab85e5e3b49304c3eadc8d14210f2a8704)

vid/src/avidm_gf2.rs
mrain: I just nuked `index` field [82d848b](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/82d848b27cf820db03e532c056b00fb891a5453b)

vid/src/avidm_gf2/proofs.rs
mrain: Also hesitating about this, don't want to change the `avidm` part but things for `avidm_gf2` is subject to change

vid/src/avidm_gf2.rs
mrain: renamed [511e5c9](https://github.com/EspressoSystems/espresso-network/pull/3691/commits/511e5c9e20f2ac9113ccbea77116424ce01341c8)
Don't want to change the `AvidM` part as I'm not sure whether it'll break some json serialization (is it case sensitive?)

alxiong (approved):

PR #3700, opened 2025-10-29
DA stake table endpoint

pls148:
Adds an endpoint for da-stake-table/:epoch and da-stake-table/current

### DIFF
--- a/crates/hotshot/testing/tests/test_da_committees.rs
+++ b/crates/hotshot/testing/tests/test_da_committees.rs
@@ -37,7 +37,7 @@ cross_tests!(
             ..TestDescription::default()
         };
 
-        let node_configs = vec![
+        let node_configs = [
             ValidatorConfig::generated_from_seed_indexed(
                 [0u8; 32],
                 0,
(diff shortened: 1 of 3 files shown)


sequencer/src/api.rs
```diff
@@ -280,6 +295,49 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence>
...
+            .await
+            .cur_epoch()
+            .await
+            .map(|e| e + 1);
+        if epoch > highest_epoch {
+            return Err(anyhow::anyhow!(
+                "requested DA stake table for epoch {epoch:?} is beyond the current epoch + 1 \
+                 {highest_epoch:?}"
+            ));
+        }
```
ss-es: for the DA stake table by epoch, I think this is not necessary. we can just query for the future DA stake table (and actually, it would be useful to not block requests far in the future, since we can use this to see that the upgrade took effect and the DA change was correctly applied)

sequencer/src/api.rs
```diff
@@ -280,6 +295,49 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence>
...
+        let mem = self
+            .consensus()
+            .await
+            .read()
+            .await
+            .membership_coordinator
+            .stake_table_for_epoch(epoch)
+            .await?;
```
ss-es: this is not necessary, we don't need to trigger catchup and should just query the membership coordinator directly

PR #3708, opened 2025-10-31
Go SDK: support streaming & decoding blocks

QuentinI:
Closes #<ISSUE_NUMBER>

### This PR:
- Makes ws multi-stream Next method parallel
- Adds support for streaming block payloads
- Adds support for decoding block payloads through rust lib

### Key places to review:

---
  - https://app.asana.com/0/0/1211836510432043

### DIFF
--- a/sdks/crypto-helper/src/payload.rs
+++ b/sdks/crypto-helper/src/payload.rs
@@ -58,11 +58,12 @@ impl DecodingResult {
 }
 
 #[no_mangle]
-// Decode a payload from payload bytes and namespace table bytes.
-//
-// Safety:
-// payload_ptr and ns_table_ptr must be valid pointers to initialized slices,
-// valid for the duration of the call. It is okay to pass null pointers if the length is zero.
+/// Decode a payload from payload bytes and namespace table bytes.
+///
+/// # Safety
+///
+/// payload_ptr and ns_table_ptr must be valid pointers to initialized slices,
+/// valid for the duration of the call. It is okay to pass null pointers if the length is zero.
 pub unsafe extern "C" fn decode_payload(
     mut payload_ptr: *const u8,
     payload_len: usize,


sdks/crypto-helper/src/payload.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Caller is responsible for ensuring this is called only on TransactionVecResult allocated
+/// by this library.
+///
+/// # Safety
+/// Caller promised this is was allocated by us,
+#[no_mangle]
+pub unsafe extern "C" fn free_transaction_vec_result(result: DecodingResult) {
+    // Free the error string if present
+    if !result.error.is_null() {
+        let _ = CString::from_raw(result.error);
```
Ayiga, line 144: `CString`s are `null` terminated, in this case the `byte 0`. Will this be a problem at all for this?  (Assuming that this is just a collection of byte data).

sdks/crypto-helper/src/payload.rs
```diff
@@ -0,0 +1,165 @@
...
+/// This function will free:
+/// - The error string if present
+/// - The transactions array
+/// - Each transaction's payload within the array
+///
+/// Caller is responsible for ensuring this is called only on TransactionVecResult allocated
+/// by this library.
+///
+/// # Safety
+/// Caller promised this is was allocated by us,
```
shenkeyao: Nit: remove "is".

sdks/crypto-helper/src/payload.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub transactions: *mut Transaction,
+    // Number of transactions in the array
+    pub transactions_len: usize,
+    // Capacity of the transactions array (for freeing)
+    pub transactions_cap: usize,
+}
+
+impl DecodingResult {
+    fn err(msg: &str) -> DecodingResult {
+        let ptr = CString::new(msg).unwrap().into_raw();
```
shenkeyao: Will this `unwrap` cause any issue? Should we do something like `unwrap_or_else(|_| CString::new("error").unwrap())` or is it unnecessary?

sdks/go/go.sum
shenkeyao, line 1: In this file and `go.mod`, there were some version downgrades. Were them intentional?

sdks/go/types/common/types.go
```diff
@@ -484,6 +484,14 @@ func (b *BlockPayload) BuilderCommitment() (*BuilderCommitment, error) {
...
 }
 
+type PayloadQueryData struct {
+	Height 		 uint64 	   `json:"height"`
+	BlockHash    *TaggedBase64 `json:"block_hash"`
+	Hash 		 *TaggedBase64 `json:"hash"`
+	Size 		 uint64 	   `json:"size"`
+	BlockPayload *BlockPayload `json:"data"`
+}
+
```
shenkeyao, line 494: Nit: I just happended notice this typo "Sting", which wasn't introduced here, but could you update it as well? No worries if it's too much trouble.

shenkeyao (commented):
Looks good overall. Would be good if one of the code owners could also review this.

sdks/go/go.sum
QuentinI: Yes, this is what allows us not to upgrade go versions in Celo. This commit was already referenced by  optimism-espresso-integration, but wasn't in the main branch here.

sdks/crypto-helper/src/payload.rs
QuentinI: This is an impossible unwrap, the error could only be if msg contains null-bytes, which Rust guarantees it wouldn't.

sdks/crypto-helper/src/payload.rs
QuentinI: By construction it was our CString (see `DecodingResult::err()`), so we should be fine by caller's promise not to call this on anything other than our `DecodingResult`s.
Could add some defensive programming here, but considering it's only ever called in a single place and there's no reason to ever call it from anywhere else, I think it's fine.

sdks/crypto-helper/src/payload.rs
QuentinI: (the reason the return type is Result in the first place is that CString::new() accepts anything that can be converted into Vec<u8>, which of course includes things that are allowed to contain null-bytes)

Ayiga: > This is an impossible unwrap, the error could only be if msg contains null-bytes, which Rust guarantees it wouldn't.

I don't think Rust has this guarantee.  Looking at the documentation for [std::ffi::CString](https://doc.rust-lang.org/std/ffi/index.html) it specifies the following:

> From Rust to C: [CString](https://doc.rust-lang.org/std/ffi/struct.CString.html) represents an owned, C-friendly string: it is nul-terminated, and has no internal nul characters. Rust code can create a [CString](https://doc.rust-lang.org/std/ffi/struct.CString.html) out of a normal string (provided that the string doesn’t have nul characters in the middle), and then use a variety of methods to obtain a raw *mut [u8](https://doc.rust-lang.org/std/primitive.u8.html) that can then be passed as an argument to functions which use the C conventions for strings.

The text "(provided that the string doesn’t have nul characters in the middle)" seems to imply that it's perfectly valid for `Rust` to have `nul` bytes in the middle of rust strings.

Based on my reading, if you provide it a `string` that contains a `nul` character in the middle, that string would be prematurely terminated.  Based on my testing, this would result in an error.

[Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3a6bb180cceaac4a552a31bad037ac33)

QuentinI: My bad! TIL \0 is a valid UTF-8 character.

QuentinI: Fixed, thank you for the catch! I was very confidently incorrect 😅

PR #3709, opened 2025-10-31
CDN -> TCP

rob-maron:
### This PR:

- Changes the CDN connection to be over TCP (client side)
- Changes the CDN to use both protocols on the same port (UDP/QUIC & TCP)

The backwards compatibility was tested against a running network


Sabrina902-prog:
Ok

Sabrina902-prog:
Ok

PR #3712, opened 2025-11-03
expose staking-cli registrations and delegations

imabdulbasit:
I need access to the registrations and delegations that the staking_cli::create() method internally creates, so they can be used in the staking-ui-service tests.

### DIFF
--- a/staking-cli/src/demo.rs
+++ b/staking-cli/src/demo.rs
@@ -89,6 +89,11 @@ impl fmt::Display for DelegationConfig {
     }
 }
 
+/// Validator registration info used by staking UI service tests.
+///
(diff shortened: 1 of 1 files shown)


staking-cli/src/demo.rs
```diff
@@ -309,6 +323,54 @@ impl<P: Provider + Clone> StakingTransactions<P> {
         let pending = self.processor.send_next(tx).await?;
         Ok(Some(pending.assert_success().await?))
     }
+
+    /// Get the list of pending registrations
```
sveitser: Since the consumer code is not in this repo (at least for now) please mention why this code is added, including where, how and why it's being used.

PR #3714, opened 2025-11-03
CI: fix slow test with rust 1.91

sveitser:
- Create two new crates: dev node, slow tests. The advantage of having crates is that we can filter with `cargo -p` and `nextest --exclude` (applies _before_ nextest compile everything). This reduces compilation time (from about 40 to about 20 minutes) for the slow tests job and decreases memory usage during compilation. I'm aware that this might not be the best grouping for the crates but I also don't want to spend too much time on it right now. If anyone has a good suggestion on how to organize this instead I would prefer to do that as a follow up PR.
- Disable `CARGO_INCREMENTAL` for all builds: a bit faster compile, less memory usage, less disk usage. If the rust-cache is used it is already disabled but since we sometimes change which builds use caches I think it's best to set explicitly. We disable the cache for many jobs because github only allows 10GB of cache and IIUC trims that every day.
- Re-enable swap for the slow-test build.
- Split slowtest CI jobs with matrix
- Be a bit more consistent about ordering of steps in CI.
- Factor out some common code into actions: installation of process-compose, free-ing up space quickly.
- There's something weird about the latest version of process-compose so we pin the one before (that's also currently used in the nix shell and in the CI on `main`, where we install it with nix profile).

A lot of code is moved (all the slow tests) best looked at locally (not on github) with

```
git diff --patience -C -M --find-copies-harder --ignore-all-space --color-moved=plain --color-moved-ws=ignore-all-space origin/main
```

It will use different colors for code that is moved.

It works pretty well but for the dev-node it somehow does not work that well.

To check the main function is fine (and only tests were removed and some types moved):
```
diff --color=always -w espresso-dev-node/src/main.rs <(git show origin/main:sequencer/src/bin/espresso-dev-node.rs)|less
```
To check tests are okay (and only the main function is missing):
```
diff --color=always -w espresso-dev-node/tests/dev_node_tests.rs <(git show origin/main:sequencer/src/bin/espresso-dev-node.rs)|less
```

Should show that nothing significant changed in the test module.

To check that no tests were removed, the quick way is to run `rg 'fn slow_.*test'`

on HEAD

```
$ rg 'fn slow_.*test'|wc -l
> 24
$ rg 'fn slow_.*test'|sort
espresso-dev-node/tests/dev_node_tests.rs:async fn slow_dev_node_multiple_lc_providers_test() {
espresso-dev-node/tests/dev_node_tests.rs:async fn slow_dev_node_test(
slow-tests/tests/api_tests.rs:async fn slow_test_merklized_state_api() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_1_da_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_1_da_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_1_regular_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_1_regular_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_minus_1_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_minus_1_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_plus_1_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_plus_1_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_2f_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_all_da_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_all_da_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_all_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_all_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_minus_1_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_minus_1_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_plus_1_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_plus_1_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_with_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_f_without_cdn() {
slow-tests/tests/restart_tests.rs:async fn slow_test_restart_staggered() {
```
on `main`:
```
$ rg 'fn slow_.*test'|wc -l
> 24
$ rg 'fn slow_.*test'|sort
sequencer/src/api.rs:    async fn slow_test_merklized_state_api() {
sequencer/src/bin/espresso-dev-node.rs:    async fn slow_dev_node_multiple_lc_providers_test() {
sequencer/src/bin/espresso-dev-node.rs:    async fn slow_dev_node_test(
sequencer/src/restart_tests.rs:async fn slow_test_restart_1_da_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_1_da_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_1_regular_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_1_regular_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_minus_1_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_minus_1_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_plus_1_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_plus_1_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_2f_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_all_da_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_all_da_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_all_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_all_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_minus_1_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_minus_1_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_plus_1_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_plus_1_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_with_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_f_without_cdn() {
sequencer/src/restart_tests.rs:async fn slow_test_restart_staggered() {
```

For the new CI runs:

https://github.com/EspressoSystems/espresso-network/actions/runs/19114607075/job/54620301757?pr=3714
https://github.com/EspressoSystems/espresso-network/actions/runs/19114607075/job/54620301767?pr=3714

On `main` due to the `slow_` filter in nextest a few tests were wrongly put into the slow tests:
```
        PASS [   4.634s] (20/23) staking-cli demo::test::test_setup_with_slow_blocks::case_1_equal_amounts
        PASS [   4.674s] (21/23) staking-cli demo::test::test_setup_with_slow_blocks::case_4_no_self_delegation
        PASS [   4.686s] (22/23) staking-cli demo::test::test_setup_with_slow_blocks::case_2_variable_amounts
        PASS [   4.831s] (23/23) staking-cli demo::test::test_setup_with_slow_blocks::case_3_multiple_delegators
```
this is fixed now because we only filter by package instead of test name.

### DIFF
--- a/sequencer/src/state.rs
+++ b/sequencer/src/state.rs
@@ -7,7 +7,8 @@ use espresso_types::{
     traits::StateCatchup,
     v0_3::{ChainConfig, RewardAccountV1, RewardMerkleTreeV1},
     v0_4::{Delta, RewardAccountV2, RewardMerkleTreeV2},
-    BlockMerkleTree, EpochVersion, FeeAccount, FeeMerkleTree, Leaf2, ValidatedState,
+    BlockMerkleTree, DrbAndHeaderUpgradeVersion, EpochVersion, FeeAccount, FeeMerkleTree, Leaf2,
+    ValidatedState,
 };
 use futures::{future::Future, StreamExt};
 use hotshot::traits::ValidatedState as HotShotState;
@@ -80,7 +81,7 @@ pub(crate) async fn compute_state_update(
         },
     }
 
-    state
+    let (state, delta, total_rewards_distributed) = state
         .apply_header(
             instance,
             peers,
@@ -89,7 +90,32 @@ pub(crate) async fn compute_state_update(
             header.version(),
             proposed_leaf.view_number(),
         )
-        .await
+        .await?;
+
+    if header.version() >= DrbAndHeaderUpgradeVersion::version() {
+        let Some(actual_total) = total_rewards_distributed else {
+            bail!(
+                "internal error! total_rewards_distributed is None for version {:?}",
+                header.version()
+            );
+        };
+
+        let Some(proposed_total) = header.total_reward_distributed() else {
+            bail!(
+                "internal error! proposed header.total_reward_distributed() is None for version \
+                 {:?}",
+                header.version()
+            );
+        };
+
+        ensure!(
+            proposed_total == actual_total,
+            "Total rewards mismatch: proposed header has {proposed_total} but actual total is \
+             {actual_total}",
+        );
+    }
+
+    Ok((state, delta))
 }
 
 async fn store_state_update(
--- a/types/src/v0/impls/state.rs
+++ b/types/src/v0/impls/state.rs
@@ -173,6 +173,11 @@ pub enum ProposalValidationError {
     NoEpochHeight,
     #[error("No First Epoch Configured")]
     NoFirstEpoch,
+    #[error("Total rewards mismatch: proposed header has {proposed} but actual is {actual}")]
+    TotalRewardsMismatch {
+        proposed: RewardAmount,
+        actual: RewardAmount,
+    },
 }
 
 impl StateDelta for Delta {}
@@ -506,10 +511,18 @@ pub(crate) struct ValidatedTransition<'a> {
     expected_chain_config: ChainConfig,
     parent: &'a Header,
     proposal: Proposal<'a>,
+    total_rewards_distributed: Option<RewardAmount>,
+    version: Version,
 }
 
 impl<'a> ValidatedTransition<'a> {
-    pub(crate) fn new(state: ValidatedState, parent: &'a Header, proposal: Proposal<'a>) -> Self {
+    pub(crate) fn new(
+        state: ValidatedState,
+        parent: &'a Header,
+        proposal: Proposal<'a>,
+        total_rewards_distributed: Option<RewardAmount>,
+        version: Version,
+    ) -> Self {
         let expected_chain_config = state
             .chain_config
             .resolve()
@@ -519,6 +532,8 @@ impl<'a> ValidatedTransition<'a> {
             expected_chain_config,
             parent,
             proposal,
+            total_rewards_distributed,
+            version,
         }
     }
 
@@ -536,6 +551,7 @@ impl<'a> ValidatedTransition<'a> {
     /// self.validate_l1_finalized()?;
     /// self.validate_l1_head()?;
     /// self.validate_namespace_table()?;
+    /// self.validate_total_rewards_distributed()?;
     /// ```
     pub(crate) fn validate(self) -> Result<Self, ProposalValidationError> {
         self.validate_timestamp()?;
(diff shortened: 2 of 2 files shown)


sveitser:
Unfortunately it's still flaky

.github/actions/install-process-compose/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    default: latest
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: |
+        set -euxo pipefail
+        if [ "${{ inputs.version }}" = "latest" ]; then
+          URL="https://github.com/F1bonacc1/process-compose/releases/latest/download/process-compose_linux_amd64.tar.gz"
```
sveitser, line 17: I don't exactly know why the "download" and version component of the URL need to be inverted, but this seems to work. And I couldn't find a consistent URL scheme that also works for both "latest" and explicit version.

imabdulbasit (commented):
Why does disabling `CARGO_INCREMENTAL` makes compilation faster?

imabdulbasit (approved):

sveitser:
> Why does disabling `CARGO_INCREMENTAL` makes compilation faster?

https://doc.rust-lang.org/cargo/reference/profiles.html?highlight=incremental#incremental

It's doing extra work and writing extra files such that **re-**compilation of crates is faster. On CI we don't recompile any crates. The cached dependency crates are not recompiled. The workspace crates are not cached.

PR #3720, opened 2025-11-04
Fix Tests and Make Them Pass without Retries

bfish713:
### This PR:

- Makes the test membership fetcher check storage before asking other nodes
- Reduces the timeout for requesting leafs from other nodes, since the test is local if a node has a leaf it should return it very quickly
- Fixes the DA logic where we determine if the DA proposal should have 2 VID commitments.  Previously we might consider a block in transition when we shouldn't
- Sets a default timeout for tests which many test use
- Adjust timeouts in slower tests to not be flaky
- Disables retrying for hotshot tests
- Clears receivers in `new_from_channels` so restarted nodes don't get events from before it was restarted.
- Reduce the number of nodes in some tests from 20-14, because 20 nodes all running in a single or couple threads are very slow at epoch transition when a lot has to happen.

The gist of the PR is that it fixes a few issues that were causing some tests to fail often, then adjust timeouts to make tests which otherwise were working fine to complete successfully.

### This PR does not:

Remove any tests or change the behavior of tests.

### Key places to review:

The main places to review are the changes in da.rs, and fetcher.rs.  Also skimming the tests to make sure nothing material was changed, just timing settings, and changing the number of nodes.

### DIFF
--- a/crates/hotshot/types/src/drb.rs
+++ b/crates/hotshot/types/src/drb.rs
@@ -105,7 +105,7 @@ pub async fn compute_drb_result(
     store_drb_progress: StoreDrbProgressFn,
     load_drb_progress: LoadDrbProgressFn,
 ) -> DrbResult {
-    tracing::warn!("Beginning DRB calculation with input {}", drb_input);
+    tracing::warn!("Beginning DRB calculation with input {:?}", drb_input);
     let mut drb_input = drb_input;
 
     if let Ok(loaded_drb_input) = load_drb_progress(drb_input.epoch).await {


crates/hotshot/hotshot/src/lib.rs
```diff
@@ -266,8 +266,18 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
...
         let instance_state = initializer.instance_state;
 
-        let (internal_tx, mut internal_rx) = internal_channel;
-        let (mut external_tx, mut external_rx) = external_channel;
+        let (internal_tx, internal_rx) = internal_channel;
+        let (mut external_tx, external_rx) = external_channel;
+
+        let mut internal_rx = internal_rx.new_receiver();
+
+        let mut external_rx = external_rx.new_receiver();
```
ss-es, line 274: why do we do this? we were handed a receiver for both channels (that we own)

if it's to ensure that we start consensus with empty channels, wouldn't it be better to do this at the call site (i.e. pass a fresh `new_receiver()` into `new_from_channels()`)?

crates/hotshot/hotshot/src/lib.rs
```diff
@@ -391,13 +395,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
...
                 .await;
+            // If we already have an epoch root, we can trigger catchup for the epoch
+            // which that root applies to.
+            if let Some(high_qc_block_number) = high_qc_block_number {
+                if is_ge_epoch_root(high_qc_block_number, config.epoch_height) {
+                    let _ = membership_coordinator
+                        .stake_table_for_epoch(Some(epoch + 2))
+                        .await;
+                }
+            }
```
ss-es, line 418: was this needed for a test to pass?

I don't think it's harmful to attempt catchup on startup so not objecting, just wondering. I feel like we would immediately try to do this catchup as soon as as consensus resumes (when we transition into `epoch+1`) anyway

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -535,6 +539,7 @@ fn spawn_catchup<T: NodeType>(
     epoch: T::Epoch,
     epoch_tx: Sender<Result<EpochMembership<T>>>,
 ) {
+    tracing::error!("Spawning catchup for epoch {epoch}");
```
ss-es: I think this should be `warn`

also, can we move it to just before the (2) `spawn_catchup` calls? that way we can add more information for `Spawning stake table catchup for epoch` vs. `Spawning randomized stake table catchup for epoch`

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -313,6 +316,7 @@ where
         .await
         {
             Ok(drb_result) => {
+                tracing::error!("Adding DRB result for epoch {epoch:?}");
```
ss-es: I think this should should be `warn`

it would be nice to be a little more descriptive about the path here, e.g. `Retrieved DRB result for epoch {} from peers. Updating membership.`

optionally, it might be worth also logging the `drb_result` here. it's not very big and it could help us catch e.g. off-by-one errors with the epoch numbers used in catchup. I think you can do something like e.g. `tracing::warn!(?drb_result, ...)`

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -268,6 +268,8 @@ where
                 }
             };
         }
+        let epochs = fetch_epochs.iter().map(|(e, _)| e).collect::<Vec<_>>();
+        tracing::error!("Fetching epochs: {epochs:?}");
```
ss-es:
```suggestion
        tracing::warn!("Fetching stake tables for epochs: {epochs:?}");
```

crates/hotshot/types/src/drb.rs
```diff
@@ -205,6 +206,8 @@ pub async fn compute_drb_result(
         difficulty_level: drb_input.difficulty_level,
     };
 
+    tracing::error!("DRB calculated for epoch {}", drb_input.epoch);
```
ss-es:
```suggestion
    tracing::warn!("Completed DRB calculation. Result: {:?}", final_drb_input);
```

crates/hotshot/types/src/drb.rs
```diff
@@ -105,6 +105,7 @@ pub async fn compute_drb_result(
     store_drb_progress: StoreDrbProgressFn,
     load_drb_progress: LoadDrbProgressFn,
 ) -> DrbResult {
+    tracing::error!("Computing DRB result for epoch {}", drb_input.epoch);
```
ss-es:
```suggestion
    tracing::warn!("Beginning DRB calculation with input {}", drb_input);
```

crates/hotshot/testing/tests/test_epoch_unequal_stake.rs
```diff
@@ -67,11 +68,13 @@ cross_tests!(
...
+        let views = (3..50).collect::<Vec<_>>();
+
         // We're going to have a lot of view failures of course, but with equal stake we should stop making progress at view 3
-        metadata.overall_safety_properties.num_successful_views = 10;
+        metadata.overall_safety_properties.num_successful_views = 20;
+        metadata.overall_safety_properties.decide_timeout = Duration::from_secs(60);
         metadata.overall_safety_properties.expected_view_failures = vec![5];
-        metadata.overall_safety_properties.possible_view_failures = vec![4, 6, 11, 12, 13, 14, 15];
-        metadata.overall_safety_properties.decide_timeout = Duration::from_secs(30);
+        metadata.overall_safety_properties.possible_view_failures = views;
```
ss-es, line 77: is this test really so flaky that we can't say which views from 3-50 will fail but it could be all of them?

crates/hotshot/testing/tests/test_epoch_catchup.rs
```diff
@@ -163,8 +142,7 @@ cross_tests!(
             );
         metadata.overall_safety_properties = OverallSafetyPropertiesDescription {
             num_successful_views: 50,
-            possible_view_failures: vec![33, 34, 39, 40],
-            decide_timeout: Duration::from_secs(15),
+            possible_view_failures: vec![11,12,13,33,34,39, 40, 41, 42, 43],
```
ss-es, line 145: I'm a bit confused about the view failures here (both before and after) -- I'm guessing this test is meant to have node 13 not start until view 35, but a) I don't see where we're stating that node 13 should be offline at the start and b) we're using `RandomizedStakeTable` here which feels weird for this kind of test

was it really failing these views sporadically? if it was failing 11, 12, 13 those should be guaranteed failures shouldn't they?

crates/hotshot/testing/tests/test_epoch_catchup.rs
```diff
@@ -60,13 +54,12 @@ cross_tests!(
...
                 TimeBasedCompletionTaskDescription {
-                    duration: Duration::from_secs(240),
+                    duration: Duration::from_secs(300),
                 },
             );
         metadata.overall_safety_properties = OverallSafetyPropertiesDescription {
             num_successful_views: 50,
-            possible_view_failures: vec![18, 19, 38, 39],
-            decide_timeout: Duration::from_secs(15),
+            possible_view_failures: vec![12, 13, 26, 27, 35, 40, 41],
```
ss-es, line 62: similarly confused here, shouldn't at least some of these be guaranteed failures? e.g. 12, 13, 26, 27 (assuming the test is somehow keeping the node offline until view 35, I'm still not sure where that's configured)

crates/hotshot/testing/src/spinning_task.rs
```diff
@@ -259,11 +258,6 @@ where
                                         node.handle.hotshot.config.epoch_height,
                                     ),
                                 ));
-                                let memberships = EpochMembershipCoordinator::new(
-                                    membership,
-                                    node.handle.hotshot.config.epoch_height,
-                                    &node.handle.storage().clone(),
-                                );
```
ss-es, line 266: just to confirm -- this isn't actually changing any logic right? because we were just constructing `memberships` but not actually using it except for `memberships.membership()` which was the `membership` constructed immediately above?

crates/hotshot/testing/src/overall_safety_task.rs
```diff
@@ -100,7 +100,7 @@ impl Default for OverallSafetyPropertiesDescription {
             transaction_threshold: 0,
             expected_view_failures: vec![],
             possible_view_failures: vec![],
-            decide_timeout: Duration::from_secs(4),
+            decide_timeout: Duration::from_secs(60),
```
ss-es, line 103: do we really want a 60s decide timeout by default? not against it, just seems surprising

crates/hotshot/task-impls/src/quorum_vote/mod.rs
```diff
@@ -748,7 +748,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> QuorumVoteTaskS
                 // Check that the signature is valid
                 ensure!(
                     sender.validate(&share.signature, payload_commitment.as_ref()),
-                    error!(
+                    warn!(
```
ss-es, line 751: this should still be `error` right?

crates/hotshot/task-impls/src/da.rs
```diff
@@ -437,17 +441,40 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
...
+                        }
                     } else {
-                        EpochTransitionIndicator::InTransition
-                    }
-                } else {
-                    EpochTransitionIndicator::NotInTransition
-                };
+                        EpochTransitionIndicator::NotInTransition
+                    };
+
```
ss-es, line 477: I still need to look at this more carefully

cc @lukaszrzasik

crates/hotshot/hotshot/src/lib.rs
bfish713: I don't think we would trigger this catchup.  This is saying if we're about to enter epoch `e` but haven't yet, then we should try to catchup to epoch `e+2`.  I don't see why we would otherwise start this catchup until the end of epoch `e+1`.    It's an unlikely edge case where we saw a QC for the epoch root but restarted before it was decided

crates/hotshot/types/src/epoch_membership.rs
bfish713: yeah I added this just for debugging and didn't remove

crates/hotshot/testing/tests/test_epoch_catchup.rs
bfish713: There should be some guaranteed failure, but I just left it how it was

crates/hotshot/testing/tests/test_epoch_catchup.rs
bfish713: It might fail all these views, but I'm not sure, I think the concept of a "failed" view in test is a bit flawed and that's why we need multiple possible views.  I could try to narrow down exactly which views should fail but it's kind of a long process and I think the idea of the test is to just see there is progress.

crates/hotshot/testing/src/overall_safety_task.rs
bfish713: Most of our test already bump this out.  It's because this is a sort of last resort to halt the test if it's hung.  Any test which has timeouts would need to set this number significantly higher.  If the timeout for a view is 5s it means that one timeout would delay the next decide by 5s + the time of the next 2 views, roughly

crates/hotshot/task-impls/src/quorum_vote/mod.rs
bfish713: I lowered this because it just shows up a lot in tests and in production.  I have an idea how to fix, but until then i'm not sure what to set this as

crates/hotshot/hotshot/src/lib.rs
bfish713: We can leave it to the caller but there really should never be a reason to be calling this function where the receiver is not empty.  The reason this function exists is mostly so the test tasks don't somehow have to get new receivers if we recreate the channels.  

Maybe it's best this function just takes the sending side of the channel and creates a receiver from that.

crates/hotshot/task-impls/src/quorum_vote/mod.rs
ss-es: yeah the VID fix should make this never show up

and I think I now agree with `warn` anyway for the kind of thing

crates/hotshot/task-impls/src/da.rs
ss-es: I think this looks right to me, I'll let Lucas also take a look though

crates/hotshot/hotshot/src/lib.rs
ss-es: I did look at this a bit more and the only place we seem to use `new_from_channels` in the spinning_task does already call `new_receiver()` on it:

`                                let context =
                                    TestRunner::<TYPES, I, V, N>::add_node_with_config_and_channels(
                                        node_id,
                                        generated_network.clone(),
                                        Arc::clone(&membership),
                                        initializer,
                                        config,
                                        validator_config,
                                        storage.clone(),
                                        internal_chan,
                                        (
                                            node.handle.external_channel_sender(),
                                            node.handle.event_stream_known_impl().new_receiver(),
                                        ),
                                    )
                                    .await;
`

so I think this change doesn't do anything right? I don't think it matters that much either way, I'm just a bit worried that clearing the receiver is somewhat unexpected if someone ends up using this in a year for some other reason

crates/hotshot/hotshot/src/lib.rs
ss-es: in the consensus task we always attempt to catchup for the next epoch on view change, so as soon as we enter epoch `e+1` we will start trying to catchup to `e+2`

crates/hotshot/types/src/epoch_membership.rs
ss-es: ah, we already log at warn right after both `spawn_catchup` calls. I just removed this line

crates/hotshot/task-impls/src/da.rs
lukaszrzasik: This looks good to me as well.

PR #3724, opened 2025-11-06
stake table: improved events for delegators

sveitser:
Problems:

- `Undelegated` and `ValidatorExit` events do not tell the consumer when the unlock happens. Getting that information requires an archive node to the original `exitEscrowPeriod` (which requires an archive node) and track events after that.
- `Withdrawal` emitted for both claiming of undelegations and validator exits, also doesn't contain enough information.

Solutions:

- Add `unlocksAt` to events.
- Emit `ValidatorExitClaimed` and `WithdrawalClaimed` instead of `Withdrawal`.
- Add a unique `undelegationId` starting at `1`. Undelegations from the V1 stake table that are claimed in V2 have Id `0`.

New events:

https://github.com/EspressoSystems/espresso-network/blob/bd2214943d4a7f3550074f69f7092a13c864f056/contracts/src/StakeTableV2.sol#L164-L202


contracts/src/StakeTableV2.sol
```diff
@@ -291,27 +305,51 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
 
         validators[validator].status = ValidatorStatus.Exited;
-        validatorExits[validator] = block.timestamp + exitEscrowPeriod;
+        uint256 unlocksAt = block.timestamp + exitEscrowPeriod;
+        validatorExits[validator] = unlocksAt;
         // in v2, the delegatedAmount is not updated until withdrawal
 
         activeStake -= validators[validator].delegatedAmount;
-        emit ValidatorExit(validator);
+        emit ValidatorExitV2(validator, unlocksAt);
```
alysiahuggins, line 421: is there a use for emitting the `validators[validator].delegatedAmount` at this point too? @Ayiga 
i can imagine a network health monitoring benefitting from this for example

Ayiga: You *could*.  But I think it might be informational rather than anything critical.

If you are rebuilding the state externally from watching the `Events` alone, you would presumably already know this value with the information provided, and the extra `amount` field would not be required.

That being said, I'm not against having extra information, if for nothing else you can use it as a verification check of your current state.

The most important information to have in the event are the explicit arguments (method parameters), and the implicit (sender if necessary).

PR #3735, opened 2025-11-11
revert to manual serialize/deserialize implementation for staketablee…

pls148:
…vents

Closes #<ISSUE_NUMBER>


imabdulbasit (commented):
I think if the serde derive implementation is able to successfully deserialize the legacy events. Then we can remove all of the legacy types, and as well as the stake table events migration

PR #3738, opened 2025-11-11
3725 governance split due to unsync default admin role and owner

alysiahuggins:
Closes #3275 

### This PR:
- removes `ownable` from `RewardClaim` and just uses `accesscontrol` as there's no need for both
- since staketable v1 is already deployed with ownable and accesscontrol is introduced in v2 to handle pauser roles and the growing governance, both owners (`Ownerable.owner()` and `AccessControl:DEFAULT_ADMIN_ROLE` had to be synchronized - thus the transferOwnership and grantRole functions were overridden to handle this)
- `Ownerable.renounceOwnership()`, `AccessControl.renounceRole()` and `AccessControl.revokeRole()` were overridden to revert to prevent any governance split and make sure that there is always an owner 

### Key places to review:
- StakeTableV2
- RewardClaim

`

---
  - https://app.asana.com/0/0/1212197384935099
  - https://app.asana.com/0/0/1212197389245827
  - https://app.asana.com/0/0/1211879674309436


contracts/test/StakeTableV2Governance.invariant.t.sol
```diff
@@ -0,0 +1,611 @@
...
+        vm.stopPrank();
+    }
+
+    /// @notice Fuzzed action: Grant PAUSER_ROLE
+    function grantPauserRole(address newPauser) public {
+        newPauser = _boundAddress(newPauser);
+
+        vm.startPrank(currentAdmin);
+
+        try proxy.grantRole(proxy.PAUSER_ROLE(), newPauser) {
```
sveitser: Do we need these try/catch? We keep counters for failed calls but we don't seem to do anything with them.

sveitser: The problem I think is currently if all calls were failing we would still pass the test but we wouldn't be testing what we want.

contracts/src/StakeTableV2.sol
```diff
@@ -300,6 +325,54 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
         _unpause();
     }
 
+    /// @notice Transfers ownership and keeps DEFAULT_ADMIN_ROLE in sync
+    /// @dev Grants the role to new owner and revokes from old owner.
+    function transferOwnership(address newOwner) public virtual override 
+    // onlyRole(DEFAULT_ADMIN_ROLE)
```
sveitser: Is this commented out on purpose? I guess the call to `super` below would also revert but it feels like keeping the modifier here would make sense.

contracts/src/StakeTableV2.sol
```diff
@@ -685,4 +770,17 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    /// @param newImplementation The address of the new implementation
+    /// @dev This function is overridden to use AccessControl instead of Ownable
+    /// Only addresses with DEFAULT_ADMIN_ROLE can authorize upgrades
+    function _authorizeUpgrade(address newImplementation)
+        internal
+        virtual
+        override
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
+        newImplementation; // Silence empty block warning
```
sveitser: Warning from sol-hint? We can add a comment instead.

sveitser (commented):
@alysiahuggins I haven't finished but some preliminary comments.

contracts/src/RewardClaim.sol
```diff
@@ -237,7 +248,12 @@ contract RewardClaim is
...
     }
 
-    function _authorizeUpgrade(address newImplementation) internal virtual override onlyOwner {
+    function _authorizeUpgrade(address newImplementation)
+        internal
+        virtual
+        override
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
         emit Upgrade(newImplementation);
```
sveitser: I guess we can remove this Upgrade event too?

contracts/src/StakeTableV2.sol
```diff
@@ -300,6 +325,54 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    {
+        if (newOwner == address(0)) {
+            revert OwnableInvalidOwner(address(0));
+        }
+        address oldOwner = owner();
+
+        super.grantRole(DEFAULT_ADMIN_ROLE, newOwner);
+        super.transferOwnership(newOwner);
+        // Only revoke if transferring to a different address to prevent self-revocation
+        if (oldOwner != newOwner) {
```
sveitser: What's the use case of calling this function with `newOwner == currentOwner`? I think this should be be documented.

alysiahuggins: documented here https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ff26c6769df86356b8dfe2cff064d7c90be0d4b5

contracts/src/RewardClaim.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ed6adc6cbda44bae727ebca6ef325a165b89b20a

contracts/src/StakeTableV2.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ed6adc6cbda44bae727ebca6ef325a165b89b20a

contracts/src/StakeTableV2.sol
alysiahuggins: re-added here https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ed6adc6cbda44bae727ebca6ef325a165b89b20a

contracts/test/StakeTableV2Governance.invariant.t.sol
alysiahuggins: that's true, i've refactored the tests https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ddb1d9b67502d06de791e08851606e01a9841761

contracts/rust/deployer/src/builder.rs
```diff
@@ -383,26 +383,44 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+                            )?,
+                            pauser: multisig_pauser,
+                            dry_run: Some(dry_run),
+                        },
                     )
                     .await?;
                 } else {
+                    // Pick admin from config. StakeTable uses OpsTimelock for faster
+                    // emergency updates since it handles critical staking ops.
+                    let admin = if let Some(use_timelock_owner) = self.use_timelock_owner {
```
alysiahuggins, line 409: since `initializeV2` takes an admin for the new `DEFAULT_ADMIN_ROLE`, the thought process is that the admin should be set as the new admin via the `upgradeToAndCall` method when `initializeV2` is called

contracts/rust/deployer/src/builder.rs
```diff
@@ -501,31 +490,38 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
-                    //   to react if they do not agree with an upgrade.
-                    tracing::info!("Transferring ownership to SafeExitTimelock");
+                // RewardClaimProxy only needs one pauser
+                // Default to deployer address if pauser not explicitly set (for local demos)
+                let deployer_addr = provider.default_signer_address();
+                let pauser = self.multisig_pauser.unwrap_or(deployer_addr);
+
+                // RewardClaim uses SafeExitTimelock (longer delay) since it can mint tokens
+                // and users need time to react to upgrades. Can be paused in emergencies.
+                let admin = if let Some(use_timelock_owner) = self.use_timelock_owner {
```
alysiahuggins, line 500: same for the reward claim contract, we know that we want it to be own by the timelock so we avoid the transfer ownership transaction by setting the timelock as the admin when handling the deployment

contracts/rust/deployer/src/lib.rs
```diff
@@ -995,7 +995,12 @@ pub async fn deploy_reward_claim_proxy(
         reward_claim_proxy.getVersion().call().await?,
         (1, 0, 0).into()
     );
-    assert_eq!(reward_claim_proxy.owner().call().await?, owner);
+    // Verify admin has DEFAULT_ADMIN_ROLE
+    let admin_role = reward_claim_proxy.DEFAULT_ADMIN_ROLE().call().await?;
```
alysiahuggins, line 999: since reward claim no longer uses ownable, we have to get the default admin role from AccessControl

contracts/rust/deployer/src/proposals/multisig.rs
```diff
@@ -587,15 +587,22 @@ pub async fn upgrade_esp_token_v2_multisig_owner(
...
     }
 
     Ok(result)
 }
 
+pub struct StakeTableV2UpgradeParams {
+    pub rpc_url: String,
+    pub multisig_address: Address,
+    pub pauser: Address,
+    pub dry_run: Option<bool>,
```
sveitser: I sometimes wonder why we use Optional types since we could just default to false this kind of flag.

contracts/src/FeeContract.sol
```diff
@@ -86,6 +87,13 @@ contract FeeContract is Initializable, OwnableUpgradeable, UUPSUpgradeable {
         emit Upgrade(newImplementation);
     }
 
+    /// @notice Cannot renounce ownership
+    /// @dev Override renounceOwnership() to revert, preventing accidental or malicious ownership
+    /// renunciation
+    function renounceOwnership() public virtual override onlyOwner {
```
sveitser, line 93: Do we currently have code to deploy a patch upgrade for the fee contract? If not, maybe a TODO/issue somewhere.

sveitser: Also feels like this should involve some kind of version bump.

contracts/src/StakeTableV2.sol
```diff
@@ -269,17 +281,28 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
-        _grantRole(DEFAULT_ADMIN_ROLE, admin);
         _grantRole(PAUSER_ROLE, pauser);
+        _grantRole(DEFAULT_ADMIN_ROLE, admin);
+
+        // Transfer ownership to admin if it's not the current owner
+        // This ensures owner() and DEFAULT_ADMIN_ROLE remain synchronized
+        // preventing governance drift.
+        address previousOwner = owner();
+        if (admin != previousOwner) {
+            _transferOwnership(admin);
```
sveitser, line 310: Why do we call `_transferOwnership` here and not `transferOwnership`?

contracts/rust/deployer/src/builder.rs
```diff
@@ -501,30 +491,38 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+                        .expect("fail to get SafeExitTimelock address");
+                    tracing::info!("Granting RewardClaim DEFAULT_ADMIN_ROLE to SafeExitTimelock");
+                    let reward_claim = RewardClaim::new(reward_claim_proxy_addr, &provider);
+                    let admin_role = reward_claim.DEFAULT_ADMIN_ROLE().call().await?;
+                    reward_claim
+                        .grantRole(admin_role, timelock_addr)
+                        .send()
+                        .await?
+                        .get_receipt()
+                        .await?;
```
sveitser: This potentially leaves the contract in a state with two owners if the second TX fails. I wonder if it's worth making some changes to the contracts so that we can enforce single "ownership" and atomically transfer "ownership". It's probably unlikely but for example what if we somehow don't realize that the second TX failed and leave the contract with two admins?

contracts/src/StakeTableV2.sol
```diff
@@ -599,15 +680,23 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
 
     /// @notice Set the minimum interval between commission updates
     /// @param newInterval The new minimum interval in seconds
-    function setMinCommissionUpdateInterval(uint256 newInterval) external virtual onlyOwner {
+    function setMinCommissionUpdateInterval(uint256 newInterval)
+        external
+        virtual
+        onlyRole(DEFAULT_ADMIN_ROLE)
```
sveitser, line 694: For my understanding. Given the changes in this PR if we kept `onlyOwner` that should be equivalent right? Since we are enforcing a single admin in this contract should we keep calling that admin owner and use `onlyOwner` for clarity?

contracts/test/StakeTableV2Governance.invariant.t.sol
```diff
@@ -0,0 +1,537 @@
...
+        bytes32 adminRole = proxy.DEFAULT_ADMIN_ROLE();
+        address currentOwner = proxy.owner();
+        address[] memory actors = handler.getActors();
+
+        uint256 adminCount = 0;
+        address adminAddress;
+
+        // Limit iteration to prevent gas issues with large actor arrays
+        uint256 maxActors = actors.length > 50 ? 50 : actors.length;
+        for (uint256 i = 0; i < maxActors; i++) {
```
sveitser: Iteration in the invariant functions really slows down tests because they run after every step. So I think it's better to keep counters that can be checked with O(1) in invariant checks. I think it's fine though, no need to change.

contracts/test/StakeTableV2Governance.invariant.t.sol
```diff
@@ -0,0 +1,537 @@
...
+    function transferOwnershipToSelf() public {
+        vm.startPrank(currentAdmin);
+        proxy.transferOwnership(currentAdmin);
+        transferOwnershipCalls++;
+        vm.stopPrank();
+    }
+
+    /// @notice Fuzzed action: Chain of transfers
+    /// @dev Should always succeed - tests multiple sequential transfers
+    function chainOfTransfers(address[] memory newOwners) public {
```
sveitser: I think `transferOwnership` already handles this. It will be called right after each other sometimes when foundry is running the test sequence.

contracts/test/StakeTableV2Governance.invariant.t.sol
```diff
@@ -0,0 +1,537 @@
...
+    /// @notice Fuzzed action: Pause contract via pauser
+    /// @dev Only pauses if not already paused (pause() reverts if already paused)
+    function pauseContract(uint256 actorIndexSeed) public {
+        if (actors.length == 0) return;
+
+        address actor = actors[actorIndexSeed % actors.length];
+        bytes32 pauserRole = proxy.PAUSER_ROLE();
+
+        // Only try if actor has pauser role and contract is not paused
+        if (!proxy.hasRole(pauserRole, actor)) return;
```
sveitser: What's the chance here that an actor is a pauser? I'm just asking because it's easy to get into situation where this almost never happens. In that case it's better to keep a separate array of pausers that we can sample from.

contracts/test/StakeTableV2Governance.invariant.t.sol
```diff
@@ -0,0 +1,537 @@
...
+        }
+        // Avoid precompiles and special addresses
+        if (uint160(addr) < 10) {
+            addr = address(uint160(addr) + 100);
+        }
+        return addr;
+    }
+
+    function _addActor(address actor) internal {
+        for (uint256 i = 0; i < actors.length; i++) {
```
sveitser, line 390: use a mapping?

contracts/rust/deployer/src/proposals/multisig.rs
sveitser: We typically at least derive `Debug` and `Clone` on types.

contracts/test/StakeTableV2Governance.t.sol
```diff
@@ -0,0 +1,562 @@
...
+contract StakeTableV2GovernanceTest is Test {
+    StakeTableUpgradeV2Test public baseTest;
+    StakeTableV2 public proxy;
+    address public initialOwner;
+    address public originalV1Owner;
+    address public pauser;
+
+    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
+    event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
+    event RoleRevoked(bytes32 indexed role, address indexed account, address indexed sender);
```
sveitser: Why not import these?

contracts/test/StakeTableV2Governance.t.sol
```diff
@@ -0,0 +1,562 @@
...
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                IAccessControl.AccessControlUnauthorizedAccount.selector, originalV1Owner, adminRole
+            )
+        );
+        proxy.updateExitEscrowPeriod(200 seconds);
+        vm.stopPrank();
+
+        vm.prank(initialOwner);
+        proxy.updateExitEscrowPeriod(200 seconds);
```
sveitser, line 99: Ensure event emitted?

contracts/src/StakeTableV2.sol
alysiahuggins: the thinking here was to phase out onlyOwner because we may want to use onlyRole for other roles in the future, didn't want it to get confusing. But yes you're right in thinking that either can be used

contracts/src/StakeTableV2.sol
alysiahuggins: since only the `DEFAULT_ADMIN_ROLE` can [call](https://github.com/EspressoSystems/espresso-network/blob/328fa39012043496acbf3f1b2c4a94ec597398d6/contracts/src/StakeTableV2.sol#L354) `transferOwnership` when it's overriden, we need to first grant the role to the new admin. However, the overriden `transferOwnership` function also calls [grantRole](https://github.com/EspressoSystems/espresso-network/blob/328fa39012043496acbf3f1b2c4a94ec597398d6/contracts/src/StakeTableV2.sol#L372), so that method would be called twice. so i opted for calling the base internal method instead so that steps aren't redundant and don't use more gas than necessary.

contracts/rust/deployer/src/proposals/multisig.rs
alysiahuggins: good catch, https://github.com/EspressoSystems/espresso-network/pull/3738/commits/aced7e3d5602f991e644130a2db32bc59337fb9b

contracts/test/StakeTableV2Governance.t.sol
alysiahuggins: sure thing https://github.com/EspressoSystems/espresso-network/pull/3738/commits/ce823a3e54c43f318461757bd9495288149e5362

contracts/test/StakeTableV2Governance.t.sol
alysiahuggins: yea no need for this, https://github.com/EspressoSystems/espresso-network/pull/3738/commits/efa596e924bf96f8aab0e9313e0ce2a817696f1c

contracts/rust/deployer/src/builder.rs
alysiahuggins: hmm the diff you're commenting on is outdated, grantRole does not happen after the deployment, if use_timelock_owner is true, the admin is set to that before the deploy. Hopefully it refreshes soon to reflect that. https://github.com/EspressoSystems/espresso-network/pull/3738/commits/29ab5dac01e2249ed1b88b0161eab777d4f15342

contracts/src/RewardClaim.sol
```diff
@@ -17,10 +15,16 @@ import "./EspTokenV2.sol";
 import "./libraries/RewardMerkleTreeVerifier.sol";
 import "./interfaces/IRewardClaim.sol";
 
+/// @title RewardClaim - Espresso Reward Claim Contract
+/// @notice Allows validators and delegators to claim ESP token rewards based on cryptographic
+/// proofs from the Espresso network.
+/// @dev Governance Architecture:
+/// This contract uses ONLY AccessControlUpgradeable.
```
sveitser, line 22: I think we should mention the customization to single admin.

contracts/src/RewardClaim.sol
```diff
@@ -237,8 +256,58 @@ contract RewardClaim is
...
+                _revokeRole(DEFAULT_ADMIN_ROLE, oldAdmin);
+            }
+        } else {
+            super.grantRole(role, account);
+        }
+    }
+
+    /// @notice Prevent renouncing DEFAULT_ADMIN_ROLE to preserve governance control
+    /// @dev Override renounceRole() to revert when attempting to renounce DEFAULT_ADMIN_ROLE,
+    /// preventing accidental or malicious admin role renunciation
```
sveitser, line 297: For better docs, I think we should match the names of variables from the BaseContract and use `@inheritdoc`. I tested that this works with forge doc. However it stops working if we add `@dev` annotations. I'm fine not fixing this in this PR since it's just for docs.

```
    /// @notice Prevent renouncing DEFAULT_ADMIN_ROLE to preserve governance control
    /// @notice Override renounceRole() to revert when attempting to renounce DEFAULT_ADMIN_ROLE,
    /// preventing accidental or malicious admin role renunciation
    ///
    /// @inheritdoc AccessControlUpgradeable
    function renounceRole(bytes32 role, address callerConfirmation) public virtual override {
        if (role == DEFAULT_ADMIN_ROLE) {
            revert DefaultAdminCannotBeRenounced();
        }
        super.renounceRole(role, callerConfirmation);
    }
```

Image

contracts/src/RewardClaim.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/3bf661b6aa9d7965ca6d880d4023f4fd9d3d9ab2

contracts/src/StakeTableV2.sol
sveitser: Thanks. Add a code comment about this?

contracts/src/StakeTableV2.sol
```diff
@@ -318,6 +341,64 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+        override
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
+        if (newOwner == address(0)) {
+            revert OwnableInvalidOwner(address(0));
+        }
+        address oldOwner = owner();
+
+        // Handle self-transfer: OpenZeppelin allows it, so we do too. Just emit the event.
+        // In super.grantRole, self-revocation is allowed but it's a no-op and doesn't change the
```
sveitser: Should this be `self-granting`?

contracts/src/RewardClaim.sol
alysiahuggins: that's pretty neat! https://github.com/EspressoSystems/espresso-network/pull/3738/commits/bf43300cd0ec466e14c18c15a89584cbb667f63b

contracts/src/StakeTableV2.sol
alysiahuggins: meaning should it have the ability to call the function even if the address is the same as the current admin? i was in two minds about it but saw that the parent library, `Ownable` allows it and `AccessControl` makes it a no-op. So continued the pattern here

contracts/test/StakeTableV2Governance.invariant.t.sol
alysiahuggins: changed to using the currentPauser https://github.com/EspressoSystems/espresso-network/pull/3738/commits/6fef9c37dfbdc953c2a97fa29a2e7c87df8f6aa2

contracts/src/StakeTableV2.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/9df35896264de0c43e550ca415ac4f58b0f37eaa

contracts/test/StakeTableV2Governance.invariant.t.sol
alysiahuggins: this implementation has since changed but i don't see that in this diff

contracts/src/RewardClaim.sol
sveitser: it's better to use the same variable names though, currently we get this

Image

contracts/test/StakeTableV2Governance.invariant.t.sol
alysiahuggins: this function has been removed (though it's not reflected in the diff on my UI)

contracts/src/RewardClaim.sol
```diff
@@ -237,8 +260,61 @@ contract RewardClaim is
...
+            }
+
+            // Grant role to new admin first, then revoke from old admin.
+            super.grantRole(role, account);
+            _currentAdmin = account;
+            if (oldAdmin != address(0)) {
+                _revokeRole(DEFAULT_ADMIN_ROLE, oldAdmin);
+            }
+        } else {
+            super.grantRole(role, account);
```
sveitser: The branching in this functions reads a bit strange/convoluted to me. All the branches call `super.grantRole` so we could pull this out to the top and get rid of the `else` branch.

contracts/src/StakeTableV2.sol
sveitser: I mean I don't understand how `grantRole` can allow "self-revocation":

>  In super.grantRole, self-revocation is allowed ...

contracts/src/StakeTableV2.sol
```diff
@@ -318,6 +345,69 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
     }
 
+    /// @notice Transfers ownership and keeps DEFAULT_ADMIN_ROLE in sync
+    /// Grants the role to new owner and revokes from old owner.
+    /// Access control is enforced by both onlyRole(DEFAULT_ADMIN_ROLE) and
+    /// super.transferOwnership() which requires onlyOwner.
+    /// This ensures that only the current admin (who holds both ownership and DEFAULT_ADMIN_ROLE)
+    /// can transfer ownership.
+    /// @inheritdoc OwnableUpgradeable
+    function transferOwnership(address newOwner)
```
sveitser, line 358: Add a comment somewhere for why these new functions aren't pausable?

contracts/src/StakeTableV2.sol
sveitser: it's fine i think

.env
```diff
@@ -116,8 +116,7 @@ ESPRESSO_STATE_SIGNATURE_TOTAL_STAKE=5
 
 # Prover service
 ESPRESSO_PROVER_SERVICE_PORT=30050
-ESPRESSO_STATE_PROVER_UPDATE_INTERVAL_SECS=60
-ESPRESSO_STATE_PROVER_ONESHOT_RETRIES=1
+ESPRESSO_STATE_PROVER_UPDATE_INTERVAL=60s
```
sveitser: This looks like an unwanted revert.

sveitser (approved):

contracts/src/RewardClaim.sol
alysiahuggins: i agree https://github.com/EspressoSystems/espresso-network/pull/3738/commits/f70f520fd05c781dc82ea95f4bc9f2c7a90c2701

contracts/src/StakeTableV2.sol
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/f70f520fd05c781dc82ea95f4bc9f2c7a90c2701

.env
alysiahuggins: https://github.com/EspressoSystems/espresso-network/pull/3738/commits/f70f520fd05c781dc82ea95f4bc9f2c7a90c2701

contracts/src/StakeTableV2.sol
alysiahuggins: ah yes https://github.com/EspressoSystems/espresso-network/pull/3738/commits/f70f520fd05c781dc82ea95f4bc9f2c7a90c2701

contracts/src/RewardClaim.sol
alysiahuggins: oh thanks, updated https://github.com/EspressoSystems/espresso-network/pull/3738/commits/f70f520fd05c781dc82ea95f4bc9f2c7a90c2701

PR #3741, opened 2025-11-11
check view number matches the finalized leaf for accounts catchup

imabdulbasit:
This adds view number validation for catchup requests when loading reward accounts and fee accounts. When a view number is provided, the functions now verify that the fetched leaf's view matches the expected view before returning the
   account data. This prevents issues where a node requests accounts for a block height with an unfinalized view number, but the block at that height has already been finalized with a different view. This prevents running a query on the state tables, because the proof will be reject anyways by the client because of view mismatch

---
  - https://app.asana.com/0/0/1211903791798745


sequencer/src/api/sql.rs
```diff
@@ -491,12 +491,25 @@ async fn load_frontier<Mode: TransactionMode>(
...
 ) -> anyhow::Result<(RewardMerkleTreeV1, Leaf2)> {
     let leaf = tx
         .get_leaf(LeafId::<SeqTypes>::from(height as usize))
         .await
         .context(format!("leaf {height} not available"))?;
+
+    // Validate that the fetched leaf's view matches the expected view if provided
+    if let Some(expected_view) = view {
+        if leaf.leaf().view_number() != expected_view {
+            bail!(
```
bfish713: Does bailing here eventually send this message as an error response, or does the requester just get a 404?

imabdulbasit: for anyhow errors its 404
```
 state
                .get_reward_account_v1(&state.node_state().await, height, view, account.into())
                .await
                .map_err(|err| Error::catch_all(StatusCode::NOT_FOUND, format!("{err:#}")))
```

sequencer/src/api/sql.rs
```diff
@@ -491,12 +511,27 @@ async fn load_frontier<Mode: TransactionMode>(
...
+) -> Result<(RewardMerkleTreeV1, Leaf2), AccountProofError> {
     let leaf = tx
         .get_leaf(LeafId::<SeqTypes>::from(height as usize))
         .await
-        .context(format!("leaf {height} not available"))?;
+        .map_err(|err| {
+            AccountProofError::LeafNotFound(format!("leaf {height} not available: {err}"))
+        })?;
+
+    // Validate that the fetched leaf's view matches the expected view if provided
```
sveitser, line 524: Would be nice to comment a bit more on why this check is necessary (like in the PR description).

sveitser (approved):
LGTM. The error handling seems a bit verbose and it would be nice to have a regression test for this.

PR #3744, opened 2025-11-11
Remove stake table events migration

pls148:
Closes #<ISSUE_NUMBER>


sequencer/src/persistence/fs.rs
```diff
@@ -1461,70 +1459,6 @@ impl SequencerPersistence for Persistence {
     }
 
     async fn migrate_stake_table_events(&self) -> anyhow::Result<()> {
```
imabdulbasit: remove this function completely? and also wherever this is called

PR #3746, opened 2025-11-12
B01: renounceOwnership reverts when called

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- renounceownership function is overriden so that it reverts when it's called

### Key places to review:

---
  - https://app.asana.com/0/0/1211924444372471

### DIFF
--- a/contracts/test/EspToken.t.sol
+++ b/contracts/test/EspToken.t.sol
@@ -321,6 +321,7 @@ contract EspTokenUpgradabilityTest is Test {
         address currentOwner = token.owner();
         assertEq(currentOwner, admin);
 
+        vm.prank(admin);
         vm.expectRevert(EspToken.OwnershipCannotBeRenounced.selector);
         token.renounceOwnership();
 


contracts/src/EspTokenV2.sol
```diff
@@ -44,6 +47,13 @@ contract EspTokenV2 is EspToken {
         _mint(to, amount);
     }
 
+    /// @notice Cannot renounce ownership
+    /// @dev Override renounceOwnership() to revert, preventing accidental or malicious ownership
+    /// renunciation
+    function renounceOwnership() public virtual override onlyOwner {
```
sveitser: @alysiahuggins should this go into the V1 (or also go into V1)? If we first deploy the V1 on mainnet because reward claims aren't ready and V2 depends on reward claims to be deployed it would probably be better to have this in V1.

alysiahuggins: good point, will switch

PR #3747, opened 2025-11-12
A01: storage upgrade compatibility tests

sveitser:
Add check for EspToken and StakeTable to verify that the storage layout of contracts in the repo are compatible with what's deployed on sepolia.

~~The new CI job is expected to fail until we add an ETHERSCAN_API_KEY secret.~~
Edit: this has been added.

---
  - https://app.asana.com/0/0/1211924444372467


alysiahuggins (approved):
lgtm

PR #3749, opened 2025-11-13
Don't broadcast share we already have from vid request task

### LINKED ISSUE #3608: Fix `VID share signature is invalid` error
The issue is triggered by the code in `cancel_vid_request_task`. We're sending `VidShareRecv` with the unchanged signature and our own public key. We probably received vid disperse from somebody else so the signature doesn't match our public key. Hence the error.

bfish713:
Closes #3608

### This PR:
Fixes the Invalid VID signature errors we see in consnesus

### DIFF
--- a/crates/hotshot/task-impls/src/request.rs
+++ b/crates/hotshot/task-impls/src/request.rs
@@ -314,7 +314,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> NetworkRequestState<TYPES, I
             // First check if we got the data before continuing
             while !Self::cancel_vid_request_task(
                 &consensus,
-                &sender,
                 &public_key,
                 &view,
                 &shutdown_flag,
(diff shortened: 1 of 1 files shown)


crates/hotshot/task-impls/src/request.rs
```diff
@@ -376,14 +374,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> NetworkRequestState<TYPES, I
...
             tracing::debug!("Send own vid share: {vid_shares:?}, my id {id:?}");
             for vid_share in vid_shares.values() {
-                broadcast_event(
-                    Arc::new(HotShotEvent::VidShareRecv(
-                        public_key.clone(),
-                        vid_share.clone(),
-                    )),
-                    sender,
-                )
-                .await;
```
ss-es, line 386: I think removing this changes some behavior, but I don't know if we rely on it for something currently

we're re-broadcasting VID shares from consensus here. shares can only get into the `Consensus` state via `update_vid_shares`, which is called in 3 places:

- from the quorum vote task on `VidShareRecv`, which obviously we don't need to re-broadcast (especially since this path seems to be the only consumer of `VidShareRecv`)
- in the `task-impls/src/vid.rs`, and
- `types/src/consensus.rs` 

in the last 2 cases it's after we create shares to send out from a vid disperse proposal.

I think the `task-impls/src/vid.rs` call will result in us sending out vid shares to all nodes (including ourselves) so this should almost always be fine (barring weird networking edge cases)

the `calculate_and_update_vid` path in `consensus.rs` does not seem to do this though. so I think we would need to broadcast our own vid share internally as a `VidShareRecv` at the end of `calculate_and_update_vid` to not change anything?

PR #3753, opened 2025-11-14
Add DRB logs

lukaszrzasik:
Closes #<ISSUE_NUMBER>


types/src/v0/impls/reward.rs
```diff
@@ -921,6 +921,8 @@ pub async fn get_leader_and_fetch_missing_rewards(
         .leader(view, Some(epoch))
         .context(format!("leader for epoch {epoch:?} not found"))?;
 
+    tracing::info!("Selected leader: {leader}");
```
imabdulbasit: can we also add view number here

imabdulbasit: ah sorry I just realized that this would be very noisy so lets make it debug

imabdulbasit (approved):

PR #3755, opened 2025-11-14
Use root leaf's version for DRB difficulty selection

ss-es:
If a node has to recalculate a very old DRB for whatever reason, using the `UpgradeLock` may not produce the correct version for that root leaf's view.

We now take the version directly from the root leaf's block header instead

### DIFF
--- a/sequencer/src/persistence/fs.rs
+++ b/sequencer/src/persistence/fs.rs
@@ -1460,7 +1460,7 @@ impl SequencerPersistence for Persistence {
 
     async fn store_drb_input(&self, drb_input: DrbInput) -> anyhow::Result<()> {
         if let Ok(loaded_drb_input) = self.load_drb_input(drb_input.epoch).await {
-            if loaded_drb_input.iteration != drb_input.iteration {
+            if loaded_drb_input.difficulty_level != drb_input.difficulty_level {
                 tracing::error!("Overwriting {loaded_drb_input:?} in storage with {drb_input:?}");
             } else if loaded_drb_input.iteration >= drb_input.iteration {
                 anyhow::bail!(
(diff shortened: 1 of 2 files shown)


sequencer/src/persistence/fs.rs
```diff
@@ -1460,7 +1460,9 @@ impl SequencerPersistence for Persistence {
 
     async fn store_drb_input(&self, drb_input: DrbInput) -> anyhow::Result<()> {
         if let Ok(loaded_drb_input) = self.load_drb_input(drb_input.epoch).await {
-            if loaded_drb_input.iteration >= drb_input.iteration {
+            if loaded_drb_input.iteration != drb_input.iteration {
+                tracing::error!("Overwriting {loaded_drb_input:?} in storage with {drb_input:?}");
+            } else if loaded_drb_input.iteration >= drb_input.iteration {
```
bfish713, line 1465: This else if can't get hit right?

sequencer/src/persistence/sql.rs
```diff
@@ -2175,7 +2175,9 @@ impl SequencerPersistence for Persistence {
 
     async fn store_drb_input(&self, drb_input: DrbInput) -> anyhow::Result<()> {
         if let Ok(loaded_drb_input) = self.load_drb_input(drb_input.epoch).await {
-            if loaded_drb_input.iteration >= drb_input.iteration {
+            if loaded_drb_input.iteration != drb_input.iteration {
+                tracing::error!("Overwriting {loaded_drb_input:?} in storage with {drb_input:?}");
+            } else if loaded_drb_input.iteration >= drb_input.iteration {
```
bfish713, line 2180: same here

sequencer/src/persistence/fs.rs
ss-es: yeah that was a typo, I intended to use `.difficulty_level !=` not `iteration !=`. fixed

PR #3756, opened 2025-11-14
Update state: apply header, verify, store

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/sequencer/src/state.rs
+++ b/sequencer/src/state.rs
@@ -4,9 +4,11 @@ use std::{cmp::max, sync::Arc, time::Duration};
 use anyhow::{bail, ensure, Context};
 use either::Either;
 use espresso_types::{
+    traits::StateCatchup,
     v0_3::{ChainConfig, RewardAccountV1, RewardMerkleTreeV1},
     v0_4::{Delta, RewardAccountV2, RewardMerkleTreeV2},
-    BlockMerkleTree, EpochVersion, FeeAccount, FeeMerkleTree, Leaf2, ValidatedState,
+    BlockMerkleTree, DrbAndHeaderUpgradeVersion, EpochVersion, FeeAccount, FeeMerkleTree, Leaf2,
+    ValidatedState,
 };
 use futures::{future::Future, StreamExt};
 use hotshot::traits::ValidatedState as HotShotState;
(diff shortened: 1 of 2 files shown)


sequencer/src/state.rs
```diff
@@ -51,71 +41,45 @@ pub(crate) async fn compute_state_update(
     );
     ensure!(
         state.block_merkle_tree.commitment() == parent_header.block_merkle_tree_root(),
-        "internal error! in-memory block tree {} does not match parent header {}",
+        "internal error! in-memory block tree {:?} does not match parent header {:?}\nfull \
+         in-memory block tree {:?}",
```
imabdulbasit: I don't think we should display the whole tree

sequencer/src/state.rs
```diff
@@ -51,71 +41,45 @@ pub(crate) async fn compute_state_update(
...
         state.block_merkle_tree.commitment(),
-        parent_header.block_merkle_tree_root()
+        parent_header.block_merkle_tree_root(),
+        state.block_merkle_tree,
     );
     ensure!(
         state.fee_merkle_tree.commitment() == parent_header.fee_merkle_tree_root(),
-        "internal error! in-memory fee tree {} does not match parent header {}",
+        "internal error! in-memory fee tree {:?} does not match parent header {:?}\nfull \
+         in-memory fee tree {:?}",
```
imabdulbasit: same here

sequencer/src/state.rs
```diff
@@ -249,35 +213,23 @@ async fn store_state_update(
 #[tracing::instrument(
     skip_all,
     fields(
-        node_id = instance.node_id,
-        view = ?parent_leaf.leaf().view_number(),
-        height = parent_leaf.height(),
+        view = ?proposed_leaf.leaf().view_number(),
```
imabdulbasit: why do we want to remove node_id?

sequencer/src/state.rs
```diff
@@ -51,71 +41,45 @@ pub(crate) async fn compute_state_update(
...
-            instance,
-            peers,
-            parent_leaf,
-            header,
-            header.version(),
-            proposed_leaf.view_number(),
-        )
-        .await?;
-
-    if header.version() >= DrbAndHeaderUpgradeVersion::version() {
```
imabdulbasit, line 95: why is this removed?

lukaszrzasik: oh, this part was not present in the release branch. It must have got removed during the rebase. Good catch.

lukaszrzasik:
I wanted to split the code into separate logical parts: apply, verify, store. I can just move the code if that's what you prefer.

imabdulbasit (approved):

PR #3757, opened 2025-11-14
Remove old Membership::new and rename new_stake to new

pls148:
Closes #<ISSUE_NUMBER>

### This PR:

The only thing that I'm not super excited about here is the Leaf2FetcherTrait. Can't equality compare on associated types so I had to wrap Leaf2Fetcher::new in that trait. Open to ideas on how to do that better


pls148:
> What do you think about this? #3799

This is what I originally wanted to do but rust-analyzer kept complaining that equality for non-trait types was not supported. Was that something changed with a recent rust version bump or something?

lukaszrzasik:
I don't think so but I'm not 100% sure.

PR #3760, opened 2025-11-17
feat: stake table metadata URIs

sveitser:
Validators need to communicate some metadata. We are currently missing this feature. This PR allows validators to set a metadata URI from where the information can be fetched.

- Registration accepts and emits metdata URI
- Validators can update their metadata URIs
- add foundry unittests for setting and updating metadata URI
- add staking-cli integration tests for setting and updating URI
- No significant changes in espresso-types because metadata URIs are not used by consensus.

---
  - https://app.asana.com/0/0/1211962898202254

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -248,6 +248,10 @@ impl StakeTableState {
         }
     }
 
+    pub fn validators(&self) -> &ValidatorMap {
+        &self.validators
+    }
+
     pub fn into_validators(self) -> ValidatorMap {
         self.validators
     }
@@ -257,7 +261,7 @@ impl StakeTableState {
     ///
     /// This function MUST NOT modify `self` if the event is invalid. All validation
     /// checks must be performed before any state modifications occur.
-    fn apply_event(&mut self, event: StakeTableEvent) -> ApplyEventResult<()> {
+    pub fn apply_event(&mut self, event: StakeTableEvent) -> ApplyEventResult<()> {
         match event {
             StakeTableEvent::Register(ValidatorRegistered {
                 account,


contracts/src/StakeTableV2.sol
```diff
@@ -548,6 +575,28 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+    /// @param metadataUri The new metadata URI
+    ///
+    /// @dev The metadata URI is NOT stored on-chain. This function is event-sourced only. Off-chain
+    /// indexers must listen to the MetadataUriUpdated event to track the current URI. URIs should
+    /// be kept reasonably short for gas efficiency (max 2048 bytes). Only the validator
+    /// (msg.sender) can update their own metadata URI.
+    ///
+    /// @dev No format validation is performed on the URI - any non-empty string within length
+    /// limits is accepted. Consumers should validate URI format and accessibility off-chain.
+    function updateMetadataUri(string memory metadataUri) external virtual whenNotPaused {
```
alysiahuggins, line 588: i think we should rate limit this to prevent spamming. Anyone can register as a validator and try to slow down our indexers by spamming this method.

contracts/src/StakeTableV2.sol
```diff
@@ -457,6 +477,11 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
             revert InvalidCommission();
         }
 
+        uint256 metadataLength = bytes(metadataUri).length;
+        if (metadataLength == 0 || metadataLength > MAX_METADATA_URI_LENGTH) {
```
alysiahuggins, line 481: we can consider other data validation like 
string starts with https://, ipfs:// & arweave://

Also, i think the metadata uri might be too long? is 256 length enough?

sveitser: For length: I read through this: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers and figured 2048 should be enough. It's the caller that pays the gas so I think there's no need to be very strict about it.

I don't think we know yet what kind of URLs we want to allow so we can't add further validation yet. We can always add more validation to the staking-cli and UI if desired. That is much easier to update as well.

contracts/src/StakeTableV2.sol
sveitser: I don't think this is necessary. Event processing is very cheap and consensus doesn't even fetch and process this event.

alysiahuggins: when you say cheap, are you referring to compute? I'm thinking if the UI is using an RPC provider , then and eth_getLogs returns a lot of data, does that consume a lot of the RPC's compute units?

contracts/src/StakeTableV2.sol
alysiahuggins: ok that's fair

contracts/src/StakeTableV2.sol
sveitser: Yes I mean the L1 gas fee for this is orders of magnitudes higher than than what clients might pay to process the event. I also think they don't really need to process these events, especially from malicious validators:

1. The UIs should use the staking-ui-service because building the stake table state is very difficult for a UI only client.
2. I don't see why a UI should fetch events from misbehaving validators. I also think those validators wouldn't be on any "recommended" or "endorsed" lists for long.
3. Clients can limit block ranges to fetch.

So in summary I think there is very little damage such an attack would do and it would be expensive. Therefore I think it's pointless and just a waste of money for the attacker. Therefore I think doesn't warrant complicating the contract code and increasing fees for honest validators.

alysiahuggins: Good points. My only thought was that if someone ever wanted to choke our indexers or UI, spamming the events we rely on would be the easiest vector. Even though clients can filter by address/topics, a large volume of our own events over big block ranges can still create load on indexers and RPCs.

that said, I agree it’s not worth complicating the contract now. If it ever becomes a real issue, we can address it in a future upgrade or at the indexing layer.

PR #3765, opened 2025-11-18
Prepopulate all Reward Accounts with 0 after Upgrade

bfish713:
### This PR:
Populates all the accounts in the Reward Merkle tree with 0 balance.  This prevents us from sending a catchup request for every account the first time it gets some rewards.  

### This PR does not:

This doesn't prevent catchup from occurring later on, or change the amount of rewards accumulated. 

### Key places to review:

The code to actually add 0 balance to every potential leader and delegator in the current stake table.


types/src/v0/impls/reward.rs
```diff
@@ -839,6 +865,13 @@ pub async fn distribute_block_reward(
         instance_state.fixed_block_reward().await?
     };
 
+    // If we just upgraded to V4 from any version we should pre-poplute the merkle tree
```
sveitser:
```suggestion
    // If we just upgraded to V4 from any version we should pre-populate the merkle tree
```

imabdulbasit (approved):

types/src/v0/impls/reward.rs
```diff
@@ -839,6 +865,13 @@ pub async fn distribute_block_reward(
         instance_state.fixed_block_reward().await?
     };
 
+    // If we just upgraded to V4 from any version we should pre-populate the merkle tree
+    // with accounts from the stake table, giving each address 0 initial balance
+    if version >= DrbAndHeaderUpgradeVersion::version()
```
imabdulbasit, line 870: oh I think this should be before  

```
let leader = get_leader_and_fetch_missing_rewards(
        instance_state,
        validated_state,
        parent_leaf,
        view_number,
    )
    .await?;
```

so that get_leader_and_fetch_missing_rewards doesn't trigger catchup

imabdulbasit (changes requested):

PR #3771, opened 2025-11-19
release-20251106 contract changes backport

sveitser:
Note to self: DO NOT squash merge.


imabdulbasit (approved):
Because this is only cherry picking commits, so if the CI passes this PR can be merged. Although I think we should just merge main into release branch if there is no breaking change

PR #3777, opened 2025-11-19
WIP deployment prep for decaf

alysiahuggins:
### This PR:
- exposes deploy RC flag
- refactors transferOwnership function
- more to come

### Key places to review:

---
  - https://app.asana.com/0/0/1211993933154209


contracts/rust/deployer/src/builder.rs
```diff
@@ -785,19 +786,29 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
...
+        // RewardClaim uses AccessControl instead of Ownable, so we need to grant the admin role
+        // instead of transferring ownership
+        let receipt = if contract_type == Contract::RewardClaimProxy {
+            tracing::info!(
+                "Granting DEFAULT_ADMIN_ROLE for {} to {} (RewardClaim uses AccessControl, not \
+                 Ownable)",
+                target_contract,
+                new_owner
+            );
+            crate::grant_admin_role(&self.deployer, contract_type, contract_address, new_owner)
```
alysiahuggins, line 798: when using the flag `--transferOwnershipFromEoa`, if the target contract is the `RewardClaim` contract, then , use the `grant_admin_role` function to transfer ownership.

scripts/verify-pos-deployment.sh
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env bash
+# Verify governance deployment matches expected configuration
+
+# Required env vars: timelock addresses, admin/proposer/executor addresses, contract proxy addresses
```
alysiahuggins, line 4: expects these env vars to be in the shell env

justfile
```diff
@@ -240,7 +240,7 @@ gen-bindings:
 export-contract-abis:
     rm -rv contracts/artifacts/abi
     mkdir -p contracts/artifacts/abi
-    for contract in LightClient{,Mock,V2{,Mock}} StakeTable{,V2} EspToken IRewardClaim; do \
+    for contract in LightClient{,Mock,V2{,Mock}} StakeTable{,V2} EspToken{,V2} IRewardClaim; do \
```
alysiahuggins, line 251: should we export the `IRewardClaim` Abi or `RewardClaim`?

contracts/rust/deployer/src/builder.rs
alysiahuggins: full grant role support will be added in another PR, a task has been added on asana

justfile
sveitser: I think the interface is usually preferred but we should make sure the Interface is up to date.

scripts/verify-pos-deployment.sh
sveitser: Are these env vars listed anywhere?

justfile
```diff
@@ -216,7 +216,7 @@ build-docker-images:
 
 # generate rust bindings for contracts
 VERSIONED := "LightClient(Arbitrum)?(V\\d+)?(Mock)?|PlonkVerifier(V\\d+)?|StakeTable(V\\d+)?|EspToken(V\\d+)?|RewardClaim(V\\d+)?"
-EXACT := "FeeContract|ERC1967Proxy|OpsTimelock|SafeExitTimelock|OwnableUpgradeable|IRewardClaim|IPlonkVerifier"
+EXACT := "FeeContract|ERC1967Proxy|OpsTimelock|SafeExitTimelock|OwnableUpgradeable|AccessControlUpgradeable|IRewardClaim|IPlonkVerifier"
```
sveitser, line 227: I think when we're not deploying contracts we can use interfaces instead.

alysiahuggins: in order to keep the `grant_admin_role()` function general and do this https://github.com/EspressoSystems/espresso-network/blob/8d0e40a589e01602549aebf995d638f4d98ddce9/contracts/rust/deployer/src/lib.rs#L1247-L1253, i need to use the implementation, i could also hardcode the role to bytes(0) but prefer not to do that for now unless there's a strong reason I'm missing

scripts/verify-pos-deployment.sh
alysiahuggins: added it in the `--help` flag and to this readme (referencing the help menu item) https://github.com/EspressoSystems/espresso-network/commit/8d0e40a589e01602549aebf995d638f4d98ddce9

PR #3778, opened 2025-11-19
fix: arbitrum network test is flaky

sveitser:
- move to network tests
- try 2 providers

Not sure if this will help but I think it shouldn't make it any worse. The network test job is nice because it indicates that it's likely the network that has issues instead of the test itself.


sveitser:
This only modifies contract tests and those jobs passed, so merging.

PR #3781, opened 2025-11-19
fix: da-committees test is very slow

sveitser:
The test requirements are relative so I think it's fine to check that the demo is still working after the DA changes.


sveitser:
@pls148 we seem to get stuck in epoch 2 now.

https://github.com/EspressoSystems/espresso-network/actions/runs/19515924772/job/55872804338?pr=3781

PR #3782, opened 2025-11-19
test profile: opt 3 for ark crates

sveitser:
This might help with the resource constraints on the CI.


sveitser:
Doesn't help with proof generation time. Same as #3783

PR #3783, opened 2025-11-19
test profile: opt 3 for jf crates

sveitser:


sveitser:
This doesn't seem to have any effect on proof generation time.

PR #3785, opened 2025-11-20
chore: CI fixes and cleanup

sveitser:
- fix: nix env macos test times out
- util: add cargo check recipe
- fix: fuzz tests time out on main
- cleanup: reward fixtures and fixture usage

### DIFF
--- a/.github/workflows/nix-env-macos-arm.yml
+++ b/.github/workflows/nix-env-macos-arm.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Activate nix env
         run: nix develop --accept-flake-config -c echo Ok
 
-      - name: Build rust crates
+      - name: Check rust crates
         run: nix develop --accept-flake-config -c just check
 
       - name: Build contracts


.github/workflows/nix-env-macos-arm.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: nix develop --accept-flake-config -c echo Ok
 
       - name: Build rust crates
```
alysiahuggins:
```suggestion
      - name: Check rust crates compile
```

justfile
```diff
@@ -41,9 +41,17 @@ clippy *args:
     # check entire workspace (including sequencer-sqlite crate) with embedded-db feature
     cargo clippy --workspace --features "embedded-db testing" --all-targets {{args}}
 
+check *args:
+    # postgres
+    cargo check {{args}}
+    # embedded-db
+    cargo check -p sequencer-sqlite -p espresso-dev-node {{args}}
```
alysiahuggins, line 48: curious why we only check  (& build as shown below) these libs

sveitser: we have to build these separately due to cargo feature unification, if we build them with the rest it turns on the embedded-db feature for all crates

PR #3787, opened 2025-11-20
feat: optional metadata URIs

sveitser:
- Allow empty metadata in contract.
- Add --no-metadata-uri flag for staking-cli.

### DIFF
--- a/staking-cli/src/lib.rs
+++ b/staking-cli/src/lib.rs
@@ -153,7 +153,7 @@ pub struct MetadataUriArgs {
     #[clap(long, env = "METADATA_URI")]
     metadata_uri: Option<String>,
 
-    #[clap(long)]
+    #[clap(long, env = "NO_METADATA_URI")]
     no_metadata_uri: bool,
 }
 


contracts/src/StakeTableV2.sol
```diff
@@ -651,6 +648,16 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
         activeStake = initialActiveStake;
     }
 
+    /// @notice Validate metadata URI length
+    /// @param metadataUri The metadata URI to validate
+    /// @dev Public view function to allow external validation before transaction submission
+    function validateMetadataUri(string memory metadataUri) public pure {
```
sveitser, line 654: Currently this isn't used externally but I figured it doesn't hurt to expose it. Especially if will validate more later.

staking-cli/src/lib.rs
```diff
@@ -145,6 +147,32 @@ impl ValidSignerConfig {
     }
 }
 
+#[derive(ClapArgs, Debug, Clone)]
+#[group(required = true, multiple = false)]
+pub struct MetadataUriArgs {
+    #[clap(long, env = "METADATA_URI")]
+    metadata_uri: Option<String>,
+
+    #[clap(long)]
```
sveitser: missing env var

PR #3788, opened 2025-11-21
Fix reward Catchup

bfish713:
Closes #<ISSUE_NUMBER>

### This PR:

Fixes fetching the header for rewards when it's missing to use the proper stake table.  IT also makes it so that when we get an epoch root we just have to pass the epoch which the root was decided in.  This should prevent the type of error this PR fixes (where block number fetched doesn't match the epoch number of the stake table).

### This PR does not:

Shouldn't change any semantics except the reward amount catchup.

### Key places to review:

The whole PR should be looked at carefully.

### DIFF
--- a/crates/hotshot/example-types/src/membership/strict_membership.rs
+++ b/crates/hotshot/example-types/src/membership/strict_membership.rs
@@ -13,7 +13,7 @@ use hotshot_types::{
         node_implementation::{ConsensusTime, NodeImplementation, NodeType},
         signature_key::StakeTableEntryType,
     },
-    utils::{epoch_from_block_number, transition_block_for_epoch},
+    utils::{root_block_in_epoch, transition_block_for_epoch},
     PeerConfig,
 };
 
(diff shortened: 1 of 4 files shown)


crates/hotshot/types/src/traits/election.rs
```diff
@@ -219,7 +219,6 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
     fn get_epoch_root(
         _membership: Arc<RwLock<Self>>,
         _block_height: u64,
-        _epoch: TYPES::Epoch,
```
ss-es, line 222: could we remove `block_height` instead and keep `epoch`? it would make more sense for the membership implementation could calculate the root block height from the epoch number requested (rather than the caller)

crates/hotshot/example-types/src/membership/strict_membership.rs
```diff
@@ -284,13 +284,13 @@ impl<
...
         membership: Arc<RwLock<Self>>,
-        block_height: u64,
-        _epoch: TYPES::Epoch,
+        epoch: TYPES::Epoch,
     ) -> anyhow::Result<Leaf2<TYPES>> {
         let membership_reader = membership.read().await;
-        let epoch = epoch_from_block_number(block_height, membership_reader.epoch_height);
 
-        let stake_table = membership_reader.inner.stake_table(Some(epoch));
+        let block_height = root_block_in_epoch(*epoch, membership_reader.epoch_height);
```
ss-es, line 291: shouldn't this be `root_block_in_epoch(*epoch - 2)`?

I guess: what are the semantics of `get_epoch_root(e)`? I figured it would return the root leaf for epoch `e`

ss-es: I think we previously had `root_block_in_epoch(root_epoch,..)` everywhere

bfish713: It's the root block decided in the epoch, I'm just trying to make the least of changes now, and before we'd call this will epoch we want a root for - 2, and block height of the leaf in epoch - 2.  This just lets the epoch argument be passed in the same way, and removes the block height argument

PR #3789, opened 2025-11-21
Load membership from storage in `MembershipCoordinator` before triggering catchup

imabdulbasit:
Currently the `MembershipCoordinator` will trigger a catchup request if the stake table is not present in memory. We do store the stake tables in `add_epoch_root().` This PR adds fetching from storage if the stake table is not found in memory before triggering a catchup request. It adds a new trait `EpochStateStorage` in hotshot which contains all the storage methods related to membership including the DRB store/load. This allows us to remove alll the DRB callbacks from the MembershipCoordinator as well.

This also removes `membership.reload_stake(RECENT_STAKE_TABLES_LIMIT).await;` because we already reload stake tables when loading the consensus data at startup

---
  - https://app.asana.com/0/0/1212049970747844


crates/hotshot/types/src/drb.rs
```diff
@@ -98,15 +98,14 @@ pub fn difficulty_level() -> u64 {
 /// # Arguments
 /// * `drb_seed_input` - Serialized QC signature.
 #[must_use]
-pub async fn compute_drb_result(
+pub async fn compute_drb_result<TYPES: NodeType>(
     drb_input: DrbInput,
-    store_drb_progress: StoreDrbProgressFn,
-    load_drb_progress: LoadDrbProgressFn,
+    storage: Arc<dyn EpochStateStorage<TYPES>>,
```
ss-es, line 103: it's not the biggest deal, but personally I think this is a downgrade in code structure -- this is the core DRB logic and now we need a full `TestStorage` implementation to test it

crates/hotshot/types/src/traits/storage.rs
```diff
@@ -29,11 +29,51 @@ use crate::{
...
+        epoch: TYPES::Epoch,
+        block_header: TYPES::BlockHeader,
+    ) -> Result<()>;
+}
+
 /// Abstraction for storing a variety of consensus payload datum.
 #[async_trait]
-pub trait Storage<TYPES: NodeType>: Send + Sync + Clone + 'static {
+pub trait Storage<TYPES: NodeType>:
+    Send + Sync + Clone + 'static + EpochStateStorage<TYPES>
```
ss-es, line 73: if `Storage` requires `EpochStateStorage`, why do we have a separate trait? why not just combine them?

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -199,6 +206,75 @@ where
         ))
     }
 
+    async fn insert_from_storage(
+        &self,
+        epoch: TYPES::Epoch,
+        need_randomized: bool,
+    ) -> Result<bool> {
```
ss-es: it's not clear from the name + return type what the semantics of this function are. I think this could really use a comment explaining what it does and what it returns

also I really think `Result<bool>` is a bad return type, it's very hard to reason about. I think it would be better to make a custom enum or something here for the return type of `insert_from_storage`

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -199,6 +206,75 @@ where
         ))
     }
 
+    async fn insert_from_storage(
+        &self,
+        epoch: TYPES::Epoch,
+        need_randomized: bool,
```
ss-es: I think an enum would make help clarity here, e.g. `NeedsRandomized` and `SkipRandomized`

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -184,6 +182,15 @@ where
...
         }
+        if self
+            .insert_from_storage(epoch, false)
+            .await
+            .inspect_err(|e| {
+                let _ = warn!("loading from storage for epoch {epoch:?} failed: {e}");
+            })?
+        {
+            return Ok(ret_val);
+        }
```
ss-es: I don't think this is correct. if we error from storage we should continue, not return

ss-es: I would also say it's not actually important that `insert_from_storage` is told whether we `need_randomized`. It should always query the DRB and load it if it exists. We can check the result afterwards, since `has_randomized_stake_table()` costs nothing to call.

imo, it would be clearer if:
- `insert_from_storage` should be called `catchup_from_storage`, returning `()`
- after `self.catchup_from_storage`, perform the `membership.read().await.has_stake_table` or `has_randomized_stake_table()` check again to decide whether to return early

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -152,6 +141,15 @@ where
...
         }
+        if self
+            .insert_from_storage(epoch, true)
+            .await
+            .inspect_err(|e| {
+                let _ = warn!("loading from storage for {epoch:?} failed: {e}");
+            })?
+        {
+            return Ok(ret_val);
+        }
```
ss-es, line 160: similar comment here

ss-es (commented):
partial review on the hotshot side, I'll keep looking later

crates/hotshot/types/src/drb.rs
imabdulbasit: I think this function is only used in epoch coordinator, and coordinator already has storage now so I don't see the point of having callbacks?

crates/hotshot/types/src/traits/storage.rs
imabdulbasit: `Storage` is not Object Safe. Passing storage to Coordinator would require a generic parameter and I wanted to avoid that

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -137,6 +126,24 @@ where
...
             return Ok(ret_val);
         };
+        if self
+            .membership
+            .read()
+            .await
+            .has_randomized_stake_table(epoch)
+            .map_err(|e| {
+                error!(
+                    "membership_for_epoch failed while called with maybe_epoch {maybe_epoch:?}: \
```
ss-es: oh this is weird that we have a different return type with `has_randomized_stake_table` and `has_stake_table`. we should fix that (in a separate PR)

PR #3793, opened 2025-11-21
Remove `epoch` from `add_epoch_root`

ss-es:
Removes the epoch argument from `add_epoch_root`, instead calculating it from the provided header's block height. Also logs at `ERROR` if we attempt to add a block with the wrong block height

### DIFF
--- a/types/src/v0/impls/stake_table.rs
+++ b/types/src/v0/impls/stake_table.rs
@@ -45,7 +45,9 @@ use hotshot_types::{
         node_implementation::{ConsensusTime, NodeImplementation, NodeType},
         signature_key::StakeTableEntryType,
     },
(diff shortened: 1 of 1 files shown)


types/src/v0/impls/stake_table.rs
```diff
@@ -2169,16 +2172,32 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+
+        let membership_reader = membership.read().await;
+        let epoch_height = membership_reader.epoch_height;
+
+        let epoch =
+            Epoch::new(epoch_from_block_number(block_number, membership_reader.epoch_height) + 2);
+
+        tracing::info!(?epoch, "adding epoch root. height={:?}", block_number);
+
+        if !is_epoch_root(block_number, epoch_height) {
```
imabdulbasit: this check should be at the start of this function

PR #3794, opened 2025-11-24
Add `reward-amounts` endpoint to fetch all the reward accounts

imabdulbasit:
This PR adds an endpoint to get all the reward accounts at a block height. This is used by the staking UI to catchup when the service reboots. This reason for adding this endpoint is to catchup quickly, otherwise we need to query all the epoch stake tables and then hit the reward-balance endpoint for each account

---
  - https://app.asana.com/0/0/1212087089992284

### DIFF
--- a/sequencer/src/api/sql.rs
+++ b/sequencer/src/api/sql.rs
@@ -526,6 +526,17 @@ impl CatchupStorage for DataSource {
     async fn get_leaf_chain(&self, height: u64) -> anyhow::Result<Vec<Leaf2>> {
         self.as_ref().get_leaf_chain(height).await
     }
+
+    async fn get_all_reward_accounts(
+        &self,
+        height: u64,
+        offset: u64,
+        limit: u64,
+    ) -> anyhow::Result<Vec<(RewardAccountV2, RewardAmount)>> {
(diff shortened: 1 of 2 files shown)


sequencer/src/api/sql.rs
```diff
@@ -250,6 +251,70 @@ impl CatchupStorage for SqlStorage {
...
+        let header = leaf.header();
+
+        if header.version() < DrbAndHeaderUpgradeVersion::version() {
+            return Ok(Vec::new());
+        }
+
+        let rows = query_as::<(Value, Value)>(&format!(
+            "SELECT idx, entry
+               FROM {}
+              WHERE created = $1
```
jbearer: Maybe I'm misremembering how the Merklized storage works, but I thought this would only return accounts which were create or updated on exactly block `height`, but exclude accounts that were updated earlier whose values were still valid/non-zero at block `height`. So this would need to be something more like the `get_path` query where we select the largest value of `created` which is `<= height` for each distinct `idx`

imabdulbasit: yeah I had a commit doing this change. I haven't pushed it yet

PR #3801, opened 2025-11-26
redefine minExitEscrowPeriod 

alysiahuggins:
Closes #<ISSUE_NUMBER>
The lower bound is dynamically set over a function call to lightClient.blocksPerEpoch() . If lightClient.blocksPerEpoch() returns a value where blocksPerEpoch() * 15 > 86400 * 14; // 14 days , the minimum would exceed the maximum, causing all calls to updateExitEscrowPeriod() to revert with ExitEscrowPeriodInvalid()

### This PR:
makes the maximum dynamic based on light client parameters.

### Key places to review:

---
  - https://app.asana.com/0/0/1212197389245811


contracts/src/StakeTableV2.sol
```diff
@@ -762,15 +762,22 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
     /// @dev This function ensures that the exit escrow period is within the valid range
     /// @dev This function is not pausable so that governance can perform emergency updates in the
-    /// presence of system
+    /// presence of systemLightClientV2
     function updateExitEscrowPeriod(uint64 newExitEscrowPeriod)
         external
         virtual
         onlyRole(DEFAULT_ADMIN_ROLE)
     {
         uint64 minExitEscrowPeriod = lightClient.blocksPerEpoch() * 15; // assuming 15 seconds per
```
alysiahuggins: I'm thinking we should add a buffer, example 10 blocks. If we do this, it would be updated in the initializer as well. What do you think @sveitser ?
 ```suggestion
        uint64 minExitEscrowPeriod = (lightClient.blocksPerEpoch() + 10) * 15; // assuming 15 seconds per
```

File: contracts/src/StakeTableV2.sol [hunk 2]
```diff
@@ -762,15 +762,22 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @param newExitEscrowPeriod The new exit escrow period
     /// @dev This function ensures that the exit escrow period is within the valid range
     /// @dev This function is not pausable so that governance can perform emergency updates in the
-    /// presence of system
+    /// presence of systemLightClientV2
```
sveitser: Seems like this comment is missing something

contracts/src/StakeTableV2.sol
sveitser: What does that help with? I think if we have very low blocks per epoch we are either testing or something is off. So maybe reverting initialization for very low blocks per epoch would be preferable?

contracts/src/StakeTableV2.sol
```diff
@@ -762,15 +762,22 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
     function updateExitEscrowPeriod(uint64 newExitEscrowPeriod)
         external
         virtual
         onlyRole(DEFAULT_ADMIN_ROLE)
     {
         uint64 minExitEscrowPeriod = lightClient.blocksPerEpoch() * 15; // assuming 15 seconds per
             // block
-        uint64 maxExitEscrowPeriod = 86400 * 14; // 14 days
+        uint64 absoluteMaxExitEscrowPeriod = 86400 * 14; // 14 days, reasonable upper bound
```
sveitser: Isn't it more important that it cannot exceed the timelock delay?

contracts/src/StakeTableV2.sol
```diff
@@ -762,15 +762,22 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
+
+        // Calculate dynamic maximum: ensure it's always greater than minimum
+        // Use the larger of: absolute max (14 days) or 2x the minimum
+        // This ensures valid bounds even when epochs are very long
+        uint256 maxExitEscrowPeriod = absoluteMaxExitEscrowPeriod > minExitEscrowPeriod * 2
```
sveitser: Seems a bit complex but I don't think there's any problem with it.

alysiahuggins: i have changed it, i simplified it so that it reverts if min is > max https://github.com/EspressoSystems/espresso-network/pull/3801/commits/74694771f4a79fd5db3863889d30f6ace2c967e2

contracts/src/StakeTableV2.sol
alysiahuggins: i was thinking the min escrow period was a bit exact, but on second thought, the light client is updated once every 3 hours so we already have a 3 hour buffer between when the last epoch ended in consensus and when a user can withdraw

contracts/src/StakeTableV2.sol
alysiahuggins: "Good point. Since OpsTimelock delay is less than 14 days, we should cap the max at `TimelockController.getMinDelay()` to ensure users can exit before governance changes execute." Would you agree? @sveitser

contracts/src/StakeTableV2.sol
```diff
@@ -763,20 +763,29 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
     {
         uint64 minExitEscrowPeriod = lightClient.blocksPerEpoch() * 15; // assuming 15 seconds per
             // block
-        uint64 maxExitEscrowPeriod = 86400 * 14; // 14 days
+        uint64 maxExitEscrowPeriod = 14 days; // 14 days, reasonable upper bound
 
+        // Validate bounds before proceeding, since the minimum bound is based on the light client's
+        // blocks per epoch
+        // we should check that it has not exceeded the maximum exit escrow period
+        if (minExitEscrowPeriod >= maxExitEscrowPeriod) {
```
sveitser: I'm not sure I understand. This effectively blocks the function from being called because neither min nor max are the value supplied?

alysiahuggins: it's a check on the bounds which the function defines, and since the bounds can change if the `blocksPerEpoch`  value from the light client contract can change, this function makes this check and reverts in that scenario so that we're aware.

contracts/src/StakeTableV2.sol
alysiahuggins: after internal discussions, we decided to set the min to 2 days with comments to explain how to derive a valid `exitEscrowPeriod` value

contracts/src/StakeTableV2.sol
alysiahuggins: after internal discussions, we decided to set the min to 2 days with comments to explain how to derive a valid exitEscrowPeriod value

contracts/src/StakeTableV2.sol
alysiahuggins: after internal discussions, we decided to set the min to 2 days with comments to explain how to derive a valid exitEscrowPeriod value

.env
```diff
@@ -117,6 +117,7 @@ ESPRESSO_STATE_SIGNATURE_TOTAL_STAKE=5
 # Prover service
 ESPRESSO_PROVER_SERVICE_PORT=30050
 ESPRESSO_STATE_PROVER_UPDATE_INTERVAL_SECS=60
+ESPRESSO_STATE_PROVER_UPDATE_INTERVAL=20s
```
alysiahuggins: i had to change this because the test-demo-pr ci test was failing, the default in the code is 10 mins. This changes makes it so that the prover update fits within the smoke test 120 second timeout @sveitser

sveitser: This issue is fixed on main so we can remove the env var here.

contracts/rust/deployer/src/lib.rs
```diff
@@ -1826,7 +1826,7 @@ mod tests {
         )
         .await?;
         let lc_addr = deploy_light_client_contract(&provider, &mut contracts, false).await?;
-        let exit_escrow_period = U256::from(1000);
+        let exit_escrow_period = U256::from(172800);
```
sveitser: Define a constant?

alysiahuggins: good idea https://github.com/EspressoSystems/espresso-network/commit/483c086fcdf392dc69ee79e0306bd1968066786c

.env
alysiahuggins: ah ok great, updated here, let's see if CI builds now https://github.com/EspressoSystems/espresso-network/pull/3801/commits/dd8344cb68d82f4b6dae8cc432346047e43b7d27

PR #3808, opened 2025-12-01
derive target epoch from proposal in proposal prelim validated handler

sveitser:


crates/hotshot/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -651,14 +651,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>
...
+                // of a new epoch before ViewChange has updated our cur_epoch.
+                // If the QC is for the last block of an epoch (EQC), the next view is in epoch + 1.
+                let justify_qc = proposal.data.justify_qc();
+                let qc_epoch = justify_qc.data.epoch;
+                let target_epoch = if justify_qc
+                    .data
+                    .block_number
+                    .is_some_and(|bn| is_last_block(bn, self.epoch_height))
+                {
+                    qc_epoch.map(|e| e + 1)
```
lukaszrzasik, line 671: To make sure that it's a valid extended QC we should also check whether the proposal has the corresponding `next_epoch_justify_qc`. You can get it with the method `next_epoch_justify_qc()` and you can check whether the two certificates correspond to each other by comparing their leaf commits. They should be the same.

lukaszrzasik: I think you should actually be able to use `handle_eqc_formed` method to also properly store the eQC received in this way.

PR #3813, opened 2025-12-02
Deployer tech debt

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- adds support for reward claim in the timelock deployer lib
- removed `--timelock-address` flag, users no longer need to specify timelock addresses manually, it is derived from the already deployed timelock as timelock selection is automatic
- add some integration tests to validate deployer arg behaviour

### Key places to review:

---
  - https://app.asana.com/0/0/1212270282811483

### DIFF
--- a/contracts/rust/deployer/src/builder.rs
+++ b/contracts/rust/deployer/src/builder.rs
@@ -155,9 +155,8 @@ impl<P: Provider + WalletProvider> DeployerArgs<P> {
                     );
                     // deployer is the timelock owner
                     if use_timelock_owner {
-                        let timelock_addr = contracts
-                            .address(Contract::OpsTimelock)
-                            .expect("fail to get OpsTimelock address");
+                        let timelock_addr =
+                            derive_timelock_address_from_contract_type(target, contracts)?;
                         crate::transfer_ownership(provider, target, addr, timelock_addr).await?;
                     }
                 } else if let Some(multisig) = self.multisig {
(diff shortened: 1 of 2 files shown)


contracts/rust/deployer/src/proposals/timelock.rs
```diff
@@ -98,6 +98,9 @@ impl TimelockContract {
                 let tx_hash = *pending_tx.tx_hash();
                 tracing::info!(%tx_hash, "waiting for tx to be mined");
                 let receipt = pending_tx.get_receipt().await?;
+                if !receipt.inner.is_success() {
```
sveitser, line 101: There is `receipt.status()`

sequencer/src/bin/deploy.rs
```diff
@@ -128,46 +128,85 @@ struct Options {
...
     #[clap(long, default_value = "false")]
     upgrade_esp_token_v2: bool,
     /// Option to deploy StakeTable V1 and proxy
+    // (backward compatibility: --deploy-stake-table is an alias for --deploy-stake-table-v1)
     #[clap(long, default_value = "false")]
     deploy_stake_table: bool,
+
+    /// Option to deploy StakeTable V1
+    #[clap(long, default_value = "false")]
+    deploy_stake_table_v1: bool,
```
sveitser, line 170: I think these API changes to the deploy binary will break most deploy scripts that are used for testing and production. Is there enough benefit to justify that?

sveitser: edit: i see there some work for backwards compat. Log a warning to inform the user that they are using a deprecated argument if they do?

PR #3815, opened 2025-12-02
Parallelize reward account fetches

rob-maron:


sequencer/src/util.rs
```diff
@@ -0,0 +1,68 @@
...
+        F: Future<Output = T> + Send + 'static,
+        T: Send,
+    {
+        // Clone the semaphore for the inner task
+        let semaphore = self.semaphore.clone();
+
+        // Wrap the task, making it wait for a semaphore permit first
+        let task = async move {
+            // Acquire the permit
+            let permit = semaphore.acquire().await;
```
imabdulbasit, line 41: I think it is safe to just call unwrap() on this since this returns a result

sequencer/src/api/sql.rs
```diff
@@ -580,10 +605,12 @@ async fn load_frontier<Mode: TransactionMode>(
 }
 
 async fn load_v1_reward_accounts<Mode: TransactionMode>(
+    db: &SqlStorage,
     tx: &mut Transaction<Mode>,
```
imabdulbasit: I think it is safe to remove this tx parameter and open a new read transaction inside this function to get the leaf

sequencer/src/util.rs
rob-maron: Unwrap() where?

imabdulbasit: This one returns the result`let permit = semaphore.acquire().await;`

so if the semaphore is dropped it can return an error, which should not happen, so we can just unwrap here

rob-maron: Ah yep 👍

PR #3816, opened 2025-12-02
check merklized state height before returning reward accounts

imabdulbasit:
The query returns reward accounts where created <= requested height, but if the node's merklized state lags behind, we could return stale balances from an older height. This adds a check to ensure the reward-state-v2 has data up to the requested height before returning results, so that  we return the latest account balances as of that height.


jbearer:
This seems like a good candidate for a regression test

PR #3817, opened 2025-12-03
chore: deps update, fix build

sveitser:
Currently we can't build without lockfile, this means using our crate as library may break.

- Remove a glob import that creates conflicts.
- Hide a misguided depreciation warning.


hotshot-query-service/src/fetching/provider/query_service.rs
```diff
@@ -511,6 +511,9 @@ mod test {
         future::{join, FutureExt},
         stream::StreamExt,
     };
+    // generic-array 0.14.x is deprecated, but VidCommitment requires this version
+    // for From<Output<H>> impl in jf-merkle-tree
+    #[allow(deprecated)]
```
sveitser, line 516: @mrain do you have a better idea on how to fix this, so we don't need to use deprecated generic-array? It seems the latest `digest` still uses the deprecated version.

mrain: `digest` depends on `crypto-common` which has a rc version that migrates to `hybrid-array`. I'll take a look

PR #3818, opened 2025-12-03
Use Epoch From events in Proposal Task

bfish713:
Closes #<ISSUE_NUMBER>


crates/hotshot/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -440,12 +440,20 @@ impl<TYPES: NodeType> ConnectedNetwork<TYPES::SignatureKey> for CombinedNetworks
...
-            let mut primary_fut = self.primary().recv_message().fuse();
-            let mut secondary_fut = self.secondary().recv_message().fuse();
+            let primary_fut = async move {
+                loop {
+                    let p = self.primary().recv_message().await?;
+                    if !p.is_empty() {
+                        return Ok::<_, NetworkError>(p);
+                    }
+                    tracing::warn!("Received empty message from primary network");
+                }
```
lukaszrzasik, line 450: Is this relevant in testing only or can it happen in real scenarios as well?

PR #3819, opened 2025-12-03
Optimize reward-amounts query for postgres'

imabdulbasit:
Optimizes get_all_reward_accounts query by using `DISTINCT ON `instead of `ROW_NUMBER() ` for postgres, and by adding an index on` (idx, created)`

### DIFF
--- a/sequencer/src/api/sql.rs
+++ b/sequencer/src/api/sql.rs
@@ -316,7 +316,7 @@ impl CatchupStorage for SqlStorage {
                   WHERE created <= $1 AND idx IS NOT NULL
              ) sub
              WHERE rn = 1
-             ORDER BY idx
+             ORDER BY idx DESC
              LIMIT $2 OFFSET $3",
             RewardMerkleTreeV2::state_type()
         );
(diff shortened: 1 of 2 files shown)


sequencer/src/api/sql.rs
```diff
@@ -294,27 +294,40 @@ impl CatchupStorage for SqlStorage {
...
              LIMIT $2 OFFSET $3",
             RewardMerkleTreeV2::state_type()
-        ))
-        .bind(height as i64)
-        .bind(limit as i64)
-        .bind(offset as i64)
-        .fetch_all(tx.as_mut())
-        .await
-        .context("loading reward accounts from storage")?;
+        );
```
jbearer, line 322: I wonder if we want to just not support this endpoint at all when using sqlite, since we know this doesn't perform well. Like just return `Err("not supported")` here

imabdulbasit: yeah, but if we enable sqlite with pruning settings then this query should be fine I think

jbearer: Oh yeah good point

PR #3832, opened 2025-12-09
Min stake

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- Add minimum delegation amount to StakeTableV2 with a starting min of 1
- Add DelegateAmountTooSmall error for amounts below the minimum
- Make minimum delegation amount configurable via governance

### Key places to review:

---
  - https://app.asana.com/0/0/1212359241595288

### DIFF
--- a/staking-cli/src/delegation.rs
+++ b/staking-cli/src/delegation.rs
@@ -3,13 +3,14 @@ use alloy::{
     primitives::{utils::format_ether, Address, U256},
     providers::{PendingTransactionBuilder, Provider},
 };
-use anyhow::Result;
+use anyhow::{bail, Result};
 use hotshot_contract_adapter::{
     evm::DecodeRevert as _,
     sol_types::{
         EspToken::{self, EspTokenErrors},
         StakeTableV2::{self, StakeTableV2Errors},
     },
+    stake_table::StakeTableContractVersion,
 };
 
 pub async fn approve(
@@ -41,6 +42,19 @@ pub async fn delegate(
         format_ether(amount)
     );
     let st = StakeTableV2::new(stake_table, provider);
+
+    let version: StakeTableContractVersion = st.getVersion().call().await?.try_into()?;
+    if let StakeTableContractVersion::V2 = version {
+        let min_amount = st.minDelegateAmount().call().await?;
+        if amount < min_amount {
+            bail!(
+                "delegation amount {} ESP is below minimum of {} ESP",
+                format_ether(amount),
+                format_ether(min_amount)
+            );
+        }
+    }
+
     st.delegate(validator_address, amount)
         .send()
         .await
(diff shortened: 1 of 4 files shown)


contracts/src/StakeTableV2.sol
alysiahuggins, line 558: in another PR we should handle undelegations

I propose
```
if (balance - amount < minDelegationAmount) undelegateAll();
```

contracts/src/StakeTableV2.sol
```diff
@@ -889,6 +906,17 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
...
     }
 
+    /// @notice Set the minimum delegate amount
+    /// @param newMinDelegateAmount The new minimum delegate amount in wei
+    function setMinDelegateAmount(uint256 newMinDelegateAmount)
+        external
+        virtual
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
+        minDelegateAmount = newMinDelegateAmount;
```
alysiahuggins, line 916: didn't check if `newMinDelegateAmount` is > 0 in the event that we did want to set it to zero, we should discuss

sveitser: We could still set it to 1 (wei) which allows anything except zero.

contracts/src/StakeTableV2.sol
```diff
@@ -336,6 +346,9 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
         // Initialize undelegation IDs to start at 1 (0 is reserved for V1 undelegations)
         nextUndelegationId = 1;
 
+        // Initialize min delegate amount to 1 ESP token
+        minDelegateAmount = 1 ether;
```
sveitser, line 350: I guess theoretically we should check the decimals here but i'm also fine with a comment saying "the token is hardcoded to 18 decimals"

contracts/src/StakeTableV2.sol
```diff
@@ -134,6 +134,9 @@ contract StakeTableV2 is StakeTable, PausableUpgradeable, AccessControlUpgradeab
     /// @notice Total stake in active (not marked for exit) validators in the contract
     uint256 public activeStake;
 
+    /// @notice min delegate amount
+    uint256 public minDelegateAmount;
```
sveitser, line 138: Add this to the changelog (in this file).

contracts/src/StakeTableV2.sol
alysiahuggins: ok will follow up in another pr to avoid long CI runs for this

contracts/src/StakeTableV2.sol
alysiahuggins: ok will follow up in another pr to avoid long CI runs for this

contracts/src/StakeTableV2.sol
alysiahuggins: ok will follow up in another pr to avoid long CI runs for this

PR #3848, opened 2025-12-16
Implement light client leaf fetching and caching with fixed stake table

jbearer:
See [task](https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510082563943?focus=true)

# This PR does

* Implement leaf fetching and verification for the light client
* Support a local fixed size cache so we don't have to fetch the same leaves over and over again. The cache can be in memory or persisted in a file, either way we use a SQLite interface

# This PR does not

* Implement stake table fetching. Of course leaf verification requires a stake table. This PR just uses a fixed stake table always. The next task will be to support dynamic stake table fetching, which will depend on leaf verification (recursively)


light-client/migrations/01_schema.sql
```diff
@@ -0,0 +1,13 @@
+-- This set of SQL(ite) statements defines the schema for the local database used by the light
+-- client. Because light clients are local and (somewhat) ephemeral, we don't worry about
+-- migrations or backwards compatibility. This file always represents the latest version of the
```
Sneh1999: nit: It might still be nice to have migrations and backward compatibility in case we dont want to lose a lot of progress - if its not too hard to implement

light-client/src/consensus/leaf.rs
```diff
@@ -44,6 +45,53 @@ pub enum FinalityProof {
...
+    /// This can be used to check that the leaf chain in a [`LeafProof`] leads up to a leaf that is
+    /// finalized, and thus the whole chain is finalized.
+    Assumption(&'a Leaf2),
+}
+
+impl<'a> LeafProofHint<'a, StakeTableQuorum<EpochMembership<SeqTypes>>> {
+    /// Construct a [`LeafProofHint`] from a known-finalized leaf, where the quorum type doesn't
+    /// matter.
+    ///
+    /// If the quorum type matters (even though it will not be used in verification whenever the
```
Sneh1999, line 87: when can the case come up when quorum type would matter for assumption?

light-client/src/consensus/leaf.rs
```diff
@@ -44,6 +45,53 @@ pub enum FinalityProof {
...
+        }
+    }
+}
+
+/// A hint that allows a verifier to verify a proof.
+///
+/// The hint should be supplied by the verifier (e.g. the light client). It represents the root of
+/// trust for this proof.
+#[derive(Clone, Copy, Debug)]
+pub enum LeafProofHint<'a, Q> {
```
Sneh1999, line 68: nit: why cant Proof and hit just be part of one struct? Wouldnt that simplify a lot of logic like checking they are of the same type etc?

light-client/src/storage.rs
```diff
@@ -0,0 +1,334 @@
...
+        .bind(&hash)
+        .bind(data)
+        .fetch_one(tx.as_mut())
+        .await
+        .context("inserting new leaf")?;
+        tracing::debug!(height, hash, id, "inserted leaf");
+
+        // Delete the oldest leaves as necessary until the number of leaves stored does not exceed
+        // `num_leaves`.
+        let (num_leaves,): (u32,) = query_as("SELECT count(*) FROM leaf")
```
Sneh1999, line 164: nit: Doing garbage collection while inserting leaf might slow down insertions, maybe we could do garbage collection in the background?

Sneh1999 (commented):
Added some nits but looks good! Thanks for all the comments in the code, helps me understand

light-client/migrations/01_schema.sql
jbearer: You know what yeah. It's trivial to just treat this as the first migration of a set. I'll remove this comment

light-client/src/consensus/leaf.rs
jbearer: The proof is provided by the server, the hint is something already known and provided by the client/verifier

light-client/src/consensus/leaf.rs
jbearer: Like if you're working in some block that hard-codes Q: AlwaysTrueQuorum or something. It would only matter for making the code compile.

light-client/src/storage.rs
jbearer: I think it is just a couple extra queries against a database that is on the local file system or even in memory. It doesn't involve any network round trips or anything. I'm hoping this will be fine.

I would wait to see that it's actually a problem before adding the complexity of async garbage collection

PR #3850, opened 2025-12-18
Fee contract upgrade

alysiahuggins:
Closes #<ISSUE_NUMBER>

### This PR:
- Adds a patch upgrade for the FeeContract (v1.0.1)
- Forces fresh implementation deployment by clearing the contracts cache when upgrading to a patch
- Adds tests covering fee contract and light client upgrade behavior

### Key places to review:
- FeeContract.sol version bump (`getVersion`)
- `upgrade_fee_v1` and multisig upgrade helpers
- Cache-clearing logic before redeploying implementations
- New upgrade and redeploy tests

---
  - https://app.asana.com/0/0/1212418778879470

### DIFF
--- a/contracts/rust/deployer/src/lib.rs
+++ b/contracts/rust/deployer/src/lib.rs
@@ -1289,7 +1289,10 @@ pub async fn upgrade_fee_v1(
 
     let owner = fee_contract_proxy.owner().call().await?;
     if is_contract(&provider, owner).await? {
-        anyhow::bail!("FeeContract owner is not an EOA, can't upgrade");
+        anyhow::bail!(
+            "FeeContract owner ({:#x}) is not an EOA, can't upgrade",
+            owner
+        );
     }
 
     let curr_version = fee_contract_proxy.getVersion().call().await?;
@@ -1344,7 +1347,7 @@ pub async fn upgrade_fee_v1(
                 new_version.patchVersion
             );
         }
-        tracing::error!(
+        tracing::info!(
             proxy = %fee_contract_proxy_addr,
             impl = %new_fee_contract_addr,
             "FeeContract successfully upgraded to v1.0.1"
@@ -1461,7 +1464,7 @@ pub async fn get_proxy_initialized_version(
     // From openzeppelin Initializable.sol, the initialized version slot is keccak256("openzeppelin.storage.Initializable");
     let slot: B256 = "0xf0c57e16840df040f15088dc2f81fe391c3923bec73e23a9662efc9c229c6a00"
         .parse()
-        .unwrap();
+        .context("Failed to parse OpenZeppelin Initializable slot")?;
     let value = provider.get_storage_at(proxy_addr, slot.into()).await?;
     let initialized = value.as_le_bytes()[0]; // `_initialized` is u8 stored in the last byte
     Ok(initialized)


contracts/rust/deployer/src/proposals/multisig.rs
```diff
@@ -653,3 +653,95 @@ pub async fn upgrade_stake_table_v2_multisig_owner(
...
+    let proxy_addr = contracts
+        .address(Contract::FeeContractProxy)
+        .ok_or_else(|| anyhow!("FeeContractProxy (multisig owner) not found, can't upgrade"))?;
+    tracing::info!("FeeContractProxy found at {proxy_addr:#x}");
+    let proxy = FeeContract::new(proxy_addr, &provider);
+    let owner_addr = proxy.owner().call().await?;
+
+    // Verify current version before upgrading
+    let curr_version = proxy.getVersion().call().await?;
+    if curr_version.majorVersion != 1 {
```
alysiahuggins, line 680: kept this simple to allow for patch upgrades within v1

contracts/rust/deployer/src/proposals/multisig.rs
```diff
@@ -653,3 +653,95 @@ pub async fn upgrade_stake_table_v2_multisig_owner(
...
+        anyhow::bail!("Proxy owner is not a contract. Expected: {owner_addr:#x}");
+    }
+
+    // Deploy new implementation (with patch version)
+    let fee_contract_addr = if !dry_run {
+        let old_fee_contract_addr = contracts.address(Contract::FeeContract);
+
+        // Remove old implementation from cache so we deploy a new one
+        contracts.0.remove(&Contract::FeeContract);
+
```
alysiahuggins, line 709: we can allow an explicit flag e.g. `--patch-upgrade` to allow for redeployment of the same contract and upgrade, in order to do a patch upgrade

sveitser: @alysiahuggins Can you explain why we need to do this? I think usually what we do is that if a contract address is set we don't deploy it. So this would be different and I think might end up being confusing / lead to errors. If we do have to override the default behaviour I would prefer the override to be more explicit.

alysiahuggins: this is to allow for the patch upgrade as it's still the same contract with one logic change. but would you prefer that we add the --patch-upgrade flag instead?

sveitser: Ok, couldn't we just **not** set the fee contract address and throw an error if it's set? I think either probably fine we should just try to be consistent where possible.

alysiahuggins: ah yea that makes sense, thanks so much for pointing that out so the behaviour remains consistent https://github.com/EspressoSystems/espresso-network/commit/2408fec4a00d5d6c9f60893275b86e48df72753a

contracts/rust/deployer/src/lib.rs
```diff
@@ -1241,6 +1274,105 @@ pub async fn upgrade_stake_table_v2(
...
+
+    // now deploy the new implementation contract
+    let new_fee_contract_addr = contracts
+        .deploy(
+            Contract::FeeContract,
+            FeeContract::deploy_builder(&provider),
+        )
+        .await?;
+
+    if let Some(old_fee_contract_addr) = old_fee_contract_addr {
```
sveitser: So if this is true, we already exited the function on line 1310. Or i might be misunderstanding.

contracts/rust/deployer/src/lib.rs
```diff
@@ -1241,6 +1274,105 @@ pub async fn upgrade_stake_table_v2(
...
+        .await?
+        .get_receipt()
+        .await
+        .context("Failed to get upgrade transaction receipt")?;
+
+    if receipt.inner.is_success() {
+        let new_version = fee_contract_proxy.getVersion().call().await?;
+        if new_version.majorVersion != 1
+            || new_version.minorVersion != 0
+            || new_version.patchVersion != 1
```
sveitser: There may be some way to write this a bit more concisely `new_version != (1, 0, 1).into()`

There is come code for this in the bindings

```rust
            #[doc(hidden)]
            type UnderlyingSolTuple<'a> = (
                alloy::sol_types::sol_data::Uint<8>,
                alloy::sol_types::sol_data::Uint<8>,
                alloy::sol_types::sol_data::Uint<8>,
            );
            #[doc(hidden)]
            type UnderlyingRustTuple<'a> = (u8, u8, u8);
            #[cfg(test)]
            #[allow(dead_code, unreachable_patterns)]
            fn _type_assertion(
                _t: alloy_sol_types::private::AssertTypeEq<UnderlyingRustTuple>,
            ) {
                match _t {
                    alloy_sol_types::private::AssertTypeEq::<
                        <UnderlyingSolTuple as alloy_sol_types::SolType>::RustType,
                    >(_) => {}
                }
            }
            #[automatically_derived]
            #[doc(hidden)]
            impl ::core::convert::From<getVersionReturn> for UnderlyingRustTuple<'_> {
                fn from(value: getVersionReturn) -> Self {
                    (value.majorVersion, value.minorVersion, value.patchVersion)
                }
            }
            #[automatically_derived]
            #[doc(hidden)]
            impl ::core::convert::From<UnderlyingRustTuple<'_>> for getVersionReturn {
                fn from(tuple: UnderlyingRustTuple<'_>) -> Self {
                    Self {
                        majorVersion: tuple.0,
                        minorVersion: tuple.1,
                        patchVersion: tuple.2,
                    }
                }
            }
```

alysiahuggins: thanks a lot for this, updated https://github.com/EspressoSystems/espresso-network/commit/cafe619ce96873d98b0a0f49138f65b62dd96e0c

contracts/rust/deployer/src/lib.rs
alysiahuggins: oh forgot to remove this, thanks! https://github.com/EspressoSystems/espresso-network/commit/651ec5fe5bf5717556920d6b4d4429940b708091

sequencer/src/bin/deploy.rs
```diff
@@ -782,6 +794,10 @@ async fn main() -> anyhow::Result<()> {
             .write(true)
             .open(out)?;
         contracts.write(file)?;
+        // Also write to stdout so users can see output immediately
+        if !opt.quiet {
+            contracts.write(stdout())?;
```
sveitser, line 799: For binaries that we also run on ECS this isn't ideal, because these won't be JSON logs so they won't parse. Probably not a huge issue here but in general we don't want to print anything to stdout that isn't JSON if `RUST_LOG_FORMAT=json` is set. For logging we handle this with the helpers in `logging.rs`.

If you want to keep the API we have here you can invert the flag instead and print to stdout if `--verbose` is passed.

PR #3851, opened 2025-12-19
Binary upgrade test

pls148:
Closes #<ISSUE_NUMBER>


scripts/binary-upgrade-tests/helpers
```diff
@@ -0,0 +1,60 @@
...
+export ESPRESSO_SEQUENCER_PERMISSIONED_STAKE_TABLE_ADDRESS=0x8ce361602b935680e8dec218b820ff5056beb7af
+export ESPRESSO_SEQUENCER_FEE_CONTRACT_PROXY_ADDRESS=0xa15bb66138824a1c7167f5e85b957d04dd34e468
+export ESPRESSO_SEQUENCER_STAKE_TABLE_PROXY_ADDRESS=0xed1db453c3156ff3155a97ad217b3087d5dc5f6e
+
+
+cleanup_docker_compose() {
+    # do some cleanup
+    just down -v || true
+    docker stop $(docker ps -aq) || true
+    docker rm $(docker ps -aq) || true
```
sveitser, line 23: Can we do something like in the native demo script where we try avoid taking down everything running in docker?

scripts/binary-upgrade-tests/helpers
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
```
sveitser, line 2: probably want something like `set -euo pipefail` to exit with error in case of failure

PR #3852, opened 2025-12-19
Jb/light client header

jbearer:
See issues 
* https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510082563925?focus=true
* https://app.asana.com/1/1208976916964769/project/1210128724229719/task/1211510082563949?focus=true

### This PR:
* Introduces a `HeaderProof` type for verifying headers via Merkle proof
* Implements the `/light-client/header` endpoint
* Implements a method on the light client to fetch and verify headers

### DIFF
--- a/light-client/src/state.rs
+++ b/light-client/src/state.rs
@@ -3,6 +3,7 @@
 use std::borrow::Cow;
 
 use anyhow::{ensure, Context, Result};
+use committable::Committable;
 use espresso_types::{Header, Leaf2, SeqTypes};
 use hotshot_query_service::{
     availability::{LeafId, LeafQueryData},
@@ -100,7 +101,21 @@ where
                 // header, we can ask the server for an inclusion proof for the requested header
                 // relative to the Merkle root in the upper bound leaf.
                 let proof = self.server.header_proof(leaf.height(), id).await?;
-                return proof.verify(leaf.header().block_merkle_tree_root());
+                let header = proof
+                    .verify(leaf.header().block_merkle_tree_root())
+                    .context("invalid header proof")?;
+
+                // The server has given us a header and correctly proved it finalized, but we still
+                // need to verify that it actually gave us the header we requested.
+                ensure!(
+                    header_matches_id(&header, id),
+                    "server returned a valid header proof for the wrong header (requested header \
+                     {id}, got header {} with hash {})",
+                    header.height(),
+                    header.commit(),
+                );
+
+                return Ok(header);
             }
         }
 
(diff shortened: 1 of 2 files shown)


light-client/src/consensus/header.rs
```diff
@@ -0,0 +1,126 @@
...
+        let mt = BlockMerkleTree::from_elems(
+            Some(BLOCK_MERKLE_TREE_HEIGHT),
+            [leaves[0].block_hash(), leaves[1].block_hash()],
+        )
+        .unwrap();
+        let proof = HeaderProof::new(
+            leaves[0].header().clone(),
+            mt.lookup(1).expect_ok().unwrap().1,
+        );
+        proof.verify(mt.commitment()).unwrap_err();
```
Sneh1999: nit: we dont have an assert statement here, so we dont know if this is the expected error we want. Maybe should add an assert here?

light-client/src/consensus/header.rs
```diff
@@ -0,0 +1,126 @@
...
+                    [leaves[0].block_hash(), leaves[1].block_hash()],
+                )
+                .unwrap()
+            })
+            .collect::<Vec<_>>();
+        let proof = HeaderProof::new(
+            leaves[0].header().clone(),
+            mts[0].lookup(1).expect_ok().unwrap().1,
+        );
+        proof.verify(mts[1].commitment()).unwrap_err();
```
Sneh1999: same here

sequencer/src/api/light_client.rs
```diff
@@ -158,16 +181,132 @@ where
             get_leaf_proof(state, requested, finalized, fetch_timeout).await
         }
         .boxed()
+    })?
+    .get("header", move |req, state| {
```
Sneh1999, line 173: nit: not a big deal but I think we can make the code a bit smaller by introducing a helper function like:
```
fn bad_param(name: &'static str, err: impl std::fmt::Display) -> Error {
    Error::Custom {
        message: format!("{name}: {err:#}"),
        status: StatusCode::BAD_REQUEST,
    }
}

fn not_found(msg: impl Into<String>) -> Error {
    Error::Custom {
        message: msg.into(),
        status: StatusCode::NOT_FOUND,
    }
}
```

and then using it like 
```
if let Some(height) = req.opt_integer_param("height").map_err(|e| bad_param("height", e))? {
                BlockId::Number(height)
            }
```

light-client/src/state.rs
```diff
@@ -223,4 +261,81 @@ mod test {
...
+            SqliteStorage::default().await.unwrap(),
+            client.clone(),
+            client.genesis(),
+        );
+        client.return_invalid_proof(1).await;
+        lc.fetch_header(BlockId::Number(1)).await.unwrap_err();
+        lc.fetch_header(BlockId::Hash(client.leaf(1).await.block_hash()))
+            .await
+            .unwrap_err();
+        lc.fetch_header(BlockId::PayloadHash(client.leaf(1).await.payload_hash()))
```
Sneh1999: nit: same here

light-client/src/state.rs
```diff
@@ -223,4 +261,81 @@ mod test {
...
+        let db = SqliteStorage::default().await.unwrap();
+
+        // Start with an upper bound, so that `fetch_header` goes through the `client.header_proof`
+        // path.
+        db.insert_leaf(client.leaf(2).await).await.unwrap();
+
+        let lc = LightClient::from_genesis(db, client.clone(), client.genesis());
+        client.return_wrong_leaf(1, 0).await;
+        lc.fetch_leaf(LeafId::Number(1)).await.unwrap_err();
+        lc.fetch_leaf(LeafId::Hash(client.leaf(1).await.hash()))
```
Sneh1999: nit: and here

PR #3858, opened 2025-12-30
Implement light client stake table fetching

jbearer:


light-client/src/state.rs
```diff
@@ -122,74 +194,199 @@ where
...
+        // Replay one epoch at a time from the lower bound stake table to the requested epoch.
+        for epoch in *lower_bound + 1..=*epoch {
+            let events = self
+                .server
+                .stake_table_events(EpochNumber::new(epoch))
+                .await?;
+            tracing::debug!(epoch, num_events = events.len(), "reconstruct stake table");
+            for event in events {
+                tracing::debug!(epoch, ?event, "replay event");
+                if let Err(err) = stake_table.apply_event(event).context("applying event")? {
```
Sneh1999, line 343: why are we ignoring this error?

light-client/src/state.rs
```diff
@@ -122,74 +194,199 @@ where
...
+            prev_quorum = next_quorum;
+        }
+
+        // Finally, add the reconstructed stake table to cache and storage, and return it.
+        if let Err(err) = self.db.insert_stake_table(epoch, &stake_table).await {
+            // If this fails, we can still successfully return the stake table that we have in
+            // memory right now, so this is just a warning.
+            tracing::warn!(%epoch, "failed to cache stake table: {err:#}");
+        }
+        Ok(self.cache_stake_table(epoch, prev_quorum).await)
```
Sneh1999, line 378: nit: this condition can have `Ok(Err(err))` which is bit confusing for the client? I wonder instead of doing this we should separate out the Ok() and self.cache_stake_table call

light-client/src/state.rs
```diff
@@ -362,4 +559,132 @@ mod test {
...
+    async fn test_fetch_stake_table_invalid() {
+        let db = SqliteStorage::default().await.unwrap();
+        let client = TestClient::default();
+        let genesis = client.genesis().await;
+        let lc = LightClient::from_genesis(db.clone(), client.clone(), genesis.clone());
+
+        let epoch = genesis.first_epoch_with_dynamic_stake_table + 5;
+        client.return_invalid_quorum(epoch).await;
+        let err = lc.quorum_for_epoch(epoch).await.unwrap_err();
+        tracing::info!("quorum_for_epoch failed as expected: {err:#}");
```
Sneh1999: nit: I think you meant to add an assert here?

sequencer/src/api/light_client.rs
```diff
@@ -196,6 +201,83 @@ where
             get_header_proof(state, root, requested, fetch_timeout).await
         }
         .boxed()
+    })?
+    .get("stake_table", move |req, state| {
+        async move {
+            let epoch: u64 = req.integer_param("epoch").map_err(|err| Error::Custom {
```
Sneh1999: nit: we can follow the same pattern and use bad_param and not_found funcs here similar to what we did in last PR

light-client/src/state.rs
jbearer: `apply_event` returns a `Result<Result<_>>`. The outer result we respect (the `?` at the end of this line). The inner result indicates something was wrong with the event itself, rather than our ability to apply it, e.g. a registration event without a valid signature. Canonically we treat these events as nops so the correct behavior is to just ignore and continue applying events.

light-client/src/state.rs
jbearer: Not sure I understand, `cache_stake_table` is infallible (returns `Arc<StakeTable>`, no `Result`), where does the `Ok(Err(err))` potentially come from?

sequencer/src/api/light_client.rs
jbearer: haha you're right I wrote this first and forgot to update it with that suggestion on the last PR

light-client/src/state.rs
Sneh1999: oh true! you are right

light-client/src/state.rs
Sneh1999: Makes sense!

PR #3861, opened 2026-01-05
fix: implementation contract used as target

sveitser:
Fixes errors like

      Error: Invalid contract type for timelock
      derivation: ESPRESSO_SEQUENCER_ESP_TOKEN_V2_ADDRESS

If we do the ownership transfer after an upgrade we need to explicitly use the proxy.

Small refactoring to make some runtime errors impossible and convert some late errors into erros during clap argument parsing.

### DIFF
--- a/sequencer/src/bin/deploy.rs
+++ b/sequencer/src/bin/deploy.rs
@@ -623,7 +623,7 @@ async fn main() -> anyhow::Result<()> {
         })?;
 
         args_builder.timelock_operation_type(timelock_operation_type);
-        let target_contract = opt.target_contract.clone().ok_or_else(|| {
+        let target_contract = opt.target_contract.ok_or_else(|| {
             anyhow::anyhow!(
                 "Must provide --target-contract or ESPRESSO_TARGET_CONTRACT env var when \
                  scheduling timelock operation"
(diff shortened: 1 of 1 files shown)


contracts/rust/deployer/src/lib.rs
```diff
@@ -236,6 +236,33 @@ pub enum Contract {
...
+pub enum OwnableContract {
+    #[value(alias = "feecontract", alias = "FeeContract")]
+    FeeContractProxy,
+    #[value(alias = "lightclient", alias = "LightClient")]
+    LightClientProxy,
+    #[value(alias = "staketable", alias = "StakeTable")]
+    StakeTableProxy,
+    #[value(alias = "esptoken", alias = "EspToken")]
+    EspTokenProxy,
+    #[value(alias = "rewardclaim", alias = "RewardClaim")]
```
alysiahuggins, line 250: Just want to confirm my assumption here. This enum was created instead of using the `Ownable.sol` interface because not every contract that has the ownership structure, inherits from `Ownable.sol`. e.g. `RewardClaim.sol` doesn't inherit from `Ownable.sol` but inherits from `AccessControl.sol`. Maybe a comment incase someone thinks this is a potential refactoring opportunity when in fact it's very intentional.

sveitser: I didn't think too much about it, I just wanted an enum that only contains the contracts for which ownership is defined so that functions that needs such contract can express that requirement on the type level.

PR #3862, opened 2026-01-05
Add range namespace endpoint to sdk

Sneh1999:
# Description
Created a go endpoint to get all transactions for a given namespace within a specific range

---
  - https://app.asana.com/0/0/1212672252368095

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -896,7 +896,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -907,7 +907,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -2224,7 +2224,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.12.1",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -2978,7 +2978,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -3906,7 +3906,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -4155,7 +4155,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -6543,7 +6543,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi 0.5.2",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


sdks/go/client/client.go
```diff
@@ -105,6 +105,30 @@ func (c *Client) FetchTransactionByHash(ctx context.Context, hash *types.TaggedB
...
+	var res []types.NamespaceTransactionsRangeData
+	if err := c.get(ctx, &res, "availability/block/%d/%d/namespace/%d", from, until, namespace); err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func (c *Client) FetchLimits(ctx context.Context) (types.LimitsData, error) {
+	var res types.LimitsData
+	if err := c.get(ctx, &res, "availability/limits"); err != nil {
```
jjeangal, line 126: How does the `limits` API call work, what is this really meant for?

sdks/go/client/client_test.go
```diff
@@ -204,3 +204,41 @@ func TestExplorerFetchTransactionByHash(t *testing.T) {
...
+	namespace := uint64(22266222)
+	startHeight := uint64(6386698)
+	endHeight := uint64(6386700)
+
+	blocksWithNamespaceTransactions, err := client.FetchNamespaceTransactionsInRange(ctx, startHeight, endHeight, namespace)
+	if err != nil {
+		t.Fatal("failed to fetch namespace transactions in range", err)
+	}
+
+	if len(blocksWithNamespaceTransactions) == 0 {
```
jjeangal: Could we instead compare the length to the exact amount of transactions in that interval? To make sure that call doesn't miss any transaction.

sdks/go/client/client.go
Sneh1999: its to get limits for the range

sdks/go/client/client_test.go
Sneh1999: yes sounds good, will do that!

sdks/go/client/client.go
jjeangal: I'm not sure I really understand. What defines that limit? Is that the limit for the amount of transactions that can be queried from the api?

sdks/go/types/common/types.go
```diff
@@ -46,6 +46,22 @@ type TransactionQueryData struct {
...
+	Proof struct {
+		V1 struct {
+			NsIndex   uint64           `json:"ns_index"`
+			NsPayload NamespacePayload `json:"ns_payload"`
+			NsProof   NamespaceProof   `json:"ns_proof"`
+		} `json:"V1"`
+	} `json:"proof"`
+	Transactions []Transaction `json:"transactions"`
+}
+
```
zacshowa, line 63: Why don't we define these as named structs independently and compose them more idiomatically?

E.g.


```suggestion
type V1NamespaceTransactionsRangeProof struct{
    NsIndex   uint64                            `json:"ns_index"`
    NsPayload NamespacePayload  `json:"ns_payload"`
    NsProof   NamespaceProof         `json:"ns_proof"`
}

type NamespaceTransactionsRageProof struct{
    V1 V1NamespaceTransactionsRangeProof  `json:V1`
}

type NamespaceTransactionsRangeData struct {
    Proof V1NamespaceTransactionsRangeProof `json:"proof"`
    Transactions []Transaction `json:"transactions"`
}

```

sdks/go/client/client_test.go
```diff
@@ -204,3 +204,41 @@ func TestExplorerFetchTransactionByHash(t *testing.T) {
 		t.Fatal("failed to fetch block height", err)
 	}
 }
+
+func TestNamespaceTransactionsInRange(t *testing.T) {
+	ctx := context.Background()
+	client := NewClient("https://query-0.decaf.testnet.espresso.network")
```
zacshowa: I understand why we did this, It's easy and it's how we do it in the rest of the code. However, it seems like trying to hit decaf, even more so one query node, could make this test brittle when we could figure out some method of mocking the behavior here. 

For example, say Query-0 is down while some CI that tests this is running. Suddenly, the CI breaks and needs to be re-ran. The same goes for local testing with other similar failure points.

sdks/go/client/client.go
Sneh1999: Yes basically how many blocks you can query at a time

zacshowa (approved):
The core logic looks good to me.

I also have a comment about the types layout, and some food for thought about our testing strategy.

sdks/go/client/client_test.go
Sneh1999: we dont have a mock setup for our tests right now and that seems beyond the scope of this PR, can definitely look into this in future. I can make it point to the load balancer though so its not affected by one query node 🙌

sdks/go/client/client.go
zacshowa: Yeah it's to limit the amount of transactions that can be fetched at any given time. The limit is defined by the node via block range rather than based on data size. That allows the query node to avoid loading everything to determine what it can manage to send back to the client.

sdks/go/types/common/types.go
zacshowa: I'm only suggesting this so we don't have issues (or less than readable code) in the future when constructing, or manipulating, any of these structs in the future. For [context](https://stackoverflow.com/questions/53036159/does-go-support-nested-type-declarations-or-literal-assignments-to-anonymous-str)

Sneh1999: yes sounds good, I can change this!

PR #3867, opened 2026-01-06
Add fetch leaves by range in light client

Sneh1999:
### This PR:
Adds the ability to fetch leaves by range in light client

### Key places to review:
light-client/src/state.rs

### DIFF
--- a/light-client/src/state.rs
+++ b/light-client/src/state.rs
@@ -487,8 +487,6 @@ mod test {
             client.clone(),
             client.genesis(),
         );
-        // If there is an invalid proof for any leaf in the range,
-        // the entire fetch should fail
         client.return_invalid_proof(2).await;
         let err = lc.fetch_leaves_in_range(1, 3).await.unwrap_err();
         assert!(


light-client/src/state.rs
```diff
@@ -89,6 +89,99 @@ where
         self.fetch_leaf_from_server(id, known_finalized).await
     }
 
+    /// Fetches leaves in range [start_height, end_height]
```
jbearer: Nit: it's a bit confusing that `end_height` is inclusive here but exclusive in `get_leaves_in_range`

light-client/src/state.rs
```diff
@@ -89,6 +89,99 @@ where
...
+            start_height < end_height,
+            "invalid range: start must be < end"
+        );
+        let known_end_leaf = self.fetch_leaf(LeafId::Number(end_height)).await?;
+        ensure!(
+            known_end_leaf.height() == end_height as u64,
+            "fetched end leaf has unexpected height: expected {}, got {}",
+            end_height,
+            known_end_leaf.height()
+        );
```
jbearer: Nit: this should be unnecessary, because `self.fetch_leaf` already guarantees that on success it returns only the correct leaf

light-client/src/state.rs
Sneh1999: I see that everything is exclusive in the repo, so I will make this exclusive as well

light-client/src/state.rs
Sneh1999: Makes sense!

light-client/src/state.rs
```diff
@@ -455,4 +457,66 @@ mod test {
...
+    #[test_log::test]
+    async fn test_fetch_leaves_in_range_invalid_proof() {
+        let client = TestClient::default();
+        let lc = LightClient::from_genesis(
+            SqliteStorage::default().await.unwrap(),
+            client.clone(),
+            client.genesis(),
+        );
+        // If there is an invalid proof for any leaf in the range,
+        // the entire fetch should fail
```
jbearer: I think this comment is actually not true? It would only fail if the last leaf in the range has an invalid proof (which is the case here, since 2 is the last leaf in the exclusive range [1, 3)). But we only ask for a proof for this last leaf, the other leaves are verified by parent hash

jbearer (approved):

PR #3871, opened 2026-01-07
Fix namespace range point in multi node client

Sneh1999:
# Description
Fixes the endpoint in multi node client and adds a test for namespace range endpoint

### DIFF
--- a/sdks/go/client/multiple_nodes_test.go
+++ b/sdks/go/client/multiple_nodes_test.go
@@ -213,7 +213,7 @@ func TestApiWithSingleEspressoDevNode(t *testing.T) {
 
 func TestNamespaceTransactionsInRangeForMultiClient(t *testing.T) {
 	ctx := context.Background()
-	hotshotURLs := []string{"https://query-0.decaf.testnet.espresso.network", "https://query-0.decaf.testnet.espresso.network"}
+	hotshotURLs := []string{"https://query-0.decaf.testnet.espresso.network", "https://query-1.decaf.testnet.espresso.network"}
 	client, err := NewMultipleNodesClient(hotshotURLs)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sdks/go/client/multiple_nodes_client.go
```diff
@@ -70,9 +70,9 @@ func (c *MultipleNodesClient) FetchTransactionByHash(ctx context.Context, hash *
 	return res, nil
 }
 
-func (c *MultipleNodesClient) FetchNamespaceTransactionsInRange(ctx context.Context, namespace uint64, fromHeight uint64, toHeight uint64) ([]types.NamespaceTransactionsRangeData, error) {
+func (c *MultipleNodesClient) FetchNamespaceTransactionsInRange(ctx context.Context, fromBlock uint64, toBlock uint64, namespace uint64) ([]types.NamespaceTransactionsRangeData, error) {
 	var res []types.NamespaceTransactionsRangeData
-	if err := c.getWithMajority(ctx, &res, "availability/namespace/%d/transactions/range/%d/%d", namespace, fromHeight, toHeight); err != nil {
+	if err := c.getWithMajority(ctx, &res, "availability/block/%d/%d/namespace/%d", fromBlock, toBlock, namespace); err != nil {
```
lukeiannucci, line 75: can you help me understand the difference between the two endpoints and why we moved from the first way?

lukeiannucci (approved):
lgtm! just a question

sdks/go/client/multiple_nodes_client.go
Sneh1999: I made a mistake here, the one in single client I put was right and here for some reason I added wrong url, this url doesnt exist lol 🫠

PR #3876, opened 2026-01-09
Implement light client payload and namespace fetching and verification

jbearer:


light-client/src/consensus/namespace.rs
```diff
@@ -0,0 +1,173 @@
...
+                    "received trivial proof for missing namespace, but requested namespace \
+                     {namespace} is present at position {ix:?}"
+                );
+            }
+            return Ok(vec![]);
+        };
+
+        let (txs, ns) = proof
+            .proof
+            .verify(
```
Sneh1999, line 54: why cant the verify function also take in a namespace and check if the `namespace == ns`. This way the caller doesnt have to do 
```
ensure!(
            ns == namespace,
            "proof is for wrong namespace {ns}, expected namespace {namespace}"
        );
```

jbearer: It could. The `NamespaceProof::verify` function in the light client crate (i.e. the function that is being implemented here) does exactly this. This call is to the low-level crypto API for namespace proofs and I guess the API is just different. I don't want to go changing that one since it's used many other places

Sneh1999 (approved):
LGTM!

PR #3877, opened 2026-01-09
Add `token` API endpoint for total supply

ss-es:


sequencer/api/token.toml
```diff
@@ -0,0 +1,3 @@
+[route.total_minted_supply]
+PATH = ["total_minted_supply"]
+DOC = "Get the total supply of the espresso token, counting only tokens that have been minted. Notably, this does not include unminted tokens from (for example) unclaimed rewards, or other sources."
```
sveitser: Add a comment about the caching to make it clear that this doesn't always reflect the exact on-chain value immediately?

sveitser: And mention it's in token units with an example like "123.123"?

sequencer/src/api.rs
```diff
@@ -261,6 +282,62 @@ impl<N: ConnectedNetwork<PubKey>, D: Sync, V: Versions, P: SequencerPersistence>
     }
 }
 
+impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> TokenDataSource<SeqTypes>
+    for ApiState<N, P, V>
+{
+    async fn get_total_supply(&self) -> anyhow::Result<U256> {
```
sveitser: I think the naming could be a bit more descriptive to make it clearer that this is the total supply on L1 and distinguish it from the total supply in Espresso. For example by adding an "l1" suffix.

sequencer/api/token.toml
sveitser:
```suggestion
DOC = "Get the total supply of the espresso token, counting only tokens that have been minted on Ethereum. Notably, this does not include unminted tokens from (for example) unclaimed rewards, or other sources."
```

PR #3879, opened 2026-01-12
Switch from per block to per epoch reward distribution ( V5 Upgrade) 

imabdulbasit:
addresses: https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1212659845261620

https://www.notion.so/espressosys/Per-Epoch-reward-distribution-2cc2431b68e98056baa0f71702315204

### DIFF
--- a/tests/upgrades.rs
+++ b/tests/upgrades.rs
@@ -120,6 +120,14 @@ async fn run_upgrade_test(genesis_path: &str, upgrade: Upgrade) -> Result<()> {
         txn_count_increment: 2 * expected_block_height,
         global_timeout: Duration::from_secs(expected_block_height as u64 * 3),
         first_reward_block,
+        // For V5 (EPOCH_REWARD_VERSION), require a claim whose merkle proof was built against
+        // an LC state at or after first_reward_block, proving the new per-epoch reward tree
+        // is in use.
+        claim_after_lc_block: if upgrade.target >= EPOCH_REWARD_VERSION {
+            first_reward_block
+        } else {
+            None
+        },
         ..Default::default()
     };
 
--- a/scripts/cleanup-process-compose
+++ b/scripts/cleanup-process-compose
@@ -74,6 +74,8 @@ kill_project_processes() {
 		if [[ -n "$proc_cwd" && "$proc_cwd" == "$PROJECT_DIR"* ]]; then
 			echo "Killing $proc_name (pid $pid) from $proc_cwd"
 			kill -"$signal" "$pid" || true
+		elif [[ -n "$proc_cwd" ]]; then
+			echo "Skipping $proc_name (pid $pid): cwd $proc_cwd does not match $PROJECT_DIR"
 		fi
 	done
 }
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -71,6 +71,11 @@ pub struct TestRequirements {
     pub max_consecutive_blocks_without_tx: u64,
     /// If Some, wait until rewards_claimed > 0 before exiting. Value used for diagnostic logging.
     pub first_reward_block: Option<u64>,
+    /// If Some, the test requires observing a reward claim that is provably from the new regime:
+    /// wait until LC finalizes this block, snapshot claimed_rewards at that point, then wait for
+    /// claimed_rewards to increase. This avoids exiting early on a claim made against an older
+    /// LC state (before the new reward scheme was active).
+    pub claim_after_lc_block: Option<u64>,
 }
 
 impl Default for TestRequirements {
@@ -84,6 +89,7 @@ impl Default for TestRequirements {
             block_timeout: Duration::from_secs(60),
             max_consecutive_blocks_without_tx: 10,
             first_reward_block: None,
+            claim_after_lc_block: None,
         }
     }
 }
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -26,6 +26,7 @@ pub async fn assert_native_demo_works(requirements: TestRequirements) -> Result<
     let mut blocks_without_tx = 0;
     let mut iteration = 0u64;
     let mut prev_printed = None;
+    let mut lc_threshold_snapshot: Option<U256> = None;
 
     loop {
         match tokio::time::timeout(requirements.block_timeout, sub.next()).await {
@@ -69,8 +70,8 @@ pub async fn assert_native_demo_works(requirements: TestRequirements) -> Result<
 
         if start.elapsed() > requirements.global_timeout {
             panic!(
-                "Timeout waiting for block height, transaction count, and light client updates to \
-                 increase."
+                "Timeout waiting for block height, transaction count, light client updates, and \
+                 reward claim after LC threshold to increase."
             );
         }
 
@@ -91,6 +92,24 @@ pub async fn assert_native_demo_works(requirements: TestRequirements) -> Result<
                 new.rewards_claimed,
                 new.block_height.unwrap()
             );
+        }
+
+        if let Some(lc_block) = requirements.claim_after_lc_block {
+            // Wait for the LC to commit a state at or past the threshold block so that
+            // any subsequent claim uses a proof from the new regime.
+            if new.light_client_finalized_block_height < lc_block {
+                continue;
+            }
+            let snapshot = lc_threshold_snapshot.get_or_insert_with(|| {
+                println!(
+                    "LC reached threshold block {lc_block}, snapshotting claimed_rewards={}",
+                    new.rewards_claimed
+                );
+                new.rewards_claimed
+            });
+            if new.rewards_claimed <= *snapshot {
+                continue;
+            }
         }
 
         break;


types/src/v0/impls/header.rs
```diff
@@ -676,6 +746,223 @@ impl Header {
...
+                    %epoch,
+                    %prev_epoch,
+                    "missing V6 epoch rewards at boundary, spawning calculation now"
+                );
+
+                if !reward_calculator.is_calculating(prev_epoch) {
+                    reward_calculator.spawn_background_task(
+                        prev_epoch,
+                        epoch_height,
+                        validated_state.reward_merkle_tree_v2.clone(),
```
bfish713: Do we know for sure this is the correct merkle tree to pass in here, I feel it could be arbitrarily old.  Maybe we can check the `header.reward_merkle_tree_root == validated_state.reward_merkle_tree_v2.root` and if false just pass in the empty tree with the root from the `prev_epoch_header`

imabdulbasit: made that change in this commit https://github.com/EspressoSystems/espresso-network/pull/3879/commits/bcbc39d7732ed251ebe3ab491503f1d9b40687cf

crates/espresso/types/src/v0/v0_5/header.rs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        counts.as_slice().serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<LeaderCounts, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let vec: Vec<u16> = Vec::deserialize(deserializer)?;
+        if vec.len() != MAX_VALIDATORS {
```
sveitser, line 34: Why the indirection via Vec? So that we output a nicer error message?

types/src/v0/v0_5/header.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub(crate) ns_table: NsTable,
+    pub(crate) block_merkle_tree_root: BlockMerkleCommitment,
+    pub(crate) fee_merkle_tree_root: FeeMerkleCommitment,
+    pub(crate) fee_info: FeeInfo,
+    pub(crate) builder_signature: Option<BuilderSignature>,
+    pub(crate) reward_merkle_tree_root: RewardMerkleCommitmentV2,
+    pub(crate) total_reward_distributed: RewardAmount,
+    pub(crate) next_stake_table_hash: Option<StakeTableHash>,
+    #[serde(with = "leader_counts_serde")]
+    pub(crate) leader_counts: [u16; 100],
```
sveitser: Maybe a constant for 100? and a comment, why it's 100.

sveitser: There is already a type alias elsewhere

crates/espresso/types/src/v0/v0_5/header.rs
imabdulbasit: yes

sequencer/src/request_response/request.rs
```diff
@@ -61,6 +63,8 @@ pub enum Response {
     VidShare(VidShare),
     /// A response for a state certificate at a given epoch
     StateCert(LightClientStateUpdateCertificateV2<SeqTypes>),
+    /// A response for the header at the last block of a given epoch
+    EpochHeader(Box<Header>),
```
ss-es: I don't think we should respond with a `Box`

types/src/v0/traits.rs
```diff
@@ -80,6 +80,23 @@ pub trait StateCatchup: Send + Sync {
             .await
     }
 
+    /// Fetch the header at the given height without retrying on transient errors.
+    ///
+    /// This is used for epoch rewards catchup where we need the header
+    /// of the last block of the previous epoch.
+    async fn try_fetch_header(&self, retry: usize, height: u64) -> anyhow::Result<Header>;
```
ss-es: why do we need a new method/catchup path for this? don't we already have the ability to fetch the full leaf chain?

sequencer/src/request_response/catchup/state.rs
```diff
@@ -56,6 +56,29 @@ impl<I: NodeImplementation<SeqTypes>, N: ConnectedNetwork<PubKey>, P: SequencerP
...
     }
 
+    async fn try_fetch_header(&self, _retry: usize, height: u64) -> anyhow::Result<Header> {
+        let timeout_duration = self.config.request_batch_interval * 3;
+
+        let response_validation_fn = move |_request: &Request, response: Response| async move {
+            let Response::EpochHeader(header) = response else {
+                return Err(anyhow::anyhow!("expected epoch header response"));
+            };
+            Ok(*header)
```
ss-es: for example, are we just blindly trusting whatever header we receive here? `try_fetch_header` doesn't have information to validate it (compare with `fetch_leaf`, which has the stake table and requires a full leaf chain)

but since the header is available via `fetch_leaf` anyway, I don't think we should re-implement the validation logic anywhere -- I think we should just use `fetch_leaf`

sequencer/src/api/sql.rs
ss-es, line 1456: if I understand correctly, this whole function is useless now right?

(not saying delete it in this PR because I think it's good to minimize the diff, but maybe in another PR)

types/src/v0/traits.rs
imabdulbasit: the leaf chain requires the stake table and we don't have that information where I am fetching it. Header is needed to get the leader counts.

sequencer/src/api/sql.rs
imabdulbasit: hmm I still see merklized state reconstructed in the tests and native demo and it uses this function. I'll double check

crates/hotshot/types/src/traits/storage.rs
```diff
@@ -169,7 +169,20 @@ pub async fn store_drb_input_impl<TYPES: NodeType>(
     storage: impl Storage<TYPES>,
     drb_input: DrbInput,
 ) -> Result<()> {
-    storage.store_drb_input(drb_input).await
+    for attempt in 1..=3 {
+        match storage.store_drb_input(drb_input.clone()).await {
+            Ok(()) => (),
```
ss-es: this should `return` shouldn't it?

crates/hotshot/types/src/traits/storage.rs
```diff
@@ -197,7 +210,24 @@ async fn store_drb_result_impl<TYPES: NodeType>(
     epoch: EpochNumber,
     drb_result: DrbResult,
 ) -> Result<()> {
-    storage.store_drb_result(epoch, drb_result).await
+    for attempt in 1..=3 {
+        match storage.store_drb_result(epoch, drb_result).await {
+            Ok(()) => (),
```
ss-es: same here

crates/versions/src/lib.rs
```diff
@@ -13,9 +13,10 @@ pub const VERSION_0_1: Version = version(0, 1);
 pub const FEE_VERSION: Version = version(0, 2);
 pub const EPOCH_VERSION: Version = version(0, 3);
 pub const DRB_AND_HEADER_UPGRADE_VERSION: Version = version(0, 4);
-pub const DA_UPGRADE_VERSION: Version = version(0, 5);
-pub const VID2_UPGRADE_VERSION: Version = version(0, 6);
-pub const MAX_SUPPORTED_VERSION: Version = DA_UPGRADE_VERSION;
+pub const EPOCH_REWARD_VERSION: Version = version(0, 5);
+pub const DA_UPGRADE_VERSION: Version = version(0, 6);
+pub const VID2_UPGRADE_VERSION: Version = version(0, 7);
```
ss-es, line 19: cc @mrain for the VID2 version change; do we want to change it or should it stay tied to the DA upgrade?

data/genesis/staging-new.toml
```diff
@@ -1,10 +1,28 @@
-base_version = "0.2"
-upgrade_version = "0.4"
-genesis_version = "0.2"
+base_version = "0.4"
+upgrade_version = "0.6"
```
ss-es: not that it particularly matters (I think @lukaszrzasik would change this anyway), but we probably only want the 0.4 -> 0.5 upgrade in staging right?

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -791,6 +791,18 @@ impl SqlStorage {
...
+    where
+        Types: NodeType,
+        Header<Types>: QueryableHeader<Types>,
+    {
+        let mut tx = self.read().await.map_err(|err| QueryError::Error {
+            message: err.to_string(),
+        })?;
+        let header = tx.load_header(block_id).await?;
+        Ok(header)
+    }
```
ss-es: are we still using this? I thought we were using the `fetch_leaf` path everywhere

sequencer/src/api/sql.rs
```diff
@@ -725,6 +693,23 @@ impl ChainConfigPersistence for Transaction<Write> {
...
+        self.upsert(
+            "reward_merkle_tree_v2_data",
+            ["height", "balances"],
+            ["height"],
+            [(height as i64, data)],
+        )
+        .await
+    }
+}
+
```
ss-es: I'm not sure about this (using the same transaction). Maybe it's fine, but this whole thing was very delicate with performance when I was testing originally. 

I really really think we shouldn't change it unless we have to, and if we do change it we should do it in a separate PR

sequencer/src/api.rs
```diff
@@ -1307,89 +1302,91 @@ pub(crate) trait RewardMerkleTreeDataSource: Send + Sync + Clone + 'static {
...
+                .load_reward_merkle_tree_v2(finalized_hotshot_height)
+                .await
+            {
+                Ok(tree) => tree,
+                Err(err) => {
+                    tracing::warn!(
+                        finalized_hotshot_height,
+                        "failed to load reward merkle tree: {err:#}"
+                    );
                     return Ok(());
```
ss-es, line 1354: the garbage collection must happen before this point

sequencer/src/api.rs
```diff
@@ -1307,89 +1302,91 @@ pub(crate) trait RewardMerkleTreeDataSource: Send + Sync + Clone + 'static {
...
+                    let gc_epoch = current_epoch.saturating_sub(2);
+                    (gc_epoch * epoch_height).min(finalized_hotshot_height)
+                } else {
+                    finalized_hotshot_height
                 }
+            } else {
+                finalized_hotshot_height
+            };
+
+            if let Err(err) = self.garbage_collect(gc_height).await {
```
ss-es, line 1369: if we're editing this function, we should move garbage collection to the right spot. on `release-mainnet-1.0.1-rc` we garbage collect based on `min(height, finalized_hotshot_height)` right after saving the tree

sequencer/src/api/sql.rs
ss-es: I think the separation proofs + garbage collection is fine though, I just think we should continue using `persist_tree` as previously implemented

crates/espresso/node/src/state.rs
```diff
@@ -392,18 +402,6 @@ where
...
-            .load_reward_merkle_tree_v2(parent_leaf.height())
-            .await
-            .context(
-                "Error starting the state storage update loop: failed to load RewardMerkleTreeV2 \
-                 for the previous height",
-            )?;
-
-        parent_state.reward_merkle_tree_v2 = reward_merkle_tree_v2.tree;
-    }
-
```
ss-es, line 406: I think we still need this? otherwise we restart the loop without a full reward tree

types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1019,358 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+    pub changed_accounts: HashSet<RewardAccountV2>,
+    /// Timestamp when this result was inserted into the cache.
+    pub inserted_at: std::time::Instant,
+}
+
+/// Manages epoch-based reward calculations in the background.
+#[derive(Debug, Default)]
+pub struct EpochRewardsCalculator {
+    /// Cached results by epoch
+    pub results: HashMap<EpochNumber, EpochRewardsResult>,
```
ss-es: this will be less concerning later but we're holding multiple (full?) merkle trees here. I think we should at the very least have some `Arc<RwLock<..>`s around here

types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1019,358 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+        if let Some(result) = self.results.get(&epoch) {
+            return Some(result.clone());
+        }
+
+        // Await pending calculation if exists
+        if let Some(handle) = self.pending.remove(&epoch) {
+            match handle.await {
+                Ok(Ok(result)) => {
+                    tracing::info!(%epoch, total = %result.total_distributed.0, "epoch rewards calculation completed");
+                    self.results.insert(epoch, result.clone());
```
ss-es: how many live copies of the tree do we need at a given time?

crates/espresso/types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1019,358 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+                .as_ref()
+                .fetch_leaf(epoch_last_block_height, stake_table, success_threshold)
+                .await
+                .with_context(|| {
+                    format!(
+                        "failed to fetch leaf at height {epoch_last_block_height} for epoch \
+                         {epoch}"
+                    )
+                })?;
+            let header = leaf.block_header();
```
ss-es, line 1215: I think you should expect the leaf doesn't have the full reward merkle tree in it, and you probably need to do catchup for the tree separately

iirc the only thing you can expect in the leaf is the merkle root commitment

crates/espresso/types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1019,358 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+                        .chain(v.delegators.keys().map(|d| RewardAccountV2(*d)))
+                })
+            })
+            .collect();
+
+        let missing_accounts: Vec<_> = accounts_to_update
+            .iter()
+            .filter(|account| reward_tree.lookup(**account).expect_not_in_memory().is_ok())
+            .cloned()
+            .collect();
```
ss-es, line 1274: not a strong opinion, but you could also do something like

`        let balances: Vec<_> = self
            .iter()
            .map(|(account, balance)| (*account, *balance))
            .collect();

        if balances.len() as u64 == num_leaves { no catchup needed } else { catchup }
`
I expect this would be faster than looking up every individual account that you want to update, because in this case you will likely be updating most or a near majority of the accounts (and it makes the catchup condition easier to understand)

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -1776,6 +1776,41 @@ impl EpochCommittees {
...
+    /// Returns the full Validator config if found.
+    ///
+    /// The index corresponds to the position in the `leader_counts` array in V6 headers.
+    pub fn get_validator_by_index(
+        &self,
+        epoch: &EpochNumber,
+        index: usize,
+    ) -> Option<AuthenticatedValidator<PubKey>> {
+        let committee = self.state.get(epoch)?;
+        let (bls_key, _) = committee.stake_table.get_index(index)?;
```
ss-es: how is the `committee.stake_table` ordered? is it canonical across all nodes/rpc providers etc?

the drb::election module has a canonical order that we could expose, but if this is guaranteed to be the same it's probably fine

sequencer/src/api/sql.rs
imabdulbasit: I think all the merklized state updates should be be one transaction.

crates/espresso/node/src/state.rs
imabdulbasit: we do the catchup anyway so think its not needed

crates/espresso/types/src/v0/impls/stake_table.rs
imabdulbasit: Its an indexmap and the order should always be the same

sequencer/src/api/sql.rs
imabdulbasit: but yeah I can do that in separate PR

types/src/v0/impls/reward.rs
imabdulbasit: just 2. we garbage collect everything else

crates/espresso/types/src/v0/impls/reward.rs
imabdulbasit: this only checks the number of accounts and not exact accounts?

data/genesis/staging-new.toml
imabdulbasit: I think I'll delete this file because we should have only one staing.toml

types/src/v0/impls/reward.rs
imabdulbasit: actually the whole calculation is wrapped in Mutex so and we only access it once so I don't think there is any point of wrapping this into RwLock. Also, I think get_result() can just remove the result from this map because we only use it once and don't need it otherwise

scripts/claim-rewards-loop
```diff
@@ -1,12 +1,26 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Load .env if present (process-compose may not propagate all vars)
+if [ -f .env ]; then
+  set -a; source .env; set +a
```
sveitser: Usually we don't want to do this because then we can't overwrite env vars anymore. Do we really need it? It feels weird that process compose may not propagate variables.

For the problem with the config file I think we can solve it in a better way.

sequencer/src/api/sql.rs
```diff
@@ -243,32 +243,62 @@ impl RewardMerkleTreeDataSource for SqlStorage {
...
-                "#,
+            let leaf = tx
+                .get_leaf(LeafId::<SeqTypes>::from(height as usize))
+                .await
+                .context(format!("leaf {height} not available"))?;
+
+            // For V5+, resolve to the last epoch boundary where proofs were stored.
+            // For V4, proofs exist at every height.
+            let proof_height = if leaf.header().version() < EPOCH_REWARD_VERSION
+                || epoch_height == 0
```
sveitser: Why do we need this `|| epoch_height = 0`?

crates/espresso/types/src/v0/impls/header.rs
```diff
@@ -678,6 +752,247 @@ impl Header {
         Ok(header)
     }
 
+    /// Calculate leader_counts for V5+ headers.
+    pub fn calculate_leader_counts(
+        parent_header: &Header,
+        height: u64,
+        leader_index: usize,
+        epoch_height: u64,
```
sveitser, line 768: This is a public function that would panic if `leader_index >= 100`.

crates/espresso/types/src/v0/impls/header.rs
```diff
@@ -678,6 +752,247 @@ impl Header {
...
+        let parent_counts = parent_header.leader_counts();
+
+        // If parent was the last block of an epoch, current block is epoch start
+        let is_epoch_start = is_last_block(height.saturating_sub(1), epoch_height);
+
+        if is_epoch_start || parent_counts.is_none() {
+            leader_counts[leader_index] = 1;
+        } else if let Some(parent_counts) = parent_counts {
+            leader_counts = *parent_counts;
+            leader_counts[leader_index] += 1;
```
sveitser, line 782: I think this is safe from overflow for now but it still feels like we shouldn't write the code such that it's correct only for our current choices of certain parameters that aren't explicitly mentioned anywhere here.

scripts/claim-rewards-loop
imabdulbasit: ah I didn't mean to commit this. I'll revert the changes

sequencer/src/api/sql.rs
```diff
@@ -372,6 +405,108 @@ impl RewardMerkleTreeDataSource for SqlStorage {
...
+    fn persist_reward_proofs(
+        &self,
+        node_state: &NodeState,
+        height: u64,
+        version: Version,
+    ) -> impl Send + Future<Output = anyhow::Result<()>> {
+        async move {
+            // For V4 (per-block rewards), rate-limit since this is called every block.
+            // For V5+ (epoch rewards), this is only called at epoch boundaries.
+            if !cfg!(any(test, feature = "testing"))
```
sveitser, line 474: I think this could have a comment why we need it and when it's safe to remove? IIRC we needed this at some point for reward claim e2e tests to work. cc @ss-es I would rather rewrite the test to work with epoch based rewards if that's the only why we need this gate.

imabdulbasit: https://app.asana.com/1/1208976916964769/project/1210098463466349/task/1213746664174854

crates/espresso/types/src/v0/impls/stake_table.rs
sveitser: @imabdulbasit I was looking into this and got quite confused. But I think this function is never used, let's just remove it?

types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1013,352 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+            *header
+                .leader_counts()
+                .expect("V6+ header must have leader_counts")
+        };
+
+        let membership = coordinator.membership().read().await;
+        let validators: Vec<_> = membership
+            .stake_table(Some(epoch))
+            .iter()
+            .sorted_by(|a, b| a.stake_table_entry.key().cmp(b.stake_table_entry.key()))
```
sveitser: @imabdulbasit I wonder if we could do something on the type level to avoid having to sort by keys in multiple places. The code code here and below where we construct `accounts_to_update` reach quite deep into types and it's difficult to reason about why the code is written like this, and if it's all fine and dandy.

types/src/v0/impls/reward.rs
```diff
@@ -1009,6 +1013,352 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
+    ) -> anyhow::Result<EpochRewardsResult> {
+        let mut total_distributed = U256::ZERO;
+        let mut changed_accounts = HashSet::new();
+
+        for (index, &count) in leader_counts.iter().enumerate() {
+            if count == 0 {
+                continue;
+            }
+
+            let validator = validators.get(index).context(format!(
```
sveitser: Similarly here, when passing around leader counts and validators separately the caller is reponsible for ensuring they are correct for each other. I think preferably we would have one place in the code where we build a variable that combines the two so that we only need to worry about how to do it correctly in one place. It's basically a mapping.

crates/espresso/node/src/api.rs
```diff
@@ -4630,6 +4632,467 @@ mod test {
...
+                assert!(total_distributed.0 > epoch4_last_reward.0,);
+                break;
+            }
+        }
+
+        Ok(())
+    }
+
+    // test actual rewards
+    // todo: test each account rewards by querying merklized state api
```
sveitser, line 4875: @imabdulbasit Do we currently have any tests to check that the correct leaders get the rewards with the new computation? I think it's worth having tests at least at one layer for this.

sveitser: I mean tests for the mapping from leader counts to validator.

tests/upgrades.rs
```diff
@@ -147,3 +154,12 @@ async fn test_native_demo_fee_to_drb_header_upgrade() -> Result<()> {
     )
     .await
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_native_demo_epoch_reward_upgrade() -> Result<()> {
+    run_upgrade_test(
+        "data/genesis/demo-epoch-reward-upgrade.toml",
+        Upgrade::new(DRB_AND_HEADER_UPGRADE_VERSION, EPOCH_REWARD_VERSION),
```
sveitser, line 170: I think there is currently nothing in the test that would assert anything about the epoch rewards being distributed. It's probably fine but would be nice to have reward claim E2E tested with the epoch rewards.

crates/espresso/types/src/v0/impls/stake_table.rs
imabdulbasit: removed

PR #3882, opened 2026-01-13
Add support for fetching headers in range

Sneh1999:
# Description
Add support for fetching headers in range

### DIFF
--- a/light-client/src/state.rs
+++ b/light-client/src/state.rs
@@ -9,7 +9,7 @@ use espresso_types::{
     select_active_validator_set, Header, Leaf2, PubKey, SeqTypes, StakeTableState,
 };
 use hotshot_query_service::{
-    availability::{HeaderQueryData, LeafId, LeafQueryData},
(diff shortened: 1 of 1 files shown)


light-client/src/state.rs
```diff
@@ -184,6 +184,25 @@ where
         Ok(leaves)
     }
 
+    /// Fetches header in range [start_height, end_height)
+    pub async fn fetch_headers_in_range(
+        &self,
+        start_height: usize,
+        end_height: usize,
+    ) -> Result<Vec<HeaderQueryData<SeqTypes>>> {
```
jbearer: Elsewhere in this interface we just use `Header`, not `HeaderQueryData`. I don't have a preference since `HeaderQueryData` appears to be just a wrapper around `Header`, but we should be consistent. Can you either change this to `Header` or change other uses of `Header` to `HeaderQueryData`?

Sneh1999: Makes sense, I will just change this to Header

PR #3886, opened 2026-01-14
Implement light client block height method

jbearer:


light-client/src/storage.rs
```diff
@@ -142,6 +150,14 @@ impl Storage for SqliteStorage {
         SqliteOptions::default().connect().await
     }
 
+    async fn block_height(&self) -> Result<u64> {
+        let mut tx = self.pool.begin().await?;
+        let (height,) = query_as("SELECT COALESCE(max(height) + 1, 0) FROM leaf")
```
Sneh1999, line 155: why are adding + 1 here? because if we do ` let latest_known = self.db.block_height().await?;` in the code above and this condition is false ` if latest_from_server > latest_known {` this means no new blocks have been produced it will return the finalized height to be block_height + 1 even though it might have not been finalized yet? Is my understanding correct?

jbearer: The block height is defined as one more than the highest block _number_. Because block numbers start at 0. So e.g. if `max(height)` is 10 that means there are 11 total blocks

Sneh1999: I see, makes sense!

PR #3894, opened 2026-01-15
feat(staking-cli): add private key signing and calldata export

sveitser:
Add --private-key flag for raw key signing alongside mnemonic/ledger.

Add --export-calldata mode for Safe multisig users to output transaction data instead of executing. Validates via eth_call simulation using --sender-address (skip with --skip-simulation). Supports JSON, TOML, and human-readable output formats.

Consolidate all state-changing operations into Transaction enum with unified calldata generation for both execute and export modes.

Other changes:
- Move main.rs logic into cli.rs module for better encapsulation
- Simplify ValidSignerConfig::wallet() to return EthereumWallet directly
- Add AddressExt trait for cleaner address resolution from wallet
- Use trimmed format for ESP token display (e.g. "100 ESP" not "100.0")
- Reduce public API surface: make internal modules pub(crate)
- Remove funding.rs module, use Transaction::Transfer for ESP transfers

A simulated `RegisterValidatorV2` transaction can be seen here: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/22046676-7878-4e76-8782-be9bf66b8af7

Sorry, this is quite a big diff but we have very good test coverage for the staking-cli. The main two reasons for refactoring a lot are:

1. It's better to have calldata generation and execution go through the same code paths instead of having 2 code paths. The reasoning here is that it's difficult to test with calldata end-to-end but if we use the same code path for execution we can be confident we have implemented everything and it works.
2. Due to having a lot of code in `main.rs` we had to have a really big public API in the staking-cli. So moving this to the lib (cli.rs) we can avoid this and avoid exposing most of the crate.


staking-cli/src/cli.rs
```diff
@@ -0,0 +1,634 @@
...
+    match config.commands {
+        Commands::Init {
+            mnemonic,
+            private_key,
+            account_index,
+            ledger,
+        } => {
+            let mut config = toml::from_str::<Config>(include_str!("../config.decaf.toml"))?;
+            config.signer.mnemonic = mnemonic;
+            config.signer.private_key = private_key;
```
alysiahuggins, line 230: so this private key is stored on disk? anyway that can be avoided or provide a clear warning about it?

staking-cli/src/claim.rs
```diff
@@ -179,13 +125,39 @@ pub async fn unclaimed_rewards(
...
+        claimer_address,
+    )
+    .await?
+    else {
+        return Ok(None);
+    };
+
+    Ok(Some(Transaction::ClaimRewards {
+        reward_claim: data.reward_claim_address,
+        lifetime_rewards: data.claim_input.lifetime_rewards,
```
alysiahuggins, line 148: do we have to check that this is none zero or if data is returned it's because it is def non-zero?

staking-cli/src/cli.rs
sveitser: It can be stored in the config file, provided as argument or env var. It was a feature request we got from node operators to also support private keys instead of only mnemonics and  the ledger. In terms of security it's no worse than mnemonics that are already supported. It's not uncommon or bad to store secrets on disk or in env vars / secret managers as long as other security measures are taken to secure the infra and avoid keys leaking.

So it's mostly up to the user / node operators to secure their keys, or use a hardware wallet for better security. I think if someone isn't careful with their keys there's not much we can do here. We could potentially only support hardware Ethereum wallets but I think this would be inconvenient for us and for node operators and for our consensus keys we don't have hardware wallet support anyway.

alysiahuggins: ok got it @ feature request. It would be good to note (if not already done), a warning about choosing to use private keys

sveitser: aa13c9b4ce56c35bab0c1bbcb054dd5f2aa2ea32

staking-cli/src/claim.rs
sveitser: Here's the flow when a user calls staking-cli claim-rewards with no rewards balance:

  Summary

  The CLI fails early with a user-friendly error before calling the contract. The API returns 404 if the reward balance is zero.

  Check Locations

  1. API Endpoint (first check)

  Returns 404 NOT_FOUND if zero reward balance:

  https://github.com/EspressoSystems/espresso-network/blob/92e8026a03bec2f19d682c16857ee2acfe22d57a/sequencer/src/api/endpoints.rs#L247-L253

  2. staking-cli (handles 404)

  Returns Ok(None) when API returns 404:

  https://github.com/EspressoSystems/espresso-network/blob/92e8026a03bec2f19d682c16857ee2acfe22d57a/staking-cli/src/claim.rs#L80-L82

  3. staking-cli (converts to error)

  Converts None to user-facing error:

https://github.com/EspressoSystems/espresso-network/blob/aa13c9b4ce56c35bab0c1bbcb054dd5f2aa2ea32/staking-cli/src/cli.rs#L549-L556

  4. Contract (never reached in this case)

  If somehow a zero amount reached the contract, it would revert:

  https://github.com/EspressoSystems/espresso-network/blob/92e8026a03bec2f19d682c16857ee2acfe22d57a/contracts/src/RewardClaim.sol#L247

  And for already claimed rewards:

  https://github.com/EspressoSystems/espresso-network/blob/92e8026a03bec2f19d682c16857ee2acfe22d57a/contracts/src/RewardClaim.sol#L249


A test is here:

https://github.com/EspressoSystems/espresso-network/blob/aa13c9b4ce56c35bab0c1bbcb054dd5f2aa2ea32/staking-cli/src/claim.rs#L243-L269

alysiahuggins: ok thanks a lot for this

PR #3897, opened 2026-01-18
Add `total_distributed_supply` endpoint

ss-es:


sequencer/api/token.toml
```diff
@@ -1,3 +1,7 @@
 [route.get_total_minted_supply]
 PATH = ["total-minted-supply"]
 DOC = "Get the total supply of the espresso token, counting only tokens that have been minted on Ethereum. Notably, this does not include unminted tokens from (for example) unclaimed rewards, or other sources. The value is returned as a decimal in token units, equal to the raw `totalSupply` retrieved from the contract divided by 10^18. Calls are cached for an hour after retrieval, so this endpoint should not be used if realtime information from the contract is required."
+
+[route.get_total_distributed_supply]
+PATH = ["total-distributed-supply"]
```
sveitser, line 6: I'd like us to briefly think through the naming here since this will be used externally by parties not necessarily familiar with how Espresso works. I think it may prevent some errors if we chose the names of these endpoints carefully.

For an ERC20 token on Ethereum the token `total_supply` is the supply on Ethereum. 

Now in our case we have 2 different endpoints, 

1. return total_supply in the contract (currently called "total-minted-supply"): initial supply plus everything claimed via reward claim in L1.
2. ~initial supply plus everything created in Espresso~ (edit, this is currently only the distributed rewards) (currently called "total-distributed-supply")

I think "total-minted-supply" could be misleading. I think it's more correct to say that we are minting in Espresso because this is where the tokens are created.

So I think for endpoint 1 (L1 total supply) one we could use: `l1-total-supply` or `claimed-supply`.
And for Endpoint 2 (Espresso-created total): `total-minted-supply`, or if we purposely want to use non-standard terminology to avoid confusion: `total-created-supply` or `espresso-total-supply`

cc @veilgets @clu8 @ss-es

sequencer/src/api.rs
```diff
@@ -5553,6 +5587,15 @@ mod test {
...
         assert_eq!(total_minted_supply, "100000.0");
 
+        let total_distributed_supply = client
+            .get::<String>("token/total-distributed-supply")
+            .send()
+            .await
+            .expect("failed to get total_distributed_supply");
+        tracing::info!("total_distributed_supply={total_distributed_supply:?}");
+
+        assert_eq!(total_distributed_supply, "0.0");
```
sveitser, line 5597: So this currently doesn't include the initial supply. Is this intentional?

sveitser (approved):
Code looks fine to me I'm just not sure about 1. naming of API endpoints. 2. if we should also return the reward amount plus the initial amount instead. I thought we wanted the total.

PR #3900, opened 2026-01-20
Update L1 finalized block even when storing empty events list

jbearer:
Currently, `store_events` does nothing when given an empty events list (in fact, `fetch_and_store_stake_table_events` does not even call it). But this means that the `stake_table_events_l1_block` does not get updated when we enter a new epoch with no new stake table events. This makes it impossible to distinguish between two very different scenarios:

1. The node has successfully processed events through the latest L1 finalized block, but there are no new events from the last epoch.
2. The node is lagging behind the latest L1 finalized block, and is possibly missing some new events.

In scenario 1, the light client should be able to treat the empty list of stake table events as authoritative, and derive the stake table for the next epoch (which will end up being the same as the previous one. However, in scenario 2, the light client needs to wait, because we don't yet know whether there could be any events that modify the stake table. Thus, distinguishing these two scenarios is important.

This change ensures that even if there are no new events, at least the `stake_table_events_l1_block` column gets updated. We can then distinguish the two scenarios using the `EventsPersistenceRead` return value from `load_events`.

### DIFF
--- a/sequencer/src/persistence/sql.rs
+++ b/sequencer/src/persistence/sql.rs
@@ -3597,6 +3597,28 @@ mod test {
         storage.migrate_storage().await.unwrap();
     }
 
+    /// Regression test for an ambiguous behavior in `store_events`/`load_events`.
+    ///
+    /// Previously, `store_events` did nothing when given an empty events list (in fact,
+    /// `fetch_and_store_stake_table_events` was not even calling it). But this means that the
+    /// `stake_table_events_l1_block` column does not get updated when we enter a new epoch with no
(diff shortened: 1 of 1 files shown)


sequencer/src/persistence/sql.rs
```diff
@@ -3597,6 +3596,25 @@ mod test {
 
         storage.migrate_storage().await.unwrap();
     }
+
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
+    async fn test_store_events_empty() {
```
sveitser, line 3623: I think the PR description is great but it would be nice to add some of that information to this regression test to make it more visible than just git history.

PR #3902, opened 2026-01-20
Implement a non-consensus query node that reads data using the light client

jbearer:
### This PR:
Adds an implementation of the hotshot-query-service which is not participating in consensus, reads its data from another instance of the query service (e.g. one of the staked nodes), but uses the light client to verify all the data it reads, without trusting the upstream provider.

### Key places to review:
* `light-client-query-service/src/main.rs`

### How to test this PR:
* Run the service with `cargo run --features embedded-db --bin light-client-query-service -- --light-client-genesis light-client-query-service/genesis/decaf.toml --light-client-espresso-url https://query-1.decaf.testnet.espresso.network --api-port 8080 --path /tmp/db/query --light-client-db-path /tmp/db/lc`
* Check out http://localhost:8080/node/block-height, it should be increasing
* The genesis file `light-client-query-service/genesis/decaf.toml` was generated using: `cargo run -p light-client-query-service --bin genesis -- -e https://cache.decaf.testnet.espresso.network/v1/ --decaf`

### DIFF
--- a/node-metrics/src/api/node_validator/v0/mod.rs
+++ b/node-metrics/src/api/node_validator/v0/mod.rs
@@ -615,9 +615,8 @@ where
         let self_mut = self.get_mut();
 
         // Next, do we already have a connection?
-        if self_mut.connection.is_some() {
+        if let Some(connection_mut) = &mut self_mut.connection {
             // Alright, then we'll want to retrieve the next entry
-            let connection_mut = self_mut.connection.as_mut().expect("unreachable");
             pin_mut!(connection_mut);
 
             return match connection_mut.poll_next(cx) {
@@ -667,9 +666,8 @@ where
         }
 
         // Do we have an attempt to retrieve a connection in progress?
-        if self_mut.connection_future.is_some() {
+        if let Some(connection_future) = &mut self_mut.connection_future {
             tracing::debug!("waiting for connection to be established");
-            let connection_future = self_mut.connection_future.as_mut().expect("unreachable");
             pin_mut!(connection_future);
             match connection_future.poll(cx) {
                 std::task::Poll::Pending => {


light-client-query-service/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+            .fetch_leaf(LeafId::Number(block as usize))
+            .await
+            .context(format!("fetching leaf {block}"))?;
+        let (payload, vid_common) = lc
+            .fetch_block_and_vid_common(BlockId::Number(block as usize))
+            .await
+            .context(format!("fetching block {block}"))?;
+        ds.append(BlockInfo::new(leaf, Some(payload), Some(vid_common), None))
+            .await
+            .context(format!("storing data for block {block}"))?;
```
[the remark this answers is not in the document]
jbearer: This is kinda not true because the leaf should be cached after the call to `fetch_leaf`, so that `fetch_payload_and_vid_common` will hit the cache and not have to fetch the leaf from the server again. But I guess this suggestion does at least remove a call to the local DB.

light-client-query-service/src/bin/genesis.rs
```diff
@@ -0,0 +1,166 @@
...
+                .send()
+                .await
+                .context("getting chain height")?
+        } else {
+            // On Mainnet, PoS was only enabled after the version 0.4 upgrade, so we know the
+            // upgrade occurred before the first epoch.
+            config.hotshot_config().epoch_start_block()
+        };
+
+        // Through binary search, find the first block where the upgrade to version 0.4 occurred.
```
Sneh1999, line 87: why do we need to binary search? Is this block number not something which we can find like using a script or something outside this codebase and then we just put the block number here?

light-client-query-service/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+    S: Client,
+{
+    let new_height = loop {
+        let new_height = lc.block_height().await.context("getting block height")?;
+        if new_height > height {
+            break new_height;
+        }
+
+        tracing::debug!("waiting for height to increase");
+        sleep(Duration::from_secs(1)).await;
```
Sneh1999: should this ` sleep(Duration::from_secs(1)).await;` time be something which is configurable?

light-client-query-service/src/bin/genesis.rs
jbearer: That's kind of what this is, a script to find this block number and then put it in the genesis file. Once you've done this for a particular deployment then the service itself will just reference the genesis file

light-client-query-service/src/main.rs
jbearer: ah yes

PR #3903, opened 2026-01-21
fix(types): store unauthenticated validators

sveitser:
Previously, validators with invalid Schnorr signatures were rejected during RegisterV2 processing and not stored in state. This caused subsequent events targeting those validators (delegate, undelegate, key update, exit) to fail with "validator not found" errors, breaking stake table reconstruction.

Now all registered validators are stored with an `authenticated` flag:
- Valid signatures: authenticated=true, eligible for consensus
- Invalid signatures: authenticated=false, stored but filtered from active set

This provides a single code path for processing all stake table events. Events targeting unauthenticated validators succeed normally - the validator just won't be selected for the active consensus set.

Assumption: There are no unauthenticated validators on decaf, therefore computation of the existing stake tables and stake table hash are unaffected.

Tables migrated:
  1. stake_table_validators - JSON column validator
  2. epoch_drb_and_root - bincode blob column stake

  Change: Adds "authenticated": true field to all existing validator records.

  Before:
```json
  {
    "account": "0x...",
    "stake_table_key": "...",
    "state_ver_key": "...",
    "stake": "...",
    "commission": 1234,
    "delegators": {...}
  }
```

  After:
```json
  {
    "account": "0x...",
    "stake_table_key": "...",
    "state_ver_key": "...",
    "stake": "...",
    "commission": 1234,
    "delegators": {...},
    "authenticated": true // NEW
  }
```

  All validators in the database passed full signature validation before
  storage. Only after this change do we track unauthenticated validators.

Types:
- RegisteredValidator: all validators from L1 (may be unauthenticated)
- AuthenticatedValidator: newtype guaranteeing authenticated=true
- Validator (deprecated): kept for DB migration compatibility

Tests:
- Commitment equality: authenticated validators match old Validator commitment
- Commitment difference: unauthenticated validators have distinct commitment
- Parametrized test for events targeting unauthenticated validators
- Test for DB migration to new validator types

### DIFF
--- a/crates/hotshot/task-impls/src/consensus/mod.rs
+++ b/crates/hotshot/task-impls/src/consensus/mod.rs
@@ -153,10 +153,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
             HotShotEvent::ViewChange(new_view_number, epoch_number) => {
                 // Request the randomized stake table for the subsequent epoch,
                 // to trigger catchup and the DRB calculation if it happens to be missing.
-                let _ = self
-                    .membership_coordinator
-                    .membership_for_epoch(epoch_number.map(|e| e + 1))
-                    .await;
+                //
+                // the frequency is dynamic, depending on the epoch height. if the epoch height is low
+                // (e.g. like it is in tests), we do this every view
+                let frequency = (self.epoch_height / 30).clamp(1, 100);
+                if **new_view_number % frequency == 0 {
+                    let _ = self
+                        .membership_coordinator
+                        .membership_for_epoch(epoch_number.map(|e| e + 1))
+                        .await;
+                }
 
                 if let Err(e) =
                     handle_view_change(*new_view_number, *epoch_number, &sender, self).await


sequencer/src/persistence/fs.rs
```diff
@@ -1747,20 +1754,17 @@ impl MembershipPersistence for Persistence {
...
+            AuthenticatedValidatorMap,
+            Option<RewardAmount>,
+            Option<StakeTableHash>,
+        );
+        let stake: StakeTuple = bincode::deserialize(&bytes).with_context(|| {
+            format!(
+                "failed to deserialize stake table at {}",
+                file_path.display()
+            )
+        })?;
```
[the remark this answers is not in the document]
sveitser: /gemini check again?

types/src/v0/traits.rs
```diff
@@ -908,12 +912,17 @@ pub trait SequencerPersistence:
...
+        self.migrate_validator_authenticated().await?;
 
         tracing::warn!("consensus storage has been migrated to new types");
 
         Ok(())
     }
 
+    async fn migrate_validator_authenticated(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
```
bfish713: Why is this implemented for fs storage but not for sql storage.  Can we make this not have a default implementation, and move the same named function in the sql impl into the trait impl?

sequencer/src/persistence/sql.rs
```diff
@@ -706,6 +725,156 @@ impl Persistence {
...
+    /// All existing validators were authenticated (they passed validation before storage),
+    /// so we set authenticated=true for any records missing this field.
+    ///
+    /// # Migration Invariant
+    ///
+    /// This migration sets `authenticated=true` for all existing records. This is safe because:
+    /// - All validators in the database passed full signature validation before storage
+    /// - The normal registration path validates both BLS and Schnorr signatures
+    /// - Only after the `authenticated` field was added do we track unauthenticated validators
+    async fn migrate_validator_authenticated(&self) -> anyhow::Result<()> {
```
bfish713: Should this be in the `SequencerPersistence` trait impl instead?

sequencer/src/persistence/sql.rs
```diff
@@ -647,6 +652,7 @@ impl PersistenceOptions for Options {
             internal_metrics: PersistenceMetricsValue::default(),
         };
         persistence.migrate_quorum_proposal_leaf_hashes().await?;
+        persistence.migrate_validator_authenticated().await?;
```
bfish713: Why do we call this here and not let `migrate_storage` take care of it (implemented in `SequencerPersistence`?

sveitser: Thanks. Yeah I first implemented it only for sql because i totally forgot about the fs. So I forgot to move things around.

sequencer/src/persistence/fs.rs
```diff
@@ -2726,4 +2823,224 @@ mod test {
             );
         }
     }
+
+    #[allow(deprecated)]
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
+    async fn test_migrate_validator_authenticated() {
+        use std::collections::HashMap;
```
imabdulbasit, line 2830: I think we should remove these in function imports

sequencer/src/persistence/fs.rs
```diff
@@ -2726,4 +2823,224 @@ mod test {
...
+
+        // Still loadable
+        let result2 = storage.load_stake(EpochNumber::new(1)).await;
+        assert!(result2.is_ok());
+    }
+
+    #[allow(deprecated)]
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
+    async fn test_migrate_validator_authenticated_json_files() {
+        use std::collections::HashMap;
```
imabdulbasit, line 2901: same here

sequencer/api/migrations/postgres/V1103__data_migrations.sql
```diff
@@ -0,0 +1,11 @@
...
+    name TEXT NOT NULL,
+    table_name TEXT NOT NULL,
+    completed BOOLEAN NOT NULL DEFAULT false,
+    migrated_rows BIGINT NOT NULL DEFAULT 0,
+    PRIMARY KEY (name, table_name)
+);
+
+INSERT INTO data_migrations (name, table_name) VALUES
+    ('validator_authenticated', 'epoch_drb_and_root'),
+    ('validator_authenticated', 'stake_table_validators');
```
imabdulbasit, line 11: why do we have `validator_authenticated` twice here

sequencer/api/migrations/sqlite/V903__data_migrations.sql
```diff
@@ -0,0 +1,11 @@
...
+    name TEXT NOT NULL,
+    table_name TEXT NOT NULL,
+    completed BOOLEAN NOT NULL DEFAULT false,
+    migrated_rows BIGINT NOT NULL DEFAULT 0,
+    PRIMARY KEY (name, table_name)
+);
+
+INSERT INTO data_migrations (name, table_name) VALUES
+    ('validator_authenticated', 'epoch_drb_and_root'),
+    ('validator_authenticated', 'stake_table_validators');
```
imabdulbasit, line 11: same here

sequencer/src/persistence/fs.rs
```diff
@@ -1422,6 +1426,124 @@ impl SequencerPersistence for Persistence {
...
+                        AuthenticatedValidator::try_from(registered)
+                            .expect("migrate() sets authenticated=true"),
+                    )
+                })
+                .collect();
+
+            // Write back in new format (atomic: write to temp, then rename)
+            let new_data: StakeTuple = (migrated, legacy.1, legacy.2);
+            let new_bytes = bincode::serialize(&new_data)?;
+            let tmp_path = file_path.with_extension("txt.tmp");
```
imabdulbasit, line 1491: nit: can use the replace() function here as well

sequencer/src/persistence/fs.rs
```diff
@@ -2726,4 +2823,224 @@ mod test {
             );
         }
     }
+
+    #[allow(deprecated)]
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
+    async fn test_migrate_validator_authenticated() {
```
imabdulbasit, line 2829: I think these tests can be in persistence.rs which uses the macro so that one test can be run for both FS and SQL

sequencer/api/migrations/postgres/V1103__data_migrations.sql
sveitser: Because we migrate two tables. I added a migration table with migration name and affected tables. So this is one migration that affects 2 tables.

sveitser: Because we migrate two tables. I added a migration table with migration name and affected tables. So this is one migration that affects 2 tables.

PR #3908, opened 2026-01-26
Move RewardMerkleTreeV2 storage to memory

ss-es:
- persistence for catchup and for reward claims are separated (and neither uses the old database)
- we serve merkle proofs at the old reward claim interface, but only for the latest height in the reward claim contract. these are served from the database, similar to before (slightly simplified)
- we serve catchup requests at a new endpoint, reward-state-v2/reward-merkle-tree-v2/:height. this is a GET endpoint that returns the serialized (account, balance) pairs or (if something went really wrong) the full serialized tree as a fallback. there is a wrapper enum RewardMerkleTreeV2Data that handles these cases. for reference, at 500 delegators/node on demo-native (so 2500 total), the endpoint returns ~438KB of data (key/value pairs). the data here also comes directly from storageoverall, 
- we no longer persist anything in memory. but this is a bit misleading since we need to reconstruct a historical merkle tree whenever the light client's block height is updated. I think typically, however, we shouldn't hold more than 2-3 merkle trees in memory at a time (and we'll soon have improvements for this)


sequencer/src/api.rs
```diff
@@ -1279,7 +1279,158 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> StateSig
...
+
+    fn load_v2_reward_account_proof(
+        &self,
+        height: u64,
+        account: RewardAccountV2,
+    ) -> impl Send + Future<Output = anyhow::Result<RewardAccountQueryDataV2>> {
+        async move {
+            let registry = self.registry()?;
+            let Some((latest_height, ref latest)) = *registry.latest.read().await else {
+                bail!("No reward merkle tree in memory");
```
jbearer: Why is the case where `latest` is `None` (error) different from the case where `latest` is `Some` but the height doesn't match (fall through and check `finalized`)?

sequencer/src/api.rs
```diff
@@ -1279,7 +1279,158 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> StateSig
...
+        height: u64,
+        merkle_tree: RewardMerkleTreeV2,
+    ) -> impl Send + Future<Output = anyhow::Result<()>> {
+        async move {
+            let registry = self.registry()?;
+
+            {
+                let serialization = bincode::serialize(&merkle_tree)
+                    .context("Merkle tree serialization failed; this should never happen.")?;
+                self.persist(height, serialization).await?;
```
jbearer: So we still write to the database on the critical path for every block? If it's true that the bottleneck was write bandwidth, isn't that going to remain true with this change?

ss-es: yeah, this is an open question I have

experimentally, however, the performance right now seems much better. my guess is that this has something to do with the fact that:
- we are inserting a single row, and the data is serial
- even in pure throughput, this might still be less bytes/second overall

sequencer/src/api.rs
ss-es: as currently structured, `latest` should always be `Some()` and if it's `None` that means `finalized` is also `None`

sequencer/api/reward.toml
```diff
@@ -1,14 +1,29 @@
...
+
+[route.get_latest_reward_account_proof]
+PATH = ["proof/latest/:address"]
+METHOD = "GET"
 ":address" = "Literal"
-DOC = "Get current balance in reward state. Expected parameter is an Ethereum address in hex format."
+DOC = """The endpoint returns the Merkle proof for a reward account :address at the latest block height
+finalized by the light client contract. This endpoint only returns a `RewardAccountQueryDataV2`,
+containing the balance and a `RewardAccountProofV2`. This proof is based on `RewardMerkleTreeV2`, which uses the 
+Keccak256 hashing algorithm."""
```
jbearer, line 14: This might be a bit awkward to use, because there's no way to tell for sure which block height the returned proof corresponds to. You can separately read the finalized height from the light client contract, but you don't know if it just changed. I think it would be more ergonomic to return the block height also in the result, _or_ to take the block height as a parameter and fail if it is anything other than the finalized block height. This latter one might sound awkward but it has the benefit of being very cache-friendly since the result is deterministic for a given `:address` and `:height` parameter

sequencer/src/api/sql.rs
```diff
@@ -143,46 +141,241 @@ impl RewardAccountProofDataSource for SqlStorage {
...
+                .context("opening transaction for state update")?;
+
+            tokio::spawn(async move {
+                tx.upsert(
+                    "reward_merkle_tree_v2_bincode",
+                    ["height", "serialized_bytes"],
+                    ["height"],
+                    [(height as i64, merkle_tree)],
+                )
+                .await?;
```
jbearer: Why does this run in the background?

sequencer/src/api/sql.rs
```diff
@@ -143,46 +141,241 @@ impl RewardAccountProofDataSource for SqlStorage {
...
+                for _ in 0..20 {
+                    let Some(row) = iter.next() else {
+                        continue;
+                    };
+                    chunk.push(row);
+                }
+
+                if chunk.is_empty() {
+                    break;
+                }
```
jbearer, line 218: Nit: looks like a reimplementation of [`Itertools::chunks`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunks)

sequencer/src/api.rs
```diff
@@ -1279,7 +1251,186 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> StateSig
     }
 }
 
-pub(crate) trait RewardAccountProofDataSource: Sync {
+#[derive(Serialize, Deserialize)]
+pub(crate) enum RewardMerkleTreeV2Data {
+    Tree(RewardMerkleTreeV2),
+    Balances(Vec<(RewardAccountV2, RewardAmount)>),
+}
```
jbearer: Could use some comments explaining why we have the two separate cases

sequencer/src/api.rs
```diff
@@ -1279,7 +1251,186 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> StateSig
...
+    ) -> impl Send + Future<Output = anyhow::Result<()>> {
+        async move {
+            {
+                let serialization =
+                    bincode::serialize(&Into::<RewardMerkleTreeV2Data>::into(merkle_tree))
+                        .context("Merkle tree serialization failed; this should never happen.")?;
+                self.persist_tree(height, serialization).await?;
+            }
+
+            if height.is_multiple_of(30) {
```
jbearer, line 1311: Make `30` a configurable parameter?

sequencer/src/api/sql.rs
ss-es: this was actually quite hard to implement with `chunks` with the same characteristics (something something not `Send`)

I spent maybe 30 minutes on it and the only way I could get it to work was with collecting the iterator first which I really wanted to avoid

sequencer/api/reward.toml
ss-es: there's already an endpoint that takes the height as an argument (with the behavior you described) that I expect we will keep supporting (and I think it's intuitive, `/proof/{height}/{account}` vs `/proof/latest/{account}`)

I was thinking about possibly including a `height` in the result here, though I'm not sure if that's more valuable than keeping the output in the same format across both of these. I was thinking of adding the finalized hotshot block from the contract as another endpoint (we're already querying the value, so it's trivial to make the cached value available)

there's two reasons I had in mind for `latest`:
1. you can grab proofs that will (usually) work with the reward claim contract without having to query the light client contract first for the current height (maybe I misunderstood this @sveitser)
2. more of a personal thing: it was a bit annoying to debug whether reward proofs are generally available for an account because you may be chasing a height that gets garbage collected, and have to separately figure out what height to ask for. I really liked having this interface just to tell if we had the reward proof available

sequencer/src/api/sql.rs
ss-es: I think I was trying to avoid blocking the state update loop when I was dumping the bincode::serialized tree, but I think you're right

this is probably something we should block on. and given the size of the insert (1 row, a few megabytes of data at most) the brunt of the work was already done before this

ss-es: [dba8771](https://github.com/EspressoSystems/espresso-network/pull/3908/commits/dba8771744da3d6ce253b62b9d1c7fa3a026e8bd)

sequencer/api/reward.toml
ss-es: maybe we could respond with a temporary redirect `/proof/latest/{account}` -> `/proof/{height}/{account}`? though I'm not sure if that's possible in tide

jbearer: I think your two reasons above are good, this is fine

sequencer/src/api/sql.rs
jbearer: lmao thanks rust

sequencer/src/api.rs
ss-es: [c6ca037](https://github.com/EspressoSystems/espresso-network/pull/3908/commits/c6ca0376785680d9c331733c315f13eaf5aa3bc0)

ss-es (commented):
Update: proof generation is now decoupled from the state storage loop. specifically, no part of the logic which generates and persists proofs can cause `update_state_storage` to fail

because the reward merkle proofs are transient and only serve the reward claim contract, it's not consequential if this fails. we end up retrying it as part of the normal state update loop within a minute or so anyway

sequencer/src/state.rs
```diff
@@ -316,6 +283,18 @@ where
     }
 
     tx.commit().await?;
+
+    if proposed_leaf.header().version() > EpochVersion::version() {
```
imabdulbasit: shouldn't this be >= EpochVersion::version()?

ss-es: no -- unless the gating on `main` is wrong

this comes from the `else` condition on line 184 in this file

imabdulbasit: oh, right! we were using v1 for `EpochVersion`

sequencer/src/state.rs
```diff
@@ -316,6 +283,18 @@ where
...
+        storage
+            .save_reward_merkle_tree_v2(
```
imabdulbasit: I think we should not update the state height above if this fails. so this should like part of store_state_update() maybe

imabdulbasit: or you can also move the `tx.commit().await? `below this code block, and use the same transaction for insertion here

sequencer/src/api.rs
```diff
@@ -1279,7 +1251,208 @@ impl<N: ConnectedNetwork<PubKey>, V: Versions, P: SequencerPersistence> StateSig
...
+            // ultimately, we would retry in the next iteration anyway;
+            // all the information already exists.
+            if height.is_multiple_of(30) {
+                let Ok(finalized_hotshot_height) = node_state.finalized_hotshot_height().await
+                else {
+                    return Ok(());
+                };
+
+                // check to see whether we have proofs at that height already stored
+                if !self.proof_exists(finalized_hotshot_height).await {
```
imabdulbasit, line 1318: why don't we just use the merkle tree that we have to return the proofs instead of separately storing them

types/src/v0/impls/reward.rs
```diff
@@ -974,33 +974,30 @@ pub async fn get_leader_and_fetch_missing_rewards(
...
                 parent_height,
                 ?parent_view,
-                ?missing_reward_accts,
                 %reward_merkle_tree_root,
-                "fetching missing reward accounts from peers"
+                "fetching reward merkle tree from peers"
             );
 
-            let missing_account_proofs = instance_state
+            validated_state.reward_merkle_tree_v2 = instance_state
```
imabdulbasit, line 993: I think we should verify that we have the accounts in the fetched tree

sequencer/src/state.rs
```diff
@@ -214,38 +213,6 @@ async fn store_state_update(
         )
         .await
         .context("failed to store reward merkle nodes")?;
```
imabdulbasit, line 215: I’m not fully sure how the merklized state loop is currently working, so please correct me if I’m wrong. If we are  starting from a forgotten state then  accounts would be missing, and the merklized state loop only performs catchup from SqlStorage. In that case, are we simply returning the new tree in apply_header() catchup at that height and then using that tree going forward? What happens if that tree is missing?

imabdulbasit: I think there should be a migration that populates the merkle tree from all existing key/value pairs and stores that tree in the table at startup? And, at the start of the merklized state loop, we should load the reward tree from the table if it already exists

sequencer/src/state.rs
ss-es: using the same transaction caused noticeable issues locally so I want to avoid that, but I'll add a `commit()` and start a new transaction afterwards

ss-es: I think this addresses this, but please take a look and confirm: [10c7a90](https://github.com/EspressoSystems/espresso-network/pull/3908/commits/10c7a90c0db6f1d211149964a5309e06220f7ef7)

sequencer/src/api.rs
ss-es: the only consumer of these proofs now is the rewards claim contract, and presumably different people will request proofs for their own reward accounts separately

reconstructing the tree for every single one of them could explode our memory usage, and also cause a lot of cpu load on the node. I think it's much better to do this once only for the height that we plan to serve

types/src/v0/impls/reward.rs
ss-es: we verify this in all of the fetchers, see all the implementations of `try_fetch_reward_merkle_tree_v2` (and `fetch_reward_merkle_tree_v2`) in `sequencer/src/catchup.rs` and `sequencer/src/request_response/catchup/state.rs`

sequencer/src/state.rs
ss-es: yes you're absolutely correct, I missed this

I believe this fixes it: [6092615](https://github.com/EspressoSystems/espresso-network/pull/3908/commits/6092615e48e19210fe03f962841b38f6b9ef3fa3)

ss-es: (and of course you just implemented the migration -- thanks again for that!)

ss-es: I hooked up the migration you wrote here: [370c855](https://github.com/EspressoSystems/espresso-network/pull/3908/commits/370c8556ee7f82c94ddac0f9836e59aedac95d27)

PR #3910, opened 2026-01-26
start and wait for reward calc if it's not started at the end of epoch

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/types/src/v0/impls/header.rs
+++ b/types/src/v0/impls/header.rs
@@ -892,12 +892,31 @@ impl Header {
                     })?;
 
                 if prev_epoch_header.version() >= EpochRewardVersion::version() {
-                    anyhow::bail!(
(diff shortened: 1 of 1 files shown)


types/src/v0/impls/header.rs
```diff
@@ -892,12 +892,31 @@ impl Header {
...
+                        tracing::info!(
+                            %epoch,
+                            prev_epoch = %result.epoch,
+                            total = %result.total_distributed.0,
+                            "applying epoch rewards"
+                        );
+                        validated_state.reward_merkle_tree_v2 = result.reward_tree.clone();
+                        (result.total_distributed, result.changed_accounts)
+                    } else {
+                        tracing::error!(%epoch, %prev_epoch, "failed to get rewards for V6 epoch");
```
imabdulbasit, line 913: this should return an error and not default values

PR #3916, opened 2026-01-29
Add vote participation api

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Adds vote participation API.
The API can be used to show each's node vote participation in the current and previous epoch.

### DIFF
--- a/data/v3/decaf_stake_table.json
+++ b/data/v3/decaf_stake_table.json
@@ -7,7 +7,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x05ba535dd36641d95f35b7342baf9a2e5f767fd9": {
     "account": "0x05ba535dd36641d95f35b7342baf9a2e5f767fd9",
@@ -17,7 +18,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x39eb442a9480737456565b6a1f60014037289097": {
     "account": "0x39eb442a9480737456565b6a1f60014037289097",
@@ -27,7 +29,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x29b4ec1be1cddc1c74a4f55baf373d54d3c5bef6": {
     "account": "0x29b4ec1be1cddc1c74a4f55baf373d54d3c5bef6",
@@ -37,7 +40,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x6afbad8b8ec93d9aeac0bc0e94d3911324e44f22": {
     "account": "0x6afbad8b8ec93d9aeac0bc0e94d3911324e44f22",
@@ -47,7 +51,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x8cf56afa1b3c9b6ff148b77929678e271d6e1f5a": {
     "account": "0x8cf56afa1b3c9b6ff148b77929678e271d6e1f5a",
@@ -57,7 +62,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x1c22119fbb5eb16e3f3f11c1a162f28db7e72f67": {
     "account": "0x1c22119fbb5eb16e3f3f11c1a162f28db7e72f67",
@@ -67,7 +73,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0xb14217e8a9be87351970ce15790a853b6d7c5f8d": {
     "account": "0xb14217e8a9be87351970ce15790a853b6d7c5f8d",
@@ -77,7 +84,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0x6103c13c8b0b609a247fb8e00f088924139f6390": {
     "account": "0x6103c13c8b0b609a247fb8e00f088924139f6390",
@@ -87,7 +95,8 @@
     "commission": 1234,
     "delegators": {
       "0x0774a80189030b3673983ebd4d81dcb5af21c78a": "0x3635c9adc5dea00000"
-    }
+    },
+    "authenticated": true
   },
   "0xceaa793079b49346b7d9ce74a4aafa0623270f72": {
     "account": "0xceaa793079b49346b7d9ce74a4aafa0623270f72",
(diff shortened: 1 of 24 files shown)


process-compose.yaml
```diff
@@ -242,7 +243,9 @@ processes:
         condition: process_healthy
       stake-for-demo:
         condition: process_completed
-    availability: *run-forever
+    availability:
+      restart: no
+      exit_on_end: false
```
ss-es: was this file intended to be committed?

justfile
```diff
@@ -54,7 +54,7 @@ demo *args:
     fi
 
 
-demo-native *args: (build "test")
+demo-native *args: (build "release")
```
ss-es: also this file

flake.nix
```diff
@@ -67,7 +67,7 @@
         export CARGO_TARGET_DIR="$my_pwd/target/nix"
 
         # Add rust binaries to PATH
-        export PATH="$CARGO_TARGET_DIR/debug:$PATH"
+        export PATH="$CARGO_TARGET_DIR/release:$PATH"
```
ss-es: and this one

sequencer/src/api/endpoints.rs
```diff
@@ -493,6 +493,22 @@ where
         }
         .boxed()
     })?
+    .at("current_vote_participation", |_, state| {
+        async move {
+            Ok(state
+                .read(|state| state.current_vote_participation().boxed())
```
ss-es, line 499: not sure how a `HashMap` gets sent here, but I feel like it would be better to just return a `Vec<(key,f64)>`?

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -332,6 +368,24 @@ impl<
...
+        let real_qc_pp =
+            <TYPES::SignatureKey as SignatureKey>::public_parameter(stake_table, threshold);
+
+        <TYPES::SignatureKey as SignatureKey>::signers(
+            &real_qc_pp,
+            self.signatures.as_ref().unwrap(),
+        )
+        .wrap()
+        .context(|e| warn!("Tracing signers: {e}"))
+    }
```
ss-es, line 375: @mrain can you also double-check this?

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -175,6 +175,24 @@ impl<TYPES: NodeType, THRESHOLD: Threshold<TYPES>> Certificate<TYPES, DaData>
...
+    ) -> Result<Vec<<TYPES::SignatureKey as SignatureKey>::VerificationKeyType>> {
+        if self.view_number == TYPES::View::genesis() {
+            return Ok(vec![]);
+        }
+        let real_qc_pp =
+            <TYPES::SignatureKey as SignatureKey>::public_parameter(stake_table, threshold);
+
+        <TYPES::SignatureKey as SignatureKey>::signers(
+            &real_qc_pp,
+            self.signatures.as_ref().unwrap(),
```
ss-es: is this `unwrap()` safe? is there a sensible `unwrap_or_default()`?

crates/hotshot/types/src/simple_certificate.rs
```diff
@@ -113,6 +113,25 @@ impl<TYPES: NodeType, VOTEABLE: Voteable<TYPES>, THRESHOLD: Threshold<TYPES>>
...
         }
     }
+
+    fn signers(
+        &self,
+        stake_table: &[<TYPES::SignatureKey as SignatureKey>::StakeTableEntry],
+        threshold: U256,
+    ) -> Result<Vec<<TYPES::SignatureKey as SignatureKey>::VerificationKeyType>> {
+        if self.view_number == TYPES::View::genesis() {
+            return Ok(vec![]);
```
ss-es, line 123: in my opinion I think we shouldn't special case genesis here

I'm not sure whether this returns the full stake table or an empty one if you don't special case it, but I see arguments for both and I think it's better to just return what we normally would

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -531,6 +531,23 @@ where
 
         Ok(drb)
     }
+
+    /// Returns the stake table and success threshold for the given epoch if it is available.
+    /// Otherwise returns a default stake table and MAX U256.
+    pub async fn stake_table_and_threshold(
+        &self,
+        epoch: Option<TYPES::Epoch>,
+    ) -> (HSStakeTable<TYPES>, U256) {
```
ss-es: can this not return a `Result`?

ss-es: I feel like it would be better to propagate the error from `stake_table_for_epoch` rather than return a default (empty?) stake table

crates/hotshot/types/src/qc.rs
```diff
@@ -178,8 +186,6 @@ where
             )));
         }
```
ss-es, line 187: so if I understand, the idea is that we don't want to validate the message before returning the list of signers? is there a reason for that?

ss-es (approved):
approving but I think the flake/justfile/process-compose changes should be uncommitted (unless they were intentional for some reason, but maybe we can do that separately), and left some comments about the error handling on some functions

process-compose.yaml
lukaszrzasik: good catch, these files were committed by accident. I changed them only to test locally.

justfile
lukaszrzasik: same

flake.nix
lukaszrzasik: same

sequencer/src/api/endpoints.rs
lukaszrzasik: It's well tested and working. I followed the existing pattern in `current_proposal_participation`

crates/hotshot/types/src/simple_certificate.rs
lukaszrzasik: The genesis certificate doesn't really have signers, right? It's the only case where no signers and no signature makes sense and doesn't indicate an error. I think `Ok` and an empty vector of signers is correct here. Wdyt? Similar thing is done in `is_valid_cert`

crates/hotshot/types/src/qc.rs
lukaszrzasik: Yes, the purpose of this method is to simply return the list of the certificate's signers. The assumption is the certificate's signatures have been previously validated. Checking the correctness of the certificate's signatures introduces additional complexity, specifically we need to pass around the upgrade lock so that we know how to calculate the data's commit. That's not readily available everywhere so I decided to keep it simple.

I left the logic of the trace method intact if the check and trace signers feature is required.

crates/hotshot/types/src/epoch_membership.rs
lukaszrzasik: I have mixed feelings about this. In principle I agree with you. It seems natural for the method to return an error if it fails to get the stake table for the epoch. In fact in the previous iteration of the PR this function didn't exist. I simply called `stake_table_for_epoch` and handled the error in the calling code.

I introduced this method to get rid of the repetition. I don't have a strong opinion whether this method should be modified or even completely removed. let me know what you think.

crates/hotshot/types/src/simple_certificate.rs
lukaszrzasik: Because the `singers` method is very similar to `is_valid_cert` I followed what's there but I agree that unwrap doesn't seem correct in either of them. Maybe @mrain could have a look and say if `None` here makes sense at all. If not I would prefer to simply return `Err` in both `signers` and `is_valid_cert`.

crates/hotshot/types/src/simple_certificate.rs
ss-es: ah, I guess if `is_valid_cert` special cases it anyway it's fine

yes, I agree. it definitely doesn't matter either way

crates/hotshot/types/src/epoch_membership.rs
ss-es: I think my primary worry is the possibility of misuse (and lack of clarity), since it's a somewhat arbitrary thing it returns in that case

personally I think just using `stake_table_for_epoch` and handling the error sounds better

crates/hotshot/types/src/simple_certificate.rs
mrain: Yes I agree that it should be `unwrap_or_default()`, the signatures field is an Option

lukaszrzasik: It can't be `unwrap_or_default` because `QcType` doesn't implement the `Default` trait. I feel the best approach is to return an error if there are no signatures.

PR #3919, opened 2026-01-30
Nix shell changes

ss-es:
- Adds a `dockerShell` which has `docker` added to the packages provided by the nix-shell, and sets `DOCKER_HOST` in the shell. This allows you to run `docker-rootless` provided by the development shell for `just demo-native` without installing `docker` globally.
- Uncommits `.envrc` and adds the file to `.gitignore`. I feel like this is fundamentally a personal thing that should never have been committed e.g. I have sometimes wanted to use a different shell from the flake. The existing `.envrc` is renamed to `.envrc.example` for conveience, so that it's still available for anyone to rename `.envrc.example` -> `.envrc`

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ a dockerized Espresso Sequencer network with an example Layer 2 rollup applicati
 
 - Obtain code: `git clone git@github.com:EspressoSystems/espresso-network`.
 - Make sure [nix](https://nixos.org/download.html) is installed.
-- Activate the environment with `nix-shell`, or `nix develop`, or `direnv allow` if using [direnv](https://direnv.net/).
+- Activate the environment with `nix-shell`, or `nix develop`. If using [direnv](https://direnv.net/), copy `.envrc.example` to `.envrc` (or create your own `.envrc` file) and run `direnv allow`.
 - For installation without nix please see [ubuntu.md](./doc/ubuntu.md).
 
 ## Documentation


flake.nix
```diff
@@ -276,6 +277,9 @@
             # with rustup installations.
             export CARGO_HOME=$HOME/.cargo-nix
 
+            # Required for demo-native to run with docker-rootless
+            export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
```
ss-es: cc @sveitser : my understanding was that this shouldn't actually interefere with anyone's existing setup but maybe I'm wrong

.gitignore
```diff
@@ -3,6 +3,10 @@
 debug/
 target/
 
+# .envrc typically just has `use flake .`
+# but this should be done by the user
```
sveitser, line 7: Let's update the README section concerning the direnv setup then.

flake.nix
sveitser: Does seem to break things for me

```
$ export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
$  docker ps 
failed to connect to the docker API at unix:///run/user/1000/docker.sock; check if the path is correct and if the daemon is running: dial unix /run/user/1000/docker.sock: connect: no such file or directory
```

sveitser: But yeah a rootless docker setup seems much preferrable so let's make it work somehow.

ss-es: oh that's unfortunate

maybe the better thing to do then is to just add a separate shell for this? with the .envrc removal it shouldn't make a difference to have it in a non-default shell

ss-es: I'll look into it a bit more but I'll just update with that if I can't find anything

sveitser: we could just check if the socket exists before setting the env var?

ss-es: > But yeah a rootless docker setup seems much preferrable so let's make it work somehow.

I was only a bit worried about breaking existing workflows

PR #3923, opened 2026-02-03
[decaf] generate reward proofs for every block height

ss-es:
This PR extends #3908 for backwards compatibility with current decaf nodes. **this is a temporary bridge between binaries and should not be merged.**

on this branch, query nodes will generate all reward proofs for every block (as opposed to only the finalized block for rewards claims). this allows us to serve catchup requests for currently running nodes, while nodes migrate to the newer binary. after all nodes have migrated, we can safely revert query nodes to the behavior in the base branch (i.e. only generate proofs to serve for reward claims)

### DIFF
--- a/sequencer/src/catchup.rs
+++ b/sequencer/src/catchup.rs
@@ -439,8 +439,8 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
         reward_merkle_tree_root: RewardMerkleCommitmentV2,
         accounts: Arc<Vec<RewardAccountV2>>,
     ) -> anyhow::Result<RewardMerkleTreeV2> {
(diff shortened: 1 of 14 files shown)


sequencer/src/catchup.rs
```diff
@@ -432,36 +432,72 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
+                        .await?;
 
-                Ok::<Vec<u8>, anyhow::Error>(tree_bytes)
-            })
-            .await
-            .context("Fetching from peer failed")?;
+                    Ok::<Vec<u8>, anyhow::Error>(tree_bytes)
+                })
+                .await
+                .context("Fetching from peer failed")?;
```
bfish713: I think this `?` returns from the whole function, not the code block, which means we never hit the fallback

PR #3926, opened 2026-02-03
feat(staking-cli): add OpenMetrics support and fix flaky tests

sveitser:
Add support for fetching validator metadata from sequencer node's /status/metrics endpoint (OpenMetrics format) as an alternative to hosting a separate JSON metadata file.

Changes:
- Add --node-url CLI option for fetching metadata from node metrics
- Add OpenMetrics parsing (prometheus-parse crate)
- Auto-detect JSON vs OpenMetrics format via Content-Type header
- Add serve_on_random_port() helper to fix flaky tests caused by race conditions with portpicker

The OpenMetrics parsing code is copied from staking-ui-service and marked with TODO comments for future consolidation into a shared crate.

Manually tested the ignored tests against our mainnet node.

### DIFF
--- a/crates/cliquenet/Cargo.toml
+++ b/crates/cliquenet/Cargo.toml
@@ -13,8 +13,6 @@ bon = { workspace = true }
 bs58 = "0.5.1"
 bytes = { workspace = true }
 ed25519-compact = "2.2.0"
-# optional:
-hotshot-types = { workspace = true, optional = true }
 nohash-hasher = { workspace = true }
 parking_lot = { workspace = true }
 rand = "0.9.2"
@@ -24,6 +22,9 @@ snow = { version = "0.10.0", features = ["ring-accelerated"] }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tracing = { workspace = true }
+
+# optional:
+hotshot-types = { workspace = true, optional = true }
 
 [dev-dependencies]
 criterion = "0.8.1"


crates/cliquenet/Cargo.toml
```diff
@@ -8,11 +8,13 @@ description = "A fully connected mesh network"
...
 
 [dependencies]
-bon = { workspace = true }
-bytes = { workspace = true }
 bimap = "0.6.3"
+bon = { workspace = true }
 bs58 = "0.5.1"
+bytes = { workspace = true }
 ed25519-compact = "2.2.0"
+# optional:
```
sveitser: cargo sort

staking-cli/src/lib.rs
```diff
@@ -215,25 +222,179 @@ impl ValidSignerConfig {
...
 #[group(required = true, multiple = false)]
 pub struct MetadataUriArgs {
+    /// URL where validator metadata JSON is hosted.
     #[clap(long, env = "METADATA_URI")]
-    metadata_uri: Option<String>,
+    pub metadata_uri: Option<String>,
+
+    /// URL of the node's API. Metadata will be fetched from /status/metrics.
+    #[clap(long, env = "NODE_URL")]
+    pub node_url: Option<String>,
```
jbearer, line 231: Not sure I understand the point of this option. Whenever I would do `--node-url URL` couldn't I just as easily do `--metadata-uri URL/status/metrics`?

staking-cli/src/lib.rs
```diff
@@ -215,25 +222,179 @@ impl ValidSignerConfig {
...
+            MetadataSource::NodeUrl(url) => match url.join("status/metrics") {
+                Ok(metrics_url) => Some(metrics_url),
+                Err(e) => {
+                    tracing::warn!("failed to construct metrics URL from {url}: {e}");
+                    None
+                },
+            },
+            MetadataSource::None => None,
+        }
+    }
```
jbearer, line 293: Nit: this seems to duplicate a lot of the logic from just above. Perhaps could be implemented as `self.metadata_uri().inspect_err(|e| tracing::warn!("failed to construct metrics URL from {url}: {e})).ok()`

jbearer (commented):
Functionality looks good, but I don't understand the need for some of the added complexity with the additional `--node-url` option and the `MetadataSource` enum.

PR #3932, opened 2026-02-05
Remove portpicker race condition

sveitser:
Bind to socket then release it which will make the OS not hand out the socket for 60s by default on linux.

Run 1: https://github.com/EspressoSystems/espresso-network/actions/runs/21724072749?pr=3932 ok
Run 2: https://github.com/EspressoSystems/espresso-network/actions/runs/21724930185?pr=3932 2 failures (also on main)
Run 3: https://github.com/EspressoSystems/espresso-network/actions/runs/21724930185?pr=3932 same 2 failures (also on main)

I didn't see any addr already in use error anymore.

### DIFF
--- a/espresso-dev-node/src/main.rs
+++ b/espresso-dev-node/src/main.rs
@@ -9,7 +9,7 @@ use alloy::{
     network::EthereumWallet,
     node_bindings::Anvil,
     primitives::{Address, Bytes, U256},
-    providers::{layers::AnvilLayer, Provider, ProviderBuilder, WalletProvider},
+    providers::{Provider, ProviderBuilder, WalletProvider},
     rpc::client::RpcClient,
     signers::{
         k256::ecdsa::SigningKey,
@@ -283,8 +283,8 @@ async fn main() -> anyhow::Result<()> {
         (url, None)
     } else {
         tracing::warn!("L1 url is not provided. running an anvil node");
-        let anvil_layer = AnvilLayer::from(
-            Anvil::new().args([
+        let instance = Anvil::new()
+            .args([
                 "--slots-in-an-epoch",
                 "0",
                 "--mnemonic",
@@ -300,11 +300,10 @@ async fn main() -> anyhow::Result<()> {
                     .context("Failed to convert path to string")?,
                 "--state-interval",
                 "1",
-            ]),
-        );
-        let url = anvil_layer.endpoint_url();
+            ])
+            .spawn();
+        let url = instance.endpoint_url();
         tracing::info!("l1 url: {}", url);
-        let instance = anvil_layer.instance().clone();
         (url, Some(instance))
     };
 
--- a/tests/reward_claims_e2e.rs
+++ b/tests/reward_claims_e2e.rs
@@ -4,7 +4,7 @@ use alloy::{
     network::EthereumWallet,
     node_bindings::Anvil,
     primitives::U256,
-    providers::{layers::AnvilLayer, Provider, ProviderBuilder, WalletProvider},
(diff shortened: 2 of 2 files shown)


slow-tests/tests/restart_tests.rs
```diff
@@ -1240,27 +1235,18 @@ async fn start_marshal(dir: &Path, port: u16) -> JoinHandle<()> {
...
-                 try reducing the number of ports being used."
-            );
-        }
+    fn pick(&mut self) -> anyhow::Result<u16> {
+        let listener = TcpListener::bind("127.0.0.1:0")?;
+        let port = listener.local_addr()?.port();
+        self.listeners.push(listener);
+        Ok(port)
     }
 }
```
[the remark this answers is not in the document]
sveitser: /gemini check

crates/builder/src/lib.rs
```diff
@@ -239,18 +238,6 @@ pub mod testing {
             }
         }
 
-        // url for the hotshot event streaming api
```
sveitser, line 242: This seems to be unused, hence removed.

crates/builder/src/lib.rs
```diff
@@ -90,7 +90,6 @@ pub mod testing {
     use super::*;
     use crate::non_permissioned::BuilderConfig;
 
-    #[derive(Clone)]
```
sveitser, line 93: will add this back

crates/hotshot/hotshot/src/traits/networking/cliquenet_network.rs
```diff
@@ -151,16 +151,16 @@ impl<T: NodeType> TestableNetworkingImplementation<T> for Cliquenet<T> {
...
-            use cliquenet::Address;
+        use cliquenet::Address;
 
+        let mut parties: Vec<(Keypair, T::SignatureKey, Address)> = Vec::new();
+        for i in 0..expected_node_count {
             let secret = T::SignatureKey::generated_from_seed_indexed([0u8; 32], i as u64).1;
             let public = T::SignatureKey::from_private(&secret);
             let kpair = derive_keypair::<<T as NodeType>::SignatureKey>(&secret);
-            let port = portpicker::pick_unused_port().expect("an unused port is available");
+            let port = test_utils::reserve_tcp_port().expect("Could not reserve TCP port");
```
twittner: I would have preferred the ["expect as precondition style"](https://doc.rust-lang.org/std/error/index.html#common-message-styles), but not important.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+/// // Services can still bind to it explicitly
+/// ```
+pub fn reserve_tcp_port() -> std::io::Result<u16> {
+    let server = TcpListener::bind("127.0.0.1:0")?;
+    let addr = server.local_addr()?;
+
+    // Force TIME_WAIT by completing TCP handshake
+    let _client = TcpStream::connect(addr)?;
+    let (_accepted, _) = server.accept()?;
+    // All sockets drop here - port enters TIME_WAIT
```
twittner, line 46: It would perhaps be nice to upstream this to the port picker crate. It would neatly fit into https://github.com/Dentosal/portpicker-rs/blob/912f913ac325278a564d8828d550070db1d79373/src/lib.rs#L15. While under review the repo could be forked.

twittner (approved):
Looks good to me.

crates/hotshot/hotshot/src/traits/networking/cliquenet_network.rs
sveitser: yeah i agree i will change it

PR #3936, opened 2026-02-06
feat: storage backed reward Merkle tree

mrain:
Closes #<ISSUE_NUMBER>


types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+/// File system backed storage implementation for reward Merkle trees.
+///
+/// Stores each inner tree root as a serialized bincode file on disk.
+/// Uses a single-entry cache with RwLock for thread-safe interior mutability.
+#[derive(Debug)]
+pub struct RewardMerkleTreeFSStorage {
+    /// Directory where tree files are stored
+    storage_dir: PathBuf,
+    /// Cached inner tree with its index (uses RwLock for thread-safe interior mutability)
+    cache: RwLock<Option<(OuterIndex, InnerRewardMerkleTreeRoot)>>,
```
ss-es: can you explain what `OuterIndex` represents in the comment

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+            .expect("Failed to create temporary directory for Reward Merkle Tree FS storage.")
+            .keep();
+        // fs::create_dir_all(&storage_dir).expect("Failed to create temporary storage dir.");
+
+        Self {
+            storage_dir,
+            cache: RwLock::new(None),
+        }
+    }
+}
```
ss-es: in my opinion, we should not implement `Default` especially if it can potentially crash

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+            let cache = self.cache.read().unwrap();
+            if let Some((cached_index, _)) = &*cache {
+                if *cached_index == index {
+                    return; // Already cached
+                }
+            }
+        }
+
+        // Flush the current cache if it exists
+        if let Err(e) = self.flush_cache() {
```
ss-es: since `flush_cache` can error on save, I think this should propagate the error

ss-es: this also means the return type should probably be `Result<()>`

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+            tracing::error!("Error flushing Reward Merkle tree cache: {}", e);
+        }
+
+        // Load the tree from disk or create a new one
+        let root = self
+            .load_from_disk(index)
+            .unwrap_or(InnerRewardMerkleTreeRoot::Empty);
```
ss-es: I think we should also just propagate the error if `load_from_disk` fails (except for the file not existing)

ss-es: otherwise, `load_into_cache` could be destructive

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+    fn with_tree<F, R>(&self, index: OuterIndex, f: F) -> R
+    where
+        F: FnOnce(&InnerRewardMerkleTreeRoot) -> R,
+    {
+        // Load into cache if needed
+        self.load_into_cache(index);
+
+        // Execute the closure with the cached tree
+        let cache = self.cache.read().unwrap();
+        let (_, tree) = cache.as_ref().expect("Tree should be in cache after load");
```
ss-es: I think it's better to return `Result<()>` rather than an `expect`

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,611 @@
...
+    fn with_tree_mut<F, R>(&self, index: OuterIndex, f: F) -> R
+    where
+        F: FnOnce(&mut InnerRewardMerkleTreeRoot) -> R,
+    {
+        // Load into cache if needed
+        self.load_into_cache(index);
+
+        // Execute the closure with the cached tree
+        let mut cache = self.cache.write().unwrap();
+        let (_, tree) = cache.as_mut().expect("Tree should be in cache after load");
```
ss-es: same here

mrain:
@ss-es I think all your previous comments are addressed now

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,678 @@
...
+/// Height of each inner tree (156 bits)
+const REWARD_MERKLE_TREE_V2_INNER_HEIGHT: usize =
+    REWARD_MERKLE_TREE_V2_HEIGHT - REWARD_MERKLE_TREE_V2_OUTER_HEIGHT;
+
+/// Inner tree type: 156-bit universal Merkle tree for account balances.
+///
+/// Each inner tree corresponds to one of the 16 partitions (indexed by the first
+/// 4 bits of the account address). Stores reward amounts keyed by full account addresses.
+/// Uses standard Keccak256 hashing for both leaves and internal nodes.
+type InnerRewardMerkleTreeV2 = UniversalMerkleTree<
```
imabdulbasit, line 140: Can we just use the existing `RewardMerkleTreeV2` type?

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,678 @@
...
+/// produces identical commitments to a single-level tree.
+pub type ExpectedRewardMerkleTreeV2 = InnerRewardMerkleTreeV2;
+
+/// Outer tree type: 4-bit universal Merkle tree storing inner tree roots.
+///
+/// Contains up to 16 entries (2^4), each storing the root hash of one inner tree.
+/// Uses a custom hasher (OuterKeccak256Hasher) that treats leaves as pre-hashed values.
+type OuterRewardMerkleTreeV2 = UniversalMerkleTree<
+    KeccakNode,
+    OuterKeccak256Hasher,
```
imabdulbasit, line 160: can you explain a bit more why we need separate `OuterKeccak256Hasher`?

types/src/v0/reward_mt/mod.rs
mrain: Yes we can. And this type alias is just for clarity.

types/src/v0/reward_mt/mod.rs
mrain: Leaves for outer merkle trees are roots of the inner Merkle tree. And they should be passed thru, not be hashed. So we need a separate hasher.

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,760 @@
...
+    /// any existing content. Called during cache flush operations.
+    ///
+    /// # Arguments
+    /// * `index` - Partition index to save
+    /// * `root` - Tree root to serialize and write
+    ///
+    /// # Returns
+    /// - `Ok(())` - Successfully written to disk
+    /// - `Err(io::Error)` - Disk I/O failure or serialization error
+    fn save_to_disk(&self, index: OuterIndex, root: &InnerRewardMerkleTreeRoot) -> io::Result<()> {
```
sveitser: What happens if this write fails? Currently it looks like it can leave a corrupted file. Can we write to another file first and then do an atomic rename? This way we won't leave the file in a broken state.

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,760 @@
...
+    ///
+    /// If cache contains a tree, serializes and writes it to disk. Called automatically
+    /// before loading a different partition, when cloning, and on drop.
+    ///
+    /// # Returns
+    /// - `Ok(())` - Cache flushed successfully
+    /// - `Err(io::Error)` - Disk I/O failure
+    fn flush_cache(&self) -> io::Result<()> {
+        let mut cache = self.cache.write().unwrap();
+        if let Some((index, root)) = cache.take() {
```
sveitser: Here we take it from the cache then save to disk which can fail. Shouldn't we only remove it from the cache if flushing succeeds?

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,760 @@
...
+/// - Snapshotting current state
+///
+/// # Note
+///
+/// Both instances write to the same files. Concurrent writes from multiple instances
+/// can cause data corruption. Use external synchronization if cloning for concurrent access.
+impl Clone for RewardMerkleTreeFSStorage {
+    fn clone(&self) -> Self {
+        // Flush cache before cloning to ensure clone sees latest state
+        let _ = self.flush_cache();
```
sveitser: Errors from `flush_cache` are ignored here.

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,760 @@
...
+/// # Use Cases
+///
+/// - Creating multiple readers of the same storage
+/// - Forking state for parallel processing
+/// - Snapshotting current state
+///
+/// # Note
+///
+/// Both instances write to the same files. Concurrent writes from multiple instances
+/// can cause data corruption. Use external synchronization if cloning for concurrent access.
```
sveitser: This feels kind of dangerous esp. for `Clone`. Usually I think we don't worry about such things when we clone something.

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,678 @@
...
+                            } else {
+                                RewardMerkleNode::ForgettenSubtree {
+                                    value: node.value(),
+                                }
+                            })
+                        })
+                        .collect::<Vec<_>>(),
+                }),
+                MerkleNode::Empty => proof.proof.push(RewardMerkleNode::Empty),
+                _ => unreachable!("Proof nodes in outer tree should be branches or empty"),
```
sveitser: I think having potential panics in code handling proofs is a bit problematic. Currently it's only used internally but in general proofs are untrusted data structures that may be obtained from elsewhere and are to be verified. I think it's therefore preferable to make sure code that handles proofs doesn't panic, especially for safety in the future.

types/src/v0/reward_mt/fs_storage.rs
mrain: WDYT is the better way to handle the error here? This fn doesn't return a result

types/src/v0/reward_mt/mod.rs
mrain: Here we only feed internally generated proofs

types/src/v0/reward_mt/fs_storage.rs
mrain: Optimized the logic: [4138998](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/413899894c0c5c39fa889cc8da6e6c1c8e8d8565)

types/src/v0/reward_mt/fs_storage.rs
mrain: Good idea: [db208d7](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/db208d765939f4091a8e6a1b1bf2790935560a9b)

types/src/v0/reward_mt/fs_storage.rs
mrain: deleted the Clone

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,763 @@
...
+    ) -> io::Result<()> {
+        // Already cached — nothing to do
+        if let Some((cached_index, ..)) = cache.as_ref() {
+            if cached_index == index {
+                return Ok(());
+            }
+        }
+
+        // Evict the current entry: only write back if dirty (inline flush logic)
+        if let Some((evict_index, root, dirty)) = cache.take() {
```
sveitser: There are fallible operations after the `cache.take`, so potential data loss.

types/src/v0/reward_mt/fs_storage.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// # Warning
+    ///
+    /// The temp directory may be cleaned up by the OS. For production use, call
+    /// `new()` with an explicit directory path.
+    ///
+    /// # Panics
+    ///
+    /// Panics if temporary directory creation fails (extremely rare, indicates
+    /// system-level issues like no disk space or permission denied).
+    pub fn tempfile() -> io::Result<Self> {
```
sveitser: This looks like it should go into behind `test` gate. I think it's better to create the tempdirs in tests then you can don't have to use `keep` as long as you keep them in scope for the tests.

mrain: tempfile are moved to `test`, no more `keep()` [06a7ca7](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/06a7ca7fbeff55662facd132eb4265085e7ad6d7)

types/src/v0/reward_mt/fs_storage.rs
mrain: Missed this one, fixed here [06a7ca7](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/06a7ca7fbeff55662facd132eb4265085e7ad6d7)

types/src/v0/reward_mt/mod.rs
ss-es: could we maybe call this `InMemoryRewardMerkleTreeV2`? not a strong opinion, so if you think `Inner` is more appropriate that's fine

types/src/v0/reward_mt/mod.rs
ss-es: I also prefer we don't have a panic. especially in this case we can just log at `error`, since we don't even need to return anything

ss-es (commented):
still looking at this, if I understand correctly is the `InMemoryRewardMerkleTreeV2` the one that just stores the balances in memory (but not the full tree) at some depth in contrast to the filebacked one which stores that info in storage?

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+    /// Returns storage error if IO operation fails.
+    pub fn lookup(
+        &self,
+        index: impl Borrow<RewardAccountV2>,
+    ) -> anyhow::Result<LookupResult<RewardAmount, RewardMerkleProof, ()>> {
+        let outer_index = OuterIndex::new(index.borrow());
+        let outer_proof = match self.outer.lookup(outer_index) {
+            LookupResult::Ok(_, proof) => proof,
+            LookupResult::NotInMemory => {
+                unreachable!("Outer reward merkle tree will never be forgetten.")
```
ss-es: I would prefer that this logs at `Error` and returns an `Err`, since the return type is `Result` anyway

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+    pub fn universal_lookup(
+        &self,
+        index: impl Borrow<RewardAccountV2>,
+    ) -> anyhow::Result<LookupResult<RewardAmount, RewardMerkleProof, RewardMerkleProof>> {
+        let index = index.borrow();
+        let outer_index = OuterIndex::new(index);
+        let outer_proof = match self.outer.universal_lookup(outer_index) {
+            LookupResult::Ok(_, proof) => proof,
+            LookupResult::NotInMemory => {
+                unreachable!("Outer reward merkle tree will never be forgetten.")
```
ss-es: same comment

types/src/v0/reward_mt/mod.rs
```diff
@@ -0,0 +1,669 @@
...
+/// ```
+impl<S: RewardMerkleTreeStorage> IntoIterator for StorageBackedRewardMerkleTreeV2<S> {
+    type Item = (RewardAccountV2, RewardAmount);
+
+    type IntoIter = <std::vec::Vec<(RewardAccountV2, RewardAmount)> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.storage
+            .get_entries()
+            .expect("Storage operation failed during iteration")
```
ss-es: is there any way we can avoid the `expect` here?

types/src/v0/reward_mt/mod.rs
mrain: [6e5bdbd](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/6e5bdbda560de64f0452f42bc3ecd04a44d3c0b7)

types/src/v0/reward_mt/mod.rs
mrain: [6e5bdbd](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/6e5bdbda560de64f0452f42bc3ecd04a44d3c0b7)

types/src/v0/reward_mt/mod.rs
mrain: removed `IntoIterator` impl, move to `get_entires` instead: [6e5bdbd](https://github.com/EspressoSystems/espresso-network/pull/3936/commits/6e5bdbda560de64f0452f42bc3ecd04a44d3c0b7)

PR #3937, opened 2026-02-06
refactor(staking-cli): move skip-metadata-validation to command args

sveitser:
Move --skip-metadata-validation from global flag to MetadataUriArgs, making it a command-specific flag that appears after the subcommand.

Changes:
- Move skip_metadata_validation field to MetadataUriArgs
- Refactor metadata_uri to Option<Url> with custom value parser
- Add Clap constraints: requires, conflicts_with, required_unless_present
- Move MetadataUriArgs to metadata.rs for better organization
- Add tests for expected error messaged.

Breaking change: --skip-metadata-validation must now appear after the subcommand instead of before it.

I think it's fine to make this breaking change now instead of doing it in a backwards compatible way. Only few people have used the CLI and we haven't built anything on top of this.

### DIFF
--- a/staking-cli/src/metadata.rs
+++ b/staking-cli/src/metadata.rs
@@ -150,6 +150,10 @@ impl fmt::Display for MetadataUri {
 }
 
 /// Custom value parser for metadata URIs that enforces 2048 byte limit.
+///
+/// While HTTP specs don't mandate a URL length limit, most browsers and servers
+/// support at least 2048 characters. This practical limit ensures compatibility
+/// across different HTTP clients and servers.
 fn parse_metadata_url(s: &str) -> Result<Url, String> {
     let url = Url::parse(s).map_err(|e| e.to_string())?;
     if url.as_str().len() > 2048 {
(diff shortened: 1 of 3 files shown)


staking-cli/src/metadata.rs
```diff
@@ -148,13 +149,137 @@ impl fmt::Display for MetadataUri {
     }
 }
 
+/// Custom value parser for metadata URIs that enforces 2048 byte limit.
+fn parse_metadata_url(s: &str) -> Result<Url, String> {
+    let url = Url::parse(s).map_err(|e| e.to_string())?;
+    if url.as_str().len() > 2048 {
+        return Err("metadata URI cannot exceed 2048 bytes".to_string());
```
imabdulbasit, line 160: where is this limit coming from?>

sveitser: Just something we found that should work for everyone, anything longer is probably a mistake and will likely start to not work with some systems.

sveitser: I'll add a comment

sveitser: https://stackoverflow.com/a/417184

sveitser: added a comment here: https://github.com/EspressoSystems/espresso-network/pull/3937/commits/e76b867b60cbdd80de6403c5c852522b58b7a3b6

PR #3939, opened 2026-02-06
Remove View and Epoch types from NodeTypes

bfish713:
Closes #<ISSUE_NUMBER>


twittner (approved):
👍 This is a great simplification I think.

PR #3943, opened 2026-02-09
versions

twittner:
Removes the `Versions` trait and `TestVersions` type in preference of a `versions` crate that defines well known versions as constants. Instead of parameterising types and functions with `V: Versions`, values are passed around. This results in a substantial decrease in compilation time and disk usage.

Since previously there were some inconsistencies regarding the version number of epochs support (0.3 vs. 0.4) I had to update some of the reference data that was previously created for 0.4 to match 0.3.


data/v3/header.json
```diff
@@ -34,7 +34,7 @@
     "ns_table": {
       "bytes": "AwAAAO7/wAAcBgAAobC5EkAOAABksAWiXBQAAA=="
     },
-    "payload_commitment": "HASH~u-mEo1mwByROUhnvO7pBFitcD0UEvruK-b8WONkKoCLQ",
+    "payload_commitment": "AvidMCommit~GVev7Msgzxsj2Kfx6HfqnNoK4USAnZ602eXda3Ic0zrY",
```
sveitser, line 37: @twittner do you know why we have changes here and in messages.json?

twittner: Yes, I briefly mentioned that above. There is no universal agreement in tests about version numbers. Previously `TestVersions::Epochs` was defined to be 0.4, but epochs elsewhere was set to 0.3. Same for other versions.

In here an attempt was made to have consistent version numbers. To that extend the `versions` crate defines the version numbers once and for all (_please double check that those are correct_). However when making changes in tests, this resulted in different behaviour, e.g. the function `vid_commitment` now references `versions::EPOCH_VERSION` and will always use 0.3. Previously it used whatever the type parameter `V: Versions` provided.

Note that `TEST_VERSIONS` still uses different version numbers and should perhaps be replaced. However the changes are already so numerous in here that I did not pursue this further for now.

crates/versions/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[non_exhaustive]
+pub struct Upgrade {
+    pub base: Version,
+    pub target: Version,
+}
+
+impl Upgrade {
+    pub const fn new(base: Version, target: Version) -> Self {
+        debug_assert! {
```
sveitser: Why only debug_assert here? Wouldn't it be good to always forbid creating invalid upgrades?

crates/versions/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+// Known versions:
+
+pub const VERSION_0_0: Version = version(0, 0);
+pub const VERSION_0_1: Version = version(0, 1);
+pub const FEE_VERSION: Version = version(0, 2);
+pub const EPOCH_VERSION: Version = version(0, 3);
+pub const DRB_AND_HEADER_UPGRADE_VERSION: Version = version(0, 4);
+pub const DA_UPGRADE_VERSION: Version = version(0, 5);
+pub const VID2_UPGRADE_VERSION: Version = version(0, 6);
+pub const MAX_SUPPORTED_VERSION: Version = DA_UPGRADE_VERSION;
```
sveitser, line 18: I would expect this to be 0.6

data/v3/header.json
sveitser: I checked some headers on decaf v3 and our reference test seems to simply have been wrong. Thanks for fixing.

crates/versions/src/lib.rs
twittner: :+1: Changed to `assert!` in 5aec9d9.

crates/versions/src/lib.rs
twittner: The light client assumes it is 0.5: https://github.com/EspressoSystems/espresso-network/blob/ce0cf2ce357fc0bd39ac745e31786664ac542667/light-client/src/consensus/quorum.rs#L38

CC: @ss-es, @mrain

PR #3947, opened 2026-02-11
increase catchup timeout to 2s and make it an env variable

imabdulbasit:


crates/builder/src/non_permissioned.rs
ss-es, line 45: is this function just for tests?

imabdulbasit: no the binary uses it but only like chain config and some other stuff

PR #3948, opened 2026-02-12
prevent catchup when previous view is still catching up

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/task-impls/src/helpers.rs
+++ b/crates/hotshot/task-impls/src/helpers.rs
@@ -772,16 +772,15 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
         .context(info!("Failed to fetch proposal"))?;
     }
 
-    if wait_for_previous {
-        if wait_for_previous_view(parent_qc.view_number(), event_receiver.clone())
+    if wait_for_previous
+        && wait_for_previous_view(parent_qc.view_number(), event_receiver.clone())
             .await
             .is_some()
-        {
-            tracing::info!(
-                "Successfully waited for previous view {:?}",
-                parent_qc.view_number()
-            );
-        }
+    {
+        tracing::info!(
+            "Successfully waited for previous view {:?}",
+            parent_qc.view_number()
+        );
     }
 
     let consensus_reader = consensus.read().await;
--- a/crates/hotshot/task-impls/src/quorum_vote/handlers.rs
+++ b/crates/hotshot/task-impls/src/quorum_vote/handlers.rs
@@ -319,19 +319,18 @@ pub(crate) async fn update_shared_state<TYPES: NodeType, V: Versions>(
 
     drop(consensus_reader);
 
-    if wait_for_previous {
-        if wait_for_previous_view(parent_view_number.unwrap(), receiver.activate_cloned())
+    if wait_for_previous
+        && wait_for_previous_view(parent_view_number.unwrap(), receiver.activate_cloned())
             .await
             .is_some()
-        {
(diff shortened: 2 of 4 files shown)


crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -308,6 +308,9 @@ pub(crate) async fn update_shared_state<TYPES: NodeType, V: Versions>(
             .cloned()
     });
 
+    let wait_for_previous = parent_view_number
+        .is_some_and(|view_number| consensus_reader.is_view_validating(view_number));
+
```
ss-es, line 313: so this is true if the parent view is in the process of being validated?

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -316,6 +319,19 @@ pub(crate) async fn update_shared_state<TYPES: NodeType, V: Versions>(
 
     drop(consensus_reader);
 
+    if wait_for_previous
+        && wait_for_previous_view(parent_view_number.unwrap(), receiver.activate_cloned())
```
ss-es, line 323: this potentially blocks for up to 6s, but this only happens on the vote path after we have all dependencies so this timeout is only triggered once per view right?

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -316,6 +319,19 @@ pub(crate) async fn update_shared_state<TYPES: NodeType, V: Versions>(
...
+            .await
+            .is_some()
+    {
+        tracing::info!("Successfully waited for previous view {parent_view_number:?}");
+        maybe_validated_view = consensus
+            .read()
+            .await
+            .validated_state_map()
+            .get(&parent_view_number.unwrap())
+            .cloned()
```
ss-es, line 333: if I understand correctly, don't you want to also update `maybe_parent` here to avoid the call to `fetch_proposal`? 

as it is it seems like we will fetch the proposal no matter what happens here

crates/hotshot/task-impls/src/helpers.rs
ss-es, line 760: similarly, don't we want to avoid this `fetch_proposal`? i.e. shouldn't we try waiting for the previous view, then bind `let vsm_contains_parent_view =...` and only then check if we need to `fetch_proposal`?

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
ss-es: (and similarly the other one happens only on the proposal path?)

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
bfish713: yes

crates/hotshot/task-impls/src/quorum_vote/handlers.rs
bfish713: yeah exactly

crates/hotshot/task-impls/src/helpers.rs
bfish713: yes this is right, but I don't think it really matters, I don't see how we could even get to this point with out having stored an empty state in `validated_state_map` if `wait_for_previous` is true.  I think fetch proposal just covers the edge case where there was a timeout and parent view is old, and we never saw a proposal for that old view.  In that case I don't think `wait_for_previous` would be true.  Either way it's better to change the order like you suggest

PR #3960, opened 2026-02-16
Mainnet release fixes

ss-es:


sequencer/src/api/endpoints.rs
```diff
@@ -185,7 +188,55 @@ where
...
+                    .await
+                    .map_err(|err| merklized_state::Error::Custom {
+                        message: format!(
+                            "failed to load RewardMerkleTreeV2Data from storage at height \
+                             {height}: {err}"
+                        ),
+                        status: StatusCode::NOT_FOUND,
+                    })?;
+
+            let tree_data = bincode::deserialize::<RewardMerkleTreeV2Data>(&tree_bytes)
```
sveitser, line 220: @ss-es Does this mean anyone can trigger unbounded memory use by making parallel requests to this endpoint? It looks to me like whenever this endpoint is hit we will load a new tree into memory which will be dropped only after the request is done.

ss-es: Not really (well maybe, but it would be hard) -- note we aren't reconstructing the `RewardMerkleTreeV2` here. `RewardMerkleTreeV2Data` is just an array of account-balance pairs, which is about ~5 MB when we have 30k delegators

sequencer/src/api.rs
```diff
@@ -1308,7 +1282,51 @@ pub(crate) trait RewardMerkleTreeDataSource: Send + Sync + Clone + 'static {
...
+                                Some((serialized_account, serialized_proof))
+                            });
+
+                    let Ok(_) = self.persist_proofs(finalized_hotshot_height, iter).await else {
+                        return Ok(());
+                    };
+
+                    // tree is dropped here
+                }
+            } else if (height + node_state.node_id).is_multiple_of(30) {
```
sveitser, line 1331: I think we should have a comment as to why we add the node_id here and what we expect to achieve with that.

sequencer/src/api.rs
```diff
@@ -1308,7 +1282,51 @@ pub(crate) trait RewardMerkleTreeDataSource: Send + Sync + Clone + 'static {
...
             // ultimately, we would retry in the next iteration anyway;
             // all the information already exists.
-            if height.is_multiple_of(30) {
+            //
+            // in tests we run this check every block, while in release builds we do not. however,
+            // in either case the actual work should still not occur every block, since the proofs
+            // are not generated unless they were not already stored for the previous light client
+            // finalized height.
+
+            if cfg!(any(test, feature = "testing")) {
```
sveitser, line 1291: Is there anything we do different in this branch vs the `else if` branch? Otherwise could we combine it? If it's something that we should do but don't have time I'm also okay with a TODO comment.

tests/upgrades.rs
```diff
@@ -97,7 +97,7 @@ where
     // Calculate reward claim deadline if applicable
     let reward_claim_deadline_block_height =
         if Target::version() >= DrbAndHeaderUpgradeVersion::version() {
-            // Rewards should be claimable after 2 epochs on the upgrade version (min 200 blocks)
+            // Rewards should be claimable after 3 epochs on the upgrade version
```
bfish713, line 100: It's still 2 epochs on the upgrade version right, we're just starting the upgrade 1 epoch later now?

sequencer/src/api.rs
bfish713, line 1331: Why add the node_id in here?

ss-es: I'll add a comment -- basically I wanted to mitigate having this calculation happen at the same block height on all nodes, and this was a cheap way to get that

sequencer/src/api.rs
ss-es: yes, in the first branch:

- we store proofs every block height
- we never garbage collect proofs

and without these changes it's a huge pain to make the tests pass. I did try to gate the specific parts that were different initially but I think it's much clearer to have a separate block

tests/upgrades.rs
ss-es: I think this is just the amount we add to whenever the upgrade actually happens, not any relation to the upgrade height (which comes from genesis I think?)

sequencer/src/api.rs
ss-es: I'll add a comment -- basically I wanted to mitigate having this calculation happen at the same block height on all nodes, and this was a cheap way to sort of get that

ss-es: [40fb4f0](https://github.com/EspressoSystems/espresso-network/pull/3960/commits/40fb4f0b499887a175c44309471ba40264d941a6)

PR #3972, opened 2026-02-19
fix(test): flaky CI fire fighting

sveitser:
Retry flaky hotshot integration test up to 3 times
- test_success_with_async_delay_2_with_epochs 
- test_vid2_success intermittently 

Start the timeout timer for native demo tests only after demo is started up. Currently the timeout is 60 seconds and startup takes about 60 seconds, which leads to somewhat frequent failures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -238,7 +238,7 @@ jobs:
           cargo nextest run --archive-file nextest-archive-postgres.tar.zst  --verbose --no-fail-fast \
           --workspace-remap $PWD \
           --partition hash:${{ matrix.partition }}/10
-        timeout-minutes: 20
+        timeout-minutes: 30
 
   test-sqlite:
     needs: build-test-artifacts-sqlite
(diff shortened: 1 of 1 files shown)


tests/common/mod.rs
```diff
@@ -254,7 +254,12 @@ impl TestRuntime {
 
     pub async fn from_requirements(requirements: TestRequirements) -> Result<Self> {
         let config = TestConfig::from_env(requirements)?;
-        Self::initialize(config, Duration::from_secs(30)).await
+        tokio::time::timeout(
```
imabdulbasit: how is this going to help flaky tests? initialize() function would return an error anyway after the timeout (Duration::from_secs(30) )that we are passing?

sveitser: Let me check. I just wanted to prevent it from taking forever to initialize since the start time was moved to after initialize.

sveitser: The inner timeout is to wait for the sequencer client to return something but the tests waits for other services too, like submit transactions.

sveitser: It could definitely all use a lot of refactoring.

sveitser: cleaned it up a bit: https://github.com/EspressoSystems/espresso-network/pull/3972/commits/514604e4b3f2e683805e1d26bfdf03bd0af99037

PR #3979, opened 2026-02-20
add `--stake-table-only` flag to reset-storage binary to clear stake table data only

imabdulbasit:
Some node operators have experienced issues where stake table events fetched from L1 and stored in persistence are incomplete. On node startup, add_epoch_root() is called with the stored events, which means the stake table cannot beconstructed and the node gets stuck. The only solution is to refetch these events from L1. To do this, all stored stake table events and stake tables in persistence must be deleted.

This PR adds a `--stake-table-only` flag to the storage reset binary that deletes only the stake table data. This makes it easier for node operators and avoids having them manually delete the tables.

### DIFF
--- a/sequencer/src/persistence.rs
+++ b/sequencer/src/persistence.rs
@@ -1811,7 +1811,8 @@ mod tests {
         // Validators cleared
         let loaded = storage
             .load_all_validators(EpochNumber::new(1), 0, 10)
-            .await?;
+            .await
+            .unwrap_or_default();
         assert!(loaded.is_empty());
 
         let event3 = StakeTableEvent::Delegate(Delegated {


sequencer/src/persistence/sql.rs
```diff
@@ -2807,6 +2807,28 @@ impl MembershipPersistence for Persistence {
...
+            .await?;
+        #[cfg(feature = "embedded-db")]
+        {
+            query("DELETE FROM stake_table_events")
+                .execute(tx.as_mut())
+                .await?;
+            query("DELETE FROM stake_table_events_l1_block")
+                .execute(tx.as_mut())
+                .await?;
+            query("DELETE FROM epoch_drb_and_root")
```
sveitser, line 2827: @imabdulbasit Should we also clear the `stake_table_validators` table? I guess it would not be backfilled, so maybe not.

imabdulbasit: yes we should delete it

https://github.com/EspressoSystems/espresso-network/pull/3979/commits/681dc151d70e690edb6283253d8f48270c4293a7

PR #3989, opened 2026-02-25
 truncate legacy tables

imabdulbasit:
The transition to proof of stake required some tables to be migrated to v2 tables because of a change in serialization. Now that decaf and mainnet have upgraded to proof of stake, these old v1 tables are no longer required so we can just delete the data. Some of these tables have foreign key constraints so I think we shouldn't drop them, but only delete the data.


hotshot-query-service/migrations/sqlite/V1000__delete_old_table_data.sql
```diff
@@ -0,0 +1,4 @@
+-- Delete data from old tables
+-- The data has been migrated to the new tables (leaf2, vid2)
```
sveitser, line 2: This comment is nice, might as well also add it to the sequencer migrations.

PR #3994, opened 2026-02-27
chore: reduce build times, cleanup

sveitser:
- Remove sequencer dependency from hotshot-examples, extract shared infra code into a library crate
 - Remove `hotshot-examples` from default-members to improve compilation times
 - Remove `espresso-crypto-helper` from default-members to prevent vendored openssl feature 
  unification leaking to the workspace (~3min build overhead)
  - Exclude `espresso-crypto-helper` and `hotshot-examples` from nextest workspace builds via nextest_excludes
  - Remove workflow telemetry action because CPU and memory charts no longer render and repo seems unmaintained.
  - Add --lib --bins --tests to nextest/nextest-archive to skip compiling examples (e.g. query service simple-server, which takes ~ 2minutes)
  - Switch just clippy and CI clippy-postgres from default-members to explicit --workspace --exclude for consistency
  - Add --timings flag to all CI build and test archive steps
  - Add cargo-timing-summary script that extracts top-20 slowest crates from cargo timing HTML into a markdown table
  - Add cargo timing upload and GitHub step summary to build.yml and test.yml workflows

### DIFF
--- a/crates/espresso/node/src/api.rs
+++ b/crates/espresso/node/src/api.rs
@@ -3640,7 +3640,9 @@ mod test {
 
     #[test_log::test(tokio::test(flavor = "multi_thread"))]
     async fn test_epoch_reward_upgrade() {
-        test_upgrade_helper(Upgrade::new(
+        // Use fewer nodes: epoch mode from view 0 is resource-heavy on CI with
+        // postgres Docker containers, causing view timeouts and consensus stall.
+        test_upgrade_helper_with_nodes::<3>(Upgrade::new(
             versions::DRB_AND_HEADER_UPGRADE_VERSION,
             versions::EPOCH_REWARD_VERSION,
         ))
@@ -3648,11 +3650,13 @@ mod test {
     }
 
     async fn test_upgrade_helper(upgrade: Upgrade) {
+        test_upgrade_helper_with_nodes::<5>(upgrade).await;
+    }
+
+    async fn test_upgrade_helper_with_nodes<const NUM_NODES: usize>(upgrade: Upgrade) {
         // wait this number of views beyond the configured first view
         // before asserting anything.
         let wait_extra_views = 10;
-        // Number of nodes running in the test network.
-        const NUM_NODES: usize = 5;
         let port = reserve_tcp_port().expect("OS should have ephemeral ports available");
         let epoch_start_block = if upgrade.base >= versions::EPOCH_VERSION {
             0


sveitser:
So because we have an active release branch we don't have a cache for main right now.

Before (20 minutes to build test artifacts)

Screenshot 2026-02-27 at 12 11 23

After (13 minutes to build test artifacts)
Screenshot 2026-02-27 at 12 11 59

crates/hotshot/examples/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod infra;
```
sveitser, line 1: @ss-es are you okay with this? I think it makes examples compile faster that inlining the module and i did some other things to exclude them by default. Currently we are building them quite often and I think we only need to compile them when we actually want to run them.

.github/workflows/lint.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
       - name: Run clippy
         run: |
-          cargo clippy --features testing --all-targets --keep-going -- -D warnings
+          cargo clippy --workspace --exclude sequencer-sqlite --exclude espresso-dev-node --features testing --all-targets --keep-going -- -D warnings
```
imabdulbasit, line 49: why the dev-node is excluded?

Cargo.toml
```diff
@@ -319,16 +319,59 @@ inherits = "dev"
...
+opt-level = 1
+[profile.test.package.espresso-types]
+opt-level = 1
+[profile.test.package.vid]
+opt-level = 1
+[profile.test.package.hotshot-builder-api]
+opt-level = 1
+[profile.test.package.hotshot-builder-legacy]
+opt-level = 1
+[profile.test.package.hotshot-utils]
```
imabdulbasit, line 428: why do some packages have opt-level=1 and some have opt-level=3?

crates/versions/Cargo.toml
```diff
@@ -7,8 +7,8 @@ edition = "2024"
 [dependencies]
 bincode = { workspace = true }
 serde = { workspace = true }
-vbs = { workspace = true }
 thiserror = { workspace = true }
+vbs = { workspace = true }
```
imabdulbasit, line 11: I think this should fail cargo-sort but I don't see it failing. not sure why

scripts/cargo-timing-summary
twittner, line 1: This looks hard to maintain. Is this format even stable?

crates/versions/Cargo.toml
sveitser: i think cargo sort did this, we don't run it on CI

.github/workflows/lint.yml
sveitser: It runs with `--workspace` now so without the exclude it would enable embedded-db. This is because this PR removes crypto-helper from default members due problems with feature unification where openssl vendored gets pulled by crypto helper  iwhich takes a really long time to compile. We still want to run clippy on everything. I think for clippy the vendored openssl doesn't have a strong impact on runtime.

Cargo.toml
sveitser: There's a comment further up. In theory everything should just work with default opt-level=0 for tests but it does not. So we selectively turn on some optimizations to make tests pass at the expense of slower compilation times. I thought what is there now works but 2 hotshot tests just failed on a re-run.

For some of the compute heavy crates we turn on more optimizations. It's a bit of a guessing game what really helps because stuff only fails on the CI anyway. I did some profiling on tests that were failing a few commits ago and it seemed that for the hotshot network tests most of the compute happens in ark crates, so I bumped those up a bit too in the hopes of having more stable tests. But it's not quite working.

I started this because on CI having opt-level 1 for query service makes its test target compilation take about 300s but for opt-level 0 it's only ~100 seconds.

sveitser: Ah the new failures is flaky ark-srs download from github so not my changes. yay

justfile
```diff
@@ -67,14 +67,14 @@ lint *args:
     just clippy {{args}} -- -D warnings
 
 clippy *args:
-    # check all targets in default workspace members
-    cargo clippy --features testing --all-targets {{args}}
+    # check all targets in workspace (excluding embedded-db crates)
+    cargo clippy --workspace --exclude sequencer-sqlite --exclude espresso-dev-node --features testing --all-targets {{args}}
```
ss-es: does the `--exclude` really save us much? I find it somewhat confusing to exclude crates here; I would expect `just clippy` to check all crates

scripts/cargo-timing-summary
sveitser: I don't know. We can also output json but then we don't get the html, getting both directly seems to be impossible.

justfile
sveitser: It would turn on the embedded db feature. Previously that was turned off because of not passing `--workspace`.

scripts/cargo-timing-summary
twittner: I would prefer not to run with timings on every build. This can be done whenever needed. However if deemed necessary, why not keep the HTML as is?

sveitser: Does it take significant extra time to build with timings (in case you know)? I find it useful to have something that I can inspect easily without having to do extra work. I don't need it on every build but having to turn it on then do CI runs is a lot of overhead. Which just means longer build time, and more time goes by until issues are detected. The html fine works fine for deeper digging and is useful for checking which crates block other crates but if I'm trying to fix some stuff like slow build times of certain crates it's annoying to have to click a few times for every build and to have a format that's not easy to compare.

twittner: > Does it take significant extra time to build with timings (in case you know)?

I don't know, but I don't think so. I am mostly wary of bash scripts and think that for the (hopefully) rare cases where slow build times need to be analysed the complete HTML is more useful. And if I were to do it, I would just run `cargo build --timings` locally when needed. Anyway, just my two cents.

justfile
```diff
@@ -67,14 +67,14 @@ lint *args:
     just clippy {{args}} -- -D warnings
 
 clippy *args:
-    # check all targets in default workspace members
-    cargo clippy --features testing --all-targets {{args}}
+    # check all targets in workspace (excluding embedded-db crates)
+    cargo clippy --workspace --exclude espresso-node-sqlite --exclude espresso-dev-node --features testing --all-targets {{args}}
```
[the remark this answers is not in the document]
sveitser: clippy is fast compared to compilation

PR #4009, opened 2026-03-03
Backport some fixes to cliquenet.

twittner:
1. Add length to trailer for future extensibility.
2. Fix cancellation-safety in `Retry::receive`.
3. Allow adding peers with explicit role.
4. Add a length check in `Retry::send`.
5. Add `Network::close` to enable implementing shutdown without drop.
6. Fix spelling error in metrics.
7. Fix message size check when receiving frames.

### DIFF
--- a/crates/hotshot/types/src/data.rs
+++ b/crates/hotshot/types/src/data.rs
@@ -417,7 +417,15 @@ impl From<VidCommon2> for VidCommon {
     }
 }
 
-impl VidCommon {
+/// Borrowed view of [`VidCommon`] that avoids cloning large common data.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum VidCommonRef<'a> {
+    V0(&'a VidCommon0),
+    V1(&'a VidCommon1),
+    V2(&'a VidCommon2),
+}
+
+impl<'a> VidCommonRef<'a> {
     pub fn is_consistent(&self, comm: &VidCommitment) -> bool {
         match (self, comm) {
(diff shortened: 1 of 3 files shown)


crates/cliquenet/src/retry.rs
```diff
@@ -175,20 +183,26 @@ where
     }
 
     pub async fn receive(&self) -> Result<(K, Bytes)> {
-        let pending = self.inner.pending.lock().pop_first();
-        if let Some((_, Pending { src, data, trailer })) = pending {
+        let first_entry = self.inner.pending.lock().pop_first();
+        if let Some((key, val)) = first_entry {
+            let pending = &self.inner.pending;
+            // Put the pending value back if the future is dropped:
+            let guard = guard((key, val.clone()), |(k, v)| {
```
akonring, line 190: Alternatively, one could just put the value back immediately and remove it again after successful send?

crates/cliquenet/src/retry.rs
akonring, line 228: Should we do the same here?

crates/cliquenet/src/retry.rs
twittner: Yes, this is possible. A concurrent `receive` could then pick up the value and we deliver the message multiple times. Downstream needs to be able to handle this anyway, but using a scope guard here we can avoid that.

crates/cliquenet/src/retry.rs
twittner: I think the scope guard is the better approach as it avoids redundant message deliveries.

Issue #4011, opened 2026-03-03
Create a client for submitting transactions to some block builders.


zacshowa:
In some situations it would be beneficial for users to submit transactions directly to the networks block builders.

To support this, we should implement a Go client that can submit data to a block builder directly, while also reading from a single separate query endpoint.

PR #4012, opened 2026-03-03
Add EspressoBuilderClient

zacshowa:
Closes:

Asana Ticket: [Create a new client in the go SDK to send txn's directly to builders](https://app.asana.com/0/home/1209393650288774/1212077732927260)
&
Issue #4011 

### This PR:

- Modifies the `EspressoClient` to follow a composable interface pattern.
- Refactors the existing client to abstract submitting to endpoints to an interface, allowing the client to utilize the newly introduced `builderSubmitter` and `querySubmitter`
- Refactors the tests to have less duplicate code

### This PR does not:
- Use this client in the nitro integration (This will be done at a later time.)

### Key places to review:
`client.go`
`builder_submitter.go`
`query_submitter.go`
`client_test.go`

The error retrieval functionality is intended for use in logging IMO. Assuming that we only care if one transaction makes it to one of many builders, we don't need to necessarily retry or throw an error if one submit request fails.
Therefore, we may want to record these failures in some method for use in logging/warning alerting if some builders are consistently down. 

The error retrieval functionality does this by storing an individual submit requests error logs in a state variable until the next call. This list can be retrieved by calling `client.Submitter.GetPreviousSubmissionErrors()`. These can then be handled however the client user would like.

### How to test this PR:
In `espresso-network/sdks/go`, run `go test ./... -v`

You will need to ensure you have the dynamic library from the espresso-crypto-helper installed to run these tests (I figured these could be fixed in a later PR). I will be adding instructions on how to properly install this to the readme before this is merged.

---
  - https://app.asana.com/0/0/1212077732927260


sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	if !strings.HasSuffix(url, "/") {
+		url += "/"
+	}
+	return url
+}
+
+func NewBuilderClient(queryUrl string, builderUrls []string) *BuilderClient {
+	sanitizedQueryUrl := sanitizeUrl(queryUrl)
+	builderClients := make([]*http.Client, len(builderUrls))
+	for i, url := range builderUrls {
```
lukeiannucci: do we need to make sure urls are present?

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	var res types.LimitsData
+	if err := c.get(ctx, &res, "availability/limits"); err != nil {
+		return types.LimitsData{}, err
+	}
+	return res, nil
+}
+
+func (c *BuilderClient) SubmitTransaction(ctx context.Context, tx types.Transaction) (*types.TaggedBase64, error) {
+	c.previousSubmitErrors = make([]error, 0)
+	for clientIdx, url := range c.builderUrls {
```
lukeiannucci: i do wonder if every txn we always submit to every builder what overhead this can cause on espresso. im wondering if we should submit to one builder, if resubmission comes and not finalized, move onto the one, and repeat. Just a thought

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+
+// Apply sanitation rules to url before creating the client.
+// currently, this is just ensuring the url has the suffix `/`
+// but more rules can be applied here later.
+func sanitizeUrl(url string) string {
+	if !strings.HasSuffix(url, "/") {
+		url += "/"
+	}
+	return url
+}
```
ImJeremyHe: Nice. Can you move this function to a more general file? I think this can be used by other clients

ImJeremyHe (commented):
I think it is a bit confusing for other users to understand why a builder client is needed in some scenarios. Can you elaborate it in the file?

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return res, nil
+}
+
+func (c *BuilderClient) FetchTransactionByHash(ctx context.Context, hash *types.TaggedBase64) (types.TransactionQueryData, error) {
+	if hash == nil {
+		return types.TransactionQueryData{}, fmt.Errorf("hash is nil")
+	}
+	var res types.TransactionQueryData
+	if err := c.get(ctx, &res, "availability/transaction/hash/%s", hash.String()); err != nil {
```
Ayiga: We're hitting the `availability` API on the `BuilderClient`?  `Builder` doesn't have an `availability` API, only the `Query Service` does.

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return res, nil
+}
+
+func (c *BuilderClient) FetchExplorerTransactionByHash(ctx context.Context, hash *types.TaggedBase64) (types.ExplorerTransactionQueryData, error) {
+	if hash == nil {
+		return types.ExplorerTransactionQueryData{}, fmt.Errorf("%w: hash is nil", ErrPermanent)
+	}
+	var res types.ExplorerTransactionQueryData
+	if err := c.get(ctx, &res, "explorer/transaction/hash/%s", hash.String()); err != nil {
```
Ayiga: We're hitting the `explorer` API on the `BuilderClient`?  `Builder` doesn't have an `availability` API, only the `Query Service` does.

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (c *BuilderClient) getRawMessage(ctx context.Context, format string, args ...any) (json.RawMessage, error) {
+	res, err := c.tryGetRequest(ctx, c.queryUrl, format, args...)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrPermanent, err)
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != 200 {
```
Ayiga:
```suggestion
	if res.StatusCode != http.StatusOK {
```

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	for clientIdx, url := range c.builderUrls {
+		response, err := c.tryPostRequest(ctx, url, clientIdx, tx)
+
+		if err != nil {
+			c.previousSubmitErrors = append(c.previousSubmitErrors, err)
+			return nil, fmt.Errorf("%w: %v", ErrPermanent, err)
+		}
+
+		defer response.Body.Close()
+		if response.StatusCode != 200 {
```
Ayiga:
```suggestion
		if response.StatusCode != http.StatusOK {
```

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+
+}
+
+func (c *BuilderClient) tryPostRequest(ctx context.Context, baseUrl string, clientIndex int, tx types.Transaction) (*http.Response, error) {
+	marshalled, err := json.Marshal(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	request, err := http.NewRequestWithContext(ctx, "POST", baseUrl+"txn_submit/submit", bytes.NewBuffer(marshalled))
```
Ayiga: `tryPostRequest` seems to want to be a generic `POST` handling function.  Yet we're hard-coding a `path` here.  This path should be resolved in the calling function instead of here.

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+	builderClients       []*http.Client
+	queryUrl             string
+	queryClient          *http.Client
+	previousSubmitErrors []error
+}
+
+// Apply sanitation rules to url before creating the client.
+// currently, this is just ensuring the url has the suffix `/`
+// but more rules can be applied here later.
+func sanitizeUrl(url string) string {
```
Ayiga: This function is a misnomer.  This isn't sanitizing a Url, it's ensuring that the passed string has a trailing `/` .

This `function` is ultimately being utilized to try and ensure a base `URL` that has a `trailing slash` in it, for the sole purpose of making `URL` sub-path construction easier.

`URL` / `path` manipulation can be more easily resolved using [path](https://pkg.go.dev/path) or [net/url](https://pkg.go.dev/net/url) already available in go's `stdlib`.

sdks/go/client/builder_client.go
```diff
@@ -0,0 +1,324 @@
...
+
+func NewBuilderClient(queryUrl string, builderUrls []string) *BuilderClient {
+	sanitizedQueryUrl := sanitizeUrl(queryUrl)
+	builderClients := make([]*http.Client, len(builderUrls))
+	for i, url := range builderUrls {
+		builderUrls[i] = sanitizeUrl(url)
+	}
+
+	return &BuilderClient{
+		queryUrl:       sanitizedQueryUrl,
```
Ayiga: I think in general it would be better to embed a separate `Query` client rather than re-implement the `Query` client here.  That being said, it's probably even better to keep them separated for cleaner separation of concerns.

If we want a unified `Client` that combines a `Builder` and a `Query` client (`EspressoClient`), then we could easily create a new `type` / `interface` that embeds both.

sdks/go/client/builder_client.go
zacshowa: Yeah the naming on this client isn't super clear. The client takes one or more builder URLs, and a query service URL. It then uses the builder url's for specifically post requests, while using the query url for get requests. I could probably make that more clear in some of the naming conventions though?

Anyways, I think this will change quite a bit. we've since decided that I should instead add builder url's as an optional configuration option to the normal client.

sdks/go/client/builder_client.go
zacshowa: Hmm. I agree at a large level. The naming isn't great, but it was intended to capture the spirit of "here is where we do the pre-processing of the URLs, as this was something that was repeated across go clients. I will think on how we can used the linked libraries to accomplish this better as I refactor as requested.

sdks/go/client/builder_client.go
zacshowa: Yeah we could probably do some more checking on the URL. I'll add this while I work on some of the feedback provided by theo and jeremy in this regard. 

I think that I also do need to initialize the builder clients. I somewhat naively expected the default of the `*http.Client` array to be `http.DefaultClient`. I realize this won't be the case. I am a bit curious if it would be better to have a single client, it doesn't seem like we ever give these http clients a url directly, but I also don't know if they store any intermediate state that would be important for any particular connection.

sdks/go/client/builder_client.go
zacshowa: Yeah I think this is what we will likely end up doing in this PR. I'm getting inspiration from Toralf's [robusta](https://github.com/EspressoSystems/timeboost/tree/c24b1297ace7e14a8f8b831e8dd8ffc541aad55d/robusta) client from timeboost.

sdks/go/client/builder_client.go
zacshowa: Yeah I agree the naming is a bit off.  I think we can work on this as I refactor this PR as referenced [here](https://github.com/EspressoSystems/espresso-network/pull/4012/changes#r2887356969)

submitter-context.txt
```diff
@@ -0,0 +1,259 @@
+# Espresso Submitter Context
```
ImJeremyHe: What is this file for?

zacshowa: Aah I didn't mean to commit this. It was something I did locally to pass context to claude from one repo to another. With the tool I use for sandboxing, claude can't read outside of the repo I start it in, and I wanted it to gather context on the way something was written in the nitro repo.

sdks/go/client/builder_submitter.go
```diff
@@ -0,0 +1,102 @@
...
+		builderUrls:    builderUrls,
+		builderClients: builderClients,
+	}, nil
+}
+
+// This error is meant to signal to consumer code (via the client methods), that despite the txn submission being successful,
+// some attempts to submit were unsuccessful. This indicates that the user should check the cached errors in previousSubmitErrors
+var ErrAllBuildersFailed = errors.New("submission to all builders failed, check previousSubmitErrors")
+
+// SubmitTransaction:
```
Sneh1999, line 47: nit: these comments only mention what the params are and not what the function is supposed to do

sdks/go/client/query.go
```diff
@@ -27,6 +27,8 @@ type QueryService interface {
 	FetchVidCommonByHeight(ctx context.Context, blockHeight uint64) (types.VidCommon, error)
 	// Get the transaction by its hash from the explorer.
 	FetchExplorerTransactionByHash(ctx context.Context, hash *types.TaggedBase64) (types.ExplorerTransactionQueryData, error)
+	// Stream payloads starting from the given height
```
Sneh1999, line 30: why are we adding this one? I dont think we use this anywhere

sdks/go/client/query_submitter.go
```diff
@@ -0,0 +1,69 @@
...
+	url := formatUrl(baseUrl)
+
+	return &QuerySubmitter{
+		baseUrl: url,
+		client:  http.DefaultClient,
+	}
+}
+
+// The QuerySubmitter's implementation of submit transaction, which targets the query services transaction submission endpoint.
+func (q *QuerySubmitter) SubmitTransaction(ctx context.Context, tx types.Transaction) (*types.TaggedBase64, error) {
```
Sneh1999, line 30: do we still need this when we have the builder flow?

sdks/go/client/utilities.go
```diff
@@ -0,0 +1,13 @@
+package client
+
+import "strings"
+
+// Apply formatting rules to url before creating the client.
+// currently, this is just ensuring the url has the suffix `/`
+// but more rules can be applied here later.
+func formatUrl(url string) string {
+	if !strings.HasSuffix(url, "/") {
```
Sneh1999, line 9: I dont think we need a function for this it just two lines

justfile
```diff
@@ -372,3 +372,12 @@ contracts-test-network *args='-vv':
         exit 1
     fi
     forge test --match-test test_Network_ --jobs 1 {{args}}
+
+claude-sandbox:
```
Sneh1999: why are we adding this here?

sdks/go/client/client.go
```diff
@@ -19,19 +17,76 @@ var _ QueryService = (*Client)(nil)
...
 
+type EspressoClientConfigOption func(*EspressoClientConfig)
+
+type EspressoClientConfig struct {
+	BaseUrl              string
+	TransactionSubmitter SubmitAPI
+}
+
+var DefaultEspressoClientConfig = EspressoClientConfig{
+	BaseUrl: "query.main.net.espresso.network",
```
jjeangal, line 28: Should we define this BaseUrl somewhere else or is this the only place we define that value?

sdks/go/client/client.go
```diff
@@ -19,19 +17,76 @@ var _ QueryService = (*Client)(nil)
...
+}
+
+// Validate that the espressoClientConfig is valid.
+func ValidateEspressoClientConfig(config EspressoClientConfig) error {
```
jjeangal, line 32: Is the client config simply being different than nil enough to validate it?

sdks/go/client/client.go
zacshowa: This is the only place that we give it a default, but in theory the client user should be reading in their own config.

sdks/go/client/utilities.go
zacshowa: It was a series of lines that was repeated a lot in the constructors of various clients. I made this it's own function under the assumption that if the clients have needed the same set of rules to pre-process URL's, they will likely share at least some of those rules going forward. So if we need the clients to share some set of rules, those rules can be defined in this function, and updated for each of the clients at the same time, rather than needing to update the clients individually.

sdks/go/client/query.go
zacshowa: It is a public method exposed by the clients, so it should probably be in the interface? and to refactor the tests, I needed to have it exposed in the interface.

justfile
zacshowa: This was just something I was playing around with during the retreat as a way to run claude sandboxed on the repo via some nix packages for different platforms. 

It's not relevant to this PR, and I didn't mean to include it. but perhaps it could be useful in a separate isolated PR?

sdks/go/client/client.go
zacshowa: I'm not sure there is much we can do other than checking that the transaction submitter is not nil as that's where most of the business logic is handled, but I'm totally open to ideas!

PR #4015, opened 2026-03-05
limit incomming responses by request hash, don't validate if the request is fulfilled

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/request-response/Cargo.toml
+++ b/request-response/Cargo.toml
@@ -7,7 +7,7 @@ license = "MIT"
 [dependencies]
 anyhow = { workspace = true }
 async-broadcast = { workspace = true }
-async-lock.workspace = true
+async-lock = { workspace = true }
 async-trait = { workspace = true }
 bincode = { workspace = true }
 blake3 = { workspace = true }


request-response/src/lib.rs
```diff
@@ -633,6 +637,32 @@ impl<
...
+        request_hash: RequestHash,
+        outgoing_requests: &OutgoingRequestsMap<Req>,
+    ) -> Result<NamedSemaphorePermit<RequestHash>, NamedSemaphoreError> {
+        while outgoing_requests.read().contains_key(&request_hash) {
+            let permit = incoming_responses.try_acquire(request_hash);
+            let permit = match permit {
+                Ok(permit) => permit,
+                Err(NamedSemaphoreError::PerKeyLimitReached) => {
+                    tokio::time::sleep(Duration::from_millis(100)).await;
+                    continue;
```
ss-es, line 649: why do we retry in this case but not in the global limit case?

request-response/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT"
 [dependencies]
 anyhow = { workspace = true }
 async-broadcast = { workspace = true }
+async-lock.workspace = true
```
ss-es: nit: just for consistent formatting with the other entries

```suggestion
async-lock = { workspace = true }
```

request-response/src/lib.rs
bfish713: To me the global limit is to prevent us from queuing up too many pending tasks.  The request limit is to limit entry into the validation logic

PR #4016, opened 2026-03-05
Add catchup endpoint for RewardMerkleTreeV2 to get from consensus state

imabdulbasit:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/sequencer/api/catchup.toml
+++ b/sequencer/api/catchup.toml
@@ -143,7 +143,7 @@ Return all RewardMerkleTreeV2 accounts stored for the requested height, paginate
 """
 
 [route.reward_merkle_tree_v2]
-PATH = ["/:height/:view/reward-merkle-tree-v2"]
+PATH = ["/reward-merkle-tree-v2/:height/:view"]
 ":height" = "Integer"
 ":view" = "Integer"
 DOC = """
--- a/sequencer/src/catchup.rs
+++ b/sequencer/src/catchup.rs
@@ -409,7 +409,7 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
                 // the reward-state-v2 endpoint which returns from storage decided state
                 let tree_bytes = match client
                     .inner
-                    .get::<Vec<u8>>(&format!("catchup/{height}/{}/reward-merkle-tree-v2", *view))
+                    .get::<Vec<u8>>(&format!("catchup/reward-merkle-tree-v2/{height}/{}", *view))
                     .send()
                     .await
                 {


sequencer/src/catchup.rs
```diff
@@ -398,17 +398,35 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
+                        );
+                        client
+                            .inner
+                            .get::<Vec<u8>>(&format!(
+                                "reward-state-v2/reward-merkle-tree-v2/{height}"
+                            ))
+                            .send()
+                            .await?
+                    },
+                };
```
ss-es, line 429: I think we should try them simultaneously, and just take the first one that returns successfully e.g. something like this


```
use futures::future::select_ok;

.fetch(retry, |client| async move {
    let fut1 = async {
        client
            .inner
            .get::<Vec<u8>>(&format!("reward-state-v2/reward-merkle-tree-v2/{height}"))
            .send()
            .await
            .map_err(anyhow::Error::from)
    };

    let fut2 = async {
        client
            .inner
            .get::<Vec<u8>>(&format!("catchup/reward-merkle-tree-v2/{height}/{view}"))
            .send()
            .await
            .map_err(anyhow::Error::from)
    };

    let (result, _) = select_ok(vec![fut1, fut2]).await?;
    Ok::<Vec<u8>, anyhow::Error>(result)
})
```

sequencer/api/catchup.toml
```diff
@@ -142,6 +142,15 @@ Return all RewardMerkleTreeV2 accounts stored for the requested height, paginate
 `:offset` (limit must be <= 10000). The response is a list of `(account, balance)`
 """
 
+[route.reward_merkle_tree_v2]
+PATH = ["/:height/:view/reward-merkle-tree-v2"]
```
ss-es: this should really be

`reward-merkle-tree-v2/:height/:view`

(I know we have other endpoints not like this but I think we shouldn't repeat that)

sequencer/src/catchup.rs
imabdulbasit: hmm I was thinking for catchup it would be better to try memory first because in most cases the data we need should be in memory. So that we can avoid extra db request

sequencer/api/catchup.toml
imabdulbasit: yeah this was to be consistent with other endpoints, but I can change it np

sequencer/src/catchup.rs
bfish713: I think I prefer only falling back in the error case.  If it's really hitting the DB we should already expect it to be a bit slow.  And if we do the decided version on each catchup requests we're triggering a bunch of DB hits for nothing most times

ss-es: this isn't just for catchup though -- this has to serve requests from the merklized state loop as far as I understand

ss-es: oh wait, that will always have accounts in the parent state

imabdulbasit: no merklized state loop doesn't use http catchup

bfish713: Merklized sate loop should not trigger catchup right? or am I mistaken.  I thought the whole idea was to go leaf by leaf and always have a full state.  Catchup should just come from our own storage only

imabdulbasit: only uses `SqlStateCatchup` which loads from the database here
```

 async fn try_fetch_reward_merkle_tree_v2(
        &self,
        _retry: usize,
        height: u64,
        _view: ViewNumber,
        reward_merkle_tree_root: RewardMerkleCommitmentV2,
        accounts: Arc<Vec<RewardAccountV2>>,
    ) -> anyhow::Result<PermittedRewardMerkleTreeV2> {
        let tree: PermittedRewardMerkleTreeV2 = self.db.load_reward_merkle_tree_v2(height).await?;

        ensure!(tree.tree.commitment() == reward_merkle_tree_root);
        ensure!(!forgotten_accounts_include(&tree, &accounts));

        Ok(tree)
    }

```

imabdulbasit: Its a bit hard to follow but if you look at this function, we create `SqlStateCatchup` only 
```


#[tracing::instrument(skip_all)]
pub(crate) async fn update_state_storage_loop<T>(
    storage: Arc<T>,
    instance: impl Future<Output = NodeState>,
) -> anyhow::Result<()>
where
    T: SequencerStateDataSource,
    for<'a> T::Transaction<'a>: SequencerStateUpdate,
{
    let instance = instance.await;
    let peers = SqlStateCatchup::new(storage.clone(), Default::default());
```

sequencer/api/catchup.toml
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4016/commits/f33dd9bd3c5a7ea0f4e6da75f40f1ab0db06ccbd

PR #4032, opened 2026-03-11
Add endpoint to return table sizes

ss-es:
small PR to try out claude


imabdulbasit (changes requested):
Please add some tests for this module to see if queries are working

PR #4110, opened 2026-03-31
Remove types migrations from espresso-node

imabdulbasit:
## What

Decaf and mainnet have been upgraded to proof of stake, so the V1 to V2 consensus types migration (`anchor_leaf`, `da_proposal`, `vid_share`, `quorum_proposals`, `quorum_certificate`) is no longer needed. The x25519 stake table migration has also run everywhere, so this removes that too.

Node startup no longer runs any data migration. Net 2041 lines deleted.

Rebased onto current main and reduced to a single commit; the original April commits are superseded by it.

## Removed

- `SequencerPersistence::migrate_storage` and the six `migrate_*` methods it called, plus their impls in `sql.rs`, `fs.rs` and `no_storage.rs`.
- The startup hook in `run.rs`.
- `sql.rs`: `DataMigration`, `is_migration_complete`, `mark_migration_complete`, and the `test_consensus_migration` / `test_x25519_keys_migration` tests.
- `fs.rs`: `Inner::migrated`, `update_migration`, the `migration` tracking file, and the now-unused `decided_leaf_path`, `vid_dir_path` and `quorum_proposals_dir_path` helpers.
- hotshot `Storage::migrate_storage`, which had no callers anywhere in the tree, and its `TestStorage` impl.
- Three stale doc references: a `# Panics` section on `SystemContext::new` (the function has no panics at all), a comment in `availability.rs` claiming leaves are migrated on startup, and guidance in `doc/agents/rust.md` pointing at the deleted `migrated` HashSet.

## Where to focus

**The SQL migration files are intentionally untouched.** Refinery runs with abort-on-missing and abort-on-divergent, so deleting them would break startup on every existing database. As a result these are now dead schema:

- Tracking tables: `epoch_migration`, `data_migrations`, and the query service's `types_migration`.
- Legacy data tables: `anchor_leaf`, `da_proposal`, `vid_share`, `quorum_proposals`, `quorum_certificate`.

Dropping them needs a new forward migration and is deliberately not part of this PR, since it permanently discards the legacy rows. Note `migrate_quorum_proposal_leaf_hashes` still reads `quorum_proposals`, so that table cannot be dropped without addressing it first.

## Deliberately kept

Each of these was checked and is load-bearing or unrelated to the types migration:

- The `DataBackfill` / `MigrationRegistry` background framework, the `deferred_migrations` table, and `/v1/database/migration-status`. Separate system, still wired via `hash_bigint_migrations()`.
- `migrate_quorum_proposal_leaf_hashes`, which backfills a column rather than converting types.
- `RegisteredValidatorNoX25519::migrate` (five callers) and `legacy_anchor_leaf_path` (used by `load_anchor_leaf` and pruning), both read-time fallbacks for older on-disk formats.
- V1 `append_da` / `append_vid` / `append_proposal`, called from `traits/storage.rs`, the new-protocol coordinator, and `quorum_vote/handlers.rs`.

## Unrelated issue spotted

`postgres/V1402__delete_reward_merkle_trees.sql` branches on `epoch_migration WHERE table_name = 'reward_merkle_tree_v2_data'`, but no Rust code ever writes that row. This predates this PR, so on Postgres the old `reward_merkle_tree` / `reward_merkle_tree_v2` tables are never truncated. Not addressed here.

## Testing

- `cargo check --workspace --all-targets`: clean, no warnings.
- `cargo clippy --workspace --all-targets`: clean, no warnings.
- `cargo fmt --all` applied; pre-commit hooks pass.

Since this is a pure deletion of already-completed migrations, the removed tests covered only the deleted code paths. No new tests added.


imabdulbasit:
@claude review again

crates/espresso/node/src/persistence/fs.rs
```diff
@@ -3064,239 +2420,6 @@ mod test {
...
-            commission: 0,
-            delegators: HashMap::new(),
-            authenticated: true,
-            x25519_key: None,
-            p2p_addr: None,
-        }
-    }
-
-    #[test_log::test(tokio::test(flavor = "multi_thread"))]
-    async fn test_load_stake_legacy_storage() {
```
sveitser, line 3195: @imabdulbasit I think these tests shouldn't be removed because we did not actually do a migration for the pre-option types.

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4110/changes/cf119cae47d897e8db3b892b70043cc1947b07f5

sveitser (approved):

PR #4149, opened 2026-04-08
Ensure old reward merkle tree tables get deleted

jbearer:
### This PR:

Deletes old reward Merkle tree tables. These tables are extremely large (currently over 1 TB on Decaf) and are no longer needed after the `reward_merkle_tree_v2_data` migration.

On Postgres, we ensure that the `reward_merkle_tree_v2_data` runs before the tables are deleted, whether or not the migration has already run on a particular node before this upgrade is deployed.

On Sqlite, this is difficult to enforce, and it is possible that the old tables will be deleted before the `reward_merkle_tree_v2_data` migration runs, resulting in empty `reward_merkle_tree_v2_data` and `reward_merkle_tree_v2_proofs` tables. This is acceptable since we shouldn't be relying on SQLite nodes to provide the full reward history in any case; we can leave this to the postgres nodes.

### DIFF
--- a/crates/espresso/node/src/persistence/sql.rs
+++ b/crates/espresso/node/src/persistence/sql.rs
@@ -1362,10 +1362,15 @@ impl SequencerPersistence for Persistence {
             [("reward_merkle_tree_v2_data".to_string(), true, offset)],
         )
         .await?;
-        query("TRUNCATE reward_merkle_tree_v2")
+        let truncate = if cfg!(feature = "embedded-db") {
+            "DELETE FROM"
+        } else {
+            "TRUNCATE"
+        };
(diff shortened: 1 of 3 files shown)


crates/espresso/node/src/persistence/sql.rs
```diff
@@ -1362,6 +1362,17 @@ impl SequencerPersistence for Persistence {
             [("reward_merkle_tree_v2_data".to_string(), true, offset)],
         )
         .await?;
+        let truncate = if cfg!(feature = "embedded-db") {
```
imabdulbasit, line 1365: I think everyone has migrated so it should be enough to just have the SQL migrations

imabdulbasit (approved):

PR #4158, opened 2026-04-09
fix: persist reward tree at every epoch boundary for epoch reward version

imabdulbasit:
During a V4→V5 upgrade, the first post upgrade epoch boundary skips reward distribution because the previous epoch's header is pre V5, leaving `rewards_delta` empty and causing `save_and_gc_reward_tree_v2` to never be called at that height. When the next epoch's reward calculation tries to `fetch_reward_merkle_tree_v2` at that height for catchup, it fails because no tree exists in storage. This PR fixes it by saving the tree at every boundary if version is >= v5

### DIFF
--- a/crates/espresso/node/src/state.rs
+++ b/crates/espresso/node/src/state.rs
@@ -19,6 +19,7 @@ use hotshot_query_service::{
     status::StatusDataSource,
     types::HeightIndexed,
 };
+use hotshot_types::utils::is_last_block;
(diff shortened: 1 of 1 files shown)


ss-es (approved):
looks fine to me

I'm not sure I fully understand the issue, but I'm quite confident this shouldn't cause any problems because at worst we are just persisting the tree one additional time per epoch

PR #4168, opened 2026-04-10
fix(test): reduce node count for epoch reward upgrade test

sveitser:
Epoch reward upgrade test starts in full epoch mode from view 0, making every view resource-heavy with stake table lookups. With 5 postgres Docker containers on CI this causes view timeouts and consensus stall. Reduce to 3 nodes.


sveitser:
@imabdulbasit got 1/1 pass so far, re-running now. I think it might just be a load issue and that's why it doesn't happen for sqlite.

crates/espresso/node/src/api.rs
```diff
@@ -3640,19 +3640,19 @@ mod test {
 
     #[test_log::test(tokio::test(flavor = "multi_thread"))]
     async fn test_epoch_reward_upgrade() {
-        test_upgrade_helper(Upgrade::new(
+        // Use fewer nodes: epoch mode from view 0 is resource-heavy on CI with
+        // postgres Docker containers, causing view timeouts and consensus stall.
```
imabdulbasit, line 3644: hmm I thought there would be only one postgres node and rest of nodes run file system

imabdulbasit: actually no we create SQL storage for all nodes, but have only one query node. I think we should also try using file system for other nodes persistence and see if that passes.

imabdulbasit:
```
   let storage = join_all((0..NUM_NODES).map(|_| SqlDataSource::create_storage())).await;
        let persistence: [_; NUM_NODES] = storage
            .iter()
            .map(<SqlDataSource as TestableSequencerDataSource>::persistence_options)
            .collect::<Vec<_>>()
            .try_into()
            .unwrap();
```

imabdulbasit (approved):

PR #4169, opened 2026-04-10
Refactor query service types into their own crate

jbearer:
### This PR:

Moves query service type definitions into their own crate, which becomes a lighter-weight dependency for downstream crates that want to use the query service but not import the full server backend.

In particular, this allows the `light-client` crate to not import `hotshot-query-service`, which in turn will allow (in a future PR) `hotshot-query-service` to import the light client, which it can use to implement a better fetching provider which retrieves and verifies data from a peer.


imabdulbasit (approved):
it is very time consuming to review everything in this PR because the diff is pretty big, but since this is just moving types to a crate, I am approving it

PR #4178, opened 2026-04-13
testnet deploy steps

alysiahuggins:
Closes #<ISSUE_NUMBER>
when setting up a drill i realized that i needed to allow for env vars even in localhost anvil runs

### Key places to review:

---
  - https://app.asana.com/0/0/1214112505149196


sveitser (approved):
LGTM but #4111 is going to change all env var names.

PR #4190, opened 2026-04-16
feat(stake-able): register fast finality network info

sveitser:
Adds x25519 key and p2p address registration to the stake table so validators can discover each other on-chain. This is
required for fast finality, where validators need direct peer communication. The stake table becomes the single source
of truth for peer discovery.

Design doc: [doc/stake-table-fast-finality.md](https://github.com/EspressoSystems/espresso-network/blob/ma/stake-table-fast-finality/doc/stake-table-fast-finality.md)

### This PR:

- Adds `StakeTableV3` contract with event storage for x25519 key + p2p address, plus `registerValidatorV3`,
  `updateNetworkConfig`, `updateX25519Key`, `updateP2pAddr`, and a `validateP2pAddr` helper.
- Wires V3 upgrade paths (multisig + timelock) into the deployer and `--upgrade-stake-table-v3` CLI flag.
- Adds Rust event processing for the new V3 events (including split `NetworkConfigUpdated` -> `X25519KeyUpdated` /
  `P2pAddrUpgradeded` events).
- Updates `staking-cli` to register/update with x25519 key and p2p address; uses native Rust types and tagged base64 for
  x25519 serialization.
- Updates the light client flow to trust the server for 0.3 stake tables on Decaf.
- Adds V3 invariant fuzz targets, storage-compat tests, proptest coverage for p2p address parsing, and CLI e2e tests.

### This PR does not
- Remove any V1 stake table code. We don't need all of that anymore but the PR is huge already.

### Key places to review:

**Contracts (reviewer: @alysiahuggins )** `contracts/`

- `contracts/src/StakeTableV3.sol` - new contract, storage layout, reinitializer, `validateP2pAddr`.
- `contracts/test/StakeTableV3.t.sol`, `contracts/test/StakeTableUpgradeToV3.t.sol`,
  `contracts/test/StorageUpgradeCompatibility.t.sol`, `contracts/test/StakeTableV2.invariant.t.sol`,
  `contracts/test/MockStakeTableV3.sol`.
- `contracts/rust/deployer/` - V3 upgrade proposal paths (multisig + timelock), `--upgrade-stake-table-v3` CLI.
- `contracts/rust/deployment-info/` - address registries and V3 version bump.
- `contracts/rust/adapter/` - V3 bindings, sol types, stake table copy/convert.

**Stake table event processing + node (@ss-es , @lukaszrzasik, @imabdulbasit )** `staking-cli/`, `crates/espresso`  

- Processing of the new V3 events: `RegisteredValidatorV3`, `X25519KeyUpdated`, `P2pAddrUpdated`. Verify event ->
  `ValidatorMap` transitions are correct.
- Commitments: without any new events no commitment changes, all new events part of commitments.
- No unexpected errors when handling events - an error here is a liveness failure. 
- `staking-cli/src/registration.rs`, `staking-cli/src/deploy.rs`, `staking-cli/src/lib.rs`: V2 vs V3 routing; V2 must
  not require x25519/p2p; V3 must surface a clear error when they are missing.
  
  
**Requirements**: @twittner @bfish713 please check if this matches what we need. @twittner there are some changes to address parsing code you may want to have a look at.

### How to test this PR:

- Everything runs on CI except some deployment tests:
- Mainnet-fork V3 timelock Safe TX [recipe](https://github.com/EspressoSystems/espresso-network/blob/ma/stake-table-fast-finality/doc/stake-table-fast-finality.md#testing-the-safe-tx-output-against-a-mainnet-fork) in `contracts/rust/deployer/scripts/README.md`.


contracts/rust/adapter/src/sol_types.rs
```diff
@@ -193,6 +194,66 @@ impl From<stake_table_v2::BN254::G1Point> for G1PointSol {
     }
 }
 
+// Transmute conversions for StakeTableV3 types
+use crate::bindings::stake_table_v3;
+
+impl From<stake_table_v3::BN254::G2Point> for G2PointSol {
+    fn from(v: stake_table_v3::BN254::G2Point) -> Self {
+        unsafe { std::mem::transmute(v) }
+    }
```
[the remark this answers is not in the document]
sveitser: /gemini check

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -118,75 +123,111 @@ impl DisplayLog for Log {
...
+/// The V3 ABI is a superset of V2. Shared events have structurally identical generated types
+/// in both binding modules. This transmute bridges the two for events that predate V3.
+fn transmute_v3_to_v2<V3, V2>(v: V3) -> V2 {
+    assert_eq!(
+        std::mem::size_of::<V3>(),
+        std::mem::size_of::<V2>(),
+        "V3 and V2 event types must have the same size"
+    );
+    unsafe { std::mem::transmute_copy(&std::mem::ManuallyDrop::new(v)) }
+}
```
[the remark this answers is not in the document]
sveitser: /gemini check

crates/hotshot/types/src/addr.rs
```diff
@@ -116,10 +116,23 @@ impl std::str::FromStr for NetAddr {
...
+            let a = if a.starts_with('[') && a.ends_with(']') {
+                &a[1..a.len() - 1]
+            } else {
+                a
+            };
             IpAddr::from_str(a)
                 .map(|a| Self::Inet(a, p))
                 .or_else(|_| Ok(Self::Name(a.to_string().into(), p)))
         };
+        // Handle bracketed IPv6 like `[::1]:8080` or `[::1]` (no port).
```
sveitser, line 129: cc @twittner

crates/hotshot/types/src/x25519.rs
```diff
@@ -208,6 +210,27 @@ impl TryFrom<&str> for SecretKey {
     }
 }
 
+const X25519_PUBLIC_KEY: &str = "X25519_PK";
+
+impl TryFrom<TaggedBase64> for PublicKey {
```
sveitser, line 215: @twittner regarding the change from base58 tagged base64 i pinged you about.

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -612,6 +635,126 @@ impl StakeTableState {
...
+                    stake_table_key,
+                    state_ver_key,
+                    stake: U256::ZERO,
+                    commission: *commission,
+                    delegators: HashMap::new(),
+                    authenticated,
+                    x25519_key: Some(x25519_key),
+                    p2p_addr,
+                });
+            },
```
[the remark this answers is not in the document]
sveitser: What should we do if in rust we get an address that is invalid? @twittner @bfish713 @akonring 

I think it shouldn't really happen but better make it a conscious decision.

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -612,6 +635,126 @@ impl StakeTableState {
...
+                    .ok_or(StakeTableError::ValidatorNotFound(validator))?;
+
+                let key = x25519::PublicKey::try_from(x25519Key.0.as_slice())
+                    .map_err(|e| StakeTableError::InvalidX25519Key(format!("{e}")))?;
+                if self.used_x25519_keys.contains(&key) {
+                    return Err(StakeTableError::X25519KeyAlreadyUsed(key.to_string()));
+                }
+                self.used_x25519_keys.insert(key);
+                val.x25519_key = Some(key);
+            },
```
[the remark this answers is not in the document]
sveitser: Yes it's on purpose and documented.

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -612,6 +635,126 @@ impl StakeTableState {
...
+                    .ok_or(StakeTableError::ValidatorNotFound(validator))?;
+
+                match p2pAddr.parse::<NetAddr>() {
+                    Ok(addr) => val.p2p_addr = Some(addr),
+                    Err(e) => {
+                        tracing::warn!(%e, validator = ?validator, "Failed to parse p2p addr");
+                        val.p2p_addr = None;
+                    },
+                }
+            },
```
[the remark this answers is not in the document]
sveitser: @bfish713 @twittner @akonring similar to the other question about bad p2p addresses.

crates/hotshot/types/src/addr.rs
```diff
@@ -189,4 +202,28 @@ mod tests {
         assert_eq!("sub.domain.com", &h);
         assert_eq!(1234, p);
     }
+
+    #[test]
+    fn test_parse_bracketed_ipv6() {
```
twittner, line 207: Maybe we can generalise those tests into a single property test:

```rust
#[cfg(test)]
mod tests {
    use std::{iter::repeat_with, net::IpAddr};

    use quickcheck::{Arbitrary, Gen, quickcheck};

    use super::NetAddr;

    impl Arbitrary for NetAddr {
        fn arbitrary(g: &mut Gen) -> Self {
            let port = u16::arbitrary(g);
            if bool::arbitrary(g) {
                let len = u8::arbitrary(g);
                let host: String = repeat_with(|| char::arbitrary(g))
                    .filter(|c| !"[]".contains(*c))
                    .take(len.into())
                    .collect();
                NetAddr::Name(host.into(), port)
            } else {
                let ip = IpAddr::arbitrary(g);
                NetAddr::Inet(ip, port)
            }
        }
    }

    quickcheck! {
        fn prop_to_string_parse_identity(a: NetAddr) -> bool {
            a.to_string().parse().ok() == Some(a)
        }
    }

    #[test]
    fn empty_is_invalid() {
        assert!("".parse::<NetAddr>().is_err())
    }
}
```

crates/hotshot/types/src/addr.rs
```diff
@@ -116,10 +116,23 @@ impl std::str::FromStr for NetAddr {
...
+            } else {
+                a
+            };
             IpAddr::from_str(a)
                 .map(|a| Self::Inet(a, p))
                 .or_else(|_| Ok(Self::Name(a.to_string().into(), p)))
         };
+        // Handle bracketed IPv6 like `[::1]:8080` or `[::1]` (no port).
+        if s.starts_with('[') {
+            return match s.find("]:") {
```
twittner, line 131:
```suggestion
            return match s.rfind("]:") {
```

crates/hotshot/types/src/addr.rs
```diff
@@ -116,10 +116,23 @@ impl std::str::FromStr for NetAddr {
...
+                &a[1..a.len() - 1]
+            } else {
+                a
+            };
             IpAddr::from_str(a)
                 .map(|a| Self::Inet(a, p))
                 .or_else(|_| Ok(Self::Name(a.to_string().into(), p)))
         };
+        // Handle bracketed IPv6 like `[::1]:8080` or `[::1]` (no port).
+        if s.starts_with('[') {
```
twittner, line 130:
```suggestion
        if s.is_empty() {
            return Err(InvalidNetAddr(()))
        }
        if s.starts_with('[') {
```

crates/espresso/types/src/v0/impls/stake_table.rs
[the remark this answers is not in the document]
twittner: An invalid address should be handled like a rejected X25519 key I think. The validator is not usable and should not be registered.

contracts/rust/deployer/src/proposals/timelock.rs
```diff
@@ -505,3 +510,266 @@ async fn perform_timelock_operation_via_multisig(
...
+            U256::ZERO.to_string(),
+            upgrade_calldata.data.to_string(),
+            B256::ZERO.to_string(),
+            params.salt.to_string(),
+            params.delay.to_string(),
+        ],
+        U256::ZERO,
+    )?
+    .with_description(format!(
+        "Schedule StakeTable -> V3 upgrade via timelock {timelock_addr:#x} (proxy \
```
alysiahuggins: nit: could be good to say what the version we're upgrading from is, just incase to catch any mistakes

contracts/rust/deployer/src/lib.rs
```diff
@@ -233,6 +237,8 @@ pub enum Contract {
     StakeTable,
     #[display("ESPRESSO_STAKE_TABLE_V2_ADDRESS")]
     StakeTableV2,
+    #[display("ESPRESSO_SEQUENCER_STAKE_TABLE_V3_ADDRESS")]
```
alysiahuggins: since we're no longer using the term 'sequencer', perhaps `ESPRESSO_STAKE_TABLE_V3_ADDRESS`

contracts/src/StakeTableV3.sol
```diff
@@ -0,0 +1,249 @@
...
+        emit ValidatorRegisteredV3(
+            validator,
+            blsVK,
+            schnorrVK,
+            commission,
+            blsSig,
+            schnorrSig,
+            metadataUri,
+            x25519Key,
+            p2pAddr
```
alysiahuggins, line 184: would there be onchain callers who want to know the p2pAddr without reading the events offchain?

alysiahuggins (approved):
approved with small nits, thanks @sveitser !

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -243,6 +250,17 @@ impl Committable for StakeTableState {
             builder = builder.var_size_bytes(&schnorr_key_bytes);
         }
 
+        // Only include used_x25519_keys when non-empty to preserve the pre-fast-finality commitment
+        // value for stake tables that have no x25519 keys yet. Same pattern as x25519_key /
+        // p2p_addr on RegisteredValidator. Backward compatibility is cross-checked by
+        // REFERENCE_V4_HEADER_COMMITMENT in reference_tests.rs.
+        if !self.used_x25519_keys.is_empty() {
```
imabdulbasit, line 256: why we don't track p2p address, and only x25519?

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -612,6 +635,126 @@ impl StakeTableState {
                     .ok_or(StakeTableError::ValidatorNotFound(validator))?;
                 val.commission = newCommission;
             },
+
+            StakeTableEvent::RegisterV3(ref reg) => {
+                let authenticated = reg.authenticate().is_ok();
+                if !authenticated {
```
imabdulbasit, line 640: why we are not skipping the validator if authentication fails?

crates/espresso/types/src/v0/impls/stake_table.rs
sveitser: For keys the space is huge and if you ever got a used one you can just re-generate. Therefore I think using a key that is used before is almost certainly a mistake. For addresses you can't choose or generate them freely. Especially for ipv4 addresses there really aren't that many available anymore. We can also disallow re-use, I don't have a strong opinion.

cc @bfish713 @twittner

crates/espresso/types/src/v0/impls/stake_table.rs
sveitser: So that we can still track delegations etc. towards the unauthenticated validator which the stake table cannot prevent. This should be the same for V2.

contracts/src/StakeTableV3.sol
sveitser: Not as far as I know.

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -668,6 +813,15 @@ pub fn select_active_validator_set(
                         tracing::info!("Validator {address:?} does not have any stake");
                         return None;
                     }
+                    if !cv.is_eligible(protocol_version) {
```
imabdulbasit, line 814: so one of the issue with using the protocol version from the header is that header version can be < CliqueNet version but the protocol can be in cliquenet version because root of the epoch is always E - 2.  So I think the first two epochs for new version will have no keys.

imabdulbasit: or actually we will just skip the check, but validators can have the keys if they have registered the keys

light-client/src/consensus/quorum.rs
```diff
@@ -223,6 +223,29 @@ impl StakeTablePair for (Arc<StakeTable>, Arc<StakeTable>) {
     }
 }
 
+/// A `StakeTablePair` that promises only the previous-epoch stake table will be consulted.
```
sveitser, line 259: link

imabdulbasit (approved):
I reviewed espresso types changes. I also looked at light client changes, and they look fine but I might not be the best person to review that

sveitser:
Thanks @imabdulbasit, I asked @jbearer to take a look yesterday and he said it seems fine.

So basically the staking-cli changes are still not reviewed. @lukaszrzasik would you have time for that? I don't think it needs super detailed review since it's mainly boilerplate similar to registration / updating consensus keys.

staking-cli/src/claim.rs
sveitser, line 1: @lukaszrzasik changes in this crate.

staking-cli/src/demo.rs
```diff
@@ -940,10 +967,15 @@ impl StakingTransactions<HttpProviderWithWallet> {
 
             let payload =
                 NodeSignatures::create(address, &validator.bls_key_pair, &validator.state_key_pair);
+            let p2p_addr = format!("127.0.0.1:{}", 8080 + address.0[19] as u16)
+                .parse()
+                .expect("valid net addr");
```
lukaszrzasik: Does that mean we always use `127.0.0.1` address?

staking-cli/src/demo.rs
```diff
@@ -1082,11 +1114,12 @@ pub(crate) async fn stake_for_demo(
...
         ))?)?;
-        validator_keys.push((
+        validator_keys.push(StakingKeySet {
             signer,
-            consensus_private_key,
-            StateKeyPair::from_sign_key(state_private_key),
-        ));
+            bls: consensus_private_key,
+            state: StateKeyPair::from_sign_key(state_private_key),
+            x25519: x25519::Keypair::generate().expect("failed to generate x25519 keypair"),
```
lukaszrzasik: Is this always a newly generated key? Can't we use an existing one?

lukaszrzasik (commented):
Just two questions.

staking-cli/src/demo.rs
sveitser: yes, this wouldn't always work, will fix

staking-cli/src/demo.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/4190/commits/e4795c781ef237d1b28bca496a9705f7e8919904

staking-cli/src/demo.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/4190/commits/e4795c781ef237d1b28bca496a9705f7e8919904

lukaszrzasik (approved):

PR #4193, opened 2026-04-16
Restart Tests and Fixes

bfish713:
Closes #<ISSUE_NUMBER>


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -603,6 +667,16 @@ impl<T: NodeType, N: ConnectedNetwork<T::SignatureKey>> Coordinator<T, N> {
...
+
+                // Proactively fetch DRBs for the next epoch so
+                // late-starting nodes have them before they need to
+                // propose or verify certs in a new epoch. The dedup
+                // in request_drb_result makes repeated calls free.
+                let next_epoch = epoch + 1;
+                if next_epoch > EpochNumber::genesis() + 1 {
+                    self.epoch_manager.request_drb_result(next_epoch);
+                    self.epoch_manager.request_drb_result(next_epoch + 1);
+                }
```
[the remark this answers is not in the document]
bfish713: The point of this on every view is to make sure it gets triggered for a node that restarts, the check in the map should be fast so I don't think it's an issue

crates/hotshot/new-protocol/src/storage.rs
```diff
@@ -1 +1,147 @@
...
+            };
+            loop {
+                match storage.append_vid(&proposal).await {
+                    Ok(()) => return,
+                    Err(err) => {
+                        warn!(%err, "failed to append VID share, retrying");
+                        sleep(RETRY_DELAY).await;
+                    },
+                }
+            }
```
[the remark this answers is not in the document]
bfish713: It will be obvious in the logs that storage is broken.  Also we cancel these tasks every time we GC, this will prevent infinite retries

crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -1087,6 +1206,52 @@ impl<T: NodeType> Consensus<T> {
...
+        }
+    }
+
+    /// Retry verification of pending certificates whose epoch may now be available.
+    async fn retry_pending_certs(&mut self, outbox: &mut Outbox<ConsensusOutput<T>>) {
+        // Retry pending cert1s.
+        let pending = std::mem::take(&mut self.pending_certs1);
+        for (view, cert) in pending {
+            self.handle_certificate1(cert.clone(), outbox).await;
+            self.maybe_vote_2_and_update_lock(view, outbox).await;
```
imabdulbasit, line 1245: so I am trying to understand the flow here. How can we vote if we received the certificate some time ago and consensus has progressed?

crates/hotshot/new-protocol/bench/src/node.rs
```diff
@@ -169,6 +170,14 @@ async fn build_coordinator(
         .epoch_manager(epoch_manager)
         .block_builder(block_builder)
         .proposal_validator(proposal_validator)
+        .storage(hotshot_new_protocol::storage::Storage::new(
+            TestStorage::default(),
+            private_key,
+        ))
+        .external_events({
```
imabdulbasit: I don't fully understand why we need this. The Leaf external message is already handled in application layer?

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -69,6 +72,15 @@ pub struct Coordinator<T: NodeType, N> {
...
     proposal_validator: ProposalValidator<T>,
+    storage: Storage<T, S>,
+    /// Broadcast sink for external messages forwarded to the membership.
+    ///
+    /// Every incoming `MessageType::External(data)` is re-published as a
+    /// `hotshot_types::event::Event` with `ExternalMessageReceived` so the
+    /// membership's `Leaf2Fetcher` (fed via
+    /// `EpochMembershipCoordinator::set_external_channel`) can answer leaf
+    /// requests during catchup exactly as it does in the old protocol.
+    external_events: BroadcastSender<Event<T>>,
```
imabdulbasit: I only see it being used in tests. If that's the case, can we just gate this behind test cfg or testing feature?

crates/hotshot/new-protocol/src/tests/common/runner.rs
```diff
@@ -163,92 +211,173 @@ impl TestRunner {
...
+                                // Kill existing task if any.
+                                if let Some(handle) = node_handles[change.idx].take() {
+                                    handle.abort();
+                                }
+                                // Shut down the old network so its internal
+                                // buffers don't block broadcasts.
+                                network_state.shutdown_node(change.idx).await;
+                                // Create a fresh coordinator from genesis.
+                                let net = network_state.create_node(change.idx).await;
+                                let (membership, storage) = network_state
```
imabdulbasit, line 315: can we reuse the same membership and storage?

crates/hotshot/new-protocol/bench/src/node.rs
bfish713: This is what sends the event to the application layer when we get the data.  For example if we got a request response message, this would send it to the application.

bfish713: Ok we just direclty use the network in those places, only the strict mamebership uses this in test, and I don't even think we need it at all then.

crates/hotshot/new-protocol/src/tests/common/runner.rs
bfish713: We could reuse storage, but the membership should be recreated to make sure catchup works (this is how restarting in prod would work.  A new membership would be created.

crates/hotshot/new-protocol/src/coordinator.rs
bfish713: Moved it to the test code completely

crates/hotshot/new-protocol/bench/src/node.rs
bfish713: I moved it to the test code only now

crates/hotshot/new-protocol/src/consensus.rs
bfish713: It's always safe to vote on stuff that is old, it would get ignored by others if they've moved on, otherwise it'll contribute to the certificate.

imabdulbasit (approved):

PR #4201, opened 2026-04-20
Lr/serialization errors handling

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

The crux of the changes:
1. `DEFERRABLE` removed from the read transactions. This makes them succeed or fail immediately instead of waiting. If it fails due to serializable error it is retried. This approach should be more efficient that using deferrable which can wait for long time as it's more conservative: one it starts, it has to finish successfully.
2. Add retries for serializable errors for read and write transactions.
3. Adds a filter where clause to many queries: height > pruned_height. This is supposed to minimize serializable errors. It's not perfect due to predicate lock escalation: row -> page -> relation (table).
4. Adds diagnostic logs fetched from postgres when serialization error is encountered. This might be very often so they are debug and put into a separate internal module so they can be specifically enabled with RUST_LOG.
5. There are some formatting and clippy changes forced by rust tool chain update.
6. `delete_batch` uses the `20260407` approach. I don't know if that's correct.
7. There delays between pruning queries. These were used to simulate long pruning in the local tests and should be removed from the final version.

### DIFF
--- a/crates/espresso/node/src/persistence/sql.rs
+++ b/crates/espresso/node/src/persistence/sql.rs
@@ -1302,54 +1302,65 @@ fn is_serialization_error(err: &anyhow::Error) -> bool {
         .any(|e| e.code().as_deref() == Some(PG_SERIALIZATION_FAILURE_CODE))
 }
 
-/// Spawn a background task that queries `pg_stat_activity` to identify connections involved
-/// in a serialization conflict. Best-effort; failures are silently ignored.
-#[cfg(not(feature = "embedded-db"))]
-fn spawn_pg_stat_activity_log(pool: sqlx::Pool<Db>, op: &'static str) {
-    tokio::spawn(async move {
-        match sqlx::query(
-            "SELECT pid, COALESCE(state, 'unknown') AS state, left(COALESCE(query, ''), 200) AS \
-             query FROM pg_stat_activity WHERE pid != pg_backend_pid() AND state IS DISTINCT FROM \
-             'idle'",
-        )
-        .fetch_all(&pool)
-        .await
-        {
-            Ok(rows) if rows.is_empty() => {
-                tracing::debug!(op, "serialization conflict: no other non-idle DB sessions");
-            },
-            Ok(rows) => {
-                for row in &rows {
-                    let pid: i32 = row.try_get("pid").unwrap_or(-1);
-                    let state: String = row.try_get("state").unwrap_or_default();
-                    let query_text: String = row.try_get("query").unwrap_or_default();
-                    tracing::warn!(
-                        op,
-                        pid,
-                        state,
-                        "serialization conflict: concurrent session: {query_text}",
-                    );
-                }
-            },
-            Err(e) => {
-                tracing::debug!(op, "failed to query pg_stat_activity: {e:#}");
-            },
-        }
-    });
+/// Diagnostics for PostgreSQL serialization conflicts.
+///
+/// Enable targeted logging with `RUST_LOG=sequencer::persistence::sql::pg_stat_diag=debug`.
+mod pg_stat_diag {
+    /// Spawn a background task that queries `pg_stat_activity` to identify connections involved
+    /// in a serialization conflict. Best-effort; failures are silently ignored.
+    #[cfg(not(feature = "embedded-db"))]
+    pub(super) fn spawn_pg_stat_activity_log(pool: sqlx::Pool<super::Db>, op: &'static str) {
+        use sqlx::Row as _;
+        tokio::spawn(async move {
+            match sqlx::query(
+                "SELECT pid, COALESCE(state, 'unknown') AS state, left(COALESCE(query, ''), 200) \
+                 AS query FROM pg_stat_activity WHERE pid != pg_backend_pid() AND state IS \
+                 DISTINCT FROM 'idle'",
+            )
+            .fetch_all(&pool)
+            .await
+            {
+                Ok(rows) if rows.is_empty() => {
+                    tracing::debug!(op, "serialization conflict: no other non-idle DB sessions");
+                },
+                Ok(rows) => {
+                    for row in &rows {
+                        let pid: i32 = row.try_get("pid").unwrap_or(-1);
+                        let state: String = row.try_get("state").unwrap_or_default();
+                        let query_text: String = row.try_get("query").unwrap_or_default();
+                        tracing::debug!(
+                            op,
+                            pid,
+                            state,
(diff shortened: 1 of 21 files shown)


crates/espresso/node/src/persistence/sql.rs
```diff
@@ -782,141 +799,233 @@ impl Persistence {
...
-                    Ok(row) => row,
-                    Err(err) => {
-                        // If there's an error getting a row, try generating an event with the rows
-                        // we do have.
-                        tracing::warn!("error loading row: {err:#}");
-                        break;
-                    },
-                };
+            // Retry the entire read section on serialization failures (40001). Each `continue
+            // 'read` abandons the current transaction and starts fresh.
```
jbearer, line 828: Why not use the `retry_if` helper here, like in the other reads in this file? In addition to adding a lot of noise to the code of the loop, it seems like semantically we have also lost the maximum retries, so that we will retry this indefinitely on serialization errors. Was that intentional?

jbearer: If nothing else, I think it would really help the readability of this section if we pull the body of the `'read` loop into its own function that just bubbles errors. Then we can have the logic for handling serialization errors with `continue 'read` in one place, at the call site.

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -710,6 +710,10 @@ pub struct Persistence {
     internal_metrics: PersistenceMetricsValue,
 }
 
+/// PostgreSQL error code for serialization failures under SERIALIZABLE isolation.
+/// Transactions that fail with this code are safe to retry from scratch.
+const PG_SERIALIZATION_FAILURE_CODE: &str = "40001";
```
jbearer, line 715: What happens on SQLite? Does it just never return this error code, so `should_retry` is always `false`?

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -1101,52 +1226,185 @@ impl Persistence {
...
+            })
+            .await
+    }
+}
+
+/// Maximum number of retries on PostgreSQL serialization conflicts (error 40001).
+const SERIALIZABLE_RETRY_MAX: u32 = 100;
+
+/// Backoff parameters for retrying transactions aborted by PostgreSQL serialization conflicts.
+const SERIALIZABLE_BACKOFF: BackoffParams = BackoffParams::new(
```
jbearer, line 1286: It would be really nice if we could make these parameters somehow configurable. This seems like something we will pretty likely end up playing around with in a deployment

hotshot-query-service/src/data_source/storage/sql/transaction.rs
```diff
@@ -416,58 +421,43 @@ impl Transaction<Write> {
...
+
+        let res = query("DELETE FROM vid2 WHERE height <= $1")
+            .bind(height as i64)
+            .execute(self.as_mut())
+            .await
+            .context("deleting vid")?;
+        tracing::debug!(rows_affected = res.rows_affected(), "pruned vid");
+        if let Some(d) = delay {
+            sleep(d).await;
+        }
```
jbearer, line 460: This diff seems problematic. I'm a little confused with all the branches. But the before and after here is semantically very different. The deleted code assumes we have already run migration V1400__dedup_payloads, and are dealing with the new payload-hash-indexed `payload` and `vid_common` tables. The added code assumes we have not yet run this migration, and are dealing with the old height-indexed `payload` and `vid2` tables

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -132,6 +132,37 @@ use self::{
...
+                    "serialization conflict on read transaction, retrying after {delay:?}"
+                );
+                sleep(delay).await;
+                delay = min(delay * 2, max_delay);
+            },
+            Err(err) => return Err(err),
+        }
+    }
+    f().await
+}
```
jbearer, line 164: It's a bit unfortunate that we have a whole different retry implementation here vs in the node crate, would be nice if we could merge them. In particular, this implementation lacks the randomized jitter, which might actually be kind of important in this case. We also have a different way of detecting errors here, using a string comparison instead of the error code, which might be more liable to break when postgres upgrades

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -1914,42 +1954,76 @@ where
...
+        let start = match range.start_bound() {
+            Bound::Included(&n) => Bound::Included(n),
+            Bound::Excluded(&n) => Bound::Excluded(n),
+            Bound::Unbounded => Bound::Unbounded,
+        };
+        let end = match range.end_bound() {
+            Bound::Included(&n) => Bound::Included(n),
+            Bound::Excluded(&n) => Bound::Excluded(n),
+            Bound::Unbounded => Bound::Unbounded,
+        };
```
jbearer, line 1980: Why do we have this? This doesn't seem directly related to adding the retries

hotshot-query-service/src/data_source/storage/sql/queries/availability.rs
```diff
@@ -178,9 +233,25 @@ where
...
-        let sql = format!("SELECT {LEAF_COLUMNS} FROM leaf2 {where_clause} ORDER BY height ASC");
+        let ph = query.bind(pruned_height)?;
+        let sql = if where_clause.is_empty() {
+            format!("SELECT {LEAF_COLUMNS} FROM leaf2 WHERE height > {ph} ORDER BY height ASC")
+        } else {
+            format!(
+                "SELECT {LEAF_COLUMNS} FROM leaf2 {where_clause} AND height > {ph} ORDER BY \
+                 height ASC"
+            )
+        };
```
jbearer, line 254: This pattern is repeated a lot and is mostly just noise. Perhaps we factor this out into a helper function or fold it into `bounds_to_where_clause`

jbearer (commented):
The retries add a lot of noise to each call site. It also seems like an easy thing to forget when adding new queries in the future. I'm trying to think of ways to make this retry logic more transparent/automatic. I don't have any great ideas yet but at least an improvement could be to move some of the parameters (e.g. `SERIALIZABLE_BACKOFF`, `SERIALIZABLE_RETRY_MAX`, `is_serialization_error`) plus the transaction handling, into our retry function, so that the call sites look more like this:

```rust
retryable_read(&self.db, |tx: &mut Transaction<Read>| async move {
  do_stuff(tx).await?;
  Ok(result)
}).await?;
```

PR #4206, opened 2026-04-21
Enable query service backfilling in chronological order.

jbearer:
Previously, we needed the hash of a leaf before we could fetch that leaf, in order to verify that the leaf we fetched was correct. This lead to backfilling proceeding in reverse order, where we would fetch the most recent leave, then use it's `parent_hash` to fetch and verify the preceding leaf, and so on back towards genesis. This behavior is somewhat unintuitive, and it works very badly with the merklized state update loop and the aggregator task, which have to proceed in chronological order. When syncing a node from scratch, this means we fetch every single leaf in reverse order, and only when we reach genesis can the state update loop and the aggregator even begin to start syncing.

With the light client, this reverse fetching is no longer necessary. We can fetch and verify an arbitrary leaf given only its height. Thus, we now proceed to fetch things in chronological order.

### DIFF
--- a/crates/espresso/node/src/persistence/sql.rs
+++ b/crates/espresso/node/src/persistence/sql.rs
@@ -33,21 +33,21 @@ use hotshot_libp2p_networking::network::behaviours::dht::store::persistent::{
 };
 use hotshot_new_protocol::message::Certificate2;
 use hotshot_query_service::{
-    availability::{BlockId, LeafQueryData},
+    availability::BlockId,
     data_source::{
         Transaction as _, VersionedDataSource,
         storage::{
             AvailabilityStorage,
             pruning::PrunerCfg,
             sql::{
-                Config, Db, Read, SqlStorage, StorageConnectionType, Transaction, TransactionMode,
-                Write, include_migrations, query_as, syntax_helpers::MAX_FN,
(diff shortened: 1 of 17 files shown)


crates/espresso/node/src/persistence/sql.rs
```diff
@@ -3457,71 +3457,6 @@ impl Provider<SeqTypes, PayloadRequest> for Persistence {
     }
 }
 
-#[async_trait]
```
jbearer, line 3549: We can no longer fetch leaves from consensus storage, because we no longer have the hash of the requested leaf ahead of time, and so we have no way of verifying that the fetched leaf is finalized (note that consensus storage in general contains data that is not yet finalized and may never become finalized).

This is alright: the importance of these `Provider` impls is more about recovering payload data from consensus storage, since DA nodes are obligated to be able to retrieve and serve all data that they have signed for (and a signature only guarantees that the object has made it to consensus storage). Leaves, though, are sent to every node, and so losing a leaf is less of a concern. Thus, a DA node which has crashed can always fetch a missing leaf from a non-DA node, then use that leaf to recover the full payload from its consensus storage.

imabdulbasit (approved):

PR #4216, opened 2026-04-23
Add availability endpoint to axum API

ss-es:
This PR migrates the `availability` endpoint to Axum, with a copy of the old endpoint served at `v1/availability` and newer protobuf-specified endpoints served at `v2/data` and `v2/consensus`. 

You can skip `test_api.rs` (this is just to render an example UI without having to compile `espresso-node`) and `espresso.api.v2.rs` (this is automatically generated).


crates/espresso/api/src/axum.rs
```diff
@@ -88,10 +90,124 @@ async fn redirect_to_docs() -> axum::response::Redirect {
...
+struct NamespaceIdPath {
+    namespace_id: u32,
+}
+
+#[derive(serde::Serialize, schemars::JsonSchema)]
+#[serde(untagged)]
+enum NamespaceProofResponseUnion {
+    Single(serialization_api::v2::NamespaceProofResponse),
+    Range(serialization_api::v2::NamespaceProofRangeResponse),
+}
```
ss-es: this is not good -- we should use a type defined in a .proto file

crates/espresso/api/src/axum.rs
```diff
@@ -88,10 +90,124 @@ async fn redirect_to_docs() -> axum::response::Redirect {
...
+        _ => Err(ApiError::BadRequest(anyhow::anyhow!(
+            "Must specify either 'block' or both 'from' and 'to' query parameters"
+        ))),
+    }
+}
+
+#[derive(serde::Deserialize, schemars::JsonSchema)]
+struct IncorrectEncodingProofQuery {
+    block: u64,
+}
```
ss-es: this is also not good

crates/serialization/api/proto/v2/consensus.proto
```diff
@@ -0,0 +1,48 @@
...
+// Request to get stake table for an epoch
+message GetStakeTableRequest {
+  // Epoch number
+  uint64 epoch = 1;
+}
+
+// Stake table response
+message StakeTableResponse {
+  // Serialized stake table data
+  bytes stake_table = 1;
```
ss-es: not good

crates/serialization/api/proto/v2/consensus.proto
```diff
@@ -0,0 +1,48 @@
...
+// Request to get state certificate for an epoch
+message GetStateCertificateRequest {
+  // Epoch number
+  uint64 epoch = 1;
+}
+
+// State certificate response
+message StateCertificateResponse {
+  // Serialized light client state update certificate
+  bytes certificate = 1;
```
ss-es: not good

crates/espresso/api/src/axum.rs
```diff
@@ -165,6 +214,74 @@ where
...
 
+    // Availability API handlers
+    let get_namespace_proof_by_height =
+        |State(state): State<S>, Path((namespace, height)): Path<(u32, u64)>| async move {
+            state
+                .get_namespace_proof(v1::availability::BlockId::Height(height), namespace)
+                .await
+                .map(Json)
+                .map_err(ApiError::Internal)
+        };
```
ss-es, line 224: this part here feels a bit verbose and could probably be condensed. exploring options in the next PR.

crates/espresso/node/src/api/state.rs
```diff
@@ -612,3 +838,576 @@ where
...
+        use espresso_types::{NamespaceId, NsProof};
+        use futures::join;
+        use hotshot_query_service::availability::BlockId;
+
+        let ns_id = NamespaceId::from(namespace_id);
+        let block_id = BlockId::Number(block_height as usize);
+
+        // Fetch block and VID common data
+        let ds = &*self.data_source;
+        let timeout = std::time::Duration::from_millis(500);
```
ss-es, line 1010: this seems a bit weird, but the existing incorrect_encoding_proof path uses get_block_for_ns_proof which (I believe) does essentially the same thing

crates/espresso/node/src/api/state.rs
```diff
@@ -612,3 +838,576 @@ where
...
+            Some(cert) => cert,
+            None => {
+                // Not found locally, try to fetch from peers
+                const TIMEOUT: Duration = Duration::from_secs(40);
+                let cert = ds.request_state_cert(epoch, TIMEOUT).await.map_err(|e| {
+                    anyhow::anyhow!("failed to fetch state cert for epoch {}: {}", epoch, e)
+                })?;
+
+                // Store the fetched certificate
+                ds.insert_state_cert(epoch, cert.clone()).await?;
```
ss-es, line 1109: this mirrors the existing `get_state_cert` in `espresso/node/src/api/endpoints/availability.rs`, though maybe we should revisit the behaviour

crates/espresso/api/src/axum/routes.rs
```diff
@@ -61,6 +65,43 @@ pub mod v1 {
...
         "/v1/reward-state-v2/reward-merkle-tree-v2/{height}";
+
+    //=========================================================================
+    // Availability API Routes
+    //=========================================================================
+
+    /// Get namespace proof by block height
+    /// Path: GET /v1/availability/namespace/{namespace}/block/{height}
+    pub const NAMESPACE_PROOF_BY_HEIGHT_ROUTE: &str =
+        "/v1/availability/namespace/{namespace}/block/{height}";
```
sveitser: Is the change of parameter order intentional?

https://github.com/EspressoSystems/espresso-network/blob/178357d7012e2527079178e71b703515ead0c6b8/crates/espresso/node/api/availability.toml#L1-L10

crates/espresso/node/src/api/state.rs
```diff
@@ -238,6 +258,221 @@ where
...
+                    &mut serde_json::Serializer::new(&mut payload_bytes),
+                )
+                .map_err(|e| anyhow::anyhow!("failed to serialize payload: {}", e))?;
+                // Convert to string and remove quotes added by JSON serializer
+                let payload_str = String::from_utf8(payload_bytes)?
+                    .trim_matches('"')
+                    .to_string();
+
+                Ok(v2::Transaction {
+                    namespace: tx.namespace.0 as u32,
```
sveitser: It seems we have namespace ID as u64 and u32 in different places. Maybe a comment why this is fine here?

crates/espresso/node/src/api/state.rs
```diff
@@ -612,3 +848,576 @@ where
...
+                    transactions: vec![],
+                    proof: None,
+                });
+            }
+        }
+
+        Ok(proofs)
+    }
+
+    async fn get_incorrect_encoding_proof(
```
sveitser, line 992: I'm confused. This function always returns an error.

sveitser:
This PR adds the axum migration for the espresso node availability API overlay but most of the availability API is from the hotshot query service. I think we can merge this first but just wanted to mention it, it kind of confused me at first.

crates/espresso/api/src/axum/routes.rs
ss-es: no definitely not, very good catch!

crates/serialization/api/proto/v2/common.proto
```diff
@@ -4,3 +4,64 @@ package espresso.api.v2;
...
+message NetAddr {
+  oneof addr_type {
+    InetAddr inet = 1;
+    NameAddr name = 2;
+  }
+}
+
+// IP address (v4 or v6) with port
+message InetAddr {
+  string ip = 1;    // IP address string (e.g., "127.0.0.1" or "::1")
```
Ayiga: Are we expecting to be able to concatenate the `port` and the `ip` without inspecting the `ip` value itself?

If so, then we may actually want to store `[::1]` instead of just `::1` for convenience.

crates/espresso/node/src/api/state.rs
ss-es: oops, fixed: [9864160](https://github.com/EspressoSystems/espresso-network/pull/4216/commits/9864160c4b7b4121868cfdf9295899440d353e4d)

the code is a bit weird, but I think this follows the existing behaviour around line 413 in `crates/espresso/node/src/api/endpoints/availability.rs`. unfortunately I think this specific endpoint is hard to add to the test, we don't have any existing tests for it and I'm not sure if anyone is using this endpoint anyway

crates/espresso/api/src/axum/routes.rs
ss-es: added tests to ensure this doesn't happen: [9d917b1](https://github.com/EspressoSystems/espresso-network/pull/4216/commits/9d917b148e08e9f1ec95c4a8707d9eddf461fa22)

crates/serialization/api/proto/v2/common.proto
ss-es: fixed here: [9864160](https://github.com/EspressoSystems/espresso-network/pull/4216/commits/9864160c4b7b4121868cfdf9295899440d353e4d)

crates/espresso/node/src/api/state.rs
ss-es: I think `u32` was how the old Tide API was casting it in some places. But `u64` is the internal type, and I think this is actually better to just make a `u64` for consistency (especially since it only eliminates conversions)

fixed: [6ad876c](https://github.com/EspressoSystems/espresso-network/pull/4216/commits/6ad876c851f25ee1bc5dfcfb6809a589e4d91a0a)

sveitser (approved):
I think this is fine but what about the query service availability API? The comment I left here: https://github.com/EspressoSystems/espresso-network/pull/4216#issuecomment-4351555408

PR #4219, opened 2026-04-23
  fix(reconstruct_state): load latest reward tree for V5+ block height

imabdulbasit:
reward_merkle_tree_v2 is only persisted at last block of an epoch, so it fails to load the tree if height is not last block of the epoch. This PR fixes it by loading the last stored tree with height <= requested height for V5+ state reconstruction. The loaded tree should be the correct one because v5+ only updates tree at epoch boundary. We also do commitment check so it should catch the wrong tree.


ss-es (approved):
looks correct to me

PR #4224, opened 2026-04-27
[Fast Finality] State Certificate Voting + Proposing

bfish713:
### This PR:

- Moves the light client state helper functions to a common place so they are shared between old and new protocol
- Creates a vote collector for the state certificate.  
- Also collects regular vote 1s for the root certificate in new collector.  This means either we form a Cert1 + State cert of nothing at all.
- Adds the logic to propose the cert when extending the root
- Adds logic to send state cert vote when a proposal is the epoch root
- Adds logic to check for the state cert when the proposal extends the epoch root


crates/hotshot/new-protocol/src/epoch_root_vote_collector.rs
```diff
@@ -0,0 +1,244 @@
...
+        self.per_view = keep;
+        self.membership_cache = self.membership_cache.split_off(&epoch);
+    }
+}
+
+fn generate_vote_commitment<T: NodeType, V: Vote<T>>(
+    vote: &V,
+    upgrade_lock: &UpgradeLock<T>,
+) -> Option<committable::Commitment<VersionedVoteData<T, V::Commitment>>> {
+    match VersionedVoteData::new(vote.date().clone(), vote.view_number(), upgrade_lock) {
```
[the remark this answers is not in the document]
bfish713: This is the actual name of the function, it's a typo from long before this pr, the build is fine

contracts/rust/adapter/src/light_client.rs
```diff
@@ -116,3 +136,98 @@ impl<F: PrimeField> From<GenericStakeTableState<F>> for StakeTableStateSol {
...
+    });
+
+    let mut accumulated_stake = U256::from(0);
+    let signed_state_digest = derive_signed_state_digest(
+        &state_cert.light_client_state,
+        &state_cert.next_stake_table_state,
+        &state_cert.auth_root,
+    );
+    for (key, sig, sig_v2) in state_cert.signatures.iter() {
+        if let Some(stake) = state_key_map.get(key) {
```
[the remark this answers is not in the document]
mrain: This suggestion is valid, the original implementation is not good

crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -730,6 +762,33 @@ impl<T: NodeType> Consensus<T> {
...
+                return;
+            };
+            let Some(sc) = self.state_certs.get(&parent_epoch).cloned() else {
+                warn!(
+                    %view,
+                    "epoch-root parent QC without state_cert — atomicity invariant broken; skipping propose"
+                );
+                return;
+            };
+            if !check_qc_state_cert_correspondence(parent_cert, &sc, *self.epoch_height) {
```
mrain, line 891: why don't we also validate state cert here?

crates/hotshot/new-protocol/src/tests/common/utils.rs
```diff
@@ -376,6 +410,33 @@ impl TestData {
...
+                    .stake_table()
+                    .await
+                    .commitment(num_nodes)
+                    .expect("stake table commitment");
+                let state_cert = build_state_cert_for_test(
+                    &proposal.block_header,
+                    view_number,
+                    epoch,
+                    &next_stake_table_state,
+                    num_nodes as u64,
```
mrain: we could use `one_honest_threshold` here, given uniform stake distribution it should be `num_nodes / 3 + 1`

crates/hotshot/new-protocol/src/tests/common/utils.rs
```diff
@@ -463,6 +535,96 @@ pub fn key_map_with_num_nodes(num_nodes: u64) -> BTreeMap<BLSPubKey, BLSPrivKey>
...
+    let light_client_state =
+        <_ as BlockHeader<TestTypes>>::get_light_client_state(block_header, view_number)
+            .expect("get_light_client_state");
+    let auth_root = <_ as BlockHeader<TestTypes>>::auth_root(block_header).expect("auth_root");
+
+    let mut signatures = Vec::with_capacity(num_signers as usize);
+    for i in 0..num_signers {
+        let state_key_pair = StateKeyPair::generate_from_seed_indexed([0u8; 32], i);
+        let state_sign_key = state_key_pair.sign_key_ref();
+        let v2_sig =
```
mrain, line 604: nit: given that all our deployed light client contract is at V3, we can get rid of v2 signature in the state cert.  it requires some new structs, so not necessarily in this PR.

bfish713: I considered this, but yeah it's actually a bit of a pain to change this type because it changes commitments and serialization for a type we already use.  I think we should upgrade to only include the V3 signature in a follow up, but in a new PR

crates/hotshot/new-protocol/src/consensus.rs
bfish713: I don't think we need to because this would be a cert that we formed ourselves, so we should already know it's valid.

mrain (approved):

PR #4227, opened 2026-04-27
Validate genesis toml for v3 and v4

imabdulbasit:
validates that the genesis toml file has loaded the required epoch related field


crates/espresso/node/src/genesis.rs
```diff
@@ -91,6 +93,32 @@ impl Genesis {
...
+        }
+
+        if version >= DRB_AND_HEADER_UPGRADE_VERSION {
+            self.drb_difficulty
+                .context("drb_difficulty missing from genesis")?;
+            self.drb_upgrade_difficulty
+                .context("drb_upgrade_difficulty missing from genesis")?;
+        }
+
+        Ok(())
```
ss-es, line 119: this looks fine to me

if we're adding this though, can we also add a test for all the genesis files committed in data/genesis? (should be simple -- just read each file in data/genesis and make sure the validation succeeds)

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4227/changes/55979fac80e37ed004e8ca35fcb28316ac648394

ss-es (approved):
looks good to me

lukaszrzasik (approved):

PR #4231, opened 2026-04-27
[new-protocol] add cert2 storage in query service and new protocol light client finality proof

imabdulbasit:
Adds query service storage and fetching for Certificate2 and uses it to build new protocol light client finality proofs.


crates/espresso/node/src/catchup.rs
```diff
@@ -699,25 +744,53 @@ where
...
+                .context("no cert2 available for new-protocol leaf")?;
+
+            let cert2_height = cert2.data.block_number;
+            let mut leaves = vec![leaf];
+            if height < cert2_height {
+                leaves.extend(self.db.get_leaf_range(height + 1..=cert2_height).await?);
+            }
 
-        Ok(leaf)
+            verify_leaf_chain_with_cert2(
```
bfish713: I know all this verify leaf chain stuff was here, but do we actually need to do this if loading from our old DB.  I think I originally added it because I wasn't sure, but on second though I think we can do similar to the light client and just return what we got in DB since it should already be verified.

crates/hotshot/new-protocol/src/utils.rs
```diff
@@ -0,0 +1,86 @@
...
+        return Ok(leaf_chain[0].clone());
+    }
+
+    let mut current = &leaf_chain[0];
+    for leaf in leaf_chain[1..].iter() {
+        let justify_qc = current.justify_qc();
+        if justify_qc.view_number() != leaf.view_number()
+            || justify_qc.data().leaf_commit != leaf.commit()
+        {
+            continue;
```
bfish713, line 71: This is probably ok to keep, but this is a case we should never hit.  The leaf chain shouldn't have a random leaf in it not pointed to by the others.  I could see it somehow happening by mistake for a view that timed out (but we saw a proposal).  I'd say just log a warning here if we get to this `continue`

hotshot-query-service/src/data_source/update.rs
bfish713, line 1: I need to examine these changes a bit more, is there a description somewhere of exactly what we expect this needs to do for both protocol versions?

bfish713 (commented):
Generally looks good, I will look at update.rs tomorrow but I went through everything else and left just a couple comments.

crates/espresso/node/src/catchup.rs
imabdulbasit: yeah I don't see the point of verifying if we are loading from db.

hotshot-query-service/src/data_source/update.rs
imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4231/changes/21518098b54364c417a7fa9170b20beb6e0e1179

crates/espresso/node/src/request_response/data_source.rs
```diff
@@ -107,47 +107,20 @@ impl<I: NodeImplementation<SeqTypes>, N: ConnectedNetwork<PubKey>, P: SequencerP
...
             Request::Leaf(height) => {
-                // Try to get the leaves from memory first, then fall back to storage
-                let mut leaves = self.consensus_handle.undecided_leaves().await;
-                leaves.sort_by_key(|l| l.view_number());
-
-                if let Some((position, mut last_leaf)) =
-                    leaves.iter().find_position(|l| l.height() == *height)
+                // Try in-memory undecided leaves first. Fall back to storage,
+                //
+                // TODO: under the new protocol, also serve from memory
```
bfish713: I don't really think we need to do this for the new protocol

crates/espresso/node/src/catchup.rs
```diff
@@ -370,28 +371,59 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
         stake_table: HSStakeTable<SeqTypes>,
         success_threshold: U256,
     ) -> anyhow::Result<Leaf2> {
-        // Get the leaf chain
+        let upgrade_lock = UpgradeLock::<SeqTypes>::new(versions::Upgrade::trivial(EPOCH_VERSION));
+
```
bfish713: This doesn't look right.  We pass this lock in below for the fetching with cert2 but the version will be the old version

bfish713 (approved):
Besides the upgrade_lock thing this looks good to, make sure to check that before merging though

crates/espresso/node/src/catchup.rs
imabdulbasit: this is only used by signature verification which is the same for both protocols. But yeah I think it would be better to use correct version

PR #4236, opened 2026-04-29
Bf/satke-catchup

bfish713:
### This PR:

Implements initial stake table catchup to seed cliquenet properly.

The initial catchup works as follows:

1. Find the highest epoch which we already have stake table and preceding stake table for.  We need the 2 consecutive stake tables because epoch roots apply 2 epochs after the one they are decided in.  So if e is our higest stake table we need the stake table for e-1 to get the stake table for e+1
2. If there is no stake tables in storage we start from the first epoch + 1 (the epoch seeded in `set_first_epoch`)
3. Starting from this highest epoch, query for the next epoch root from all state peers, validate it and get the stake table until no root is available from any.
4. After this process in most case we should have the current previous and next epoch stake table at least.  This is because the epoch root for the next epoch is decided before entering the current epoch.  In some cases where an epoch is almost over and the root is decided, but the change hasn't happened we'll have the current and next 2 epochs.  In all cases we will be able to receive messages and connect to the network.
5. We pass the union of prev, current and next epoch peers into cliquenet and start consensus

Also when we validate a proposal we make sure to update the epoch in cliquenet to make sure it's accurate.


crates/espresso/node/src/startup_catchup.rs
```diff
@@ -0,0 +1,116 @@
...
+    // be unfillable from peers and would silently terminate the bootstrap
+    // at a stale epoch.
+    //
+    // `set_first_epoch` always seeds `first_epoch` and `first_epoch + 1`,
+    // so the scan terminates at worst at `first_epoch + 1`.
+    let mut highest = {
+        let m = membership.read().await;
+        let initial = m.highest_known_epoch().unwrap_or(first_epoch + 1);
+        let mut h = initial;
+        while h > first_epoch + 1 && !(m.has_stake_table(h) && m.has_stake_table(h - 1)) {
```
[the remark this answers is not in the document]
bfish713: `h` is an EpochNumber so changing to h -= 1 will not compile

crates/espresso/node/src/lib.rs
```diff
@@ -750,6 +751,11 @@ where
     // reintroduce CompatNetwork for legacy and spin up a separate CliqueNet network
     // for the fast finality consensus upgrade i.e Coordinator.
     let cliquenet = {
+        // Seed cliquenet with the genesis non-epoch committee. Once
+        // `SequencerContext::init` runs `load_start_epoch_info` and
+        // `bootstrap_epoch_window`, it calls `update_view` on this network
+        // which replaces these peers with the proper N-1/N/N+1 sliding
+        // window for the current epoch.
         let peers = coordinator
```
twittner: I wonder if we should initialise the network without peers at this point.

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -276,6 +276,18 @@ where
                         self.vid_reconstructor.handle_vid_share(s.clone(), m);
                         self.storage.append_vid(s);
                         self.storage.append_proposal(validated.message.proposal.data.clone());
+
+                        // Refresh the network's peer set when the proposal's
+                        // epoch advances so cliquenet keeps connections to the
+                        // surrounding stake-table window in sync.
```
twittner:
```suggestion
```
This runs on every proposal received.

crates/espresso/node/src/startup_catchup.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    // `highest` corresponds to N+1 (the leaf at root_block_in_epoch(N-1) is
+    // the last finalized one peers can serve). So current epoch N = highest - 1.
+    let current = if *highest >= 1 {
+        EpochNumber::new(highest.saturating_sub(1))
+    } else {
+        highest
+    };
```
twittner:
```suggestion
    let current = EpochNumber::new(highest.saturating_sub(1));
```
`highest.saturating_sub(1)` already ensures that `highest` does not underflow.

twittner (approved):

PR #4238, opened 2026-04-30
Add `oldest-leaf` and `oldest-block` endpoints

ss-es:
Adds endpoints for helping debug database catchup/sync issues

### DIFF
--- a/crates/espresso/node/api/database.toml
+++ b/crates/espresso/node/api/database.toml
@@ -7,3 +7,21 @@ Get the sizes of all database tables.
 Returns a list of table size information, including row counts (estimated for PostgreSQL) and disk usage.
 Works with both PostgreSQL and SQLite backends (via embedded-db feature; disk usage currently PostgreSQL only).
 """
+
+[route.get_oldest_block]
+PATH = ["oldest-block"]
+METHOD = "GET"
+DOC = """
+Get the block height of the oldest (smallest height) block present in the database.
+
+Returns null if no blocks are stored.
+"""
+
+[route.get_oldest_leaf]
+PATH = ["oldest-leaf"]
+METHOD = "GET"
+DOC = """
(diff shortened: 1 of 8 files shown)


crates/espresso/node/src/api/sql.rs
```diff
@@ -976,6 +976,52 @@ impl ChainConfigPersistence for Transaction<Write> {
...
+            .context("opening transaction to fetch oldest leaf")?;
+        let row = sqlx::query("SELECT MIN(height) AS height FROM leaf2")
+            .fetch_one(tx.as_mut())
+            .await
+            .context("failed to query oldest leaf height")?;
+        let height: Option<i64> = row.try_get("height")?;
+        match height {
+            None => Ok(None),
+            Some(h) => Ok(Some(
+                tx.get_leaf(LeafId::<SeqTypes>::from(h as usize))
```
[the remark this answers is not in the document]
ss-es: updated with try_from, but no `expect`

crates/espresso/node/api/database.toml
```diff
@@ -7,3 +7,21 @@ Get the sizes of all database tables.
 Returns a list of table size information, including row counts (estimated for PostgreSQL) and disk usage.
 Works with both PostgreSQL and SQLite backends (via embedded-db feature; disk usage currently PostgreSQL only).
 """
+
+[route.get_oldest_block]
```
imabdulbasit: I think these should be part of `node` module because these endpoints tell us state about node. Database module is more like stats for the database. WDYT?

ss-es: no opinion from me. I've moved them to `node`

it ultimately doesn't matter since these will be migrated in the next few weeks

imabdulbasit (approved):

PR #4258, opened 2026-05-06
[new-protocol] new decide event handling, demo, integration test

imabdulbasit:
end to end integration of new protocol in espresso node.

 - `Coordinator::maker` now seeds `seed_genesis` /
    `seed_state` from the `HotShotInitializer`
  - **Stake table API**: reject requests for epochs below `first_epoch`.
  - **Leaf fetcher**: compare QC by `(view, data)` instead of full commit
    (signature subsets differ across cert1s for the same leaf). 
  - **Decide events**: `decide_events_from_chain` splits chains at
    `NEW_PROTOCOL_VERSION`, emitting legacy `Decide` + `NewDecide` as needed.
- add demo file and integration test


crates/espresso/node/src/persistence/fs.rs
```diff
@@ -494,24 +509,40 @@ impl Inner {
...
+                        data: share,
+                        signature: proposal.signature,
+                        _pd: Default::default(),
+                    }),
+                    _ => None,
+                });
+                CoordinatorEvent::NewDecide(NewDecideEvent {
+                    leaves: vec![leaf.leaf],
+                    cert1: cert.qc().clone(),
+                    cert2: self.load_cert2(view)?,
```
bfish713: Do we insert a cert2 for every view, even those decided indirectly.  I.e. if cert2(V) decides V and v-1 do we store cert(v) for view v-1? or does it work some other way?  Otherwise we won't generate a decide event for indirectly decided views.

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -716,6 +716,85 @@ pub struct Persistence {
...
+                    signature: proposal.signature,
+                    _pd: Default::default(),
+                }),
+                _ => None,
+            }));
+        }
+
+        events.push(CoordinatorEvent::NewDecide(NewDecideEvent {
+            leaves,
+            cert1,
```
bfish713, line 801: I'm a little confused about the cert handling here.  We have one decide event with many leaves which has 2 certificates.  I assume some of it is a hold over from the old protocol but the cert1 here doesn't really mean anything if we have the cert 2.  No real issue, just pointing out that the code is a little confusing, especially since the fs impl does a decide event for every leaf.

crates/espresso/node/src/persistence/fs.rs
bfish713: One way to handle this is to follow claudes comment and make this function work like SQL where we generate the decides in a batch (1 event for many leafs)

crates/espresso/node/src/lib.rs
```diff
@@ -805,12 +805,27 @@ where
...
+            .config
+            .known_nodes_with_stake
+            .iter()
+            .filter(|p| p.stake_table_entry.stake_key != pub_key)
+            .filter_map(|p| {
+                p.connect_info
+                    .clone()
+                    .map(|info| (p.stake_table_entry.stake_key, info))
+            })
+            .collect();
```
bfish713: This shouldn't be needed, I removed this in recent PR and we seed cliquenet after we attempt stake table catchup.  Can we remove this?

crates/espresso/types/src/v0/traits.rs
```diff
@@ -794,34 +794,98 @@ pub trait SequencerPersistence:
...
+                    tracing::error!(
+                        leaf_count = decide.leaves.len(),
+                        vid_share_count = decide.vid_shares.len(),
+                        "new protocol decide event has mismatched leaves and VID shares"
+                    );
+                    return;
+                }
+
+                for signed in decide.vid_shares.iter().flatten() {
+                    if let Err(err) = self.append_vid(&convert_proposal(signed.clone())).await {
```
bfish713: Don't we already do this from consensus?

crates/espresso/node/src/persistence/fs.rs
bfish713: Looking at other code I would actually expect that we don't always append a cert2 before we get here and the fs nodes do not properly handle this case

crates/espresso/types/src/v0/traits.rs
```diff
@@ -794,34 +794,98 @@ pub trait SequencerPersistence:
...
+                        tracing::error!("failed to append VID share from new protocol: {err:#}");
+                    }
+                }
+
+                let leaf_infos = decide
+                    .leaves
+                    .iter()
+                    .zip(&decide.vid_shares)
+                    .map(|(leaf, vid_share)| {
+                        let state = Arc::new(ValidatedState::from_header(leaf.block_header()));
```
bfish713: How does the old protocol handle state, this doesn't seem right but idk if it matters.  I feel like we have the state directly from HS in the old version, shouldn't we do the same for this new version?

crates/hotshot/new-protocol/src/block.rs
```diff
@@ -137,6 +137,12 @@ impl<T: NodeType> BlockBuilder<T> {
         let membership = self.membership.clone();
 
         let handle = self.tasks.spawn(async move {
+            // Throttle block production: without this, block times are fast
+            // enough that the coordinator's event queue overflows downstream.
+            // TODO: not viable for production — replace with proper backpressure
+            // on the coordinator event queue (or a configurable target block time)
+            // before mainnet.
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
bfish713: I think we need to address this before merging.  We can test with the sleep but it's shouldn't be in main.  I think the best approach to fix this is to add a sleep if we have no transactions in our queue.  That would be an ok place to start and it would be ok to merge, then we can improve it later.

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -121,10 +121,29 @@ where
...
+
+        let mut state_manager = StateManager::new(
             Arc::new(initializer.instance_state.clone()),
             upgrade_lock.clone(),
         );
+        state_manager.seed_state(
+            ViewNumber::genesis(),
+            initializer.anchor_state.clone(),
+            initializer.anchor_leaf.clone(),
+        );
```
bfish713, line 155: Is some of this duplicate elsewhere?  Specifically the genesis state I feel like we seed somewhere else.

bfish713 (changes requested):
I didn't check waiting for L1 stuff too closely but I trust the tests.  I think the comments I left about generating the decide events may be valid.  Basically I'm worried that we're going to miss generating decide events when views are decided indirectly.  I'm also worried about inserting empty state into the `LeafInfo` when handling new protocol decides.

crates/espresso/types/src/v0/traits.rs
imabdulbasit: actually I don't think this state is used anywhere. We can let hotshot pass the state in the event if we want to

crates/espresso/node/src/lib.rs
imabdulbasit: hmm I am pretty sure that the v8 base demo wasn't working if I removed it, but I'll check again

crates/espresso/node/src/persistence/fs.rs
imabdulbasit: so I think we will append cert2 on the hotshot side.` self.load_cert2(view)?` will return None if the cert2 doesn't exist for the view.

crates/espresso/node/src/persistence/sql.rs
imabdulbasit: so the only reason we have cert1 in the decide event is for the query service to store it like qc, because the query service stores qc for every leaf

crates/hotshot/new-protocol/src/coordinator.rs
imabdulbasit: yeah I copied this code from the bench binary

crates/espresso/node/src/persistence/fs.rs
imabdulbasit: I know this is confusing because we are storing cert1 for every leaf. We actually don't use FS for query service but I'll see if I can make a batch event like SQL

crates/espresso/types/src/v0/traits.rs
imabdulbasit: hmm I am now thinking if it would make be cleaner to just pass `LeafInfo` struct like old protocol does

crates/espresso/node/src/persistence/fs.rs
bfish713: But a node will never store a cert 2 for a leaf that it didn't form a cert2 for, and this line will cause the function to return

crates/espresso/node/src/persistence/sql.rs
bfish713: Yeah that's what I figured, so cert1 in this case is just the cert of the newest leaf, and then we can later get the other cert1s from the justify qc

crates/espresso/node/src/lib.rs
bfish713: If it doesn't work we just have to make sure we  update the peers after set first epoch is called

crates/hotshot/new-protocol/src/coordinator.rs
bfish713: ah ok that's fine then

bfish713: But does this mean we can remove this code from the bench binary?

imabdulbasit: so there is a build_coordinator() in bench binary which uses test types
https://github.com/EspressoSystems/espresso-network/blob/1cb7f44f11a2eb9110b84a65689be301b23fdbe0/crates/hotshot/new-protocol/bench/src/node.rs#L156-L160

crates/espresso/node/src/lib.rs
imabdulbasit: so the problem was this condition is true for genesis epoch.

```
  if epoch <= self.epoch {
            info!(%epoch, ours = %self.epoch, "epoch already seen");
            return Ok(());
        }
```

bfish713 (approved):

PR #4259, opened 2026-05-06
fix(bootstrap epoch window): make bootstrap step timeout configurable

imabdulbasit:
The startup `bootstrap_epoch_window` walk chased the live tip indefinitely on busy networks (chain produced finalized roots faster than each per step fetch). This PR makes the per step timeout as env variable `ESPRESSO_NODE_BOOTSTRAP_EPOCH_CATCHUP_TIMEOUT` (default 30s, 2s in slow-tests).
 
 Slow-tests use epoch_height=15, so a new finalized root lands every few seconds which is faster than 10s per step fetch so this is stuck infinitely


sveitser:
@imabdulbasit can you leave a comment and open an issue describing why this is not okay as a final solution and then we can merge it to unblock the CI? If the CI is broken everyone has to force merge and if they break more tests they would probably miss that.

cc @bfish713 @twittner

sveitser (approved):

bfish713 (approved):

PR #4260, opened 2026-05-06
fix(libp2p): make AutoNAT log messages actionable for operators

sveitser:
- Per-probe failures ([autonat::OutboundProbeEvent::Error](url)) demoted from warn! to debug!. Each failed probe is a single peer momentarily unable to dial back; the operator cannot fix any individual probe and the `WARN` spam drowns out the actually-useful aggregate signal.

- `StatusChanged` events promoted from debug! to actionable messages keyed off `NatStatus`:

```
    Public(addr): info!  - confirms reachability, includes the address
    Private:      error! - operator MUST act, message names the flag and the
                           network-level checks (firewall, NAT, security group)
    Unknown:      debug! - genuinely uncertain, not actionable
```

- Before this change, an operator behind a misconfigured NAT/NodePort saw a flood of opaque `"AutoNAT Probe failed to peer ... with error: OutboundRequest( Io(Kind(UnexpectedEof)))"` lines and no clear signal that their node was unreachable. After this change they see a single `ERROR` line naming `--libp2p-advertise-address` (env `ESPRESSO_NODE_LIBP2P_ADVERTISE_ADDRESS`) and the firewall checks they need to perform.

- Drive-by: fix a "fales" typo in a comment and whitelist the `bimap` crate name in .typos.toml so the pre-commit spell-check passes.


sveitser:
One of the things I don't really know due to not being familiar with libp2p is when the status changes. I will check that soon.

crates/hotshot/libp2p-networking/src/network/node.rs
```diff
@@ -687,14 +687,31 @@ impl<T: NodeType, D: DhtPersistentStorage> NetworkNode<T, D> {
...
+                                    error!(
+                                        "AutoNAT: this node is NOT publicly reachable. Peers \
+                                         cannot direct-message us, so leader views will fail and \
+                                         we will accumulate missed slots. Fixes: set \
+                                         --libp2p-advertise-address (env \
+                                         ESPRESSO_NODE_LIBP2P_ADVERTISE_ADDRESS) to a publicly \
+                                         reachable host:port, and ensure inbound UDP at that port \
+                                         is open from the public internet (firewall/NAT/security \
+                                         group). Status transitioned {old:?} -> Private."
+                                    );
```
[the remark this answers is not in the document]
sveitser: I think given this code is going to removed soon this is acceptable

sveitser:
> One of the things I don't really know due to not being familiar with libp2p is when the status changes. I will check that soon.

So it seems this would fire if the first check fails which is bad, will fix.

twittner (approved):

crates/hotshot/libp2p-networking/src/network/node.rs
```diff
@@ -687,16 +696,46 @@ impl<T: NodeType, D: DhtPersistentStorage> NetworkNode<T, D> {
...
+                                    self.autonat_private_logged = false;
+                                },
                             },
                         };
+                        let autonat = &self.swarm.behaviour().autonat;
+                        if matches!(autonat.nat_status(), autonat::NatStatus::Private)
+                            && autonat.confidence() >= AUTONAT_CONFIDENCE_MAX
+                            && !self.autonat_private_logged
+                        {
+                            error!(
```
bfish713, line 733: I think this error is too harsh, we can still perform as leader (and should) without libp2p.  If the CDN is also down or we aren't connect then we will miss slots.

bfish713 (approved):

crates/hotshot/libp2p-networking/src/network/node.rs
sveitser: But it's just a log message. If this autonat stuff works correctly then we are in a degraded state and operators should fix it ASAP. We are still continuing.

sveitser: @bfish713 updated message in b9010fd

PR #4261, opened 2026-05-06
fix(libp2p): plumb advertise address to swarm.add_external_address

sveitser:
- The CLI flag `--libp2p-advertise-address` (env `ESPRESSO_NODE_LIBP2P_ADVERTISE_ADDRESS`)
  was only used in the orchestrator-bootstrap branch (lib.rs:446) and never reached
  the libp2p Swarm in the persistence-loaded or peer-fetch branches. The flag's
  docstring promised that it "advertises to other nodes" but in production paths
  it was a silent no-op.

- Effect: validators behind NAT, K8s NodePort, or Docker bridge networking can only
  be reached by peers they have already established outbound connections to. Identify
  announces their bind-derived addresses (loopback / VPC-internal IPs), so peers that
  learn about them via DHT lookup, or that lose and try to re-establish a connection,
  cannot dial them. Coverage of the validator set is incomplete and degrades over
  time as connections drop. Observed-address from existing connections is not
  propagated by identify to third parties, and AutoNAT cannot self-confirm the
  public address because the bind-derived addresses it offers as probe candidates
  are not reachable.

- Fix: thread Vec<Multiaddr> through Libp2pNetwork::from_config (libp2p_network.rs)
  into NetworkNodeConfig.announce_addresses (config.rs), and call
  Swarm::add_external_address for each in NetworkNode::new (node.rs). The existing
  SwarmEvent::ExternalAddrConfirmed handler at node.rs already populates the DHT
  self-record once the address is confirmed.

- Also drop the default_value = "localhost:1769" on the CLI flag. Loopback was
  actively wrong as a default once the flag is wired through to identify; absent
  is now valid (no advertise) and bootstrapping the orchestrator without it
  errors explicitly with a clear message.

- Drive-by: fix a "fales" typo in a comment and whitelist the bimap crate name
  in .typos.toml so the pre-commit spell-check passes.

- References:

  rust-libp2p Swarm::add_external_address (libp2p-swarm 0.47):
    https://docs.rs/libp2p/0.56.0/libp2p/swarm/struct.Swarm.html#method.add_external_address
    "Add a confirmed external address for the local node. This function should
    only be called with addresses that are guaranteed to be reachable."

  libp2p Identify spec - the listenAddrs field that peers receive during
  identify exchange:
    https://github.com/libp2p/specs/blob/master/identify/README.md
  Without add_external_address, only listen addresses are populated, which for
  bind=0.0.0.0 means loopback and any local interface IPs.


sveitser:
@claude review

bfish713 (approved):

PR #4262, opened 2026-05-06
[Fast Finality] Protocol upgrade + tests

bfish713:
## Summary

  Implements the legacy → new-protocol (HotShot 0.8 / cliquenet) upgrade cutover. Both stacks run side-by-side per node;  at the decided upgrade view, the new-protocol coordinator takes over from legacy with no observable gap and without losing decided/undecided state.

  ## How the cutover works

  1. **Legacy stack runs alone** until an `UpgradeCertificate` for `CLIQUENET_VERSION` is decided. Each existing
     task (`consensus`, `da`, `quorum_proposal{,_recv}`, `quorum_vote`, `vid`, `view_sync`, `transactions`) early-returns
     from its event loop once `upgrade_lock.new_protocol_active(view)` is true. The wire layer was already
     version-gated; this stops legacy from producing protocol traffic past the boundary.
  2. **The new-protocol coordinator is spawned at startup** but stays effectively idle behind a `CutoverGate`
     (`cutover.rs`). Every routing call (`current_view`, `submit_transaction`, `update_leaf`, ...) checks the gate;
     while inactive, calls go to legacy.
  3. **At cutover**, the first `cutover_active()` call that observes legacy's `cur_view >= CLIQUENET_VERSION` extracts a
     `PreCutoverSeed` from the legacy `SystemContextHandle` (decided anchor, undecided chain, `high_qc`, validated
     states, `new_version_first_view`) and dispatches it into the coordinator via `ClientApi::seed_pre_cutover`. The
     gate latches; subsequent calls short-circuit.
  4. **`Consensus::apply_pre_cutover_seed`** anchors `last_decided_view`, installs undecided legacy leaves as
     `proposals` (marked `pre_cutover_views`), registers each `justify_qc` as a Cert1 (bumping `locked_cert`),
     advances `current_view`/`timeout_view` to `cutover_view - 1`. Pre-cutover views are V1 AvidM dispersal so
     `maybe_vote_2` skips them; their children verify against the parent proposal but skip V2 block-reconstruction.
     The coordinator then refreshes peers for the cutover epoch and kicks the leader (either resuming a forwarded-TC2
     path or calling `start()` to propose).
  5. **Two background bridges** run for the lifetime of the node:
     - `forward_legacy_timeout_votes` — legacy now emits a `LegacyTimeoutVoteEmitted` event whenever it signs a
       `TimeoutVote2` (gated on `decided_upgrade_cert().is_some()`). The bridge forwards each into the new-protocol
       timeout collectors + rebroadcasts on the new wire, so the first new leader can form TC2 even if cutover happens
       mid-timeout.
     - `forward_legacy_epoch_changes` — forwards legacy `Decide` events as `bump_network_epoch` so cliquenet peers stay
       current while legacy is still driving.

  ## Key places to review

  - `crates/hotshot/new-protocol/src/cutover.rs` — gate, seed extraction, the two forwarders.
  - `crates/hotshot/new-protocol/src/consensus.rs` — `PreCutoverSeed`, `apply_pre_cutover_seed`, `register_legacy_qc`,
    and the `pre_cutover_views` carve-outs in proposal validation / `maybe_vote_2`.
  - `crates/hotshot/new-protocol/src/coordinator.rs` — `SeedPreCutover` handler (state-manager seeding, network epoch
    apply, resume-after-TC vs `start()` dispatch) and the rewritten `start()` that now works from any view, not just
    view 1.
  - `crates/espresso/node/src/consensus_handle.rs` — `CutoverGate` integration; rename
    `new_protocol`/`new_protocol_at` → `cutover_active`/`at_or_past_cutover` to disambiguate "stateful: has the gate
    latched?" from "stateless: is this view past the version boundary?".
  - `crates/hotshot/task-impls/src/*` — early-return guards on every legacy task.
  - `crates/hotshot/new-protocol/src/tests/legacy_cutover.rs` — full E2E (legacy cluster runs, hits upgrade view,
    new-protocol takes over). `tests/cutover.rs` covers the seed-apply unit cases including all timeout-near-boundary
    combinations.


crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -250,6 +262,117 @@ impl<T: NodeType> Consensus<T> {
...
+                max_seeded_epoch = epoch;
+            }
+        }
+
+        // Advance current_view past the highest seeded view so the local
+        // node's leader logic targets the first POST-cutover view rather
+        // than re-running the seeded ones.
+        if max_seeded_view > self.current_view {
+            self.current_view = max_seeded_view;
+            self.current_epoch = Some(max_seeded_epoch);
```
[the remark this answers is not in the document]
bfish713: new protocol always should have epochs

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -266,27 +263,61 @@ where
...
+                        parent_proposal,
                     },
                 ));
         }
     }
 
+    /// Kick the leader after the seed lands when a forwarded TC2 had
+    /// already advanced `current_view`. No-op unless leader and all
+    /// prerequisites are present.
+    async fn resume_after_cutover_tc(&mut self) {
```
twittner:
```suggestion
    fn resume_after_cutover_tc(&mut self) {
```

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -1004,6 +1038,105 @@ where
...
+                    tracing::error!(
+                        %cutover_epoch,
+                        %err,
+                        "network on_epoch_change failed during seed_pre_cutover",
+                    );
+                }
+
+                let cur_view = self.consensus.current_view();
+                if self.consensus.timeout_cert_at(cur_view).is_some() {
+                    self.resume_after_cutover_tc().await;
```
twittner:
```suggestion
                    self.resume_after_cutover_tc();
```

twittner (approved):

PR #4264, opened 2026-05-06
Migrate hash sequence key to bigint

ss-es:
We hit the sequence limit for the hash table on a decaf node, and it's likely we'll hit the same limit on mainnet/other decaf nodes in the next few months. This PR switches the postgres id to a bigint, which should resolve the issue.

No sqlite migration is needed since that is already a u64.


jbearer (approved):

ss-es:
This seems to be extremely slow to run -- do not merge

PR #4265, opened 2026-05-06
Backport #4264

ss-es:


sveitser:
@ss-es close this, since we're doing the background migration?

PR #4266, opened 2026-05-06
Make upsert robust to large inputs

jbearer:
Both sqlite and postgres restrict the number of query parameters allowed in a single operation. With bulk catchup inserts, we are now starting to hit this limit when we insert large batches of blocks each containing a large number of transactions. This change prevents this scenario by breaking large upserts into chunks.


Ancient123 (approved):

jbearer:
I don't think the `insert_vid_range` found by Claude is an issue. Shares are never backfilled, which means they will never be inserted in bulk. They will only be triggered by a Decide event, with at most a few blocks. Whenever `insert_vid_range` is called with a large batch from backfilling, the shares will be `None` and only the VID common will be inserted (via upsert)

PR #4270, opened 2026-05-07
Migrate query-service availability endpoints

ss-es:
Migrates remaining query-service availability endpoints to `axum`. 

Notably, this PR does not include new v2 endpoints for any of these. I will add those in a separate PR.

There are tests in `crates/espresso/node/src/api.rs` that should ensure parity between the axum and tide APIs for all endpoints.


crates/espresso/node/src/api.rs
```diff
@@ -7555,6 +7555,55 @@ mod test {
...
+
+        assert!(
+            common >= 2,
+            "v1/{path}: expected ≥2 messages in common between tide ({} msgs) and axum ({} msgs), \
+             got {common}",
+            tide_msgs.len(),
+            axum_msgs.len(),
+        );
+        Ok(())
+    }
```
ss-es, line 7784: it's a bit hard to test streaming websocket endpoints for parity conceptually, since there may be subtle timing differences between them (neither returns deterministic data)

to make this test a bit more resilient, we wait for 10 messages from each and just check that the streams had at least 2 in common

crates/espresso/node/src/api.rs
```diff
@@ -7654,272 +7654,790 @@ mod test {
...
+        std::thread::Builder::new()
+            .stack_size(32 * 1024 * 1024)
+            .spawn(move || {
+                tokio::runtime::Builder::new_multi_thread()
+                    .enable_all()
+                    .build()
+                    .unwrap()
+                    .block_on(async move {
+                        const EPOCH_HEIGHT: u64 = 10;
+                        const NUM_NODES: usize = 5;
```
ss-es: it took a lot of experimentation to get this test to not hit a stack overflow. to me this seems innocuous so I didn't try to dig much further to figure out exactly where the line was

crates/cliquenet/Cargo.toml
```diff
@@ -14,16 +14,16 @@ bytes           = { workspace = true }
 ed25519-compact = { workspace = true }
 parking_lot     = { workspace = true }
 rand            = "0.10.0"
+
+# optional:
```
sveitser: undo?

crates/espresso/node/src/api/state.rs
```diff
@@ -1302,6 +1302,48 @@ where
         Ok(proofs)
     }
 
+    async fn stream_namespace_proofs(
+        &self,
+        from: usize,
+        namespace: u32,
+    ) -> anyhow::Result<futures::stream::BoxStream<'static, Self::NamespaceProofQueryData>> {
+        use espresso_types::{NamespaceId, NamespaceProofQueryData, NsProof};
```
sveitser: Move imports to top?

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1501,572 @@ where
...
+        hotshot_query_service::availability::VidCommonQueryData<espresso_types::SeqTypes>;
+    type Transaction =
+        hotshot_query_service::availability::TransactionQueryData<espresso_types::SeqTypes>;
+    type TransactionWithProof = hotshot_query_service::availability::TransactionWithProofQueryData<
+        espresso_types::SeqTypes,
+    >;
+    type BlockSummary =
+        hotshot_query_service::availability::BlockSummaryQueryData<espresso_types::SeqTypes>;
+    type Limits = hotshot_query_service::availability::Limits;
+    type Cert2 = hotshot_new_protocol::message::Certificate2<espresso_types::SeqTypes>;
```
sveitser: import some of these types for readability?

light-client/Cargo.toml
```diff
@@ -14,18 +14,23 @@ alloy = { workspace = true }
 anyhow = { workspace = true }
 async-lock = { workspace = true }
 async-trait = { workspace = true }
+
+# Testing dependencies
+bitvec = { workspace = true, optional = true }
```
sveitser: accidental cargo sort without `-g`?

sveitser:
@ss-es I think my main concern is that we should do some equivalence checks for common errors where they are informative, like range limits / not found. To ensure we are returning reasonable errors from Axum for clients. At least for errors that they probably care about.

crates/espresso/node/src/api.rs
```diff
@@ -7700,272 +7700,863 @@ mod test {
...
+                            )
+                            .collect::<Vec<_>>()
+                            .try_into()
+                            .unwrap();
+
+                        let mut api_opts = Options::with_port(api_port).catchup(Default::default());
+                        api_opts.http.axum_port = Some(axum_port);
+
+                        let config = TestNetworkConfigBuilder::with_num_nodes()
+                    .api_config(SqlDataSource::options(&storage[0], api_opts))
```
sveitser: what's going on with rustfmt 👀

crates/espresso/node/src/api.rs
```diff
@@ -7700,272 +7700,863 @@ mod test {
...
-                axum_port,
-                &format!("reward-state-v2/reward-merkle-tree-v2/{height}"),
-            )
-            .await?;
+                            let mut network = TestNetwork::new(config, upgrade).await;
 
-            // Availability v1 parity: verify the axum v1 routes return the same JSON as tide.
+                            // wait for 4 epochs
+                            let mut events = network.server.event_stream();
+                            wait_for_epochs(&mut events, EPOCH_HEIGHT, 4).await;
```
sveitser: What are we doing here exactly? I mean why do we spin up the network twice for the V2 branch?

crates/espresso/node/src/api.rs
```diff
@@ -7700,272 +7700,863 @@ mod test {
...
+    }
 
-                match res.proof.proof {
-                    RewardMerkleProofV1::Presence(p) => {
-                        assert_eq!(
-                            p, expected_proof,
-                            "Proof mismatch for V1 at {height}, addr={address}"
-                        );
+    /// Connect to both tide-disco and axum WebSocket endpoints, collect up to 10 messages each,
+    /// and assert that at least 2 messages appear in both streams.
```
sveitser, line 7738: why?

crates/espresso/node/src/api.rs
sveitser: ah okay, could inline this comment

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1506,579 @@ where
...
+        let ds = self.data_source.clone();
+        let stream = (*ds)
+            .subscribe_blocks(from)
+            .await
+            .flat_map(move |block| {
+                let ns_filter = namespace.map(NamespaceId::from);
+                let txs: Vec<Self::Transaction> = block
+                    .enumerate()
+                    .enumerate()
+                    .filter_map(|(i, (idx, _))| {
```
sveitser: Can we name these indices?

crates/espresso/node/src/api.rs
ss-es: oh this is strange -- I actually saw this before and could've sworn I removed that

crates/espresso/node/src/api.rs
ss-es: technically they're different websockets that you're subscribing to at different points in time, so there's no guarantee you'll receive exactly the same messages on both, so I worry naively asserting that the streams are identical will lead to flakiness.

so instead I opted to test that the streams have some overlap

crates/espresso/node/src/api.rs
ss-es: should be fixed: [eb12a40](https://github.com/EspressoSystems/espresso-network/pull/4270/commits/eb12a4044abf1723e19f9516e8ba143c19a70b05)

crates/espresso/node/src/api/state.rs
ss-es: [306d82a](https://github.com/EspressoSystems/espresso-network/pull/4270/commits/306d82a0c9f990a65d26af7014e1631a7d34be1e)

crates/espresso/node/src/api.rs
ss-es: I think rustfmt gives up at a deep enough nesting level; this might help: [69b399e](https://github.com/EspressoSystems/espresso-network/pull/4270/commits/69b399e5c0f256a277c2309be529e36742367b7f)

crates/espresso/node/src/api/state.rs
```diff
@@ -594,3 +833,1099 @@ where
...
+        let ns_id = NamespaceId::from(namespace);
+
+        // Validate range
+        if until <= from {
+            return Err(anyhow::anyhow!(
+                "invalid range: until ({}) must be greater than from ({})",
+                until,
+                from
+            ));
+        }
```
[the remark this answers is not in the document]
sveitser: @ss-es this sounds like a valid concern?

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1506,579 @@ where
...
+    type TransactionWithProof = TransactionWithProofQueryData<espresso_types::SeqTypes>;
+    type BlockSummary = BlockSummaryQueryData<espresso_types::SeqTypes>;
+    type Limits = HsLimits;
+    type Cert2 = Certificate2<espresso_types::SeqTypes>;
+
+    async fn get_leaf(
+        &self,
+        id: espresso_api::v1::availability::LeafId,
+    ) -> anyhow::Result<Self::Leaf> {
+        use std::time::Duration;
```
sveitser: Do imports once on top?

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1506,579 @@ where
         Ok(espresso_types::StateCertQueryDataV2(cert))
     }
 }
+
+// ============================================================================
+// v1::HotShotAvailabilityApi implementation
+// ============================================================================
+
+fn not_found(msg: impl Into<String>) -> anyhow::Error {
```
sveitser, line 1470: I'm wondering if there's a better way to do this to avoid these free functions? Even just inlining the function body at the call site seems reasonable?

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1506,579 @@ where
...
+                    })
+                    .collect();
+                futures::stream::iter(txs)
+            })
+            .boxed();
+        Ok(stream)
+    }
+}
+
+fn block_id_to_hs(
```
sveitser, line 1922: This looks like a `TryFrom`?

crates/espresso/node/src/api/state.rs
```diff
@@ -1459,3 +1506,579 @@ where
...
+        espresso_api::v1::availability::BlockId::PayloadHash(h) => {
+            let payload_hash = h
+                .parse()
+                .map_err(|_| bad_request(format!("invalid payload hash: {}", h)))?;
+            Ok(HsBlockId::PayloadHash(payload_hash))
+        },
+    }
+}
+
+fn payload_id_to_hs(
```
sveitser, line 1942: This also looks like a `TryFrom`?

crates/espresso/node/src/api/state.rs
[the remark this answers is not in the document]
ss-es: I don't think it is

this matches the existing semantics in `espresso/node/src/api/light_client.rs:241`. I don't think there's any reason to change (well, especially not in this PR since the goal is to not break anything), since there's nothing technically wrong with these semantics (even if they're a bit weird)

crates/espresso/node/src/api/state.rs
ss-es: I think you'd have to implement `TryFrom` in either espresso-api or hotshot-query-service. espresso-api would need to import hotshot-query-service (which is very bad), or hotshot-query-service would have to import espresso-api (less bad but I'd rather keep the dependency chain clean)

crates/espresso/node/src/api/state.rs
ss-es: I don't think there's anything that bad about this since they're completely local (not `pub`) and self-contained to the module. we could inline it, I think that was a bit messier for some of these but the error coercions probably don't need to be separate functions

crates/espresso/node/src/api/state.rs
ss-es: also on the subject of putting it in e.g. hotshot-query-service -- part of the goal here is to try to keep the serialization interface in one place. the v1 interface is not defined in a proto file so technically this is a lost cause for v1, but as a general rule I'd really like for any conversions involving espresso-api or serialization types to happen in this file (which I think essentially precludes a `TryFrom` implementation)

crates/espresso/node/src/api/state.rs
ss-es: cleaned up imports here: [9d15a1c](https://github.com/EspressoSystems/espresso-network/pull/4270/commits/9d15a1c9c5df688fcab678be3e66b48f87e66e61)

sveitser (approved):

PR #4278, opened 2026-05-08
Stop printing raw merkle proofs in logs

ss-es:
In logging the leaf chain (and possibly other things), we print the full merkle proofs from the raw VID shares. This information is probably not useful in any situation.


mrain (approved):
LGTM!

PR #4281, opened 2026-05-10
feat: add node telemetry

sveitser:
Our side of the telemetry code is in https://github.com/EspressoSystems/espresso-node-telemetry

Opt-in, in-process export of node-operator logs and metrics to the Espresso
telemetry proxy. Both signals are off by default and enabled independently.

## New crate: `espresso-telemetry`

- **Logs**: `tracing` events passing an `EnvFilter` are bridged to an OTLP/HTTP
  batch exporter (`opentelemetry-appender-tracing`).
- **Metrics**: the node's `prometheus::Registry` is scraped on an interval,
  encoded as Prometheus remote-write protobuf, and POSTed. Counters, gauges, and
  histograms are supported; summary/untyped are rejected at encode.
- **Auth**: one BLS-BN254 JWT, signed by the staking key and minted at startup,
  sent as `Authorization: Bearer` on both signals. Optional `node_name` /
  `company_name` claims. TTL six months, not refreshed mid-process.

## Wire format

| Signal  | Protocol          | Path            | Encoding             | Cadence                   |
| ------- | ----------------- | --------------- | -------------------- | ------------------------- |
| Logs    | OTLP/HTTP         | `/v1/logs`      | protobuf, gzip       | `BatchLogProcessor`       |
| Metrics | Prom remote-write | `/api/v1/write` | protobuf, snappy     | `metrics_interval_secs`   |

- Logs export retries transient failures via opentelemetry-otlp's built-in
  `experimental-http-retry`.
- Metrics run on a dedicated thread with its own current-thread runtime; one
  final flush on shutdown.
- HTTP 429 logs a single operator-facing ERROR per process (`compare_exchange`
  latch) carrying the active log filter and `Retry-After`.
- Neither path uses disk; neither blocks consensus or the node runtime. All
  errors are `warn!`-only.

## Configuration

CLI flags flattened into the node `Options`; each has an env var:

| Flag                          | Env                                       | Default        |
| ----------------------------- | ----------------------------------------- | -------------- |
| `--telemetry-logs-enable`     | `ESPRESSO_NODE_TELEMETRY_LOGS_ENABLE`     | `false`        |
| `--telemetry-metrics-enable`  | `ESPRESSO_NODE_TELEMETRY_METRICS_ENABLE`  | `false`        |
| `--telemetry-endpoint`        | `ESPRESSO_NODE_TELEMETRY_ENDPOINT`        | by chain id    |
| `--telemetry-log`             | `ESPRESSO_NODE_TELEMETRY_LOG`             | `warn`         |
| `--telemetry-metrics-interval`| `ESPRESSO_NODE_TELEMETRY_METRICS_INTERVAL`| `60`           |

Default endpoint resolves from the genesis chain id: mainnet and decaf testnet
have built-in URLs; other chains require an explicit endpoint or telemetry stays
off with a warning.

## Node integration

- `run::main` builds the telemetry pipeline before logging init so the OTel layer
  attaches to the global subscriber. The `prometheus::Registry` is built later by
  the API setup, deposited via `telemetry::set_registry`, and the metrics push is
  attached once consensus is about to start. The handle is shut down on exit.
- `Config::init_with_otel` (espresso-utils) and `initialize_logging_with`
  (hotshot helpers) compose an optional extra `Layer` over the existing stderr
  fmt layer.
- `PrometheusMetrics::registry` exposes the underlying registry for scraping.

## Supporting changes

- `proto/{remote,types}.proto` vendored from prometheus/prometheus, built via
  `prost-build`. `scripts/update-prometheus-protos.sh` refreshes them at a pinned
  tag and strips gogoproto annotations.
- Demo wiring: `demo/vector.yaml`, `docker-compose.yaml`, `process-compose.yaml`,
  `.env`.

## Tests

Unit and integration tests cover JWT mint/verify, remote-write encoding (counter,
gauge, histogram, `+Inf` dedup, label sort, unsupported type), per-signal enable
flags, the logs/metrics lifecycle against a mock OTLP server, and 429 rate-limit
handling.


crates/espresso/telemetry/tests/encode.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+fn round_trip(req: &WriteRequest) -> WriteRequest {
+    let bytes = req.encode_to_vec();
+    WriteRequest::decode(&*bytes).expect("decode WriteRequest")
+}
+
+// TEST:metrics-counter-encode-ok
+#[test]
+fn metrics_counter_encode_ok() {
```
sveitser, line 48: These tests would be easier to read if they asserted human readable output so the human can directly see what output it produces.

crates/espresso/telemetry/tests/flags.rs
```diff
@@ -0,0 +1,171 @@
...
+    let key = make_staking_key();
+    let opts = TelemetryOptions {
+        logs_enable: false,
+        metrics_enable: false,
+        ..Default::default()
+    };
+    let endpoint = local_endpoint(reserve_port());
+    let (handle, warnings) =
+        init(&opts, &key, None, None, &endpoint, None).expect("disabled init never errors");
+    assert!(handle.is_none(), "all flags off must return None");
```
sveitser: useless custom assert messages, remove all the ones that aren't helpful (probably all or almost all of them)

crates/espresso/telemetry/tests/flags.rs
```diff
@@ -0,0 +1,171 @@
+//! Unit tests for per-signal telemetry enable flags.
+//!
+//! Tests serialize on `TEST_LOCK` (same pattern as `rate_limit.rs`) because
+//! `init` reads process-global env vars.
```
sveitser: Can we remove this now? I think we fixed this problem where we needed to read env vars in tests.

crates/espresso/telemetry/src/push_task.rs
```diff
@@ -0,0 +1,132 @@
...
+        .header(AUTHORIZATION, format!("Bearer {jwt}"))
+        .header(CONTENT_TYPE, "application/x-protobuf")
+        .header("Content-Encoding", "snappy")
+        .body(body)
+        .send()
+        .await;
+
+    match resp {
+        Ok(r) if r.status().is_success() => {},
+        Ok(r) if r.status().as_u16() == 429 => {
```
lukaszrzasik: It would be nice to use a self-describing const instead of the magic number `429`

crates/espresso/telemetry/src/lifecycle.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// Flush the push thread, then shut down the OTel logger provider.
+    /// Best-effort; failures are logged, never bubbled. The provider shutdown
+    /// can deadlock on a current-thread runtime, so it runs on a detached OS
+    /// thread bounded by [`LOGGER_SHUTDOWN_TIMEOUT`].
+    pub fn shutdown(self) {
+        if let Some(MetricsPushHandle { shutdown, thread }) = self.metrics_push {
+            let _ = shutdown.send(());
+            if thread.join().is_err() {
+                tracing::warn!("telemetry: metrics push thread panicked");
+            }
```
lukaszrzasik: Can we mirror `logger_provider` approach? `thread.join()` can potentially wait indefinitely and we're on the main thread here.

lukaszrzasik: Can we mirror the shutdown mechanism for the logger? We block the main thread here. It would be nice to have a timeouts here as well.

crates/espresso/telemetry/src/lifecycle.rs
```diff
@@ -0,0 +1,414 @@
...
+        let mut deferred = Vec::new();
+        self.attach_metrics_push_buffered(registry, &mut deferred);
+        for w in deferred {
+            tracing::warn!("{w}");
+        }
+    }
+
+    /// Like [`attach_metrics_push`], but buffers warnings into `deferred`
+    /// instead of emitting them, for use during [`init`] before a subscriber is
+    /// installed.
```
lukaszrzasik: Nit: the comment is a bit misleading in sense that the main logic lives here in `attach_metrics_push_buffered` and `attach_metrics_push` is just a wrapper that extracts the deferred warnings. Probably a leftover of the refactors.

crates/espresso/telemetry/src/lifecycle.rs
```diff
@@ -0,0 +1,414 @@
...
+//! externally supplied `prometheus::Registry` is scraped on a tokio interval and
+//! POSTed to `/api/v1/write` as snappy-compressed protobuf.
+//!
+//! Failure modes:
+//! - Proxy down: `BatchLogProcessor` queue fills (~2k records) then drops;
+//!   metrics push `warn!`s and retries next tick. Neither uses disk nor blocks
+//!   consensus.
+//! - JWT misconfig: [`init`] returns `Err`; the caller continues without telemetry.
+//! - Token TTL expiry mid-process is not handled; the six-month TTL outlasts
+//!   typical restart cadence.
```
lukaszrzasik, line 13: Not for this PR but we should plan for it: the token rotation will become a necessity when the espresso network enters a more stable period. Maybe a followup task?

lukaszrzasik (approved):
Just a few nits, nothing blocking.

crates/espresso/telemetry/src/lifecycle.rs
sveitser: I think it's very unlikely that we'll ever have 6 months without updates, but yes, in that case we can add rotation.

PR #4284, opened 2026-05-11
Add deferred hash_id migration

ss-es:


crates/espresso/node/api/migrations/postgres/V1302__hash_id_bigint_expand.sql
```diff
@@ -0,0 +1,52 @@
...
+-- The backfill sets id_big = id (1..2147483647), so the two ranges never overlap.
+-- GREATEST(..., 1) guards against an empty table where MAX(id) is NULL and COALESCE
+-- would return 0, which is below the sequence's MINVALUE of 1.
+SELECT setval('hash_id_big_seq', GREATEST(COALESCE((SELECT MAX(id) FROM hash), 0), 1));
+ALTER TABLE hash ALTER COLUMN id_big SET DEFAULT nextval('hash_id_big_seq');
+
+-- 2. Replace the exhausted INT sequence for id with a placeholder sequence.
+--    Placeholder values are always negative; new inserts that omit id receive a unique
+--    negative value. The PK constraint and its index are preserved.
+CREATE SEQUENCE hash_id_placeholder_seq AS INT
```
imabdulbasit, line 27: what is happening here? can you please explain a bit more

crates/espresso/node/src/persistence/migrations/hash_id_bigint.rs
```diff
@@ -0,0 +1,519 @@
...
+    fn order(&self) -> u32 {
+        1
+    }
+}
+
+#[async_trait]
+impl DeferredSchemaChange for CreateIndex {
+    async fn run(&self, storage: &SqlStorage) -> anyhow::Result<()> {
+        sqlx::query(
+            "CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS hash_id_big_unique ON hash(id_big)",
```
imabdulbasit, line 202: I think these indexes should be part of SQL migrations but also refinery doesn't support `CONCURRENTLY` so not sure if there is better option

crates/espresso/node/src/persistence/migrations/hash_id_bigint.rs
```diff
@@ -0,0 +1,519 @@
...
+            "reward_merkle_tree",
+            "reward_merkle_tree_v2",
+        ] {
+            sqlx::query(&format!("ALTER TABLE \"{table}\" DROP COLUMN hash_id"))
+                .execute(tx.as_mut())
+                .await?;
+        }
+
+        // Rename new columns to canonical names.
+        sqlx::query("ALTER TABLE hash RENAME COLUMN id_big TO id")
```
imabdulbasit, line 306: where are we setting the primary key for this column?

hotshot-query-service/src/data_source/storage/sql/queries/state.rs
```diff
@@ -73,27 +74,55 @@ where
...
-        // Hashmap will be used to get the hash value of the nodes children and the node itself.
-        let hashes = if !hash_ids.is_empty() {
-            let (query, sql) = build_where_in("SELECT id, value FROM hash", "id", hash_ids)?;
-            query
-                .query_as(&sql)
-                .fetch(self.as_mut())
-                .try_collect::<HashMap<i32, Vec<u8>>>()
-                .await?
+        // Find all the hash values and create a hashmap.
+        // During the migration window some rows have id_big (new) and some have only id (legacy).
```
imabdulbasit, line 90: I think this is very messy. Why don't we just use legacy hash_id and also insert to the new and old column until the backfilling is done? Also what is the plan for cleaning up all of this code after the backfilling is done?

crates/espresso/node/api/migrations/postgres/V1302__hash_id_bigint_expand.sql
```diff
@@ -0,0 +1,52 @@
...
+ALTER TABLE hash ALTER COLUMN id SET DEFAULT nextval('hash_id_placeholder_seq');
+
+-- 3. Drop FK constraints and NOT NULL on hash_id so new rows can omit it.
+--    New merkle rows will store only hash_id_big; hash_id is left NULL until the backfill.
+ALTER TABLE fee_merkle_tree       DROP CONSTRAINT fee_merkle_tree_hash_id_fkey;
+ALTER TABLE block_merkle_tree     DROP CONSTRAINT block_merkle_tree_hash_id_fkey;
+ALTER TABLE reward_merkle_tree    DROP CONSTRAINT reward_merkle_tree_hash_id_fkey;
+ALTER TABLE reward_merkle_tree_v2 DROP CONSTRAINT reward_merkle_tree_v2_hash_id_fkey;
+
+ALTER TABLE fee_merkle_tree       ALTER COLUMN hash_id DROP NOT NULL;
```
imabdulbasit, line 43: why drop the `NOT NULL `constraint?

crates/espresso/node/src/persistence/migrations/hash_id_bigint.rs
```diff
@@ -0,0 +1,519 @@
...
+
+        // Re-add FK constraints pointing to the new PK.
+        for table in &[
+            "fee_merkle_tree",
+            "block_merkle_tree",
+            "reward_merkle_tree",
+            "reward_merkle_tree_v2",
+        ] {
+            sqlx::query(&format!(
+                "ALTER TABLE \"{table}\" ADD CONSTRAINT {table}_hash_id_big_fkey FOREIGN KEY \
```
imabdulbasit, line 275: same here. I don't like that these constraints are not part of SQL migrations

imabdulbasit: This is after backfilling so it would lock the table and take lot of time.

imabdulbasit (commented):
I will look at this PR in detail next week. However, I feel like all the schema changes, indexes, and constraints should be part of SQL migrations that refinery runs, because refinery already keeps the versions table and tracks the migrations that have been run. I do see the problem this PR is trying to solve, but I feel like it also introduces changes to the queries which is very messy. Also, if I am not mistaken, lot of schema changes are now executed as rust statements, which means the schemas between different nodes can be out of sync because we are not tracking the order, content of these statements

I wonder if this hash id migration could be solved by just creating new state tables and a new hash table. We could rename the old tables to something like `fee_merkle_tree_old` and rename the new tables to the original names, so there would be no need to change the queries. Then we could backfill the data from the old tables to the new ones in the background. The node might not build the merklized state loop correctly during the migration, but I think we only want consensus to progress so it should be fine.

ss-es:
Closed in favor of #4325

PR #4293, opened 2026-05-13
Reject all connections to self.

twittner:
Previously we filtered out connect attempts to self, however when a party is added we did not do the same check, which may lead to connect tasks that establish a connection to the same node. This is fixed by checking the target key in `spawn_connect` and *not* actually spawning the task if it is equal to the node's key.

In addition should we for whatever reason accept a connection from somewhere with the same key, we tell the remote to backoff essentially forever (`Duration::MAX`).

The hello protocol is improved by considering not just the remote's hello package but also our own, so that we only spawn a peer task if both are considered to be okay.

Finally we make sure that `parties` only contains remote peers.


lukaszrzasik (approved):
LGTM although I'm probably not the person to approve this.

bfish713 (approved):

PR #4296, opened 2026-05-13
fix(light-client): require LIGHT_CLIENT_DB_PATH for catchup queries

sveitser:
[test(light-client): regression for in-memory schema visibility](https://github.com/EspressoSystems/espresso-network/commit/d933b1b75c99e2775dcf0b16fccc51424b286536)

Demonstrates that the current `:memory:` fallback used when
`LIGHT_CLIENT_DB_PATH` is unset gives every SQLite connection its own
private database, so a multi-connection pool's schema migration is
invisible to all but the first connection.

- Route `connect()`'s in-memory branch through an `in_memory_uri()`
  helper (no behavior change; helper returns `:memory:`).
- Add a regression test that opens two raw connections to the URI and
  asserts shared state. It fails on this commit and will pass once the
  helper switches to a shared-cache URI in the next commit.
  
[fix(light-client): require LIGHT_CLIENT_DB_PATH for catchup queries](https://github.com/EspressoSystems/espresso-network/commit/aa059a2a637b3ed5e4beaf45a0839bd620fdf032)

- Bail at startup if the `query` API module is enabled without
  `LIGHT_CLIENT_DB_PATH`; without persistence, catchup state is lost on
  restart and the multi-connection pool migration would otherwise be
  invisible to all but one connection.
- Replace bare `:memory:` fallback with a shared-cache URI
  (`file:lc-mem-N?mode=memory&cache=shared`) generated by an internal
  `LightClientDb` enum, so multi-connection in-memory pools share one
  database. Makes the regression test added in the previous commit pass.
- Auto-create the parent directory for file-backed databases.
- Set `LIGHT_CLIENT_DB_PATH` on the docker-compose / process-compose
  nodes that run the query module but were missing it.


crates/espresso/node/src/run.rs
```diff
@@ -22,6 +22,16 @@ pub async fn main(migrated_envs: Vec<(&str, &str)>) -> anyhow::Result<()> {
...
 
+    if let Some(query) = &modules.query
+        && query.light_client_db.lc_path.is_none()
+    {
+        anyhow::bail!(
+            "LIGHT_CLIENT_DB_PATH must be set when the query module is enabled; without a path \
+             the light client cache lives in memory and is destroyed on restart, so catchup \
+             cannot persist"
+        );
+    }
```
jbearer: I don't think this is necessary. Catchup should work with an in-memory DB (as long as the in-memory DB is configured correctly, with the shared cache). It just might take longer to get going

sveitser: Ah ok let's remove it then.

jbearer (approved):

PR #4298, opened 2026-05-13
Check if epoch < first_epoch before triggering catchup

ss-es:
We had two methods that would trigger catchup in the Membership object for a given epoch. `membership_for_epoch` would check if the requested epoch was smaller than the first epoch before triggering catchup, while `stake_table_for_epoch` would not.

This PR fixes `stake_table_for_epoch`.


twittner (approved):
👍

PR #4304, opened 2026-05-14
versions: collapse post per-epoch reward upgrades into V6 (0.6)

imabdulbasit:
Collapses the previously separate `DA_UPGRADE_VERSION` (0.6), `VID2_UPGRADE_VERSION` (0.7), and `CLIQUENET_VERSION` / `NEW_PROTOCOL_VERSION` (0.8) into a single V6 version.


crates/espresso/types/src/v0/mod.rs
```diff
@@ -161,7 +161,7 @@ pub type EpochVersion = StaticVersion<0, 3>;
 pub type DrbAndHeaderUpgradeVersion = StaticVersion<0, 4>;
 pub type EpochRewardVersion = StaticVersion<0, 5>;
 pub type DaUpgradeVersion = StaticVersion<0, 6>;
-pub type Vid2UpgradeVersion = StaticVersion<0, 7>;
+pub type Vid2UpgradeVersion = StaticVersion<0, 6>;
```
sveitser: @imabdulbasit shouldn't we remove this for less confusion?

crates/versions/src/lib.rs
```diff
@@ -16,9 +16,9 @@ pub const DRB_AND_HEADER_UPGRADE_VERSION: Version = version(0, 4);
 pub const EPOCH_REWARD_VERSION: Version = version(0, 5);
 pub const DRB_FIX_VERSION: Version = version(0, 5);
 pub const DA_UPGRADE_VERSION: Version = version(0, 6);
-pub const VID2_UPGRADE_VERSION: Version = version(0, 7);
-pub const CLIQUENET_VERSION: Version = version(0, 8);
-pub const NEW_PROTOCOL_VERSION: Version = version(0, 8);
+pub const VID2_UPGRADE_VERSION: Version = version(0, 6);
+pub const CLIQUENET_VERSION: Version = version(0, 6);
```
sveitser: Also this?

sveitser (approved):
LGTM but shouldn't we delete type defs for versions that no longer exist?

PR #4313, opened 2026-05-18
 fix: Retain previous epoch trees for staking api startup

imabdulbasit:
reward trees were garbage collected up to `min(height - 1, finalized_hotshot_height)`deleting any tree at heights below the L1 finalized. This broke staking api service at startup. The service starts from the current epoch's start block and queries `reward-amounts  at at `epoch_start - 1` but with node height and finalized height >  epoch's start block the previous epoch's last block tree had already been garbage collected so the staking api service fails to load. We must retain at least one whole epoch of trees even when those heights are below the finalized cursor.


sveitser (approved):
Seems fine to me but should really have a regression test.

PR #4323, opened 2026-05-18
Remove immediate timer reset.

twittner:
Messages are reliably delivered until garbage collection. Additional retries may just fill up the outbound message buffers of peers.

Change `Timer` to fire only once to avoid repeated timeouts. Thus after a timer expires it needs to be reset to make it fire again.


twittner:
@claude: review once

bfish713 (approved):

PR #4325, opened 2026-05-18
Add hash-id migration (reworked)

ss-es:
This PR:

- Adds a `DataBackfill` trait for asynchronous, batched data migrations
- Adds a `MigrationRegistry` to collect, organize and run these data migrations
- Creates new _bigint tables for the hash and fee/block merkle tree tables. In postgres this is a brand new table, while in sqlite this is a simple rename.
- The reward and reward v2 merkle tree tables are dropped, since they are no longer in use. Some associated migration code and tests have consequently been deleted.


crates/espresso/node/api/migrations/postgres/V1501__hash_bigint_expand.sql
```diff
@@ -0,0 +1,56 @@
+-- Expand phase: rename old tables to _legacy, recreate with BIGINT keys.
+-- This is an online schema change; the contract phase (dropping _legacy tables)
+-- is a separate follow-up migration.
+
+-- Step 1: rename tables that have FK deps on hash first.
+-- reward_merkle_tree and reward_merkle_tree_v2 are unused (always empty); drop them.
+ALTER TABLE fee_merkle_tree   RENAME TO fee_merkle_tree_legacy;
+ALTER TABLE block_merkle_tree RENAME TO block_merkle_tree_legacy;
+DROP TABLE reward_merkle_tree;
+DROP TABLE reward_merkle_tree_v2;
```
ss-es, line 9: these tables are now unused and should be empty on both decaf and mainnet

crates/espresso/node/api/migrations/sqlite/V1501__hash_bigint_expand.sql
```diff
@@ -0,0 +1,9 @@
+-- Rename tables to *_bigint for consistency with the postgres schema.
+-- SQLite integers are already 64-bit so this is a naming-only change.
+ALTER TABLE hash              RENAME TO hash_bigint;
+ALTER TABLE fee_merkle_tree   RENAME TO fee_merkle_tree_bigint;
+ALTER TABLE block_merkle_tree RENAME TO block_merkle_tree_bigint;
+
+-- Drop reward merkle tree tables — unused and always empty across all deployments.
+DROP TABLE IF EXISTS reward_merkle_tree;
+DROP TABLE IF EXISTS reward_merkle_tree_v2;
```
ss-es, line 9: these tables should now be unused on both decaf and mainnet

crates/espresso/node/src/api/sql.rs
```diff
@@ -1862,396 +1850,6 @@ mod tests {
...
-            tx,
-            proofs_and_paths,
-            block_height,
-        )
-        .await
-        .expect("failed to batch insert proofs");
-    }
-
-    #[test_log::test(tokio::test(flavor = "multi_thread"))]
-    async fn test_reward_accounts_batch_insertion() {
```
ss-es, line 1898: as far as I can tell this test is just testing storing reward proofs with the old merkle tree storage mechanism

since the old reward merkle tree tables are now dropped, I think this test is not relevant anymore

crates/espresso/node/src/api.rs
```diff
@@ -6044,7 +6044,6 @@ mod test {
     }
 
     #[rstest]
-    #[case(POS_V3)]
```
ss-es, line 6050: POS_V3 is no longer running on decaf or mainnet, and we will never deploy this version again -- so the test case can be dropped

crates/espresso/node/src/state.rs
```diff
@@ -195,41 +191,6 @@ async fn store_state_update(
         .context("failed to store block merkle nodes")?;
     }
 
-    if version <= EPOCH_VERSION {
```
ss-es, line 198: `EPOCH_VERSION` is 0.3 so this code is no longer relevant

slow-tests/tests/state.rs
```diff
@@ -115,123 +99,3 @@ async fn slow_test_merklized_state_api() {
...
-                LookupResult::NotFound(proof) => proof,
-            };
-            let traversal_path = <RewardAccountV2 as ToTraversalPath<
-                { RewardMerkleTreeV2::ARITY },
-            >>::to_traversal_path(account, reward_tree.height());
-            (proof, traversal_path)
-        })
-        .collect();
-
-    UpdateStateData::<SeqTypes, RewardMerkleTreeV2, { RewardMerkleTreeV2::ARITY }>::insert_merkle_nodes_batch(
```
ss-es, line 183: I think the test here is fundamentally testing this call, which is no longer relevant since we don't use the old reward tree tables anymore

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -1321,202 +1323,6 @@ async fn prune_to_view(tx: &mut Transaction<Write>, view: u64) -> anyhow::Result
 
 #[async_trait]
 impl SequencerPersistence for Persistence {
-    async fn migrate_reward_merkle_tree_v2(&self) -> anyhow::Result<()> {
```
ss-es, line 1335: the migration has already been performed so this code is no longer needed

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,276 @@
...
+            }
+
+            batch_count += 1;
+
+            if done {
+                tracing::warn!(name, batches = batch_count, "deferred migration complete");
+                return true;
+            }
+
+            offset = next.unwrap() as u64;
```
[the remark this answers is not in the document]
ss-es: we probably should rewrite this to avoid the `unwrap`, I'll look into it

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -698,6 +692,14 @@ impl PersistenceOptions for Options {
         };
         persistence.migrate_quorum_proposal_leaf_hashes().await?;
         self.pool = Some(persistence.db.pool());
+
+        #[cfg(not(feature = "embedded-db"))]
+        {
+            let registry = super::migrations::hash_bigint_migrations();
+            let db = persistence.db.clone();
+            tokio::spawn(registry.run_all_migrations(db));
```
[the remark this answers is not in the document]
ss-es: I'm not sure this is terribly useful given how we log in the migration loops anyway, but I'm open to it -- will think about it

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,276 @@
...
+                },
+            };
+
+            let next = match m.run_batch(&mut tx, offset).await {
+                Ok(next) => next,
+                Err(e) => {
+                    tracing::error!(name, offset, "migration batch failed: {e:#}");
+                    return false;
+                },
+            };
```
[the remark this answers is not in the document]
ss-es: I disagree with this suggestion -- this kind of failure would require manual intervention to fix, and I think we should log repeatedly until it is fixed. besides which, the log frequency is so low I don't think there's any downside

crates/espresso/node/src/persistence/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+        let n = rows.len();
+
+        for (id, value) in rows {
+            sqlx::query(
+                "INSERT INTO hash_bigint (id, value) VALUES ($1, $2) ON CONFLICT DO NOTHING",
+            )
+            .bind(id as i64)
+            .bind(&value)
+            .execute(tx.as_mut())
+            .await?;
```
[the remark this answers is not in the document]
ss-es: [8e80bc7](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/8e80bc754a4a7ee7e8a3abf7ebf10e92ed1dc48c)

crates/espresso/node/src/persistence/migrations.rs
```diff
@@ -0,0 +1,141 @@
...
+        "hash_bigint_backfill_hash"
+    }
+
+    async fn run_batch(
+        &self,
+        tx: &mut Transaction<Write>,
+        offset: u64,
+    ) -> anyhow::Result<Option<u64>> {
+        let rows: Vec<(i32, Vec<u8>)> =
+            sqlx::query_as("SELECT id, value FROM hash ORDER BY id LIMIT $1 OFFSET $2")
```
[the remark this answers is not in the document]
ss-es: [66b4c55](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/66b4c55a36e5a2f150f9d54b93e0ec448271f7d7)

hotshot-query-service/src/migration.rs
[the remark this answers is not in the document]
ss-es: rewrote this part

ss-es:
@claude review

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,517 @@
...
+                tracing::warn!("all deferred migrations complete");
+                break;
+            }
+
+            tokio::time::sleep(RETRY_INTERVAL).await;
+        }
+    }
+
+    /// Run all batches for a single migration to completion. Returns `true` if the migration
+    /// finished successfully, `false` if it encountered an error.
```
sveitser: Add type instead of doc string?

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,517 @@
...
+    /// Number of rows to process per batch.
+    fn batch_size(&self) -> usize {
+        1_000
+    }
+
+    /// Number of batches between progress log lines.
+    fn log_frequency(&self) -> usize {
+        10
+    }
+
```
sveitser, line 31: I think we should add a global configurable parameter for sleep between statements. I'm just worried that a busy loop might put too much strain on the DB for some operators.

crates/espresso/node/api/migrations/postgres/V1501__hash_bigint_expand.sql
```diff
@@ -0,0 +1,47 @@
+-- Expand phase: create new BIGINT-keyed tables alongside the existing ones.
```
sveitser, line 1: The original PR had a doc with some explanations for why the DB migrations are done like this. Currently I don't think there's anything like that. I think we should have such a document and then put into the AGENTS.md that migrations must follow this pattern to not affect consensus and link to the document. Or alternatively have this documentation inline in rust.

Right now if someone I feel if wants to add a migration who didn't see this change, the past of least resistance is still to just do what we previously did. We may only notice very late, like in decaf that we actually added another blocking migration.

hotshot-query-service/src/data_source/storage/sql/queries/state.rs
```diff
@@ -366,16 +456,17 @@ pub(crate) fn build_hash_batch_insert(
...
     if hashes.is_empty() {
         return Ok(HashMap::new());
     }
 
-    // Use UNNEST-based batch insert (more efficient and avoids parameter limits)
-    let sql = "INSERT INTO hash(value) SELECT * FROM UNNEST($1::bytea[]) ON CONFLICT (value) DO \
-               UPDATE SET value = EXCLUDED.value RETURNING value, id";
+    // Use UNNEST-based batch insert (more efficient and avoids parameter limits).
+    // Cast id to BIGINT in RETURNING so the result maps directly to i64.
+    let sql = "INSERT INTO hash_bigint(value) SELECT * FROM UNNEST($1::bytea[]) ON CONFLICT \
```
sveitser, line 466: Do we have the offset anywhere to ensure we are not writing live data with IDs that the data uses that we migrate? It's not clear to me.

ss-es: I believe this does that: https://github.com/EspressoSystems/espresso-network/pull/4325/changes#diff-95639b6dfbb3dd0860e440f2fe9afa0a9a0fadb2974bf282a062343a16b220e6R19-R22

my understanding is that a (value) insert will take the next available number in sequence starting at the maximum hash_id on the old table, whereas a (value, id) insert will just write directly to that id

hotshot-query-service/src/migration.rs
ss-es: set it to a default 50ms between batches, configurable by env var: [6a7db49](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/6a7db4915b95feb4df0c36a129057957bf11bd29)

hotshot-query-service/src/migration.rs
ss-es: not sure what you meant here

ss-es:
@sveitser we discussed the potential (significant) increase in memory during the migration here, and I think I've mitigated most of it in this commit: [a9563bc](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/a9563bcf65093c6f7126ade09dae6168e782c488)

the change here deletes rows from the old fee/block merkle tree tables as we fill in the new table -- this is destructive, but I believe all of this is reconstructable by restarting the merklized state loop from scratch anyway. and I believe this should prevent us seeing any noticeable increase in storage from the new versions of these tables

the hash table is a little different, because (I think) we may still need entries in the old hash table while the fee/block merkle tree migrations are in progress. fortunately this is the smallest table of the 3, and I've added another migration to delete the data in it after all the other migrations have finished (which should enable resizing the database in aws etc)

unrelatedly, I've also added an endpoint to monitor migration progress: [77c5419](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/77c5419f5172faa9965efd09bf1fa3b3df9c1d0b)

crates/espresso/node/api/migrations/postgres/V1501__hash_bigint_expand.sql
ss-es: [a41f2cf](https://github.com/EspressoSystems/espresso-network/pull/4325/commits/a41f2cfd2ecf0a0cc5326e8231cf466d83c658db)

hotshot-query-service/src/migration.rs
sveitser: Instead of returning a boolean.

crates/espresso/node/src/api/data_source.rs
```diff
@@ -388,12 +388,26 @@ pub struct TableSize {
     pub total_size_bytes: Option<i64>,
 }
 
+/// Status of a single deferred background migration.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct MigrationStatus {
```
sveitser, line 393: nice

AGENTS.md
```diff
@@ -315,6 +315,50 @@ For node operator details, see https://docs.espressosys.com/network/guides/node-
 4. Update `SequencerPersistence` trait implementation for all backends
 5. Test: `cargo test -p espresso-node persistence`
 
+### ⚠️ Refinery Migrations Block Startup — Avoid Large Data Operations
```
sveitser: Suggest to put in doc/agents/... and mention here so it's not always pulled into context.

I started this in #4339

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,532 @@
...
+        &[]
+    }
+
+    /// Number of rows to process per batch.
+    fn batch_size(&self) -> usize {
+        1_000
+    }
+
+    /// Number of batches between progress log lines.
+    fn log_frequency(&self) -> usize {
```
sveitser:
```suggestion
    fn log_interval(&self) -> usize {
```

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,532 @@
...
+        self
+    }
+
+    /// Validate the registry.
+    ///
+    /// Checks that:
+    /// - All migration names are unique.
+    /// - Every name listed in `requires()` refers to a migration that appears earlier in the list.
+    pub fn validate(&self) -> anyhow::Result<()> {
+        let mut seen: std::collections::HashSet<&'static str> = std::collections::HashSet::new();
```
sveitser: nit: reads a bit verbose

hotshot-query-service/src/migration.rs
```diff
@@ -0,0 +1,532 @@
...
+                break;
+            }
+
+            tokio::time::sleep(RETRY_INTERVAL).await;
+        }
+    }
+
+    /// Run all batches for a single migration to completion. Returns `true` if the migration
+    /// finished successfully, `false` if it encountered an error.
+    async fn run_migration(db: &SqlStorage, m: &dyn DataBackfill) -> bool {
```
sveitser: return Result?

hotshot-query-service/src/sqlite_options.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// Default [`SqliteConnectOptions`] for SQLite databases in this workspace.
+///
+/// WAL journaling is the load-bearing choice: under the default rollback journal, any writer
+/// holds an exclusive lock that blocks readers on other connections, which produces spurious
+/// `SQLITE_BUSY` ("database is locked") errors under concurrent access. WAL lets readers and
+/// the single writer proceed in parallel.
+///
+/// Callers add `.filename(...)` (or use `:memory:`) on top.
+pub fn sqlite_options() -> SqliteConnectOptions {
```
sveitser, line 19: @ss-es i'm a bit confused why this is in the diff? shouldn't this be in a separate PR?

sveitser (approved):
lgtm, some nits

hotshot-query-service/src/sqlite_options.rs
ss-es: I was making all these changes as fixes on #4384 during testing, so it was easier to just pull them all over into this PR than to disentangle them.

PR #4326, opened 2026-05-18
test(binary-upgrade): add scenario matrix

sveitser:
- Add `--scenario` flag with six choices: vanilla, catchup-from-old-{fs,pg}, catchup-from-new-{fs,pg}, first-start
- Catchup scenarios wipe a node's storage (FS volume or postgres data) and restart it on UPGRADE with `ESPRESSO_NODE_CONFIG_PEERS` so it bootstraps from a still-running peer instead of the orchestrator
- `first-start` brings up a fresh espresso-node-5 on BASE_TAG against the upgraded network via `/v0/config/hotshot`, guarding against config-endpoint deserialization regressions
- Name FS + postgres data volumes explicitly so per-node wipes are deterministic across compose versions
- Run all scenarios as a CI matrix in binary-upgrade-test.yml and the PR / non-PR jobs in build.yml; per-job timeout bumped to 2400s


binary-upgrade-tests/README.md
```diff
@@ -19,23 +19,40 @@ tag, asserting the network keeps producing and serving blocks.
...
+| ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `vanilla`         | Roll each node 0..4 to UPGRADE one at a time, then bulk-upgrade the rest.                                                                                                                                                         |
+| `new-from-old-fs` | Roll node 4 (FS-backed) to UPGRADE, wipe its storage, restart with `ESPRESSO_NODE_CONFIG_PEERS`, wait for it to catch up while the other 4 nodes are still on BASE. Then finish vanilla. Tests UPGRADE binary against BASE peers. |
+| `new-from-old-pg` | Same shape but wipes node 1 (postgres-backed) plus its `espresso-node-db-1` data.                                                                                                                                                 |
```
ss-es, line 39: is there a reason we need to wipe storage? I feel like the postgres migration between versions is actually useful to test (and we would rarely be wiping storage between deployments in production)

binary-upgrade-tests/run.py
```diff
@@ -380,6 +821,20 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def run_scenario(compose: Compose, config: Config, scenario: Scenario) -> None:
+    match scenario:
+        case Scenario.VANILLA:
+            compose.vanilla(config)
+        case Scenario.NEW_FROM_OLD_FS:
```
ss-es: this feels a bit ad-hoc to me. I think it's fine but I wonder if there's a way we could parametrize these tests to make it easier to add/modify tests in the future?

I haven't fully thought about it, but couldn't we give each node a list of actions to perform in given views (e.g. node_0: (view 0, start on <tag>), (view 20, shut down), (view 20, clear storage), (view 50, start on tag))

binary-upgrade-tests/README.md
sveitser: This is a special case where we wipe. All other upgrades don't wipe.

binary-upgrade-tests/run.py
sveitser: For now this covers the scenarios that were mentioned on slack. I think it's okay since it's not much code and claude can add new scenarios rather easily. I'm not exactly sure how to best parametrize things for future test scenarios. But I agree we should do that eventually.

sveitser: Rewritten now.

binary-upgrade-tests/run.py
```diff
@@ -56,6 +70,158 @@
...
+        SmokeTest(tag_source="base"),
+        Roll(WIPE_FS_NODE),
+        Wipe(WIPE_FS_NODE, backend="fs"),
+        Roll(0),
+        Roll(1),
+        Roll(2),
+        Roll(3),
+        UpgradeSupportServices(),
+        AssertImagesUpgraded(),
+        SmokeTest(),
```
ss-es, line 149: I didn't read the rest of the python code in much depth, but this at least looks very nice

ss-es (approved):

PR #4333, opened 2026-05-20
[new-protocol] store genesis vid in query service

imabdulbasit:


hotshot-query-service/src/data_source/update.rs
```diff
@@ -281,7 +281,25 @@ where
                             )),
                             Some(VidShare::V2(share.share.clone())),
                         ),
-                        Some(_) | None => (None, None),
+                        Some(_) => (None, None),
+                        None => {
+                            if leaf.view_number().u64() == 0 {
```
twittner:
```suggestion
                            if leaf.view_number() == ViewNumber::genesis() {
```

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4333/changes/e3cb1b7f1d0cf33bad2c404360bc4a2267bfcf81

twittner (approved):

PR #4336, opened 2026-05-20
Use separate tasks for sending and receiving.

twittner:
Replaces the single-task `select!` event loop in `peer.rs` with a three-task topology -- the peer task owning retry/dequeue/GC/clock, plus a `Sender` and `Receiver` each owning their respective TCP stream half and running in separate tasks each. Noise state is `Arc<StatelessTransportState>` with per-direction `send_nonce`/`recv_nonce` owned by their respective tasks, so encryption and decryption run on different runtime workers without a mutex, i.e. we do not need to share the state via `Arc<Mutex<TransportState>>`.

- `Connection` carries explicit `send_nonce` / `recv_nonce`; `Peer` no longer owns the `Connection`.
- `Peer::start(conn, cancel)` takes the connection and cancellation token as arguments; `set_connection` and the internal cancel field are gone.
- Peer task <-> `Sender` task /`Receiver` task communication uses three channels (outbound messages, outbound ACKs, inbound ACKs); the previous local `VecDeque<Ack>` is replaced by these.
- `Countdown` is now `Clone` so the `Sender` can `start()` it on the last frame of a retry-policy send and the `Receiver` can `stop()` it on incoming bytes.

The design, which goes back a little to the original implementation, allows send and receive operations to run in parallel on different cores which improves full duplex throughput.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -475,3 +475,7 @@ opt-level = 3
 inherits = "test"
 strip = true
 debug = "none"
+
+[profile.bench]
+lto = true
+codegen-units = 1
--- a/crates/cliquenet/benches/bench1.rs
+++ b/crates/cliquenet/benches/bench1.rs
@@ -3,7 +3,7 @@ use std::{
 };
 
 use cliquenet::{
-    Config, Network, RetryPolicy, SendAction, SendCommand, Slot,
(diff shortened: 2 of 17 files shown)


twittner:
@claude: review once

twittner:
@claude: review once

PR #4346, opened 2026-05-21
fix(light-client): persist used_x25519_keys in stake table storage

sveitser:
- Add stake_table_x25519_key table to the schema, mirroring BLS/Schnorr
- Insert/load used_x25519_keys in SqliteStorage::{insert,lower_bound}
- Extend StakeTableState::new to take used_x25519_keys
- Add FromStr impl on hotshot_types::x25519::PublicKey (TaggedBase64)

Without this the LC reconstructs the stake table from an empty x25519 set on DB resume, so commit() diverges from the proposer's cumulative hash once any V3 events have been processed.

### DIFF
--- a/light-client/migrations/01_init_schema.sql
+++ b/light-client/migrations/01_init_schema.sql
@@ -53,6 +53,16 @@ CREATE TABLE stake_table_schnorr_key (
 );
 CREATE INDEX stake_table_schnorr_key_epoch ON stake_table_schnorr_key (epoch);
 
+-- This table tracks used x25519 keys for each stake table.
+--
+-- It is cumulative, meaning each key is only added once, with the epoch number where it is first
+-- used, but it belongs to the `used_x25519_keys` set in each stake table after that epoch as well.
+CREATE TABLE stake_table_x25519_key (
+    key   TEXT   PRIMARY KEY,
+    epoch BIGINT NOT NULL
+);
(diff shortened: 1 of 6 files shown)


light-client/migrations/01_init_schema.sql
```diff
@@ -53,6 +53,16 @@ CREATE TABLE stake_table_schnorr_key (
 );
 CREATE INDEX stake_table_schnorr_key_epoch ON stake_table_schnorr_key (epoch);
 
+-- This table tracks used x25519 keys for each stake table.
+--
+-- It is cumulative, meaning each key is only added once, with the epoch number where it is first
+-- used, but it belongs to the `used_x25519_keys` set in each stake table after that epoch as well.
+CREATE TABLE stake_table_x25519_key (
```
imabdulbasit, line 60: can we create a new migration file for this? It wouldn't work with existing deployment

sveitser: @imabdulbasit since we only have ephemeral testnet deployments and the light client state can be easily wiped (or we can use in-memory wipe on restart) i think it's not worth it

imabdulbasit: yeah I was thinking that this migration would not work on existing deployment on milk. But we have to reset the storage anyway because otherwise we would compute wrong hash

imabdulbasit (approved):

PR #4401, opened 2026-05-27
[Fast Finality] Generate Decide events for Query Service off Critical Path

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/espresso/node/src/context.rs
+++ b/crates/espresso/node/src/context.rs
@@ -735,6 +735,12 @@ async fn process_decided_events_task<P, C>(
                 if let Some(v) = processed {
                     last_processed = last_processed.max(v.u64());
                 }
+                // reset latest if we have processed all the decided leaves
+                if let Some((view, _)) = latest.clone()
+                    && last_processed >= view.u64()
+                {
+                    latest = None;
+                }
                 metrics.last_processed.set(last_processed as usize);
                 metrics
                     .backlog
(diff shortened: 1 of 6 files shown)


crates/espresso/node/src/context.rs
```diff
@@ -602,6 +662,99 @@ async fn handle_events<N, P>(
...
+    let mut last_processed = anchor_view.map(|v| v.u64()).unwrap_or(0);
+
+    // Process leaves persisted before a previous shutdown but not yet handled.
+    if let Some(view) = anchor_view {
+        match persistence
+            .process_decided_events(view, None, consumer.as_ref())
+            .await
+        {
+            Ok(processed) => {
+                if let Some(v) = processed {
```
imabdulbasit, line 692: I don't think this is needed because last_processed is set to anchor_view at the start of the function

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -1606,23 +1617,25 @@ impl SequencerPersistence for Persistence {
...
 
-        // Garbage collect data which was not included in any decide event, but which at this point
-        // is old enough to just forget about.
+        // Best-effort GC of data not included in any decide event; runs again at the next decide.
         if let Err(err) = self.prune(view).await {
             tracing::warn!(?view, "pruning failed: {err:#}");
         }
 
-        Ok(())
+        self.load_processed_view().await
```
imabdulbasit, line 1638: I think this call can be avoided if `generate_decide_events() `returns the `load_processed_view`.

imabdulbasit (approved):

PR #4404, opened 2026-05-28
Add optional RLP serialization to light client genesis

jbearer:
The Espresso stack will use this

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,13 @@ authors = ["Espresso Systems <hello@espressosys.com>"]
 edition = "2024"
 
 [workspace.dependencies]
-aide = { version = "0.15.1", features = ["axum", "axum-json", "scalar", "swagger", "redoc"] }
+aide = { version = "0.15.1", features = [
+    "axum",
+    "axum-json",
+    "scalar",
+    "swagger",
+    "redoc",
+] }
 alloy = { version = "1.0", features = [
     "eips",
     "provider-ws",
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -124,6 +130,7 @@ alloy = { version = "1.0", features = [
     "signer-ledger",
 ] }
 alloy-compat = { path = "alloy-compat" }
+alloy-rlp = "0.3"
```
sveitser, line 133: Why not use the `rlp` feature of alloy?

I also wonder why we need to declare a new cargo feature. I think for all our uses of alloy alloy-rlp is already included transitively so we already compile it. The extra amount of code to compile seems small so I feel like it doesn't justify having a feature. Maybe I'm missing something.

sveitser (approved):
LGTM one question about the cargo feature.

Cargo.toml
jbearer: This dependency is needed because even if you import `RlpEncodable`/`RlpDecodable` from `alloy::rlp`, the generated code from the derive macros explicitly references `alloy_rlp`. So the crate needs to be in scope, not just transitively included, for the derived impls to work.

As for why we would conditionally compile this whole feature, we don't really have to. It just felt a bit weird to me to add this RLP serialization for just a couple of types, when we don't use it anywhere else, just for the purposes of a single new downstream crate. Felt like collecting that ad hoc functionality into a feature flag made sense

PR #4407, opened 2026-05-29
perf(catchup): single-flight to stop block_merkle_tree query stampede

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

  Under high catchup concurrency, many peers issued identical block_merkle_tree
  frontier lookups simultaneously, piling up on Postgres IPC:BufferIO waits (6-10
  min queries). Add a SingleFlight utility (hotshot-query-service/src/util) and use
  it in CatchupStorage::get_frontier so N concurrent callers for the same height
  share one DB query and result.


sveitser (approved):
This looks fine to me. Although not sure I would spot a problem if there was one.

PR #4409, opened 2026-05-29
chore: add retries limit for fetching the stake table

mrain:
Closes #<ISSUE_NUMBER>


sveitser (approved):
@mrain lgtm but expected and handled retry logs should not be Error. Error level is for things that can't be handled, like exhausting all retries.

PR #4410, opened 2026-05-29
Add staking-ui-service to monorepo

ss-es:


sveitser:
@ss-es @imabdulbasit @jbearer @lukaszrzasik it seems the staking-ui-service uses slightly different code for decaf and mainnet using long running branches. How necessary is this still? Can we unite the code to make things easier going forward?

For CI I suggest we combine it with the existing builds in the repo wherever that makes sense. Nextest should build and run all the rust tests already.

PR #4415, opened 2026-06-01
[new-protocol] garbage collect consensus data for views less than decided view

imabdulbasit:


bfish713 (changes requested):
We should completely remove the garbage collection certificate if we are doing this.  I think this will trigger voting for a checkpoint every decide, which is an extra cert per view we don't actually need.

bfish713 (approved):

PR #4417, opened 2026-06-01
Add metrics to coordinator.

twittner:


crates/hotshot/new-protocol/src/coordinator/metrics.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl Drop for Measurement {
+    fn drop(&mut self) {
+        self.hist.add_point(self.start.elapsed().as_secs_f64());
+    }
+}
+
+pub fn finish_measurement(_: Option<Measurement>) {}
```
bfish713, line 74: This is just supposed to be a way to drop the `Measurement` right?

bfish713 (approved):

crates/hotshot/new-protocol/src/coordinator/metrics.rs
twittner: Yes, I thought it reads better.

PR #4418, opened 2026-06-02
fix(libp2p): add network discriminators

sveitser:
- Add `network_discriminator: Option<String>` to `NetworkNodeConfig`.
  `None` keeps mainnet's libp2p protocol strings byte-identical to
  today; `Some(d)` appends `/{d}` to gossipsub's `protocol_id_prefix`,
  the kademlia `StreamProtocol`, and the request-response direct
  message `StreamProtocol`. Identify's `protocol_version` is left
  untouched since it does not gate substream negotiation.
- Plumb the discriminator through `Libp2pNetwork::from_config`.
- At the espresso-node call site, derive the discriminator from
  `genesis.chain_config.chain_id`: `None` when chain_id is 1
  (mainnet), else `Some(chain_id.to_string())`. Decaf and other
  non-mainnet networks now run on distinct libp2p protocols and
  cannot merge with mainnet at the gossipsub or kademlia layers.
- The snapshot test added in the previous commit confirms the
  mainnet path is unchanged.


crates/espresso/node/src/lib.rs
```diff
@@ -760,6 +760,9 @@ where
 
     let combined_network = {
         info!("Initializing Libp2p network");
+        // Mainnet keeps today's libp2p protocol strings byte-identical.
+        let chain_id = genesis.chain_config.chain_id;
+        let network_discriminator = (chain_id.0 != U256::ONE).then(|| chain_id.to_string());
```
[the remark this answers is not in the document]
sveitser: /gemini check

crates/espresso/node/src/lib.rs
```diff
@@ -760,6 +760,9 @@ where
 
     let combined_network = {
         info!("Initializing Libp2p network");
+        // Mainnet keeps today's libp2p protocol strings byte-identical.
+        let chain_id = genesis.chain_config.chain_id;
+        let network_discriminator = (chain_id.0 != U256::ONE).then_some(chain_id.0);
```
lukaszrzasik: Nit: Could we use a self-describing const instead of the magic `U256::ONE`?

lukaszrzasik (approved):

crates/espresso/node/src/lib.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/4418/commits/59ed8f601d303660cc8d2360fd6bc8d2abf2055d

PR #4419, opened 2026-06-02
fix: reject short trailers in cliquenet parse 

sveitser:
- malicious peer can send a short message ending in the `STD` type byte
  and a trailer length of 16
- `from_bytes` slices `bytes[len - 18..]` which underflows usize for
  `len < 18` and panics on attacker-controlled input
- guard the `STD` and `NO_ACK` branches with explicit length checks before
  slicing
- prevents usize underflow and out-of-bounds slice panic on short,
  attacker-controlled messages

### DIFF
--- a/crates/cliquenet/src/msg/trailer.rs
+++ b/crates/cliquenet/src/msg/trailer.rs
@@ -178,4 +178,10 @@ mod tests {
         let mut b = Bytes::copy_from_slice(&[STD, 16]);
         assert_eq!(Trailer::from_bytes(&mut b), None);
     }
+
+    #[test]
+    fn regression_no_ack_trailer_short_input_does_not_panic() {
+        let mut b = Bytes::copy_from_slice(&[NO_ACK, 8]);
+        assert_eq!(Trailer::from_bytes(&mut b), None);
+    }
 }
--- a/crates/cliquenet/src/msg/trailer.rs
+++ b/crates/cliquenet/src/msg/trailer.rs
@@ -33,10 +33,7 @@ impl Trailer {
         let trailer_typ = bytes[len - 2];
         match trailer_typ {
             STD => {
-                if trailer_len != 16 {
-                    return None;
-                }
-                if len < 18 {
+                if trailer_len != 16 || len < 18 {
                     return None;
                 }
                 let id = u64::from_be_bytes(bytes[len - 10..len - 2].try_into().ok()?);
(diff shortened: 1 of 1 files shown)


crates/cliquenet/src/msg/trailer.rs
```diff
@@ -48,6 +51,9 @@ impl Trailer {
                 if trailer_len != 8 {
                     return None;
                 }
+                if len < 10 {
+                    return None;
+                }
```
twittner:
```suggestion
                if trailer_len != 8 || len < 10 {
                    return None;
                }
```

crates/cliquenet/src/msg/trailer.rs
```diff
@@ -36,6 +36,9 @@ impl Trailer {
                 if trailer_len != 16 {
                     return None;
                 }
+                if len < 18 {
+                    return None;
+                }
```
twittner:
```suggestion
                if trailer_len != 16 || len < 18 {
                    return None;
                }
```

crates/cliquenet/src/msg/trailer.rs
```diff
@@ -124,11 +130,58 @@ mod tests {
...
+
     quickcheck! {
         fn prop_to_bytes_from_bytes_id(t1: Trailer) -> bool {
             let mut b = Bytes::copy_from_slice(t1.to_bytes().as_ref());
             let t2 = Trailer::from_bytes(&mut b);
             Some(t1) == t2
         }
+
+        // `from_bytes` parses untrusted bytes, so it must never panic.
+        fn prop_from_bytes_arbitrary_does_not_panic(bytes: Vec<u8>) -> bool {
```
twittner:
```suggestion
        fn prop_from_bytes_arbitrary_does_not_panic(bytes: Vec<u8>) -> bool {
```

crates/cliquenet/src/msg/trailer.rs
```diff
@@ -124,11 +130,58 @@ mod tests {
...
+            let mut b = Bytes::copy_from_slice(&bytes);
+            let _ = Trailer::from_bytes(&mut b);
+            true
+        }
+
+        // Same, with trailer-shaped inputs that hit the slicing paths.
+        fn prop_from_bytes_trailer_like_does_not_panic(t: TrailerLike) -> bool {
```
twittner:
```suggestion
        fn prop_from_bytes_trailer_like_does_not_panic(t: TrailerLike) -> bool {
```

twittner (approved):
Good catch!

sveitser:
@twittner @bfish713 should  we backport this to the ff release branch? In case you will branch off main later again anyway I think we don't need the backport

PR #4427, opened 2026-06-03
Introduce `GcScope`.

twittner:
1. **Cliquenet no longer suppresses inbound messages below its own GC threshold**. The slot is a local value and no assumptions should be made that peers agree on the value.
2. **Revert the view number of external messages to 1 and ignore it**. Instead send external messages with the current view. The changes in 1. ensure they are delivered when received and this change ensures they are sent because they do not sit below the GC threshold.
3. **Add `GcScope`**. Some collections can (and should) be pruned on local view changes, others have to be kept around until a view has been decided. This enum encodes the distinction and `Coordinator::gc` and `Consensus::gc` are changed to clean as soon as possible.
4. **Always use the current view number for sending**. This ensures that every message we intend to send is not ignored because it sits below our own GC threshold which is especially relevant for replying to received messages as the remote's message view may be below our own.

### DIFF
--- a/crates/hotshot/new-protocol/src/coordinator.rs
+++ b/crates/hotshot/new-protocol/src/coordinator.rs
@@ -615,7 +615,10 @@ where
                         ProposalMessage::validated(proposal.clone()),
                     )),
                 };
-                if let Err(err) = self.network.broadcast(self.view(&message), &message) {
+                if let Err(err) = self
+                    .network
+                    .broadcast(self.consensus.current_view(), &message)
+                {
                     let err = CoordinatorError::from(err).context("proposal broadcast");
                     if err.severity == Severity::Critical {
                         return Err(err);
@@ -634,7 +637,7 @@ where
                     };
                     if let Err(err) =
                         self.network
-                            .unicast(self.view(&message), &recipient, &message)
+                            .unicast(self.consensus.current_view(), &recipient, &message)
                     {
                         let err = CoordinatorError::from(err).context("vid share unicast");
                         if err.severity == Severity::Critical {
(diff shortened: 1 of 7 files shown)


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -1211,6 +1218,7 @@ where
                 recipient,
                 respond,
             } => {
+                let view = max(view, self.gc_view);
```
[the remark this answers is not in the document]
twittner: External messages pretend to be in view 1. When GC moved past that point, external messages are no longer sent or received, which is why we increase the view number.

bfish713 (approved):

twittner:
@claude review

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -1366,25 +1366,46 @@ where
...
-        self.cached_validated_proposals = self.cached_validated_proposals.split_off(&view);
-        self.cached_vid_shares = self.cached_vid_shares.split_off(&view);
+    fn gc(&mut self, epoch: EpochNumber, scope: GcScope) -> Result<(), CoordinatorError> {
+        self.consensus.gc(scope);
+        match scope {
+            GcScope::Local(view) => {
+                self.block_builder.gc(view);
+                self.cached_validated_proposals = self.cached_validated_proposals.split_off(&view);
+                self.cached_vid_shares = self.cached_vid_shares.split_off(&view);
+                self.network.gc(view.saturating_sub(1).into())?;
```
imabdulbasit, line 1387: why is this view - 1?

twittner: When we *enter* a new view, we do not want to GC enqueued messages for the previous view yet.

imabdulbasit (approved):

bfish713 (approved):

PR #4450, opened 2026-06-08
Retry harness for Postgres serialization error

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* Adds `SerializableRetry` trait implemented by `FileSystemStorage` and `SqlStorage`. The trait's purpose is to run a transaction with the retry harness. The user provides the code that creates a transaction and performs some queries. If the transaction fails, it is automatically retried with the harness. `FileSystemStorage` implementation is a simple pass-through.
* Adds `SerializableRetryOptions` which provides options for fine-tuning the retry config. Noteworthy: `ESPRESSO_NODE_SERIALIZABLE_PG_STAT_DIAG` flag controls whether to print logs helping to debug the serialization errors.
* Adds two helper macros: `serializable_retry` simply calls `serializable_retry` method passing the function name for debugging. The other macro `function_name` retrieves the current's function unqualified name.
* Boilerplate: moving all the `write` and `read` calls into the retry harness (`serializable_retry` macro). One noteworthy place that is a bit more complex is the `generate_decide_events` method.

### This PR does not:

The following are not included in the PR but I'm planning to add them in the follow-up changes:
* Make `write` and `read` methods private to make them impossible to call outside the retry harness.
* Unify error types for the retry harness. Currently different parts of code running under the retry harness return different types of errors. This is why `is_serialization_conflict_err` needs to rely on string matching to recognize the serialization error because the postgres error type is lost. If we unify the errors types to `anyhow::Error` which wraps the underlying errors without overwriting the types, we can switch to strongly-typed check.


lukaszrzasik:
@claude review again

lukaszrzasik:
@claude try again

lukaszrzasik:
@claude try reviewing again

hotshot-query-service/src/data_source/storage.rs
```diff
@@ -89,6 +89,61 @@ use crate::{
...
+/// `40001` ("could not serialize access"). These aborts are expected and safe to retry from
+/// scratch, so every read or write should run inside this harness rather than calling
+/// [`read`](VersionedDataSource::read) / [`write`](VersionedDataSource::write) directly.
+///
+/// `f` is re-invoked from scratch on each attempt, so it must (re)open its own transaction.
+/// Backends that cannot produce serialization conflicts (e.g. the file system) implement this as a
+/// single pass-through call.
+///
+/// The operation is generic over its error type `E` (e.g. [`QueryError`], `anyhow::Error`, or an
+/// explorer error); conflicts are detected from the error's [`Display`](std::fmt::Display) output.
```
ss-es, line 104: I'm somewhat apprehensive about this approach. Why not just retry unconditionally rather than try to read the error?

hotshot-query-service/src/data_source/storage.rs
```diff
@@ -89,6 +89,61 @@ use crate::{
...
+            ::std::any::type_name::<T>()
+        }
+        let full: &'static str = type_name_of(__f);
+        // Strip the trailing "::__f" (5 bytes) and take the last "::" segment.
+        let trimmed: &'static str = &full[..full.len() - 5];
+        match trimmed.rfind("::") {
+            Some(idx) => &trimmed[idx + 2..],
+            None => trimmed,
+        }
+    }};
```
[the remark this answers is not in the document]
ss-es: this is worrying to me. maybe on the plus side: do I understand correctly that this is a compile-time panic (if it panics)?

also could we not achieve something like this more simply with e.g. `#[instrument]` annotations?

hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -1950,47 +1960,81 @@ impl<Types, S, P> NodeDataSource<Types> for FetchingDataSource<Types, S, P>
...
+        let start = match range.start_bound() {
+            Bound::Included(&n) => Bound::Included(n),
+            Bound::Excluded(&n) => Bound::Excluded(n),
+            Bound::Unbounded => Bound::Unbounded,
+        };
+        let end = match range.end_bound() {
+            Bound::Included(&n) => Bound::Included(n),
+            Bound::Excluded(&n) => Bound::Excluded(n),
+            Bound::Unbounded => Bound::Unbounded,
+        };
```
ss-es: this match feels unnecessary. could we e.g. add + Clone to the bound on `range`, or even just give it a concrete type? (I'm not sure where all it's used, but I feel like we shouldn't have to write this explicit match to dereference)

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,375 @@ impl SqlStorage {
...
+            ms * rng.gen_range(0..jitter_num) / jitter_den
+        };
+        min(Duration::from_millis(ms + jitter), self.max)
+    }
+}
+
+/// Returns `true` if `err`'s [`Display`](std::fmt::Display) output identifies it as a PostgreSQL
+/// serialization conflict (SQLSTATE 40001, message "could not serialize access").
+fn is_serialization_conflict_err<E: std::fmt::Display>(err: &E) -> bool {
+    format!("{err:#}").contains("could not serialize access")
```
ss-es, line 808: is there a strong reason to not just always retry? in my opinion the predictability of "always retry on error" is much better than the complexity of doing a text match on the error the database returns

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,375 @@ impl SqlStorage {
...
+        } else {
+            false
+        }
+    }
+}
+
+/// Spawn a background task that queries `pg_stat_activity` and `pg_locks` to identify the
+/// connections and predicate locks involved in a serialization conflict, logging them at `warn`.
+#[cfg(not(feature = "embedded-db"))]
+fn spawn_pg_stat_activity_log(pool: Pool<Db>, op: &'static str) {
```
ss-es, line 836: I can see why this would be useful, though would this not be better to e.g. return at the `database/` API endpoint? (potentially in addition to logging)

do you have a sense of how intensive this query is to run in general?

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,375 @@ impl SqlStorage {
...
+        match sqlx::query(
+            "SELECT pid, COALESCE(state, 'unknown') AS state, left(COALESCE(query, ''), 200) AS \
+             query FROM pg_stat_activity WHERE pid != pg_backend_pid() AND state IS DISTINCT FROM \
+             'idle'",
+        )
+        .fetch_all(&pool)
+        .await
+        {
+            Ok(rows) if rows.is_empty() => {
+                tracing::warn!(op, "serialization conflict: no other non-idle DB sessions");
```
ss-es, line 849: this should probably be `info`

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,375 @@ impl SqlStorage {
...
+                    tracing::warn!(
+                        op,
+                        pid,
+                        state,
+                        "serialization conflict: concurrent session: {query_text}",
+                    );
+                }
+            },
+            Err(e) => {
+                tracing::warn!(op, "failed to query pg_stat_activity: {e:#}");
```
ss-es: this could maybe even be `error`? (I guess I don't know, I'm sort of assuming it's low frequency and would only get hit if something is really actually wrong)

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,375 @@ impl SqlStorage {
...
+             END AS relation, l.page, l.tuple, left(COALESCE(a.query, ''), 100) AS query FROM \
+             pg_locks l JOIN pg_stat_activity a ON a.pid = l.pid LEFT JOIN pg_class c ON c.oid = \
+             l.relation WHERE l.mode = 'SIReadLock' AND a.state IS DISTINCT FROM 'idle' AND l.pid \
+             != pg_backend_pid() ORDER BY l.pid, c.relname",
+        )
+        .fetch_all(&pool)
+        .await
+        {
+            Ok(rows) if rows.is_empty() => {
+                tracing::warn!(op, "serialization conflict: no SIReadLocks held");
```
ss-es, line 881: also info?

hotshot-query-service/src/data_source/storage/sql.rs
```diff
@@ -682,10 +699,371 @@ impl SqlStorage {
...
+            self.serializable_retry_config
+                .retry_if(serialization_conflict_with_diag(self.pool(), op), f)
+                .await
+        } else {
+            let _ = op;
+            self.serializable_retry_config
+                .retry_if(is_serialization_conflict_err, f)
+                .await
+        }
+    }
```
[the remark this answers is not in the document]
ss-es: not sure what claude is referring to here

ss-es (commented):
left some comments, I think overall it makes sense/looks fine to me but my main question is why we need to inspect the error before retrying

hotshot-query-service/src/data_source/storage/sql.rs
sveitser: @lukaszrzasik how much of a lift is it to use the error code like in

```
                                e.code().as_deref() == Some(PG_SERIALIZATION_FAILURE_CODE)
```

sveitser (approved):

lukaszrzasik:
@claude try reviewing again

hotshot-query-service/src/data_source/storage/sql.rs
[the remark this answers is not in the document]
lukaszrzasik: The claude's comment is no longer relevant. I've modified the `retry_if` function to accept the `op` argument as well.

hotshot-query-service/src/data_source/storage.rs
lukaszrzasik: I agree matching the error in this way is far from perfect. I mentioned in the PR description I want to refactor the code so we can match on the actual postgres error code.

I don't think retrying every transaction error would be a good approach. Not every error is worth immediate retries and postgres docs are clear about the serialize error. It should be retried because this is a temporary error stemming from two or more transactions operating on the same data in an ambiguous way.

hotshot-query-service/src/data_source/storage.rs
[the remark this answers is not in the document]
lukaszrzasik: no, it could have panicked during runtime. I've refactored it. It should be safe now even with non-ascii characters.

yes, `#[instrument]` can be an alternative. It has some downsides as well though:
1. We'll lose the information about the calling function in case of the spawn diagnostic task. This is quite important because that information will allow us to correlate the serialization errors with the specific diagnostic logs.
2. It would require modifying all the calling functions to include `#[instrument]`.

Point 2 is not that important I think but we'll lose some functionality because of point 1.
I can change it if you have a strong opinion about it.

hotshot-query-service/src/data_source/fetching.rs
lukaszrzasik: I agree. I've modified it.

hotshot-query-service/src/data_source/storage/sql.rs
lukaszrzasik: Re: the text match: I addressed it in the comment above.

We can't use the error code because we don't always have the error code. Some methods drop the original error from postgres and I use their own preserving only the text.
A follow up PR will refactor the error handling so that the postgres error is always preserved.

hotshot-query-service/src/data_source/storage/sql.rs
lukaszrzasik: I'm not entirely convinced exposing it as api is a good idea. The log and queries are disabled by default. The idea is to enable them only when we want to actively debug the serialization errors. They're also rate limited, meaning they are executed once per given serialization error. The following retries won't trigger them.

The first query should be super cheap. The second might be a bit more intensive because of the joins. Both are only executed on demand and rate limited as mentioned above. They're also spawned in a separate task so should not interfere with other parts of the system.

hotshot-query-service/src/data_source/storage/sql.rs
lukaszrzasik: I would leave at warn. The reason is this: the logs in this function are only emitted if we specifically enable the postgres serialization diagnostics. The typical log level setting is warn. We might be prone to forgetting that we have to enable the diagnostics and change the log level.

hotshot-query-service/src/data_source/storage/sql.rs
lukaszrzasik: same reasoning as above

hotshot-query-service/src/data_source/storage/sql.rs
lukaszrzasik: yes, and only if the diagnostics are enabled. I can change it.

hotshot-query-service/src/data_source/storage.rs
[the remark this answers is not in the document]
ss-es: I'm still a bit uncomfortable with the indexing into the string. at the very least could we avoid the indexing (possibly panic) with something like e.g.   `let trimmed: &'static str = full.strip_suffix("::__f").unwrap_or(full);`

hotshot-query-service/src/data_source/storage.rs
```diff
@@ -89,6 +89,57 @@ use crate::{
...
+}
+
+/// Calls [`SerializableRetry::serializable_retry`] with `op` set to the unqualified name of the
+/// enclosing function via [`function_name!`](crate::function_name).
+#[macro_export]
+macro_rules! serializable_retry {
+    ($self:expr, $f:expr) => {
+        $self.serializable_retry($crate::function_name!(), $f)
+    };
+}
```
ss-es, line 141: maybe alternatively (sorry I didn't think that much about this earlier), why not just inline this because I don't think the macro is really saving us anything (beyond the minor repetition of the function name)? I feel like in general it's easier to read anyway without the macro

PR #4451, opened 2026-06-08
[new-protocol]: Disambiguate blocks

twittner:


crates/hotshot/new-protocol/src/coordinator.rs
bfish713: I think this block of code needs to move now, otherwise we'll append the block twice for a view and only one will be correct.  I think the storage rejects the second block for the same view and in the current case in devnet it's the second one that is the actual one proposed.

twittner: This might have been a problem already as we built multiple blocks prior to this PR. In https://github.com/EspressoSystems/espresso-network/pull/4451/commits/975e213f228f2aea9026152b62c97ac16ab8c45f I keep the blocks in memory until consensus issues a `SendProposal` command at which point we know what to propose and what to store.

bfish713: yes this was an existing problem

bfish713 (approved):

PR #4471, opened 2026-06-10
[new-protocol] finish storage tasks when coordinator is stopped

imabdulbasit:
Shut down the new protocol coordinator gracefully so pending storage writes finish instead of being aborted mid-flight. The coordinator loop now stops on a `CancellationToken` and calls `coord.stop(),` which flushes inflight storage tasks. The binary also now registers a signal handler which handles SIGINT/SIGTERM and calls `ctx.shut_down()`


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -338,7 +338,8 @@ where
     }
 
     pub async fn stop(mut self) {
-        self.network.shutdown().await
+        self.network.shutdown().await;
+        self.storage.flush().await;
```
bfish713: nit: Might be best to do a join! on these two things so if one hangs at least the other has a chance.

bfish713 (approved):

PR #4472, opened 2026-06-10
feat(new-protocol): require durable storage before voting

bfish713:
All consensus storage writes (proposal, VID share, certs, vote/propose
actions) for views up to the vote view must complete before a vote1 or
vote2 is broadcast. Payload writes (append_da) remain best-effort and
never block voting.

Storage write tasks retry until success, so waiting on them at vote
time provides durability without serializing writes into the hot path.
This closes the restart hole where a vote could leave the node before
its action record (or the data the vote attests to) was persisted.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/new-protocol/src/coordinator.rs
+++ b/crates/hotshot/new-protocol/src/coordinator.rs
@@ -513,7 +513,7 @@ where
         self.consensus.apply(input, &mut self.outbox)
     }
 
-    pub fn process_consensus_output(
+    pub async fn process_consensus_output(
         &mut self,
         output: ConsensusOutput<T>,
     ) -> Result<(), CoordinatorError> {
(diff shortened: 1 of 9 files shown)


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -700,6 +700,12 @@ where
                     "send vote1"
                 );
                 self.storage.record_action(view, HotShotAction::Vote);
+                // Everything the vote attests to (proposal, VID share, the
+                // vote action itself) must be durable before the vote leaves
+                // this node.
+                self.storage.wait_until_durable(view).await.map_err(|e| {
```
twittner, line 706: This blocks the event loop, potentially for a very long time.

PR #4473, opened 2026-06-11
ci(claude-review): add backport-correctness review mode

sveitser:
Problem: claude's review on backports is just noise right now.

- detect backports via `backport-<n>-to-<target>` branch name
- fetch original PR comments and both diffs into ./tmp
- prompt reviews backport fidelity, skips issues already raised on original

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -35,6 +35,28 @@ jobs:
             "${{ github.event.pull_request.number }}" \
             > ./tmp/pr-review-comments.json
 
+      - name: Prepare backport context
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_REF: ${{ github.head_ref }}
+          OWNER: ${{ github.repository_owner }}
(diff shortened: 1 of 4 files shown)


.github/workflows/claude-review.yml
```diff
@@ -44,7 +66,21 @@ jobs:
...
+              the original change, adapted correctly to the target branch (API/type
+              differences between branches, conflict-resolution edits, dropped or extra
+              hunks).
+            - `./tmp/original-pr-comments.json` holds the original PR's review comments.
+              Do NOT re-flag anything already raised there. Only flag problems introduced
+              by the backport itself (bad merge, lost hunk, wrong adaptation).
+            - Skip the same exclusions listed below.
+            Then stop; do not run the standard review.
+
+            Otherwise, perform a comprehensive code review with the following focus areas:
```
sveitser: maybe a second prompt would be better to make the actual prompt more straightforward

lukaszrzasik (approved):

PR #4477, opened 2026-06-11
Optimise serialisation of VID shares.

twittner:
Use `serde_bytes` to avoid serialising individual bytes. The result is compatible with exisiting bincode and JSON artefacts that do not use `serde_bytes`.


mrain (approved):
really cool. my microbench shows a massive over 30x improvement

PR #4480, opened 2026-06-12
fix(backport): install claude code CLI instead of using claude code action

sveitser:
- The conflict-resolution step only edits files locally, so the action's
  GitHub context/MCP/OIDC machinery was pure overhead (and the OIDC
  exchange 401s on pull_request_target)
- Install Claude Code CLI and run claude -p with the API key directly;
  no GitHub token or id-token permission needed
- Use --dangerously-skip-permissions for non-interactive CI; model opus 4-8

Testing here: https://github.com/EspressoSystems/espresso-network/actions/runs/27407282554/job/80999461205


sveitser:
@claude review

lukaszrzasik (approved):

PR #4482, opened 2026-06-12
fix(deps): pin time < 0.3.48 fix build without lockfile

sveitser:
- See https://github.com/time-rs/time/issues/783
- time 0.3.48 introduced a ModifierValue trait with conflicting From<HourBase> impls, exposed by Rust 1.96.0 stricter coherence check
- time = ">=0.3, <0.3.48" forces resolution to 0.3.47 which compiles cleanly
- cargo update

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,7 +148,7 @@ dependencies = [
  "axum 0.8.9",
  "bytes 1.11.1",
  "cfg-if",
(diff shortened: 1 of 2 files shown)


sveitser:
Supposedly this is fixed upstream and the affected time-rs crate version has been yanked.

PR #4498, opened 2026-06-12
[Fast Finality] Verify VID Commitment on Reconstructed Block

bfish713:
AVID-M recovery does not bind the decoded bytes to the payload
commitment: shares verify against the disperser's merkle tree, but a
Byzantine disperser can commit to a non-codeword, and a bad share
poisons the erasure decoding into silently producing garbage. The
reconstructor then shipped whatever commitment the shares claimed,
so the coordinator stored and emitted unverified payloads.

After recovery, recompute the VID commitment from the recovered bytes
(mirroring calculate_vid_disperse) and fail reconstruction on mismatch.
Failures now carry the view so the reconstructor drops its in-flight
marker; accumulated shares are kept, letting a later share trigger a
retry.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/new-protocol/src/coordinator.rs
+++ b/crates/hotshot/new-protocol/src/coordinator.rs
@@ -536,7 +536,7 @@ where
                     }
                     Err(err) => {
                         finish_measurement(next_input);
-                        return Err(CoordinatorError::regular(err.to_string()).context("vid reconstruction"))
+                        return Err(CoordinatorError::regular(err).context("vid reconstruction"))
                     }
                 },
                 Some(stored) = self.storage.next() => {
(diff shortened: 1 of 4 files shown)


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -534,9 +534,9 @@ where
                         }
                         return Ok(ConsensusInput::BlockReconstructed(out.view, out.payload_commitment))
                     }
-                    Err(()) => {
+                    Err(err) => {
                         finish_measurement(next_input);
-                        return Err(CoordinatorError::unspecified().context("vid reconstruction"))
+                        return Err(CoordinatorError::regular(err.to_string()).context("vid reconstruction"))
```
twittner: Instead of mapping the error to a string, we could add a constructor for `VidReconstructError` to `ErrorSource`.

twittner (approved):

PR #4513, opened 2026-06-15
Fix epoch reward: handle pre V5 epoch in reward calculation

imabdulbasit:
At the epoch boundary, reward calculation for the previous epoch fails when that epoch's header is pre V5. `fetch_and_calculate` returns error  instead of treating it as nothing to distribute (0 rewards). This PR returns a zero `EpochRewardsResult` for headers below `EPOCH_REWARD_VERSION`. It also collapses the missing result fallback in `handle_epoch_rewards`


crates/espresso/types/src/v0/impls/header.rs
sveitser, line 860: @imabdulbasit update this comment?

sveitser (approved):

PR #4514, opened 2026-06-16
docs(agents): add AI-assisted commit disclosure guideline

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Document a kernel-style `Assisted-by:` trailer for commits created with AI assistance, replacing `Co-Authored-By` for the assistant and forbidding AI `Signed-off-by`. Follows https://docs.kernel.org/process/coding-assistants.html.

Assisted-by: Claude Code:claude-opus-4-8

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,6 +40,27 @@ Reviewing is the bottleneck. Default to changes that minimize reviewer time.
 - Tell the reviewer where to focus.
 - Link the regression test or `reference` test when touching serializable types.
 
+### AI-assisted commits
+
+When a commit was created with help from an AI coding assistant, disclose it, following the
+[Linux kernel coding-assistant guidelines](https://docs.kernel.org/process/coding-assistants.html). The AI is a tool,
+not an author: the human who submits the commit reviews the change, verifies it, and is fully accountable for it.
+
+- Add an `Assisted-by:` trailer to the commit message:
+
+  ```
+  Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]
+  ```
+
+  - `AGENT_NAME` — the assistant or framework (e.g. `Claude Code`).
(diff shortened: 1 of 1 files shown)


ss-es (approved):
seems fine, no objection

sveitser (changes requested):
@lukaszrzasik @ss-es I don't quite understand what the point of this is. A year or longer ago I can see it being useful but now we're in a world where almost everything is delegated to AI assistants in terms of writing code, committing etc. so to me it seems this is mostly just noise in commit messages, adds to the context of the model without clear benefit? If anything we should tag human only commits as "special" but I don't think that's really useful either.

lukaszrzasik:
A few counter arguments:
1. It would be nice to follow industry good practices.
2. If not instructed claude uses `Co-Authored-By` when asked to commit. My opinion is `Assisted-by` is more appropriate.
3. If nothing else, it'll allow us to track which models we use.
4. It's not meant to add any manual work: it only applies to agents, we don't need to add it manually even we did used AI to some extent.
5. I don't think it hurts us in any way, so I would argue pros outweigh cons.

ss-es:
I don't typically let agents commit code personally (so this would have no effect for my PRs), and I also typically don't inspect the commit history when reviewing a PR.

I don't have a strong opinion, but I agree that context pollution is a real concern. I also agree the benefit is not clear -- this would probably just add a significant chunk to the agent's context for something that seems cosmetic.

sveitser:
@lukaszrzasik I think it's a good idea to do something to nudge claude towards leaving good commit messages. I think we can add a commit skill with the instructions and maybe also a local claude settings json where we set

https://code.claude.com/docs/en/settings#attribution-settings

I don't know if this also loaded into context would have to look at the system prompt for that I guess.

What I'm not sure is about the whole model attribution because it's not possible in general to know what model wrote which code without loading all the involved claude session histories. I'm okay if it's not always correct though.

PR #4518, opened 2026-06-16
test(reward): add all-node restart epoch reward regression test

sveitser:
- reproduce chain stall when all nodes restart after the V5 upgrade
- upgrade lands in epoch 4, restart 5 blocks later before the next boundary
- add `TestConfigBuilder::upgrade_proposing_views` to place the upgrade
- fails with stall at block 997 prior to f0ab04f345


imabdulbasit (approved):
looks good!

PR #4520, opened 2026-06-16
Load next epoch QC when loading `load_anchor_leaf`

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* use  `load_anchor_leaf` to also load the next epoch QC if it's available and matches the main QC
* Add `test_v5_restart_before_epoch_boundary`
* Adds `test_all_restart_transition_epochs`. We already had `test_all_restart_epochs` which restarted all nodes in view 10, i.e. the last block in the epoch (epoch height is 10) but that didn't cover the case when all the nodes are restarted in the middle of the transition. Unfortunately this makes a big difference.

### DIFF
--- a/crates/espresso/node/src/api.rs
+++ b/crates/espresso/node/src/api.rs
@@ -9145,4 +9145,149 @@ mod test {
 
         Ok(())
     }
+
+    /// Start a network at V5 from genesis, restart ALL nodes just before an epoch transition block
+    /// (`boundary - 3`), and confirm the chain keeps producing blocks afterward.
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
+    async fn test_v5_restart_before_epoch_boundary() {
+        const NUM_NODES: usize = 3;
+        const EPOCH_HEIGHT: u64 = 10;
+        // Rewards start in epoch 4. Restart 3 blocks before the boundary that closes epoch 4, so
+        // the restarted network produces the boundary block (`4 * EPOCH_HEIGHT`) itself.
+        const RESTART_EPOCH: u64 = 4;
+        const RESTART_BOUNDARY: u64 = RESTART_EPOCH * EPOCH_HEIGHT;
+        const RESTART_HEIGHT: u64 = RESTART_BOUNDARY - 3;
+        // Blocks to produce after the restart before declaring success.
+        const BLOCKS_AFTER_RESTART: u64 = 5;
+
+        const V5: Upgrade = Upgrade::trivial(EPOCH_REWARD_VERSION);
+
+        let port = reserve_tcp_port().expect("OS should have ephemeral ports available");
+
+        // Slow empty-block production so we comfortably stop at the exact target height. On an idle
+        // chain the empty-block time is ~`builder_timeout`; raise `next_view_timeout` above it so
+        // the slow (but healthy) views aren't treated as failures.
+        let network_config = TestConfigBuilder::<NUM_NODES>::default()
+            .epoch_height(EPOCH_HEIGHT)
+            .builder_timeout(Duration::from_secs(3))
+            .next_view_timeout(Duration::from_secs(10))
+            .build();
+
+        let storage = join_all((0..NUM_NODES).map(|_| SqlDataSource::create_storage())).await;
+        let persistence: [_; NUM_NODES] = storage
+            .iter()
+            .map(<SqlDataSource as TestableSequencerDataSource>::persistence_options)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
+
+        let config = TestNetworkConfigBuilder::<NUM_NODES, _, _>::with_num_nodes()
+            .api_config(SqlDataSource::options(
+                &storage[0],
+                Options::with_port(port),
+            ))
+            .persistences(persistence.clone())
+            .catchups(std::array::from_fn(|_| {
+                StatePeers::<SequencerApiVersion>::from_urls(
+                    vec![format!("http://localhost:{port}").parse().unwrap()],
+                    Default::default(),
+                    Duration::from_secs(2),
+                    &NoMetrics,
+                )
+            }))
+            .network_config(network_config)
+            .pos_hook(DelegationConfig::MultipleDelegators, Default::default(), V5)
+            .await
+            .unwrap()
+            .build();
+
+        let mut network = TestNetwork::new(config, V5).await;
+
+        // Watch the decide stream and stop as soon as `boundary - 3` is decided. Consuming the
+        // stream (rather than polling `decided_leaf`) makes this independent of block timing: we
+        // see every decided leaf and stop the network the moment the target height appears, so we
+        // can never race past it regardless of how fast blocks are produced.
+        {
+            let mut events = network.server.event_stream();
+            'wait: loop {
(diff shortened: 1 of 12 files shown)


sveitser (approved):

crates/espresso/node/src/persistence/fs.rs
```diff
@@ -675,7 +675,10 @@ impl Inner {
                 .bytes()
                 .collect::<Result<Vec<_>, _>>()
                 .context("read")?;
-            return Ok(Some(bincode::deserialize(&bytes).context("deserialize")?));
+            let (leaf2, qc2): (Leaf2, QuorumCertificate2<SeqTypes>) =
+                bincode::deserialize(&bytes).context("deserialize")?;
+            let cert_pair = CertificatePair::new(qc2, None);
+            return Ok(Some((leaf2, cert_pair)));
```
ss-es, line 681: so do we just always return `None` for the next_epoch_qc for fs storage? isn't that problematic?

crates/espresso/node/src/api.rs
```diff
@@ -9145,4 +9145,149 @@ mod test {
...
+                if stalled_polls >= stall_limit {
+                    panic!(
+                        "chain stalled after restart 3 blocks before boundary {RESTART_BOUNDARY}: \
+                         no new decide for 60s at height {last_height}, unable to produce/cross \
+                         the epoch boundary (target height was {target_height})."
+                    );
+                }
+            }
+        }
+    }
```
ss-es, line 9292: I'll confess I didn't really look at the test but I trust it

crates/espresso/node/src/lib.rs
```diff
@@ -1294,6 +1294,23 @@ pub mod testing {
             self
         }
 
+        /// Override the builder timeout. On an idle chain the test builder returns no block (so it
+        /// doesn't drive empty blocks at full speed); the leader retries until this timeout elapses
+        /// and only then proposes an empty block itself. So for an empty mempool this is effectively
+        /// the empty-block time. Keep it below [`Self::next_view_timeout`] or views time out before
+        /// the leader proposes.
```
ss-es: I think these comments could be deleted

crates/espresso/types/src/v0/traits.rs
```diff
@@ -660,48 +660,64 @@ pub trait SequencerPersistence:
...
-            .load_next_epoch_quorum_certificate()
-            .await
-            .context("loading next epoch qc")?;
-        let (leaf, mut high_qc, anchor_view) = match self
+        let config = self.load_config().await.context("loading config")?;
+        let epoch_height = config
+            .as_ref()
+            .map(|c| c.config.epoch_height)
+            .unwrap_or_default();
+        let (leaf, cert_pair, anchor_view) = match self
```
ss-es, line 668: for fs the cert pair will behave the same as before right? don't we also want to fix that

crates/espresso/types/src/v0/traits.rs
```diff
@@ -660,48 +660,64 @@ pub trait SequencerPersistence:
...
+                );
+                ensure!(
+                    epoch_height == 0 || cert_pair.verify_next_epoch_qc(epoch_height).is_ok(),
+                    format!(
+                        "Next epoch QC is required but it's not present or doesn't match primary \
+                         QC\nPrimary QC: {:?}\nNext epoch QC: {:?}",
+                        cert_pair.qc(),
+                        cert_pair.next_epoch_qc()
                     )
                 );
```
ss-es, line 692: I don't think we should error here unless I'm missing something. even if we failed to load with the next epoch qc for whatever reason, someone else might propose with it (and in that case we'd be fine right?)

ss-es (commented):
I think this seems fine, but my 2 questions are:

- why not also implement this for fs storage
- is it really appropriate to error if we have an issue with the next epoch high qc

crates/espresso/node/src/persistence/fs.rs
lukaszrzasik: This code is for the `legacy_anchor_leaf_path` case. Above, in the same method, we do try to grab and deserialize `CertificatePair` that optionally includes the next epoch QC. We fall back to the legacy anchor leaf only if `decided_leaf2_path` is not available.
Per existing warning in the method:
```
        tracing::warn!(
            "Failed to find an anchor leaf in `Leaf2` storage. Checking legacy `Leaf` storage. \
             This is very likely to fail."
        );
```

crates/espresso/node/src/api.rs
lukaszrzasik: To be honest I didn't go through the test thoroughly as well. I merged it from Abdul's PR: https://github.com/EspressoSystems/espresso-network/pull/4521
It fails there with the same symptoms we saw in water. It doesn't fail here.

crates/espresso/node/src/lib.rs
lukaszrzasik: I guess so. They got merged from the test PR.

crates/espresso/types/src/v0/traits.rs
lukaszrzasik: As indicated above, I fixed the fs storage as well.

crates/espresso/types/src/v0/traits.rs
lukaszrzasik: I'm uncertain about this. My first reaction was: yes, you're right. But on the other hand, what you said is true about other checks as well, for example: the check right before this one: `leaf_view == high_qc.view_number`. Even if we have an inconsistent anchor leaf with the high qc, somebody else might have a higher leaf and qc which are correct and we should be able to progress, so we shouldn't fail.

The question is broader then: how much corrupted data do we want to tolerate, knowing that other nodes can have the correct data.

Because having an inconsistent next epoch qc with the current epoch qc indicates that something went really wrong and this breaks the epoch transition protocol invariants.

crates/espresso/types/src/v0/traits.rs
```diff
@@ -660,48 +660,64 @@ pub trait SequencerPersistence:
...
+                    leaf_view == high_qc.view_number,
                     format!(
                         "loaded anchor leaf from view {}, but high QC is from view {}",
-                        leaf.view_number(),
-                        high_qc.view_number
+                        leaf_view, high_qc.view_number
+                    )
+                );
+                ensure!(
+                    epoch_height == 0 || cert_pair.verify_next_epoch_qc(epoch_height).is_ok(),
```
sveitser, line 685: Is `epoch_height == 0` the right gate here? For our deployed chains it doesn't matter because they are all on v4 or higher but in `demo-pos.toml` for example `epoch_height = 150` but the chain starts on v2 prior to the upgrade. Do we have a less adhoc way to check?

lukaszrzasik: epoch_height == 0 means there are no epoch (pre-PoS upgrade) so we shouldn't check the next epoch QC.

ss-es (approved):

crates/espresso/types/src/v0/traits.rs
ss-es: yeah maybe that's a question for a separate PR then. approved the PR as-is

PR #4539, opened 2026-06-19
Allow reading genesis file from URL

ss-es:


crates/espresso/node/src/genesis.rs
```diff
@@ -11,10 +13,49 @@ use espresso_types::{
...
+
+impl FromStr for GenesisSource {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        if s.starts_with("http://") || s.starts_with("https://") {
+            let url = Url::parse(s).with_context(|| format!("invalid genesis URL: {s}"))?;
+            Ok(Self::Http(url))
+        } else {
+            Ok(Self::Path(PathBuf::from(s)))
```
sveitser, line 39: Not super important but does this work with `file://` ?

sveitser (approved):
LGTM. I think a test wouldn't hurt though.

crates/espresso/node/src/genesis.rs
ss-es: it doesn't -- when I originally proposed that I didn't think we had this option to override the file with an env var.

I had the extra `file://` thing at first, but it felt silly to have both `file://` and the original which already accepted a raw filepath

PR #4540, opened 2026-06-19
Merge release-mainnet-1.2.0-rc

ss-es:


Cargo.toml
```diff
@@ -397,7 +397,7 @@ test-log = { version = "0.2", features = ["color", "trace"] }
 test-utils = { path = "test-utils" }
 thiserror = "2"
 tide-disco = "0.9"
-time = "0.3"
```
sveitser, line 400: @ss-es i think we can drop this, because the offending time version was yanked

sveitser (approved):

Cargo.toml
ss-es: reverted to `time = "0.3"`

PR #4557, opened 2026-06-22
feat(libp2p): drop peers from other networks on identify

sveitser:
- Store expected kademlia StreamProtocol on NetworkNode
- Disconnect and skip routing-table insert for peers not advertising it
- Add unit matrix + 2+1-node integration test proving selective drop


sveitser:
@claude review, check if your old concerns are addressed

ss-es (approved):
this seems fine but I definitely do not understand libp2p well enough to assess whether this is correct

PR #4570, opened 2026-06-23
fix(node): only run startup epoch-window catchup on new protocol

imabdulbasit:
Skip the startup stake-table window catchup unless the genesis max version is at or above the new protocol, since only cliquenet requires it.


sveitser (approved):

sveitser:
@imabdulbasit I removed the 1.1 backport label since we're not planning another release there

PR #4571, opened 2026-06-23
[new-protocol] support new protocol (V6) in devnode

imabdulbasit:


crates/espresso/dev-node/src/main.rs
```diff
@@ -746,12 +769,18 @@ async fn async_main(migrated_envs: Vec<(&str, &str)>) -> anyhow::Result<()> {
...
+    // dev node binary exits
     tokio::select! {
         Some(item) = handles.next() => {
             tracing::error!("exiting dev node: {item:?}");
             drop(network);
+            drop(consensus_dbs);
+        },
+        _ = espresso_utils::shutdown::wait_for_shutdown_signal() => {
+            drop(network);
+            drop(consensus_dbs);
```
twittner: Not sure why those explicit `drop` calls have been added? Right after this `tokio::select` the function returns and the values will implicitly be dropped.

imabdulbasit: yes you are right

bfish713 (approved):

PR #4575, opened 2026-06-23
Implement fetch_all_namespaces_in_range

ImJeremyHe:
### This PR:

Adds a multi-namespace variant of the existing single-namespace range query: given a set of namespace IDs and a block range, fetch a verified namespace proof for every requested namespace that is present in each block. Namespaces absent from a block are omitted from that block's result.

```
GET /light-client/namespaces/:start/:end/:namespaces
```

:namespaces is a TaggedBase64-encoded JSON array of namespace IDs (tag NS).
Returns one entry per block in [start, end), ascending by height; each entry maps a namespace ID to its proof, including only the requested namespaces actually present in that block.
Encoded as a GET with the set in the URL (rather than a POST body) so responses stay cacheable.


crates/espresso/node/src/api/light_client.rs
```diff
@@ -307,21 +315,78 @@ where
...
+}
+
+async fn get_namespace_proof_range<State>(
+    state: &State,
+    start: usize,
+    end: usize,
+    namespace: u64,
+    fetch_timeout: Duration,
+    large_object_range_limit: usize,
+) -> Result<Vec<NamespaceProof>, Error>
```
Sneh1999, line 332: nit: should we return custom errors here?

crates/espresso/node/api/light-client.toml
```diff
@@ -129,5 +129,15 @@ PATH = ["/namespace/:start/:end/:namespace"]
 ":end" = "Integer"
 ":namespace" = "Integer"
 DOC = """
-Fetch a list of namespace proofs for each block in the rang `[:start, end)`.
+Fetch a list of namespace proofs for each block in the range `[:start, end)`.
```
Sneh1999, line 132: nit: I think this description can be better

crates/espresso/node/src/api/light_client.rs
```diff
@@ -307,21 +315,78 @@ where
...
-                    status: StatusCode::INTERNAL_SERVER_ERROR,
-                })?;
-            Ok(NamespaceProof::new(ns_proof, vid_common.common().clone()))
+            let ns_proof = build_namespace_proof(&payload, &ns_index, &vid_common)?;
+            Ok(ns_proof)
+        })
+        .collect()
+}
+
+async fn get_namespaces_proof_range<State>(
```
Sneh1999, line 355: this function and the `async fn get_namespace_proof_range<State>(` above which only accepts one namespace seems to be named the same. Do we actually need two functions or can we combine them?

crates/espresso/node/api/light-client.toml
```diff
@@ -129,5 +129,15 @@ PATH = ["/namespace/:start/:end/:namespace"]
...
+"""
+
+[route.namespaces_range]
+PATH = ["/namespaces/:start/:end"]
+":start" = "Integer"
+":end" = "Integer"
+METHOD = "POST"
```
jbearer: Unfortunately `POST` is problematic here because it means we can't cache this endpoint. It's a bit ugly but we could change this to `GET` and encode the parameter as base64 in the URL

crates/espresso/node/src/api/light_client.rs
jbearer: Agree, we could probably combine them, because `get_namespace_proof_range` (singular) is equivalent to just passing a singleton list `&[namespace_id]` into this function

crates/espresso/node/api/light-client.toml
jbearer: Alternatively, quite a bit more involved, we could (finally) go in and update `tide-disco` so that `req.body_auto()` can be used to parse query parameters if the request is a `GET`. An array of namespace IDs could be passed in like `host/namespaces/:start/:end?namespaces[]=ns1&namespaces[]=ns2`

which would deserialize into an object like

```json
{
  "namespaces": ["ns1", "ns2"],
}
```

crates/espresso/node/api/light-client.toml
ImJeremyHe: I just fixed a typo; I’ll keep the description unchanged for now.

crates/espresso/node/src/api/light_client.rs
ImJeremyHe: Combined, but these two functions actually have different return types, so the LOC doesn’t really change. Previously, both reused fetch_block_data_range. Now get_namespace_proof_range uses it via a map and retrieves the result by removing the namespace key.

crates/espresso/node/api/light-client.toml
ImJeremyHe: Thanks Jeb. I found we have similar use [here](https://github.com/EspressoSystems/espresso-network/blob/98e9911b5937d02e670422dd15da8cee1d94c791/crates/espresso/node/api/catchup.toml#L36). So I use tagged based 64 to keep them aligned. LMK your thoughts

Sneh1999 (approved):

jbearer (approved):

PR #4583, opened 2026-06-24
Add release workflows

ss-es:
Adds CI workflows for a release process.

In summary, we will now have an issue tracker for each release, with workflows that automatically track the divergence from main (so it is easier to see which commits have been backported or not). There is also some UX improvements for the tagging process, with a `/tag` (and in the same vein, `/promote`) command that can be used on the issue page.

Release branches are now `release-X.Y.Z`, where `X.Y` should correspond to the protocol version intended to run from this release. Tags now take the form of `X.Y.PHASE.PATCH`, where `PATCH` is automatically incremented via the `/tag` command.

The full process is described in `RELEASES.md`


.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,131 @@
...
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+       github.event.issue.pull_request == null &&
+       contains(github.event.issue.labels.*.name, 'release-tracker') &&
+       startsWith(github.event.comment.body, '/tag'))
+    steps:
```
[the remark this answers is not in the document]
ss-es: not sure this is that important

.github/workflows/update-release-tracker.yml
```diff
@@ -0,0 +1,214 @@
...
+      - name: Dispatch /promote
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          ISSUE: ${{ github.event.issue.number }}
+          BODY: ${{ github.event.comment.body }}
+          ACTOR: ${{ github.event.comment.user.login }}
+        run: |
+          set -euo pipefail
+          stage=$(awk '{print $2; exit}' <<<"$BODY")
```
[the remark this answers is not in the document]
ss-es: should be fixed in [20a0f5c](https://github.com/EspressoSystems/espresso-network/pull/4583/commits/20a0f5c162906a6b2ed997b2c093edf379c0d3ab)

RELEASES.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Manual fallback
+
+If the tracker issue is unavailable, the same operations are available via the
+GitHub CLI:
+
+```bash
+just tag release-0.4.0                                # same as /tag
+gh workflow run promote-docker-tag.yml \              # same as /promote decaf.canary
+  -f stage=decaf.canary -f tag=0.4.0.5
```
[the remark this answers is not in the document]
ss-es: should be fixed in [20a0f5c](https://github.com/EspressoSystems/espresso-network/pull/4583/commits/20a0f5c162906a6b2ed997b2c093edf379c0d3ab)

RELEASES.md
```diff
@@ -0,0 +1,155 @@
...
+- **Experimental branches** — currently-open branches matching `release-MAJOR.MINOR.PHASE/*` with their last commit, so you can see what's being validated at a glance.
+
+## Cutting a release: end-to-end
+
+1. **Cut the branch.** From `main` at the chosen commit, push a new branch
+   `release-MAJOR.MINOR.PHASE`. Automation creates the backport label and
+   tracker issue. The tracker issue is where you'll do everything that
+   follows.
+
+2. **Run an initial tag.** Comment `/tag` on the tracker issue. This creates
```
sveitser: I think 2. can be fully automated after 1.? So can we just make this part of a workflow dispatch run or have at least a CLI convenience command? I think it's easier if everything happens in one place instead of having to run CLI commands and perform comments on github.

.github/workflows/update-release-tracker.yml
```diff
@@ -0,0 +1,217 @@
...
+
+          digest() {
+            docker buildx imagetools inspect --raw "$REGISTRY/$REF_SERVICE:$1" 2>/dev/null \
+              | jq -r '.manifests[]? | select(.platform.architecture=="amd64" and .platform.os=="linux") | .digest' \
+              || true
+          }
+
+          refresh() {
+            local issue="$1"
+            local body; body=$(gh issue view "$issue" --json body -q .body)
```
sveitser: This code is difficult to review, or even just figure out what i does exactly and there's currently no way to run anything locally, or any tests because it's inlined into the workflow. I think it could be extracted into a python script for better readability and testability.

I think there may be some concurrency issues and the workflow has many triggers.

.github/ISSUE_TEMPLATE/release-tracker.md
```diff
@@ -0,0 +1,43 @@
+---
```
sveitser, line 1: Currently the issue is used as a kind of database and it's just a string. There's no real schema, versioning, migration etc. so I'm afraid if we need to change or fix anything it may get very messy. I think there must be a better way to do this.

I haven't thought about it much but for example the state could be kept as json (or similar) in the issue body (commented) and managed by a python script.

.github/workflows/setup-release-branch.yml
```diff
@@ -0,0 +1,152 @@
...
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: read
+
+jobs:
+  on-create:
+    if: |
+      github.event_name == 'create' &&
```
sveitser, line 26: How about adding workflow dispatch for this instead of pushing a branch and then name matching on the branch? The workflow could take any rev / branch / tag as input, plus the version and then create the branch on gihub

binary-upgrade-tests/run.py
```diff
@@ -86,7 +87,8 @@
 LC_GATING_OVERLAY = REPO_ROOT / "binary-upgrade-tests" / "compose.lc-gating.yaml"
 
 
-YYYYMMDD_TAG_PATTERN = "20[0-9][0-9][0-1][0-9][0-3][0-9]"
+RELEASE_TAG_GLOB = "[0-9]*.[0-9]*.[0-9]*.[0-9]*"
+RELEASE_TAG_RE = re.compile(r"^\d+\.\d+\.\d+\.\d+$")
```
sveitser, line 91: Fallback to YYYYMMDD if it doesn't find new tags?

Not part of this PR but we should rewrite  it a bit anyway to pick up the versions that are deployed instead of tagged on github. Currently it moves the base version up too early.

sveitser (changes requested):
I think the overall ideas are great and a significant improvement. I like the sticky github issue and interaction via slash commands.

I left some comments with suggestions and concerns.

I think my biggest concern is that this adds a lot of bash-in-yaml CI code that is almost untestable so I would prefer not to do that. Whenever a workflow runs it can use all the code in the repo so I see no reason not to script the complicated parts outside github yaml files. This also allows to add some tests for it to document important behaviors that are currently implicit and difficult to determine from looking at the code. We could for example add a python script that inputs and outputs the issue body (potentially containing a more structured artifact, as suggested).

RELEASES.md
```diff
@@ -0,0 +1,155 @@
...
+a docker image you can deploy on devnet → validate → open a PR back into the
+canonical release branch. The experimental branch is deleted once merged or
+when it turns out the experiment didn't pan out.
+
+## The release tracker issue
+
+The tracker issue is the durable, branch-scoped surface for everything
+release-related. It is the place where you run release commands, see status,
+and discuss the release with the team. Its body is maintained automatically;
+the bot edits sections between `<!-- BEGIN ... -->` / `<!-- END ... -->`
```
sveitser: I think there are no such markers now?

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,131 @@
...
+          git config user.name  'github-actions[bot]'
+          git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
+
+      - name: Create and push tag
+        env:
+          TAG: ${{ steps.next.outputs.tag }}
+        run: |
+          set -euo pipefail
+          git tag -a "$TAG" -m "Release $TAG"
+          git push origin "refs/tags/$TAG"
```
sveitser, line 136: I think this workflow is missing the step to create a github release.

sveitser:
Since almost nothing on the workflow is specific to this repo why don't we first put it into a separate repo build some trivial docker images then we can test the whole workflow very quickly, including promotion, docker tag protection etc. and avoid polluting this repo and waiting for long builds.

ss-es:
@claude I've rearchitectured some of this, can you take another look?

.github/ISSUE_TEMPLATE/release-tracker.md
ss-es: I think this should be resolved now -- the issue page itself no longer maintains any state

.github/workflows/tag-release.yml
ss-es: I'm not sure we want to create a github release on every tag, since we likely will use tags for triggering builds for devnet as well now

PR #4590, opened 2026-06-25
feat: decouple vid share broadcast from vote1

mrain:
Closes #<ISSUE_NUMBER>


crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -1740,7 +1743,11 @@ impl<T: NodeType> Consensus<T> {
             debug!(%view, "dropping pending vote1 for timed-out view");
             return;
         }
+        let vid_share = self.vid_shares.get(&view).cloned();
         outbox.push_back(ConsensusOutput::SendVote1(vote1));
+        if let Some(vid_share) = vid_share {
+            outbox.push_back(ConsensusOutput::BroadcastVidShare(vid_share));
```
twittner, line 1877: ~~Instead of adding another `ConsensusOutput::BroadcastVidShare` and sending two messages I would suggest to make the VID share a parameter of `ConsensusOutput::SendVote1` and also add it to `ConsensusMessage::Vote1`. Having the VID share is a requirement for sending a vote1 and including the share in the vote1 itself is proof that it exists.~~

Sorry, I see the motivation of this PR is to move away from that, so ignore this comment.

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -804,6 +804,20 @@ where
...
+                debug!(%node, %view, "send vid share");
+                let message = Message {
+                    sender: self.public_key.clone(),
+                    message_type: MessageType::Consensus(ConsensusMessage::VidShareBroadcast(
+                        share,
+                    )),
+                };
+                self.network
+                    .sender()
+                    .broadcast(self.consensus.current_view(), &message)
```
twittner: There is now a `self.broadcast` which you might want to use.

mrain: done: [37df0f9](https://github.com/EspressoSystems/espresso-network/pull/4590/commits/37df0f9510c5f4533d09cfa45c033cb5a2ad25f1)

twittner (approved):

PR #4596, opened 2026-06-26
Adopt the view of an embedded QC from timeout votes.

twittner:
If larger than a node's current view this is used to help divergent nodes to get back to a common view.

### DIFF
--- a/crates/hotshot/new-protocol/src/consensus.rs
+++ b/crates/hotshot/new-protocol/src/consensus.rs
@@ -89,6 +89,10 @@ pub enum ConsensusInput<T: NodeType> {
     BlockReconstructed(ViewNumber, VidCommitment2),
     Certificate1(Certificate1<T>),
     Certificate2(Certificate2<T>),
+    /// A quorum certificate allows us to advance our view.
+    ///
+    /// Used to help divergent nodes re-converge on restart.
+    AdvanceView(Certificate1<T>),
     /// Atomic pair emitted by the `EpochRootVoteCollector` for epoch-root views:
     /// a `Certificate1` and its matching `LightClientStateUpdateCertificateV2`.
     /// Consensus never sees an epoch-root Cert1 without the matching state_cert.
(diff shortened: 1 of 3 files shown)


bfish713:
When we restart we use the highest actioned view +  1 as our restart view, which means we view change to whatever view that is which is likely higher than any high QC formed.  This then means nodes may be scattered.  I think it means we should only use actioned view to gate sending votes/proposals (i.e. we force timeouts for any views less than our highest actioned view)

crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -1113,6 +1125,47 @@ impl<T: NodeType> Consensus<T> {
...
+            CertVerification::EpochUnavailable => {
+                debug!(%view, %epoch, "missing epoch to verify cert1");
+                outbox.push_back(ConsensusOutput::RequestDrbResult(epoch));
+                return Protocol::Continue;
+            },
+        }
+
+        self.certs.entry(view).or_insert(cert1);
+
+        let next_view = view + 1;
```
bfish713, line 1161: I think this is the tricky part, a dishonest node could send a malicious timeout vote for an in progress view and trigger our view change to the next view, before we have sent a vote 2 in this view.

bfish713: It might actually be ok if it doesn't cause us to stop actioning in lesser views (I think it doesn't) but we should check

twittner: We could do `self.maybe_vote_2_and_update_lock(view, outbox);` before incrementing the view to ensure we submit any vote2.

bfish713: We might not yet have the requirements to vote at this point.  Block reconstruction may still be in progress, for example

twittner: Yes, this is really only for cases where data is available. The assumption is that if it is missing a later input would trigger the vote.

PR #4598, opened 2026-06-29
fix(consensus): persist high QC and restore lock on restart (V0.5)

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

* `update_high_qc2` is no longer no-op. It now uses the existing `append_high_qc2` to persist the high QC.
* `update_next_epoch_high_qc2` is also no longer no-op and now uses the existing `append_next_epoch_high_qc2`
* `append_next_epoch_high_qc2` is a new method that does what `store_next_epoch_quorum_certificate` did but atomically (using a transaction for sql and a write lock for fs)
* `store_next_epoch_quorum_certificate` has been removed
* `load_consensus_state` now loads the high QC and the next epoch QC from the persistent storage during startup, making sure the high qc and the next epoch qc correspond to each other.
* the locked view is now based on the persisted high qc
* `load_consensus_state` now builds the initializer via `HotShotInitializer::load`, which restores `undecided_leaves`/`undecided_state` from the persisted proposals. Without this, the recovered high QC pointed at leaves consensus couldn't see, so heights between the anchor and the high QC were never re-decided — the cause of the `api::test::test_restart` hang (3×360s timeout → ~17s pass).
* `HotShotInitializer::load` optionally takes the anchor's validated state and delta; the sequencer passes its full genesis state, hotshot tests pass `None` (unchanged behavior).
* `HotShotInitializer` fields are now private — construction goes through `from_genesis`/`load` only, so recovery can never again skip restoring the undecided maps; external readers use new accessors.


lukaszrzasik:
@claude review again

lukaszrzasik:
@claude review again

bfish713 (approved):

PR #4600, opened 2026-06-29
Restart at Locked QC View + 1 and Gate proposing by Timeout view

bfish713:
The goal is that we will restart and sync to the high qc view, then not act in any views after that (vote1 or propose)

Closes #<ISSUE_NUMBER>


twittner (approved):
👍

PR #4608, opened 2026-07-01
fix(shutdown): cooperatively cancel blocking drb computation

sveitser:
I believe this fixes the issue where shutdown is incomplete that I previously encountered when syncing a decaf node.

### DIFF
--- a/crates/espresso/node/src/main.rs
+++ b/crates/espresso/node/src/main.rs
@@ -46,6 +46,9 @@ pub fn main() -> anyhow::Result<()> {
     #[cfg(not(feature = "embedded-db"))]
     {
         let migrated_envs = espresso_utils::env_compat::migrate_legacy_env_vars();
-        tokio::runtime::Runtime::new()?.block_on(espresso_node::main(migrated_envs))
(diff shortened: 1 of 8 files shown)


crates/espresso/node/src/main.rs
```diff
@@ -46,6 +46,9 @@ pub fn main() -> anyhow::Result<()> {
     #[cfg(not(feature = "embedded-db"))]
     {
         let migrated_envs = espresso_utils::env_compat::migrate_legacy_env_vars();
-        tokio::runtime::Runtime::new()?.block_on(espresso_node::main(migrated_envs))
+        let rt = tokio::runtime::Runtime::new()?;
```
imabdulbasit, line 49: can we also add this for sqlite binary

imabdulbasit (approved):

crates/espresso/node/src/main.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/4608/commits/a343f150f30d2811a1b405016df2031694b42369

PR #4610, opened 2026-07-01
Reconnect on panic and rework delay queue.

twittner:
Currently, if a panic occurs when connecting, the messages to retry are lost because the `DelayQueue` that schedules the resending of messages is owned by the peer only, contrary to the message `Queue`, which is shared with the party, held in the server task.

This commit adds the `DelayQueue` to the shared state so that restarting a peer after a panic does not loose messages. It also refactores the `DelayQueue` itself slightly to make it a priority queue rather than scanning items for expiry every n seconds.

Finally, instead of storing the `Connection` in a `Peer` and replacing it with `Peer::set_connection`, the `Connection` and `CancellationToken` are given directly to `Peer::start`, the only scope that matters. If `Peer::start` returns, both values are dropped now immediately.

**Edit**: In https://github.com/EspressoSystems/espresso-network/pull/4610/commits/3e402185414cb6b94fe7071e6697ebe9da6e7c4e we replace `Queue::dequeue` with `Queue::next` and `Queue::remove` to ensure a message is at any point in one queue. The guarantee cliquenet should provide is at-least-once delivery into the remote process, for messages to known parties (assuming the default `RetryPolicy`), until GC passes the slot, the party is removed, or the network is shut down.


crates/cliquenet/src/delay.rs
```diff
@@ -0,0 +1,120 @@
...
+        if let Some(entry) = inner.map.remove(&(s, i)) {
+            inner.due.remove(&(entry.due, s, i));
+        }
+    }
+
+    pub fn reset(&self, now: Instant) {
+        let mut inner = self.inner.lock();
+        inner.due.clear();
+        let mut due = mem::take(&mut inner.due);
+        let t = timeout(&self.conf, now, 0);
```
bfish713: Is it intended to not just use `now` instead of `now + conf.send_retry_delays[0]`?  This defaults to 5s.  The old version had no delay.  I think this means previously retry messages immediately after re-connection and now we delay.

bfish713 (approved):
Looks good besides the one question about the delay after reset

crates/cliquenet/src/delay.rs
twittner: Thanks for catching this. Using `now` is indeed the right way.

PR #4611, opened 2026-07-02
feat(hotshot): widen upgrade proposal window via propose_offset

lukaszrzasik:
Increase DEFAULT_UPGRADE_CONSTANTS offsets (propose_offset 5->20, others shifted +15 to preserve the decide and transition gaps) so an upgrade proposal stays non-stale for ~20 views instead of ~6. This gives validators and the collecting leader a wider window to accumulate 90%-stake votes, addressing the decaf 0.4->0.5 upgrade-certificate stall.

Closes #<ISSUE_NUMBER>


crates/hotshot/types/src/constants.rs
```diff
@@ -47,10 +47,10 @@ pub const EXTERNAL_EVENT_CHANNEL_SIZE: usize = 10_000;
 
 /// Default values for the upgrade constants
 pub const DEFAULT_UPGRADE_CONSTANTS: UpgradeConstants = UpgradeConstants {
-    propose_offset: 5,
-    decide_by_offset: 105,
-    begin_offset: 110,
-    finish_offset: 115,
+    propose_offset: 20,
```
sveitser, line 51: @lukaszrzasik since presumably you know know what these mean, could we add a brief docstring to each of them?

sveitser (approved):

PR #4617, opened 2026-07-02
[new-protocol] decide gap views late; gap-tolerant decide event pipelines

bfish713:
Stack: [restart re-vote] · [cert2 broadcast] · **PR 3/3** ← this

Based on `feat/new-protocol-cert2-broadcast`.

## What

With Cert2s broadcast, one can arrive after a newer view already decided without its full ancestor chain, leaving a gap that previously only catchup could fill. This makes late "gap-fill" decides first-class, and makes the espresso-node event pipelines correct under them.

**Consensus (`hotshot/new-protocol`):**
- `decided_views` tracks genuinely-decided views; `last_decided_view` becomes a forward-only high-water mark. `maybe_decide` fills any still-undecided view above the decide floor (watermark − `DECIDE_BUFFER = 50`, pinned at the restart/cutover anchor).
- Decided GC retains only the decide inputs (proposals, certs, deferred certs, the decided set) down to the floor; leaves, signed proposals, and VID shares keep their old trim — late decides fill payload/VID share best-effort, both already durably persisted.
- The Cert2 relay guard moves to `!decided_views.contains(&view)` so certs for still-undecided gap views are relayed (the exact case the relay serves), while decided views stay unrelayed.
- Both cert handlers drop views at or below the floor at intake, before signature verification — bounding replayed-historical-cert cost and preventing unverified inserts into `pending_certs2` / DRB fetches for dead epochs.
- The floor pin at the restart anchor prevents replayed cert pairs from re-deciding pre-anchor views (`decided_views` is not persisted).
- Timeout GC keeps decide inputs: a timed-out view can still gather a Cert2 (vote2 is deliberately not timeout-barred); only the payload and VID share are freed.
- The deferred-Cert1 retry now attempts the decide (it can be the last missing ingredient), and the decided-view gauge reports the monotone watermark.

**Node (`crates/espresso/node`):** gap-fill decides deliver a `NewDecide` for an older view than already processed; both persistence backends assumed monotone decides and permanently lost the gap block from the event stream (sql: the strictly-forward cursor jumped the height gap via the first-row continuity bypass; fs: the pop-oldest rule discarded the late leaf as already-delivered, then pruned it).
- sql seeds the height-continuity check from the leaf at the cursor and holds the cursor at a fillable gap (bounded: legacy leaves never wait, and a gap the watermark has moved `DECIDE_GAP_FILL_HORIZON = 50` past degrades to the old skip-and-fetch behavior).
- fs replaces the pop-oldest heuristic with a persisted event cursor (view + height of the last delivered leaf), with a fallback for pre-cursor disks.
- The decide processor's watch signal keeps the max view so a gap-fill signal cannot hide a newer, unconsumed tip signal.

## Reviewer focus

- The decide floor unification: `maybe_decide`, both cert intakes, GC, and the relay all gate on `decide_floor()`.
- The sql wait-at-gap condition (`generate_decide_events`): new-protocol leaves only, within the horizon of the watermark, first row only.
- `DECIDE_GAP_FILL_HORIZON` mirrors the crate-private `DECIDE_BUFFER`; a `const` assert ties `DECIDE_BUFFER >= VID_RECONSTRUCT_GC_MARGIN`.

## Tests

- consensus: gap-fill decide of an older view (with relay and forward-only watermark), late Cert2 after view-change GC, no re-decide below the restart anchor.
- persistence (both backends, rstest): a height gap holds the event cursor without loss or duplicates; the gap-fill delivers everything in order exactly once. Both fail against the unfixed backends.
- Note: sql changes ran against SQLite locally (no docker daemon); the postgres jobs in CI exercise the same paths.


crates/espresso/node/src/persistence/fs.rs
```diff
@@ -487,6 +495,32 @@ impl Inner {
         }
     }
 
+    fn load_event_cursor(&self) -> anyhow::Result<Option<(ViewNumber, u64)>> {
```
twittner: Would it make sense to introduce a struct that is serialised? Maybe this cursor gets extended in the future and we want some backwards compatibility?

crates/espresso/node/src/persistence/fs.rs
```diff
@@ -487,6 +495,32 @@ impl Inner {
...
+    }
+
+    fn save_event_cursor(&mut self, view: ViewNumber, height: u64) -> anyhow::Result<()> {
+        let path = self.event_cursor_path();
+        self.replace(
+            &path,
+            |_| Ok(true),
+            |mut file| {
+                file.write_all(&view.u64().to_le_bytes())?;
+                file.write_all(&height.to_le_bytes())?;
```
twittner: Instead of write system calls, maybe the bytes can be created once before `self.replace` is called.

imabdulbasit (approved):

PR #4618, opened 2026-07-02
feat(deployer): add verify-proposal and auto-generate upgrade proposals 

sveitser:
- Add `deploy verify-proposal --contract <kind> --input schedule.json --input
  execute.json`: independently verify a timelock (schedule+execute) upgrade
  proposal against on-chain state and shipped bytecode, without trusting
  Etherscan or the JSON description
- Bytecode check: compare the on-chain impl to the contract binding's
  DEPLOYED_BYTECODE, normalizing trailing solc CBOR metadata and the UUPS
  __self immutable; classify FullMatch / CodeMatchMetaDiffers / Mismatch
- Governance checks: proxy owner()/currentAdmin() == mapped timelock, delay >=
  getMinDelay, value/predecessor zero, init selector, version prerequisite;
  assert schedule and execute share inner payload, salt, value, predecessor
- Compute Safe EIP-712 domain/message/tx hashes so signers can match their Ledger
- deploy writes proposals into contracts/deployments/proposals/<network>/<date>-<slug>/
  (network from chain id; --network/--proposal-slug/--proposals-root overrides):
  schedule.json, execute.json, a contract kind file, and a templated README
- Auto-fill the README Safe hashes: resolve the signing Safe from the
  deployment-info toml and read its nonce on-chain; skip with a warning when the
  signer set is ambiguous or lookup fails
- Add a CI job to verify committed proposals under contracts/deployments/proposals/**


sveitser:
@claude review again

contracts/rust/deployer/src/proposals/deployment_info.rs
```diff
@@ -0,0 +1,215 @@
...
+// ── Raw TOML types ────────────────────────────────────────────────────────────
+
+#[derive(Debug, Deserialize)]
+struct RawDeploymentInfo {
+    ops_timelock: Option<RawTimelockSection>,
+    safe_exit_timelock: Option<RawTimelockSection>,
+    esp_token: Option<RawContractSection>,
+    fee_contract: Option<RawContractSection>,
+    reward_claim: Option<RawContractSection>,
+    stake_table: Option<RawContractSection>,
```
alysiahuggins, line 30: a comment to explain why light client isn't supported at the moment

contracts/rust/deployer/src/proposals/verify.rs
```diff
@@ -0,0 +1,2312 @@
...
+                    format!(
+                        "non-empty init data with no expected selector; selector=0x{} (arbitrary \
+                         delegated call through proxy)",
+                        alloy::hex::encode(&init_data[..4.min(init_data.len())])
+                    ),
+                )
+            }
+        },
+        Some(expected) => {
+            if init_data.is_empty() {
```
alysiahuggins: I'm thinking about a case where there is a function like initializeV3() reinitializer(3) is present. Even though the param list is empty, it still has to be called during upgradeToAndCall so the OpenZeppelin library records that the proxy has been initialized for version 3.

sveitser: @alysiahuggins https://github.com/EspressoSystems/espresso-network/pull/4618/commits/9825ad88dd656916d26218b3b1229049308c8e93

alysiahuggins (approved):

PR #4625, opened 2026-07-03
feat: genesis in config API and metrics

sveitser:


crates/espresso/node/src/snapshots/espresso_node__options__tests__config_node_response_postgres.snap
```diff
@@ -22,6 +22,38 @@ config_peers:
...
+  chain_config:
+    chain_id: "35353"
+    max_block_size: "30720"
+    base_fee: "0"
+    fee_contract: ~
+    fee_recipient: "0x0000000000000000000000000000000000000000"
+    stake_table_contract: ~
+  stake_table:
+    capacity: 10
+  accounts: {}
```
sveitser, line 42: what is this?

crates/espresso/node/src/options.rs
```diff
@@ -687,6 +690,7 @@ pub struct PublicNodeConfig {
     pub config_peers: Option<Vec<Url>>,
     pub is_da: bool,
     pub genesis_file: GenesisSource,
+    pub genesis: Genesis,
```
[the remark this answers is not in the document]
sveitser: i think it's always empty, not sure what we have it for even

lukaszrzasik (approved):

crates/espresso/node/src/snapshots/espresso_node__options__tests__config_node_response_postgres.snap
sveitser: okay we actually support this but i think never use it

PR #4626, opened 2026-07-03
[new-protocol] revert fs decide-event gap hold; add storage latency metrics

bfish713:
## Motivation

After #4617, devnet (new protocol) block times degraded from sub-0.5s to runs of view timeouts, and the base demo native tests started timing out. Leaders log `send proposal` but voters never vote, reporting the proposal missing; a few voters reject proposals with timestamp drift >12s. This points at storage stalling the appends that gate proposal release (`append_proposal` + `record_action(Propose)`) and vote2 release (`append_high_qc2`, `append_vid`, `record_action(Vote)`).

The fs decide-event pipeline from #4617 is the prime suspect on fs nodes:

- `process_decided_events` holds the **exclusive persistence write lock** for the whole pass, and every consensus append needs that same lock.
- Each pass re-reads and deserializes **every retained decided-leaf file** (plus its VID share and DA proposal) before the cursor filter is applied.
- A held gap cursor (including the guaranteed post-restart hold) stops GC, so the retained set — and each pass — keeps growing while decides keep landing. This runs on *every* node, including pure validators with a `NullEventConsumer`, because the pass is also the storage GC driver.

## Changes

1. **Revert the fs half of #4617** (`facf713054a`): drop the persisted `EventCursor` and the gap hold, restoring pop-oldest delivery with eager deletion. A gap-filled leaf is skipped and pruned on fs; consumers re-fetch the missing block (pre-#4617 behavior). The sql pipeline keeps its gap tolerance; the two gap tests are now sql-only.
2. **Add storage latency metrics** (`3f93587a7f8`), since none of the existing metrics can show this failure mode (the `internal_append_*` timers start *after* the fs write lock is acquired):
   - `storage_<op>` histograms in the new-protocol `Storage` wrapper: request→completion per op, including queueing, lock waits, and error-retry loops. GC-aborted tasks record nothing.
   - `internal_process_decided_events_duration` in both backends: full decide pass (generation + GC), with the fs timer started **before** the write-lock acquisition.

Note: nodes running with neither the query nor status API module get `NoMetrics`; enable the `status` module on fs validators to scrape the new histograms.

## Reviewer focus

- The fs revert is a mechanical reverse-apply of the fs.rs hunks of #4617 plus sql-only re-wiring of the two gap tests — confirm nothing else on fs depended on the event cursor.
- In `storage.rs`, timers finish only on the success paths inside the retry loops; early signing-failure returns intentionally record nothing.

## Testing

- fs decide-event tests (`test_non_consecutive_decide`, `test_append_and_decide`, `test_decide_with_failing_event_consumer`, `test_deferred_decide_processing`, fs cases) pass after both commits.
- New-protocol storage tests (7) pass.
- sql-cased tests need docker/postgres — not runnable in my environment, expecting CI to cover them.


crates/hotshot/new-protocol/src/storage.rs
```diff
@@ -77,11 +87,70 @@ impl<T: NodeType> StorageOutput<T> {
...
+            append_high_qc: histogram("append_high_qc"),
+            append_state_cert: histogram("append_state_cert"),
+            append_proposal: histogram("append_proposal"),
+            record_action: histogram("record_action"),
+        }
+    }
+}
+
+/// Records only on [`Self::finish`], so a task aborted by GC leaves no point.
+struct OpTimer {
```
twittner: This looks like a duplicate of `Measurement` (in coordinator metrics). Could this be reused? I did not see `OpTimer::finish` being called outside of `finish_op`. Maybe I missed it.

twittner (approved):
Other than the missing `is_recording` check and the potential to reuse `Measurement` instead of `OpTimer` (unless I missed some detail), this looks good to me.

PR #4634, opened 2026-07-06
fix(contracts): reject infinity BLS points and non-canonical x25519 keys

sveitser:
- BLSSig.verifyBlsSig: revert on infinity G1 signature and all-zero G2 pubkey; the pairing precompile treats both as the point at infinity, so pk=(0,0,0,0) with sig=(0,0) passed verification without any proof of possession
- StakeTableV3: ensureCanonicalX25519Key requires the little-endian key value in (0, 2^255-19), rejecting the 19 encodings the Rust parser refuses and top-bit aliases of the same DH key, so bytes32 uniqueness equals functional key uniqueness
- regenerate bindings and ABIs (new errors, CURVE25519_P getter)


mrain (approved):
LGTM

alysiahuggins (approved):
looks good to me, @philippecamacho or @mrain should take a look before merging

PR #4669, opened 2026-07-10
[new-protocol] verify leaf chain cert2 against the cert2 epoch

imabdulbasit:
leafchain catchup terminates at the first available cert2 at or after the requested height, which can belong to a later epoch than the requested block. Certificates carry their epoch, so `verify_leaf_chain_with_cert2` now takes the EpochMembershipCoordinator and validates certificate against the stake table of the epoch it belong to

### DIFF
--- a/crates/hotshot/new-protocol/src/utils.rs
+++ b/crates/hotshot/new-protocol/src/utils.rs
@@ -1,9 +1,8 @@
-use alloy::primitives::U256;
 use anyhow::{anyhow, ensure};
 use committable::Committable;
 use hotshot_types::{
-    PeerConfig,
     data::Leaf2,
+    epoch_membership::EpochMembershipCoordinator,
     message::UpgradeLock,
     stake_table::StakeTableEntries,
     traits::node_implementation::NodeType,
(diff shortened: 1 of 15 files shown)


crates/hotshot/new-protocol/src/utils.rs
```diff
@@ -35,29 +33,37 @@ pub async fn verify_leaf_chain_with_cert2<T: NodeType>(
     leaf_chain.reverse();
 
     ensure!(!leaf_chain.is_empty(), "empty leaf chain");
+    let newest = &leaf_chain[0];
 
-    let stake_table_entries = StakeTableEntries::<T>::from(stake_table.to_vec()).0;
-
-    cert2.is_valid_cert(&stake_table_entries, success_threshold, upgrade_lock)?;
+    let membership = coordinator
+        .stake_table_for_epoch(Some(cert2.data.epoch))
```
sveitser: @imabdulbasit looks like this is attacker chosen and allows a supermajority from any previous epoch to forge certs for future epochs because we don't seem to check that epoch is correct for the height?

bfish713 (approved):

crates/hotshot/new-protocol/src/utils.rs
bfish713: Yes we should use the epoch for height function to verify the epoch here

PR #4681, opened 2026-07-14
fix(node): release coordinator network before legacy shutdown

bfish713:
## Problem

`test-postgres (10)` and `test-sqlite (10)` have failed deterministically on main since #4670, always on `api::test::test_state_reconstruction::case_1`:

```
called `Result::unwrap()` on an `Err` value: cliquenet: error binding to address 127.0.0.1:44027: Address already in use
```

The test removes a peer and re-creates it with the same node index, so the replacement binds the same pre-assigned cliquenet coordinator port. Since #4670 the coordinator parks unstarted in `ConsensusHandle`'s `Init` state, and `shut_down` disposed of it only **after** the legacy HotShot teardown. The parked coordinator owns the cliquenet network, so its listener held the port for the entire legacy drain (hundreds of ms to seconds), and the replacement node always lost the race. Before #4670 the coordinator loop was cancelled and awaited first, which released the port within microseconds.

## Fix
- `test_state_reconstruction` now awaits the removed node's `shut_down()` before re-binding its coordinator port, matching `test_merklized_state_catchup_on_restart`. With the ordering fix, the port is released microseconds into `shut_down`, long before it returns.

## Verification

`cargo nextest run -p espresso-node --features embedded-db -E 'test(test_state_reconstruction)'` — failed in <3 s on the bind before; now passes (36 s, full 7-epoch run). Clippy clean.

### DIFF
--- a/crates/espresso/node/src/api.rs
+++ b/crates/espresso/node/src/api.rs
@@ -6227,6 +6227,9 @@ mod test {
         // Adding an additional node to the test network is not straight forward,
         // as the keys have already been initialized in the config above.
         // So, we remove this node and re-add it using the same index.
+        // Await the shutdown so the node's cliquenet listener releases its
+        // port before the replacement node binds the same address.
+        network.peers[0].shut_down().await;
         network.peers.remove(0);
 
         let node_0_storage = &storage[1];
--- a/crates/espresso/node/src/consensus_handle.rs
+++ b/crates/espresso/node/src/consensus_handle.rs
@@ -449,17 +449,22 @@ where
         for t in &self.tasks {
             t.abort()
         }
+        // Release the cliquenet listener port before the slow legacy
+        // shutdown; `drop(other)` because a non-binding pattern would not
+        // move the value.
+        let new_proto = self.new_proto.write().take();
+        match new_proto {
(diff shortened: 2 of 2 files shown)


crates/espresso/node/src/api.rs
```diff
@@ -6227,6 +6227,9 @@ mod test {
         // Adding an additional node to the test network is not straight forward,
         // as the keys have already been initialized in the config above.
         // So, we remove this node and re-add it using the same index.
+        // Await the shutdown so the node's cliquenet listener releases its
+        // port before the replacement node binds the same address.
+        network.peers[0].shut_down().await;
```
twittner, line 6232: Isn't this the only required fix? The order within the handle's `shut_down` should not matter. `api::test::test_state_reconstruction::case_1` was not failing deterministically (e.g. CI in https://github.com/EspressoSystems/espresso-network/pull/4670 was green) and not awaiting the shutdown looks like the root cause of the race to bind to the port again.

bfish713: I think this is the only necessary fix, CI will confirm

sveitser (approved):

PR #4684, opened 2026-07-14
fix(test): avoid cliquenet bind race

sveitser:
test_state_reconstruction removes peer 0 then immediately re-added a node at the same index, rebinding the dropped node's cliquenet coordinator address before the OS released the socket. The race fails the faster embedded-db (sqlite) path with "address already in use", postgres seem to pass.

Give the re-added node a fresh coordinator port.

fixes: cliquenet: error binding to address 127.0.0.1:50343: Address already in use (os error 48)


sveitser:
dupe of 
- #4681

PR #4685, opened 2026-07-14
[Fast Finality] Help restarted nodes reach highest view

bfish713:
### This PR:

- Adds a message type for sending `TimeoutCertificates` to other nodes
- Adds `TimeoutCertificate` to timeout votes
- Nodes Send  highest know `TimeoutCertificate` or locked QC to nodes who timeout in a view lower than it's own.  I.e. when we receive a timeout vote for a lower view, we respond with view change evidence to pull that node up to our view
- Nodes attatch their higest TC to timeout votes
- Nodes view change higher views if they receive a valid cert for that view.

### DIFF
--- a/crates/hotshot/new-protocol/src/coordinator.rs
+++ b/crates/hotshot/new-protocol/src/coordinator.rs
@@ -784,11 +784,16 @@ where
                     }
                 }
             },
-            ConsensusOutput::SendTimeoutVote(vote, lock) => {
+            ConsensusOutput::SendTimeoutVote(vote, lock, tc) => {
                 let view = vote.view_number();
-                debug!(%node, %view, has_lock = lock.is_some(), "send timeout vote");
+                debug!(
+                    %node, %view,
+                    has_lock = lock.is_some(),
+                    has_tc = tc.is_some(),
+                    "send timeout vote"
+                );
                 self.broadcast(
-                    ConsensusMessage::TimeoutVote(message::TimeoutVoteMessage { vote, lock }),
+                    ConsensusMessage::TimeoutVote(message::TimeoutVoteMessage { vote, lock, tc }),
                     "broadcast timeout vote",
                 )?
             },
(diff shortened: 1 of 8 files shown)


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -1123,27 +1128,35 @@ where
...
-                    // the highest justified view on restart.
+                    // If a peer times out in a view at or ahead of us we adopt its
+                    // highest certificate.
                     //
-                    // TODO: Above we reject timeout votes too far ahead. A valid cert1
-                    // has precedence over that check so we need to move this up, but
-                    // first we need to verify the cert1 itself.
+                    // TODO: Above we reject timeout votes too far ahead. Valid
+                    // evidence has precedence over that check so we need to move
+                    // this up, but first we need to verify the certs themselves.
```
twittner: About this: Perhaps for testing we should move the certificate handling above the view distance check. When https://github.com/EspressoSystems/espresso-network/pull/4550 gets in we can do the actual certificate verification asynchronously first.

crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -1123,27 +1128,35 @@ where
...
+                    let lock_view = timeout_msg.lock.as_ref().map(|c| c.view_number());
+                    if let Some(tc) = timeout_msg.tc
+                        && tc.view_number() >= current_view
+                        && lock_view.is_none_or(|lv| tc.view_number() >= lv)
```
twittner: Could we not just _send_ the certificate with the higher view?

bfish713: Yes I combined the 2 certs into one evidence enum

crates/hotshot/new-protocol/src/coordinator.rs
bfish713: moved

twittner (approved):

PR #4694, opened 2026-07-16
fix(new-protocol): don't stall legacy-event forwarders pre-cutover

sveitser:
Closes the memory-soak regression introduced by #4670 (1.8 GB -> 7.8 GB total on epoch-reward, drb-header runner OOM).

## Root cause

- #4670 parks the coordinator until the V0_6 upgrade view; while parked, nothing services its `ClientRequest` queue.
- The legacy-event forwarders (`cutover.rs`) spawn at node construction. `forward_legacy_epoch_changes` called `bump_network_epoch`, which sent a request and awaited a oneshot reply. Parked coordinator: no reply, forwarder blocks forever at the first epoch boundary.
- The blocked forwarder holds an active receiver on HotShot's external broadcast stream (capacity 10,000). Broadcast retains events until all receivers consume them, so the stalled receiver pins the newest 10,000 `Decide` events, each carrying Arc'd validated state (fee/reward merkle trees, VID shares). About 1.2 GB per node on any network below V0_6.

## Changes

- `client.rs`: `submit_timeout_vote`, `submit_legacy_high_qc`, `bump_network_epoch` are now synchronous fire-and-forget (`try_send`, `respond` fields removed). Requests queue in the bounded channel until the coordinator starts; `ChannelFull` drops with an error.
- `cutover.rs`: all three forwarders share a `cutover_decided` gate (decided cert with `new_version >= NEW_PROTOCOL_VERSION`). Prevents non-V0_6 upgrade windows (exercised by the `*-upgrade` demo suites) from filling the 256-slot queue with requests the parked coordinator can't drain.
- `coordinator.rs`: dead `respond` plumbing removed from the three handler arms.

## Tests

- `forward_legacy_epoch_changes_never_blocks_on_parked_coordinator` (`tests/cutover.rs`): reproduces the stall on pre-fix code (forwarder never drains, 5s timeout), passes now.
- `forward_legacy_epoch_changes_gates_on_decided_upgrade`: nothing queues before the cert is decided, exactly one `BumpNetworkEpoch` per epoch after.
- `tests::cutover::` 7/7, `tests::legacy_cutover::` 32/32, clippy clean.

## Review focus

- Cutover behavior without the old send-then-await lockstep: handlers are order-insensitive and idempotent (stale votes dropped by view, high-QC registration idempotent, `apply_epoch` monotone); a `Full`-dropped final epoch bump is covered by the seed-time `apply_epoch`.
- Gate lifecycle across restarts: cert is persisted at decide and restored before event processing.

## Verification pending

- CI memory soak on this branch (`workflow_dispatch` runs `memory-soak-non-pr`): expect epoch-reward `Total (sum)` back at ~1.8 GB.

### DIFF
--- a/crates/hotshot/new-protocol/src/client.rs
+++ b/crates/hotshot/new-protocol/src/client.rs
@@ -13,7 +13,7 @@ use hotshot_types::{
     },
     utils::StateAndDelta,
 };
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{mpsc, mpsc::error::TrySendError, oneshot};
 
 use crate::{coordinator::error::CoordinatorError, message::Proposal, state::UpdateLeaf};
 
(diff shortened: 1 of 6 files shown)


sveitser:
This workflow dispatch may run the soak test, it's skipped on PRs now.

https://github.com/EspressoSystems/espresso-network/actions/runs/29491882458

crates/hotshot/new-protocol/src/client.rs
```diff
@@ -137,25 +137,23 @@ impl<T: NodeType> ClientApi<T> {
     }
 
     /// Forward a legacy `TimeoutVote2` into the new-protocol timeout collectors.
-    pub async fn submit_timeout_vote(&self, vote: TimeoutVote2<T>) -> Result<(), QueryError> {
-        let (respond, rx) = oneshot::channel();
-        self.call(ClientRequest::SubmitTimeoutVote { vote, respond }, rx)
-            .await
+    pub fn submit_timeout_vote(&self, vote: TimeoutVote2<T>) -> Result<(), QueryError> {
```
twittner:
```suggestion
    pub fn try_submit_legacy_timeout_vote(&self, vote: TimeoutVote2<T>) -> Result<(), QueryError> {
```

crates/hotshot/new-protocol/src/client.rs
```diff
@@ -137,25 +137,23 @@ impl<T: NodeType> ClientApi<T> {
...
+        self.try_send(ClientRequest::SubmitTimeoutVote { vote })
     }
 
     /// Forward the last legacy view's QC so the first new-protocol leader can
     /// propose on it even if the cutover seed was snapshotted before it formed.
-    pub async fn submit_legacy_high_qc(&self, qc: QuorumCertificate2<T>) -> Result<(), QueryError> {
-        let (respond, rx) = oneshot::channel();
-        self.call(ClientRequest::SubmitLegacyHighQc { qc, respond }, rx)
-            .await
+    pub fn submit_legacy_high_qc(&self, qc: QuorumCertificate2<T>) -> Result<(), QueryError> {
```
twittner:
```suggestion
    pub fn try_submit_legacy_high_qc(&self, qc: QuorumCertificate2<T>) -> Result<(), QueryError> {
```

twittner (approved):

PR #4695, opened 2026-07-16
Verify certificates asynchronously

twittner:
Continuation of https://github.com/EspressoSystems/espresso-network/pull/4550.


twittner:
@claude review

crates/hotshot/new-protocol/src/cert_verifier.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// Submit an item received from the network for verification. If the
+    /// epoch's membership isn't ready the item is held and its epoch returned
+    /// so the caller can drive that epoch's catchup. Duplicates are dropped.
+    pub fn verify(&mut self, cert: C) -> Option<EpochNumber> {
+        let view = cert.view_number();
+
+        if view < self.lower_bound
+            || self.completed.contains(&view)
+            || self.tasks.contains_key(&view)
+            || self.pending.contains_key(&view)
```
bfish713: A byzantine node can send us bad certificates and block good ones from verification.  I think the best thing to do is accept at most 1 cert per key,and  verify 1 cert per view per type at a time.  If we want to verify a cert for a view we are already verifying we queue it (if there is no other cert from that key) if cert verification fails we try the next in the queue if it succeeds we clear the queue for the view

bfish713 (approved):

PR #4708, opened 2026-07-17
refactor(demo): run native demo without Docker

sveitser:
Replace the five Docker-backed process-compose services with native processes so `just demo-native` needs no Docker daemon:

- demo-l1-network: custom geth-l1 image -> native anvil (from foundry)
- espresso-node-db-0/1: docker postgres -> scripts/run-postgres
- keydb: docker keydb -> native redis-server
- telemetry-collector (vector) and block-explorer: dropped

flake.nix adds redis and bumps postgresql_16 -> postgresql_18. cleanup-process-compose drops the docker compose teardown and reaps anvil/postgres/redis-server.

### DIFF
--- a/scripts/run-postgres
+++ b/scripts/run-postgres
@@ -1,14 +1,5 @@
 #!/usr/bin/env bash
-#
-# Start a single native PostgreSQL instance for the process-compose demo.
-#
-# The caller provides a port and a data directory; this script initializes
-# the cluster, starts postgres bound to 127.0.0.1, creates the "espresso"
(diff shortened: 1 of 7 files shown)


scripts/run-postgres
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
imabdulbasit, line 1: I use docker to run postgres. If its not an issue, could we also keep the original process compose file that uses Docker?

imabdulbasit (approved):
I haven't looked through all of the code but since the native demo works, so I'm approving it.

PR #4709, opened 2026-07-17
feat(node): abort startup on wrong or missing x25519 registration

sveitser:
From EPOCH_REWARD_VERSION onward a validator whose stake-table entry lacks connect info, or whose local x25519 key matches neither the contract head nor the three newest epoch stake tables (the window cliquenet dials), fails at startup instead of logging. Key rotation is tolerated in both directions; paths where verification is impossible (RPC failure, proxy < V3, head-fetch timeout) still warn and skip. ESPRESSO_NODE_SKIP_CLIQUENET_REGISTRATION_CHECK skips the check.

Also fixes the dead version gate: the check compared base_version with equality, so it never ran on mainnet (base 0.4, upgrade 0.5).

### DIFF
--- a/crates/espresso/node/src/lib.rs
+++ b/crates/espresso/node/src/lib.rs
@@ -35,6 +35,7 @@ use espresso_types::{
     v0::traits::SequencerPersistence,
     v0_1::{ChainId, DECAF_CHAIN_ID},
     v0_3::Fetcher,
+    validators_from_l1_events,
 };
 
 pub(crate) const MAINNET_CHAIN_ID: ChainId = ChainId(U256::ONE);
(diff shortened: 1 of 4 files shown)


crates/espresso/node/src/lib.rs
```diff
@@ -922,29 +978,76 @@ async fn check_cliquenet_info_registered(
...
-        return;
+
+    let Some(finalized) = l1_client.snapshot().await.finalized else {
+        tracing::warn!("no finalized L1 block yet; skipping network-info registration check");
+        return Ok(());
+    };
+    let head_fetch = async {
+        let events = membership
+            .fetcher()
+            .fetch_and_store_stake_table_events(addr, finalized.number)
```
sveitser, line 998: Don't really like this since this fetch can take a very long time for a new node.

PR #4714, opened 2026-07-17
feat(new-protocol): tear down legacy stack after cutover decides

bfish713:
## Tear down the legacy consensus stack after cutover

Once the coordinator has emitted `LEGACY_SHUTDOWN_DECIDE_COUNT` (100) decide events, `handle_events` triggers `ConsensusHandle::shut_down_legacy()`: the legacy→coordinator forwarders are aborted and the legacy tasks + network are shut down via a new `SystemContextHandle::shut_down_tasks_and_network()`, which — unlike `shut_down()` — leaves in-flight DRB computations on the shared membership coordinator running and legacy in-memory state readable. `CombinedNetworks::shut_down` now also drops its dedup caches and per-view channels since the network object stays alive after the cutover.

**When the base version already starts at the cutover (≥ 0.6), the legacy stack never runs at all**: `start_consensus` tears it down instead of starting legacy consensus, and `init_node` no longer blocks boot waiting for CDN/libp2p readiness (a pure-0.6 network needs no legacy infra deployed). The legacy network objects are still constructed — the production network type is fixed to `CombinedNetworks` — but are shut down before consensus starts. Upgraded networks (base < 0.6) keep the 100-decide window so lagging peers can finish crossing the boundary.

Outbound external (request-response) messages permanently switch to the coordinator's network once it's running. They're framed in the same versioned `Message` envelope as the legacy path because the coordinator's network sends external payloads verbatim and the receiver's fallback decoder expects that framing — bare payload bytes made receivers parse garbage lengths and abort on exabyte-scale allocations (caught by the new test). VID-share fetches (`request_vid_shares`) switch from `RequestType::Broadcast` to `RequestType::Batched` — the last broadcast user — because the coordinator path only carries direct external messages; batched requests reach up to 100 peers within the request's 40s window (5 every 2s), covering the full committee.

**Test**: `test_new_protocol_survives_legacy_shutdown` runs a base-0.6 network (legacy stack torn down at startup on every node), re-issues `shut_down_legacy()` mid-run to cover the decide-count trigger path being a harmless second teardown, and verifies decides continue across two epoch boundaries.

**Review focus**: the external-message framing in `send_via_coordinator`, and whether aborting the three forwarder tasks at teardown can race with a peer still crossing the cutover boundary.

---

Follow-ups tracked separately:
- `versions::decode`/ark-serialize preallocate lengths read off the wire with no bound, so any authenticated peer can abort a process with malformed bytes (pre-existing).
- Constructing the legacy networks at all on base ≥ 0.6 (libp2p binds/bootstraps briefly before teardown) could be avoided with a stub network type if desired.
- The automatic decide-count trigger in `handle_events` is not driven end-to-end by a test (the test calls `shut_down_legacy()` directly); a `#[cfg(test)]`-overridable threshold would allow that if we want the coverage.

### DIFF
--- a/crates/espresso/node/src/external_event_handler.rs
+++ b/crates/espresso/node/src/external_event_handler.rs
@@ -3,8 +3,9 @@
 use std::sync::Arc;
 
 use anyhow::{Context, Result, bail};
-use espresso_types::{PubKey, SeqTypes};
+use espresso_types::{PubKey, SeqTypes, v0::traits::SequencerPersistence};
 use hotshot::types::Message;
+use hotshot_new_protocol::client::ClientApi;
 use hotshot_types::{
     message::MessageKind,
     traits::network::{BroadcastDelay, ConnectedNetwork, Topic, ViewMessage},
@@ -14,7 +15,10 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::{Receiver, Sender, error::TrySendError};
 use vbs::{BinarySerializer, bincode_serializer::BincodeSerializer, version::StaticVersion};
 
-use crate::context::TaskList;
+use crate::{
+    consensus_handle::ConsensusHandle,
+    context::{ConsensusNode, TaskList},
+};
 
 /// An external message that can be sent to or received from a node
 #[derive(Debug, Serialize, Deserialize, Clone)]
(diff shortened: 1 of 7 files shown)


crates/hotshot/hotshot/src/types/handle.rs
```diff
@@ -284,13 +291,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandl
...
-        tracing::error!("Shutting down the network!");
+        tracing::info!("Shutting down the network!");
         self.hotshot.network.shut_down().await;
 
-        tracing::error!("Shutting down network tasks!");
+        tracing::info!("Shutting down network tasks!");
         self.network_registry.shutdown().await;
 
-        tracing::error!("Shutting down consensus!");
+        tracing::info!("Shutting down consensus!");
```
twittner:
```suggestion
        tracing::info!(target: "announce", "Shutting down the network!");
self.hotshot.network.shut_down().await;

        tracing::info!(target: "announce", "Shutting down network tasks!");
self.network_registry.shutdown().await;

        tracing::info!(target: "announce", "Shutting down consensus!");
```

crates/espresso/node/src/external_event_handler.rs
```diff
@@ -82,12 +96,30 @@ impl ExternalEventHandler {
...
+        N: ConnectedNetwork<PubKey>,
+        P: SequencerPersistence,
+    {
+        let mut network = Some(network);
+
         while let Some(message) = receiver.recv().await {
+            // Once the coordinator is running it owns the only live network;
+            // route external messages through it. The coordinator never
+            // stops once started, so drop the legacy network for good.
+            if let Some(client_api) = consensus_handle.client_api().await {
```
twittner: We could save the `ClientApi` in a local variable once we have it, then we do not need to await it again. Perhaps `network` could be `Either<N, ClientApi<SeqTypes>>`?

twittner (approved):

PR #4723, opened 2026-07-20
fix(new-protocol): verify and apply timeout certs once per view

bfish713:
## Summary

On devnet, nodes logged `apply: timeout certificate` multiple times for the same view. Correctness was never affected (consensus dedupes via `timeout_certs`), but the duplicates revealed two real inefficiencies in how certificates flow through the coordinator. The new protocol now runs on mainnet, so the same duplicates happen there.

1. **Duplicate applies.** The coordinator has two sources per certificate type racing in its select loop: the local vote collector (forms a certificate from accumulated votes) and the network verifier (certificates received from peers). Suppression was one-directional — the collector branch marked the network verifier completed, but a network certificate left the local collector running, so its certificate for the same view was applied again a moment later. This held for timeout certificates and, identically, for cert1 and cert2.

2. **Redundant verification.** Every node rebroadcasts a TC when applying it, so during a timeout each node receives ~committee-size copies in a burst. `CertBySenderVerifier` dedupes in-flight work per *sender*, so each copy arriving before the first finished spawned its own >1ms threshold-signature check; all but one result was discarded.

## Changes

- `VoteCollector` gains `mark_completed(view)`: drops the ballot box, discards buffered votes, and stops the tally. A queued tally never starts; one already running can't be cancelled (tokio `spawn_blocking`) and has its certificate discarded by `next()` instead. `signers` is left alone, so timeout diagnostics keep the pre-completion signer set. The coordinator calls it when a network TC, cert1 or cert2 wins the race, mirroring the existing calls in the opposite direction. The epoch-root collector is not suppressed: its state certificate only forms locally, and consensus dedupes the cert1 it carries.
- `CertBySenderVerifier` keeps its per-sender task keying (this is what bounds Byzantine work, since frontier certs have an unbounded view space) but now runs only one verification per view: same-view copies from other senders are parked and promoted in turn only if the in-flight one proves invalid, so a faulty sender still can't shadow a view. The struct doc states the invariant (`parked.keys() ⊆ in_flight.keys()`, one in-flight view per sender) and the resulting O(committee²) bound. Cancelled tasks skip the `Err` arm's reverse lookup, since `mark_completed` and `gc` remove the in-flight entry before aborting.
- A duplicate TC reaching consensus now logs `duplicate timeout certificate; already applied` at debug before the early return.

## For reviewers

Focus on the `CertBySenderVerifier` rework in `cert_verifier.rs`: that parking preserves the per-sender DoS bound (the struct doc spells out the argument), and that `promote_parked` can't strand a view (the failure/panic arms of `next()`, plus cleanup in `mark_completed`/`gc`). The vote.rs and coordinator.rs changes are small mirrors of existing patterns.

Left as is, on purpose: `promote_parked` discards the epoch `verify` returns when a promoted copy is re-deferred for a missing stake table. The table was available moments earlier when the in-flight copy spawned, and `retry_pending` recovers it; `CertVerifier::next` already behaves the same way.

## Testing

- Coordinator-level regression tests for the reported symptom, one per certificate type (`test_network_{tc,cert1,cert2}_suppresses_local_*_collector`): a network certificate followed by threshold votes for the same view yields exactly one `ConsensusInput` for it. All three fail deterministically with the coordinator suppression removed.
- Unit tests: same-view parking, promote-after-invalid, concurrent distinct views, same-sender resubmit, `gc` cleanup, and `mark_completed` semantics (including that its cancellation path promotes nothing) for both `CertBySenderVerifier` and `VoteCollector`.
- Full `hotshot-new-protocol` suite on the merged branch: 280/281. The one failure, `restart_all_nodes_at_epoch_boundary`, passes in isolation and was already flaky under the full parallel run before this rebase. `cargo clippy -p hotshot-new-protocol --all-targets --features testing -- -D warnings` is clean.


bfish713:
Rebased on `main` (one import conflict in `tests/integration.rs`) and addressed the review, by finding:

1. **cert1/cert2 asymmetry** — mirrored in 12aa81d: a network cert1/cert2 now marks `vote1_collector`/`vote2_collector` completed, with a comment on why the epoch-root collector is left alone (its state certificate only forms locally). Regression tests for both, failing deterministically without the change.
2. **Dropped epoch in `promote_parked`** — documented rather than plumbed (d3daf5e): the stake table was available when the in-flight copy spawned, and `retry_pending` recovers a re-deferral, as it already does for `CertVerifier::next`.
3. **`Err` arm reverse lookup** — `is_cancelled()` now short-circuits before it, and the one-in-flight-view-per-sender invariant is stated on the `in_flight` field.
4. **Doc and bound** — the struct doc states `parked.keys() ⊆ in_flight.keys()`, why it holds, and the O(committee²) worst case.
5. **`abort()` accuracy** — the description no longer claims a running tally is cancelled, and notes that `signers` is left in place.
6. **Tests** — gc, same-sender resubmit and the TC regression test landed earlier in eea6067; the cert1/cert2 regression tests are new.

Description updated to match.

crates/hotshot/new-protocol/src/cert_verifier.rs
```diff
@@ -480,6 +552,38 @@ where
...
+            self.verify(sender, cert);
+            if self.in_flight.contains_key(&key) {
+                return;
+            }
+        }
+    }
+
+    fn next_parked_sender(&mut self, key: CompletionKey) -> Option<(T::SignatureKey, C)> {
+        let map = self.parked.get_mut(&key)?;
+        let sender = map.keys().next().cloned()?;
```
twittner, line 580: Maybe we can use [randomness](https://docs.rs/rand/latest/rand/seq/trait.IteratorRandom.html#method.choose) to guard against two byzantine nodes that sit before the others in a hash table and keep sending invalid certificates, blocking others from getting a chance to run the verification.

PR #4731, opened 2026-07-21
Keep only a limited number of `EpochSnapshot`s.

twittner:
Currently, snapshots are only added but never removed, constituting a slow memory leak. Here we limit the number of snapshots ~~and DA committees~~ to the most recent `RECENT_STAKE_TABLES_LIMIT`. Evicted epochs are recovered on demand via the membership coordinator's catch-up.

### DIFF
--- a/crates/espresso/types/src/v0/impls/committee.rs
+++ b/crates/espresso/types/src/v0/impls/committee.rs
@@ -695,7 +695,7 @@ impl Membership<SeqTypes> for EpochCommittees {
     ///
     /// Everything persisted for the epoch is restored: the stake table, the
     /// epoch root header, and the randomized committee (rebuilt from the DRB
-    /// result), so lookups on the restored snapshot need no network catchup.
+    /// result).
     async fn load_stake_table(&self, epoch: EpochNumber) -> bool {
         if self.inner.read().snapshots.contains_key(&epoch) {
             return true;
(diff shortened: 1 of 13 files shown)


imabdulbasit:
One issue with garbage collecting epochs is that if the merklized state loop is far behind, it will require old memberships that may have already been garbage collected. In that case, it will have to spawn catchup requests. I think we should handle this scenario, as archival nodes catching up from genesis will face this issue

crates/espresso/types/src/v0/impls/committee.rs
```diff
@@ -674,6 +691,48 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+        let (validators, block_reward, stake_table_hash) = match loaded {
+            Ok(Some(stake)) => stake,
+            Ok(None) => return false,
+            Err(err) => {
+                warn!(%err, "failed to load stake table for epoch {epoch} from persistence");
+                return false;
+            },
+        };
+        info!(%epoch, "stake table loaded from persistence");
+        self.inner.write().put_epoch_committee(
```
imabdulbasit: I think we should load the randomized committee as well. Garbage collecting the stake tables is fine and the merklized state loop will just reload them from storage, and we only prune once per epoch. But the state loop's STF also needs the randomized committee to lookup leader when calculating the rewards ( I think it was per block rewards) so it will still spawn the catchup requests for DRB

imabdulbasit (approved):

PR #4738, opened 2026-07-22
Branch 20260722

lukaszrzasik:
Closes #<ISSUE_NUMBER>


imabdulbasit (commented):
Can we not pass the coordinator directly like main already does?

bfish713 (approved):

PR #4739, opened 2026-07-22
Verify cross-epoch chain leaf, second iteration

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR:

Important changes (I think):
* adds the fix also to post-ff-upgrade branch in `fetch_leaf`
* adds the fix also to post-ff-upgrade branch in `try_fetch_leaf`
* fetches and waits for the next epoch stake table in `fetch_and_calculate`

### DIFF
--- a/crates/espresso/node/src/catchup.rs
+++ b/crates/espresso/node/src/catchup.rs
@@ -36,11 +36,12 @@ use hotshot_types::{
     message::UpgradeLock,
     network::NetworkConfig,
     simple_certificate::LightClientStateUpdateCertificateV2,
+    stake_table::EpochStakeTables,
     traits::{
         ValidatedState as ValidatedStateTrait,
         metrics::{Counter, CounterFamily, Metrics},
     },
-    utils::{EpochStakeTable, verify_leaf_chain},
+    utils::verify_leaf_chain,
 };
 use itertools::Itertools;
 use jf_merkle_tree_compat::{ForgetableMerkleTreeScheme, MerkleTreeScheme, prelude::MerkleNode};
@@ -366,7 +367,7 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
         &self,
         retry: usize,
         height: u64,
-        stake_tables: Vec<EpochStakeTable<SeqTypes>>,
+        stake_tables: EpochStakeTables<SeqTypes>,
     ) -> anyhow::Result<Leaf2> {
         // Fetch the leaf chain. For new protocol heights this is a leaf range
         // `[height..=cert2_height]`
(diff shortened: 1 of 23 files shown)


crates/espresso/node/src/request_response/catchup/state.rs
```diff
@@ -288,20 +286,9 @@ impl<I: NodeImplementation<SeqTypes>, N: ConnectedNetwork<PubKey>, P: SequencerP
...
-                table.success_threshold,
-                height,
-                &upgrade_lock,
-                cert2,
-            )
-            .await
-            .with_context(|| "leaf chain verification with cert2 failed")?
+            verify_leaf_chain_with_cert2(leaf_chain, &stake_tables, height, &upgrade_lock, cert2)
+                .await
+                .with_context(|| "leaf chain verification with cert2 failed")?
```
lukaszrzasik, line 291: This fixes also the post-ff-upgrade branch.

crates/espresso/node/src/catchup.rs
```diff
@@ -400,20 +401,11 @@ impl<ApiVer: StaticVersionType> StateCatchup for StatePeers<ApiVer> {
...
-                &upgrade_lock,
-                cert2,
-            )
-            .await
-            .with_context(|| format!("failed to verify leaf chain with cert2 at height {height}"))
+            verify_leaf_chain_with_cert2(leaf_chain, &stake_tables, height, &upgrade_lock, cert2)
+                .await
+                .with_context(|| {
+                    format!("failed to verify leaf chain with cert2 at height {height}")
+                })
```
lukaszrzasik, line 408: This fixes also the post-ff-upgrade branch.

crates/espresso/types/src/v0/impls/reward.rs
```diff
@@ -1197,10 +1197,25 @@ impl EpochRewardsCalculator {
...
+            if let Err(err) = coordinator.membership_for_epoch(Some(next_epoch)).await {
+                tracing::info!(
+                    %next_epoch,
+                    "stake table missing for epoch, triggering catchup: {err:#}"
+                );
+                coordinator
+                    .wait_for_catchup(next_epoch)
+                    .await
+                    .context(format!("failed to catch up for epoch={next_epoch}"))?;
+            }
```
lukaszrzasik, line 1213: This might be important. If we don't have the next epoch stake table, we will silently only get the current stake table below. After the change we check if we have the next epoch stake table, fetch it if we don't and wait until we get it. Not sure if waiting here is acceptable.

bfish713: We already wait for catchup above.  I think this calculation is always spawned in a background task so it's actually best to just wait rather than let the task complete by returning an error here.

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -1762,6 +1761,30 @@ impl EpochCommittees {
...
     }
 
+    /// Stake tables for verifying a leaf chain whose expected leaf is in
+    /// `epoch`. A chain deciding a leaf near the end of an epoch contains
+    /// certificates signed by the next epoch's quorum, so include the next
+    /// epoch's stake table whenever we have it.
+    pub(crate) fn leaf_chain_stake_tables(&self, epoch: Epoch) -> EpochStakeTables<SeqTypes> {
+        let mut tables = vec![EpochStakeTable {
+            epoch: Some(epoch),
+            stake_table: self.stake_table(Some(epoch)),
```
bfish713, line 1778: This function returns an empty stake table and will later cause failed signature checks.  I think it's best to return the error here rather than return a bad value

crates/espresso/types/src/v0/impls/stake_table.rs
```diff
@@ -1762,6 +1761,30 @@ impl EpochCommittees {
         self.first_epoch
     }
 
+    /// Stake tables for verifying a leaf chain whose expected leaf is in
+    /// `epoch`. A chain deciding a leaf near the end of an epoch contains
+    /// certificates signed by the next epoch's quorum, so include the next
+    /// epoch's stake table whenever we have it.
+    pub(crate) fn leaf_chain_stake_tables(&self, epoch: Epoch) -> EpochStakeTables<SeqTypes> {
```
bfish713: This should return `Result<EpochStakeTables<SqlTypes>>`

bfish713 (approved):

PR #4740, opened 2026-07-22
fix(catchup): verify each leaf-chain QC against its certified height's epoch

bfish713:
## Problem

Leaf chains fetched in catchup (`StateCatchup::try_fetch_leaf`) were verified against a **single stake table** derived from the requested height, applied to every QC in the chain. This is wrong whenever the chain crosses an epoch boundary — which rewards catchup hits every time, since it always fetches the last block of an epoch and the deciding chain extends into the next epoch.

There was also a subtler anchoring gap: simply switching to the epoch *claimed in each QC* would be circular, because that claim is only attested by whichever quorum signed the QC. A quorum from a stale epoch (e.g. long-exited validators) could fabricate an entire chain claiming their own epoch and it would verify. The legacy (pre-V6) chain has no cert2-style anchor, so the stake-table choice is the only thing grounding it.

## Fix

Validate each QC against the stake table of the epoch **derived from the height of the leaf it certifies** (`epoch_from_block_number`), with `ensure!`s that the QC's signed `epoch` and `block_number` fields match. The certified leaf's height is trustworthy because the QC is bound to it via `leaf_commit == leaf.commit()`, and the requested height pins which epochs are acceptable end to end.

This matches how honest votes are produced in both protocols: legacy votes set the epoch from `epoch_from_block_number(leaf.height())` (`quorum_vote/mod.rs`), and new-protocol votes copy `proposal.epoch`, which is derived the same way (`new-protocol/consensus.rs`). Boundary double-quorum certs live in `next_epoch_justify_qc` and never appear as a chain `justify_qc`, so the invariant holds across epoch transitions.

- **`crates/hotshot/types/src/utils.rs`** (`verify_leaf_chain`, legacy): takes the `EpochMembershipCoordinator` instead of one stake table + threshold; per-QC epoch derivation and validation. The walk now validates each *binding* QC paired with the leaf it certifies, instead of each leaf's own QC (whose target can be outside the chain and therefore can't be anchored). The decide rule is unchanged.
- **`crates/hotshot/new-protocol/src/utils.rs`** (`verify_new_protocol_leaf_chain`): same explicit epoch↔height and block-number↔height checks in the walk. The cert2 anchor from #4669 already made this path sound by induction (leaf commitments cover embedded justify QCs), so this is explicit enforcement rather than a hole being closed.
- **`crates/espresso/node/src/catchup.rs`** (`verify_legacy_leaf_chain`): passes the coordinator through; the single stake-table lookup by requested height is gone.

Both catchup transports (StatePeers and request-response) and their callers (rewards catchup, stake-table catchup) funnel into these two functions, so all consumers are covered.

## Where to focus review

The loop restructure in `verify_leaf_chain`: which QC is validated against which leaf, and that the bottom leaf's dangling justify QC is intentionally no longer checked (its certified leaf isn't in the chain, so its epoch can't be anchored, and it contributes nothing to the authenticity of the returned leaf — which is authenticated entirely by the QCs above it).

## Testing

- `test_fetch_leaf_returns_exact_height` — fetches through `StatePeers` across three epoch boundaries (10-block epochs).
- `test_epoch_stake_table_catchup` (POS_V3 and POS_V4) — stake-table catchup fetching epoch-root leaves through the legacy verifier.


lukaszrzasik (approved):

crates/hotshot/types/src/utils.rs
```diff
@@ -141,31 +168,13 @@ pub async fn verify_leaf_chain<T: NodeType>(
...
-        .justify_qc()
-        .is_valid_cert(&stake_table_entries, success_threshold, upgrade_lock)?;
-    grand_parent.justify_qc().is_valid_cert(
-        &stake_table_entries,
-        success_threshold,
-        upgrade_lock,
-    )?;
-
     // Verify the root is in the chain of decided leaves
+    validate_qc(newest_leaf.justify_qc(), parent.height())?;
```
imabdulbasit, line 172: why don't we check all 3 certs here like before?

PR #4753, opened 2026-07-28
test(node): integration tests for radical stake table changes

bfish713:
Stacked on #4751 and #4752.

Full-application tests driving a real stake table contract on a real L1 (anvil) while the network keeps deciding, at 0.5 and 0.6:

- **Full set replacement**: the committee is replaced wholesale by a disjoint set; the 0.6 case exercises the boundary-state handoff fixed in #4751.
- **Grow and shrink**: the committee doubles when a second cohort registers and shrinks back when it deregisters; at 0.6 the grown committee's first block requires joiner votes.
- **Delegation reshuffle**: undelegation/redelegation alone reshapes the active set, plus a V5 whale phase exercising the minimum-stake threshold (max stake / 1000).
- **Full swaps across upgrades**: 0.4→0.5 and 0.5→0.6, with the swap sent before, at, or after the `UpgradeProposal`; the 0.5 cases also assert per-epoch rewards keep accruing under the replaced committee.
- **Eligibility filter**: a V2-registered validator without connect info stays in the committee through the 0.6 transition window, then drops at the first epoch whose root is a 0.6 header, while the chain stays live on 4/5 of the stake.

## Reviewer focus

- Each test's doc comment explains its scenario shape (why the query node sits in the outgoing set, the `activation_epoch <= 5` cliquenet peer-window bound, the whale threshold math).
- `.config/nextest.toml`: the new tests run in the default profile with `threads-required = 'num-cpus'` (serialized) and generous terminate-after budgets — worth a look at cumulative CI wall-clock.

### DIFF
--- a/slow-tests/tests/stake_table_changes.rs
+++ b/slow-tests/tests/stake_table_changes.rs
@@ -13,8 +13,25 @@ use alloy::{
     providers::{ProviderBuilder, ext::AnvilApi},
 };
 use espresso_contract_deployer::{Contract, upgrade_stake_table_v3};
+use espresso_node::{
+    SequencerApiVersion,
+    api::{
+        Options,
+        data_source::{SequencerDataSource, testing::TestableSequencerDataSource},
+        sql::DataSource as SqlDataSource,
+        test_helpers::{
+            TestNetwork, TestNetworkConfigBuilder, assert_node_live, assert_nodes_agree,
+            committee_is, delegate_new, deregister_validators, register_validators,
+            staking_addresses, wait_for_committee,
+        },
+    },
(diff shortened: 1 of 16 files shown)


slow-tests/tests/stake_table_changes.rs
```diff
@@ -0,0 +1,1419 @@
+//! Integration tests for radical stake table changes, driving a real stake
```
imabdulbasit, line 1: I feel like these tests should be part of`slow-tests` crate. Having them under `src/api` feels a little out of place to me.

imabdulbasit (approved):

PR #4763, opened 2026-08-01
fix(cliquenet): accept authenticated peers from any source IP

bfish713:
## Summary

Remove the source-IP check in cliquenet's inbound accept path. The server rejected connections from known, cryptographically authenticated peers whenever the TCP source IP did not match the party's registered address.

## Why this is safe

- Identity is established by the Noise IK handshake in `Connection::accept`, which authenticates the remote's static x25519 key **before** the party lookup runs. An attacker cannot impersonate a registered key from any IP.
- Promotion to a live peer additionally requires a valid encrypted `Hello` round-trip over the derived session keys, proving live possession of the key (replaying a captured handshake message is not sufficient).
- The check was already skipped for any party registered by hostname (`NetAddr::Name`), so it was never a consistent security boundary.
- It ran *after* the full Noise handshake, so it provided no DoS protection either. The unknown-party rejection (key not in the stake table) is unchanged.

## What it was breaking

Legitimate validators whose egress IP differs from their advertised ingress address were told to back off:

- NAT / cloud setups where outbound traffic leaves via a NAT gateway but the advertised `p2p_addr` is a load balancer or elastic IP
- multi-homed hosts choosing a different source interface
- dual-stack nodes, where an IPv4-mapped IPv6 source (`::ffff:x.x.x.x`) never compares equal to a registered `IpAddr::V4`

The outbound direction is unchanged: when dialing, the initiator still pins the expected key via `remote_public_key`, so we still verify we reached the registered peer.

## Reviewer focus

The IP check came in with #4200 ("Revisiting cliquenet") — flagging in case there was a motivation for it beyond what the history shows.

## Testing

`cargo nextest run -p cliquenet` — 59/59 pass; clippy clean.

### DIFF
--- a/crates/cliquenet/src/net/server.rs
+++ b/crates/cliquenet/src/net/server.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, mem, net::IpAddr, sync::Arc, time::Duration};
+use std::{collections::HashMap, mem, sync::Arc, time::Duration};
 
 use bytes::{Bytes, BytesMut};
 use tokio::{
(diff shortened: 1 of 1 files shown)


twittner:
The intention of the Cliquenet addressing scheme was to provide one way for maximum flexibility (based on DNS) and another one (based on IP address), which pins the network location. While less flexible it allows for defence in depth, because a compromised private key can not simply be copied to another machine and used from there.

I think that the dual-stack story could be improved though. For IPv4-mapped IPv6 addresses we could use `IpAddr::to_canonical`, e.g.

```diff
diff --git i/crates/cliquenet/src/net/server.rs w/crates/cliquenet/src/net/server.rs
index 7858f70c47f..65443b01a77 100644
--- i/crates/cliquenet/src/net/server.rs
+++ w/crates/cliquenet/src/net/server.rs
@@ -772,7 +772,7 @@ impl Party {
         let NetAddr::Inet(ip, _) = &self.addr else {
             return false;
         };
-        *ip != addr
+        ip.to_canonical() != addr.to_canonical()
     }
 }
```

bfish713:
The main reason to add this is because we've seen several instances of operators who have some kind of setup where their advertise IP is not the same as their nodes actual IP address.  Ideally this would be a use-case for DNS, but some don't have that setup.  The question is whether to require DNS in those situations or allow the different IPs

twittner:
> The question is whether to require DNS in those situations or allow the different IPs

My vote would be to require DNS which allows for that flexibility. Removing the IP address check on the other hand removes a feature that can not be achieved by other means.

PR #4764, opened 2026-08-03
Enable state pruning for archival nodes

imabdulbasit:
Archive nodes reconstruct pruned data rather than pruning it, so they never run the pruner and nothing bounds the merklized state tables, the largest in the database.

- Archive nodes now prune merklized state below the oldest height the L1 light client still holds a commitment for. Below that cutoff a block merkle proof can no longer be verified against L1, and anything pruned replays from the leaves an archive node keeps forever.
- `ESPRESSO_NODE_ARCHIVE_STATE_MIN_RETENTION` floors the cutoff, so a contract still holding commitments from a chain that was re-genesised underneath it cannot order the new chain's state deleted. It normally sits well inside the light client window and does not bind.
- `ESPRESSO_NODE_ARCHIVE_FULL_STATE` retains everything, as before.
- Consensus data is never pruned, and archive startup no longer clears the state pruned height.
- The batch loop lives in `SqlStorage::prune_state_below`, shared with the pruner's own state path. State deletions and the pruned marker now commit in one transaction, so a reader cannot see state missing without the marker that says it is gone.

Nodes running the pruner are unaffected: they continue to prune state by age.

On an existing archive node the first pass has a large backlog to work through. It runs on the pruner interval, in batches of `ESPRESSO_NODE_PRUNER_BATCH_SIZE`.

Review focus: the cutoff arithmetic in `prune_state_below`, where the marker must never pass the state head that the state writer resumes from, and the contract indexing in `light_client_history_start`.


crates/espresso/node/src/api/options.rs
```diff
@@ -575,6 +576,16 @@ impl Options {
             update_state_storage_loop(ds.clone(), get_node_state),
         );
 
+        // Archive mode disables the pruner, so nothing else bounds the merklized state tables.
```
sveitser, line 579: @imabdulbasit what's the point of making it configurable then? If it's useless we should just always delete it.

imabdulbasit: I would like to keep it optional at first, so if anything goes wrong then we can just disable it via an env variable. We hav had issues before with pruning so I would prefer to keep it optional

sveitser (approved):
LGTM

PR #4783, opened 2026-08-05
fix(telemetry): stop logging L1 provider credentials

sveitser:
Operator L1 RPC API keys, embedded in the provider URL path, were shipped in plaintext to the telemetry S3 bucket. Two forms leaked, sharing no common substring:

- `reqwest::Error` formats the full request URL into its own `Debug` and `Display`, so any rendering of an L1 transport error emitted it. `{err:#}` is not a safe alternative to `{err:?}` here; both print the URL.
- `url::Url`'s own `Debug` is field-wise (`path: "/v2/KEY"`) and contains no `://` at all, so it is invisible to URL-shaped text matching. It reached logs via the startup struct dump of `HotShotInitializer`.

Fixes:

- `espresso_utils::redact` provides `scrub` for rendered text, handling both forms, and `redact_url` for a `Url` in hand.
- `RedactingLogExporter` wraps the OTLP `LogExporter` and scrubs every record body and attribute. One boundary rather than one call site per log line, so a new log line cannot reintroduce the leak.
- `#[debug(...)]` on the `L1ClientOptions`, `SwitchingTransport` and `SingleTransport` URL fields keeps the field-wise form out of any dump. `SingleTransport` now carries the redacted URL, so the "L1 client error" warning names the provider that failed.
- `Redacted` at the ten L1-backed handlers in the node API, whose error bodies are served to unauthenticated callers and never pass the exporter.

Operator-local stdout is deliberately left unredacted: it is the operator's own credential on their own host, and every Rust program logging a `reqwest` error behaves the same way.

The redaction boundary is fail-closed. `ends_url_token` may only contain characters `Url::as_str()` percent-encodes in every component; anything else is legal in a path and would end the token early, leaking the tail.


sveitser:
@claude review again and check what has been fixed

lukaszrzasik (commented):
I think the biggest problem is that only the tide-disco API responses are redacted and tide-disco as of this moment is disabled.

crates/espresso/utils/src/redact.rs
```diff
@@ -0,0 +1,379 @@
...
+impl<T: fmt::Display> fmt::Display for Redacted<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let rendered = if f.alternate() {
+            format!("{:#}", self.0)
+        } else {
+            format!("{}", self.0)
+        };
+        f.write_str(&scrub(&rendered))
+    }
+}
```
lukaszrzasik: This is dead code now, is that right?

crates/espresso/utils/src/redact.rs
```diff
@@ -0,0 +1,379 @@
...
+        // Userinfo is a credential: `WsConnect` lifts `wss://user:pass@host` into an Authorization
+        // header. Keep only what follows the last `@`.
+        match text[auth_start..auth_end].rsplit_once('@') {
+            Some((_, host)) => {
+                out.push_str(REDACTED);
+                out.push('@');
+                out.push_str(host);
+            },
+            None => out.push_str(&text[auth_start..auth_end]),
+        }
```
lukaszrzasik: this can leak the password if we received a string that had failed url-parsing. This might be quite tricky to do properly.

Examples:
```
  IN : "https://user:pa/ss@host/x"      (password is "pa/ss")                                                                                              
  OUT: "https://user:pa/***"            ← "user:pa" — username + password prefix — emitted                                                                 
                                                                                                                                                           
  IN : "https://user:pa ss@host/x"      (password contains a space)                                                                                        
  OUT: "https://user:pa ss@host/x"      ← completely unchanged                                                                                             
                                                                                                                                                           
  IN : "https://u:p<q@host/x"           (password contains '<')                                                                                            
  OUT: "https://u:p<q@host/x"           ← completely unchanged                                                                                             
```

  Why the doc's defense doesn't hold here: the comment at `redact.rs:13-16` argues these characters can't occur inside a URL because `Url::as_str()` percent-encodes them. That's true for strings rendered from a parsed Url — but `scrub()` is the backstop for arbitrary rendered error text, and raw operator input reaches it without ever being normalized. Concrete path: `crates/espresso/node/src/genesis.rs:36` does `Url::parse(s).with_context(|| format!("invalid genesis URL: {s}"))` — the parse fails precisely because of the space, so the raw credentialed string is echoed verbatim into the error, which then flows to the OTLP exporter or an API error body where scrub() was supposed to be the last line of defense.

sveitser:
@claude this has been simplified to handle only patterns that seems to exist, re-review entire PR

lukaszrzasik (approved):

PR #4784, opened 2026-08-05
feat(api): add leader lookup endpoints for arbitrary views

sveitser:
Why:

- Attributing a view that timed out to the validator that should have proposed it currently requires either opt-in telemetry with 30 day retention, or reimplementing weighted-random leader selection and the DRB outside the node. The node can just answer.
- The leader is a pure function of the view, the epoch's stake table and the epoch's DRB result, so any past view is answerable with no new state and no per-view records.
- Telemetry also misattributes views near an epoch boundary: view_sync resolves the leader against the logging node's own cur_epoch, so a lagging node reports a leader from the wrong epoch. Resolving at query time from the decided chain gets it right.

What:

- GET /v1/node/leader/{view}
- GET /v1/node/leaders/{from}/{until}, inclusive, at most 10000 views
- The epoch of a view is the epoch of the first leaf decided at or after it. A view that produced no block did not advance the block height, so that leaf carries the height, and therefore the epoch, the view would have had. This matches how consensus derives a proposal's epoch from the proposed leaf's height.
- Resolution bisects epochs, not leaves. Epoch first-block heights are arithmetic, so log2(block height / epoch_height) reads suffice and always hit the same small set of rows, which stays cached. Bisecting leaves would cost log2(block height) cold reads per request.
- Reads stay within the retained height window, since pruning deletes a contiguous prefix. Views below it return 404 naming the window.
- A range crossing an epoch boundary is truncated there. Each entry carries its own view, so the caller resumes from the last one plus one.

### DIFF
--- a/crates/espresso/api/examples/test_api.rs
+++ b/crates/espresso/api/examples/test_api.rs
@@ -442,6 +442,7 @@ impl v1::NodeApi for TestApi {
     type BlockReward = serde_json::Value;
     type Block = serde_json::Value;
     type Leaf = serde_json::Value;
+    type ViewLeader = serde_json::Value;
 
     async fn block_height(&self) -> Result<u64> {
         Ok(0)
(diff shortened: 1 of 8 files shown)


sveitser:
I think I can do without this for now. closing.

PR #4822, opened 2026-08-17
fix(cliquenet-types): reject malformed bracketed p2p addresses

sveitser:
NetAddr::from_str previously parsed `[`-prefixed strings without a bracketed sockaddr shape (e.g. "[2001:db8::1:9000") as a hostname with port 0. Such strings now return an error, as do other malformed bracketed forms like "[host]:1234:5678".

This deliberately diverges from StakeTableV3.validateP2pAddr, which accepts these strings: the stake table fetcher degrades a parse error to p2p_addr = None with a warning, so an affected validator is ineligible at V0_6 until it submits a valid address via updateP2pAddr. The solidity/rust equivalence proptest exempts `[`-prefixed inputs accordingly.

Replay safety: no such address exists on chain today. All 35 P2pAddrUpdated payloads on decaf are plain host:port and mainnet still runs StakeTable V2, so old and new nodes derive identical stake tables from current L1 data. Deploy before the mainnet V3 upgrade executes.

### DIFF
--- a/contracts/rust/adapter/src/stake_table.rs
+++ b/contracts/rust/adapter/src/stake_table.rs
@@ -441,7 +441,9 @@ mod test {
 // Solidity `validateP2pAddr` and Rust `NetAddr::from_str` both parse `host:port` strings.
 // If Solidity accepts an address that Rust rejects, the validator's p2p address is silently
 // dropped and they become unreachable for cliquenet. This proptest deploys StakeTableV3
-// on anvil and fuzzes the property: Solidity accepts => Rust accepts.
+// on anvil and fuzzes the property: Solidity accepts => Rust accepts, except the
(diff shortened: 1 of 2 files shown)


twittner (commented):
I think it would be nice if we could enforce the following format:

```
`host:port` where host = IPv4 literal | "[" IPv6 literal "]" | hostname
```

with `hostname` as dot-separated ASCII letters, digits, `-` and `_` (not starting or ending with `-`). Currently the parser is still too lenient. It accepts for example

- `[]:7` -- an empty hostname
- `[[a]]:5` -- resulting in `[a]` as the hostname which is not valid.
- `[::1]:80]:90` -- another malformed hostname
- `a:080`, or `a:+80` -- non-canonical ports.
- ...

PR #4824, opened 2026-08-18
Begin formalising consensus and check conformance

twittner:
Add the begginings of a Lean formalisation of the new-protocol consensus core, and the machinery to hold the implementation to it.

The `lean` folder holds four packages.

- `new-protocol-spec` states what one node may do (`SafetySpec`, `StepSpec`) and what it may prune (`GcSpec`), and proves the results the protocol exists for, chiefly `decideSafety`: the Cert2-certified blocks form a single chain.
- `new-protocol-impl` is an executable machine proved to satisfy that specification.
- `new-protocol-diff` replays a recorded run against the machine and reports where the two part company.
- `new-protocol-docs` renders a reference from the specification's own declarations.

There are no sorries and no axioms beyond Lean's own, and `Checks.lean` re-checks that on every build, along with the clause and premise lists the prose claims.

On the Rust side, `trace.rs` records a node's inputs and the outputs they drew when the environment variable `NP_TRACE_DIR` is set, and is inert otherwise. `tests/random.rs` drives one node with a seeded random schedule over a hundred views and asserts nothing about the outcome; what checks the run is the recorded trace, replayed against the machine. `just test-lean-diff` records the suite and replays it, or replays an
existing corpus given a directory.

_One behaviour change_: `Consensus::gc` pruned `voted_2_views` at the current view while a parked vote survived, so a re-entry could cast the same vote2 twice. It now prunes at the decide floor.

The specification currently covers one static committee of crash-free nodes. Epochs, restart, block fetching and the network are outside it; the reference inventories what is left out and what each omission costs.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


sveitser (approved):
I only reviewed the rust code and that looks fine to me.

I would recommend running claude to declutter the comments, I found most of them not very helpful.

PR #4825, opened 2026-08-18
fix(stake-table): persist the epoch root header on both protocols

sveitser:
`Storage::store_epoch_root` only fires on the legacy decide path, in `decide_epoch_root`. Once a node cuts over to the V0_6 stack that path is torn down, so the new protocol reaches epoch roots through `EpochManager::handle_leaf_decided` -> `add_epoch_root` and never persists the header. `epoch_drb_and_root.block_header` stays NULL forever on a V0_6-native node, so `load_epoch_root` always returns None.

Nothing breaks outright, because every consumer falls back to fetching from peers, but restart recovery and dynamic block reward calculation quietly gain a hard dependency on reachable peers, and `load_start_epoch_info`'s header branch becomes dead code.

Introduced by df7a0ccdf34 (#4456), which replicated the membership half of `decide_epoch_root` and dropped the storage half.

`MembershipPersistence` already declared `load_epoch_root` with no matching store, which is the asymmetry that allowed this. Move `store_epoch_root` down beside it and call it from `add_epoch_root`, the one function both protocols funnel through, which already computes the same `+ 2` epoch key. On legacy nodes this is a second write of an identical header to the same key, which both the SQL upsert and the FS replace handle idempotently.

### DIFF
--- a/crates/espresso/node/src/persistence/fs.rs
+++ b/crates/espresso/node/src/persistence/fs.rs
@@ -1387,37 +1387,6 @@ impl SequencerPersistence for Persistence {
         )
     }
 
-    async fn store_epoch_root(
-        &self,
-        epoch: EpochNumber,
-        block_header: <SeqTypes as NodeType>::BlockHeader,
-    ) -> anyhow::Result<()> {
-        let mut inner = self.inner.write().await;
-        let dir_path = inner.epoch_root_block_header_dir_path();
-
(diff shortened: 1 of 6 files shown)


imabdulbasit (approved):
I checked by running the native demo. Without the PR, the block header is `NULL` and with the PR, it is populated.

PR #4841, opened 2026-08-20
feat(api): generate v2 from protobuf, serve status and token only

imabdulbasit:
The v2 API is now defined entirely in protobuf. Each rpc in crates/espresso/api/proto/v2/*.proto is the single definition site: the signature gives the request and response types, the google.api.http option gives the route, and the comments become the published documentation. From those files build.rs generates the message types, the tonic server traits, the Axum handlers that transcode HTTP/JSON onto them, and the OpenAPI document, so one trait impl serves both transports and an endpoint's URL is only ever edited in the proto. This replaces the hand-written adapter and a separate crate, and it fixes the JSON encoding, which was previously whatever prost's field layout plus a blanket serde derive happened to produce.

- JSON is canonical protoJSON (pbjson). Previously bytes serialized as an array of JSON numbers and uint64 as a bare number, silently losing precision above 2^53 in JS clients. Standard protobuf tooling can now generate a working client. tests/proto_json.rs pins the encoding.
- Deletes api/src/tonic.rs, api/src/v2/, api/src/handlers.rs, the crates/serialization/api crate, and the test_api example. Net 6,219 deletions against 4,970 insertions across 44 files, most of the insertions being committed generated code.
- Serves status and token only, nine endpoints. Everything else clients need is still on v1, so nothing loses functionality. Every route in the OpenAPI document is a route serve_axum mounts, asserted as an equality in axum.rs.
- OpenAPI document served at /v2/docs/openapi.json, with Swagger at /v2 and Scalar at /v2/scalar. The generator fails the build rather than publishing a document it cannot describe (nested request messages, duplicate operationIds or routes, colliding message names).
- One error boundary for both transports. v1 and v2 share error::classify, so they cannot drift on what counts as a 404, and to_status scrubs L1 provider credentials out of messages that reach unauthenticated callers through grpc-message and REST bodies. Query-parameter rejections are rewritten into the same Google error envelope.
- CI now checks the committed src/generated/ matches the protos. The fmt job never builds and the clippy job never diffed, so a stale artifact could previously merge silently.


API.md
```diff
@@ -1,18 +1,135 @@
...
+   the service, its rpcs, and their `google.api.http` options.
+2. Regenerate as above.
+3. Implement the generated `<name>_service_server::<Name>Service` trait on `NodeApiStateImpl`.
+4. Wire the transports in `crates/espresso/api/src/lib.rs`: add the trait bound to `serve_axum` and `serve_tonic`, merge
+   `rest::<name>_service_rest_router(...)` in `serve_axum`, and `add_service` the tonic server in `serve_tonic`.
+
+### Rules and caveats
+
+- Field and rpc numbers are frozen once merged. Only make additive changes: new fields, new rpcs, new messages. Never
+  renumber, reuse, or change the type of an existing field.
```
sveitser, line 98: I think this should be once released. Not once merged into main.

sveitser (approved):

.github/workflows/lint.yml
```diff
@@ -66,6 +66,12 @@ jobs:
         run: |
           cargo clippy --workspace --exclude espresso-node-sqlite --exclude espresso-dev-node --features testing --all-targets --keep-going -- -D warnings
 
+      # The v2 API artifacts under src/generated are committed for review visibility and rewritten
+      # by espresso-api's build script, which the clippy run above has just executed. The fmt job
+      # cannot catch a stale one because it never builds.
+      - name: Check generated API code is up to date
```
sveitser: @imabdulbasit please make this a separate job so that gets reported as a granular failure.

sveitser (approved):

PR #4842, opened 2026-08-20
fix(catchup): retry V2 reward merkle tree fetches and stop latching failed epoch reward calculations

bfish713:
## Problem

Consensus catchup of the V2 reward merkle tree — the only catchup path used by V5/V6 epoch-boundary reward calculation — effectively had **no retries**, and a failed background rewards calculation **latched its error** until the epoch boundary. Together these meant a single transient failure (a 2s per-peer timeout on the largest catchup payload, a peer that hadn't persisted the boundary tree yet, semaphore contention) could fail header validation at an epoch boundary. Nodes with sparse reward trees (recently restarted or caught-up) would then miss the boundary vote every epoch; with many such nodes at once (rolling restart/upgrade), this could stall consensus at the boundary.

## Changes

**Commit 1 — `fix(catchup): retry reward merkle tree v2 fetches from remote peers`**

`ParallelStateCatchup` overrides every retrying `fetch_*` method *except* `fetch_reward_merkle_tree_v2`, so the trait default ran with the parallel provider's disabled backoff: exactly one attempt, `retry` pinned at 0, per-peer timeout stuck at the 2s base. Add the missing override, matching `fetch_accounts` / `fetch_reward_accounts_v1`: one local attempt, then delegate to the remote providers' own retrying fetch (StatePeers retries with the configured `catchup_backoff` and an escalating per-peer timeout; the request/response protocol retries via `request_indefinitely`).

**Commit 2 — `fix(rewards): recalculate failed epoch rewards instead of latching the error`**

A background epoch rewards task that failed stayed parked in `EpochRewardsCalculator` until the boundary: `is_calculating()` treated the dead task as in-progress so it was never respawned, and at the boundary `get_result()` propagated the stale error, failing validation even when a fresh calculation would succeed. Now:

- `pending` holds `Running(JoinHandle)` / `Ready(result)`; new `reap_finished_task()` consumes a finished task's outcome without blocking — caching a success for the boundary, clearing a failure so it can be respawned.
- The eager per-block spawn in `handle_epoch_rewards` reaps first, so a failed attempt is retried on the next block instead of latching all epoch.
- At the boundary, a failed result is treated as missing and falls into the existing respawn-and-await fallback.

## Where to focus

- The new `ParallelStateCatchup::fetch_reward_merkle_tree_v2` override is line-for-line parallel to `fetch_reward_accounts_v1` directly below it.
- `reap_finished_task` in `reward.rs`: the `Running`/`Ready` split exists because a `JoinHandle` outcome can only be consumed once — confirm the caching semantics (a reaped success still counts as `is_calculating` and is delivered by `get_result`).
- Boundary error handling in `header.rs::handle_epoch_rewards`: `Some(Err)` → warn + fall through to the respawn branch.

## Testing

- New regression test `test_reward_merkle_tree_v2_fetch_retries`: a mock remote provider fails once then succeeds; verified the test fails without the override (one attempt, fatal) and passes with it.
- New unit tests for the calculator: reaped failure is cleared and respawnable; reaped success is cached and returned by `get_result`; a running task is left untouched.
- `espresso-types` suite: 256 passed; the 2 failing `l1::test::test_failover_*` tests also fail on a clean `main` checkout in this environment (need real network access) — unrelated.

## Follow-ups (not in this PR)

- Peer scoring: `StatePeers::try_fetch_reward_merkle_tree_v2` verifies the response outside the per-peer loop, so a peer serving a stale tree is scored as a success and retries keep hitting it first.
- The epoch path requests the tree with `ViewNumber::new(0)`, so the consensus-memory endpoint on peers always misses and only the storage fallback can serve.

### DIFF
--- a/crates/espresso/types/src/v0/impls/header.rs
+++ b/crates/espresso/types/src/v0/impls/header.rs
@@ -866,10 +866,10 @@ impl Header {
     ///   validation.
     ///
     /// If the previous epoch's result is missing at the boundary (e.g. after a
-    /// restart), the function spawns the calculation and awaits it before
-    /// applying. The background task fetches the epoch's leaf and recovers the
-    /// leader counts and stake table itself, returning zero rewards for epochs
-    /// whose header version is < V5
+    /// restart) or the background attempt failed, the function spawns a fresh
+    /// calculation and awaits it before applying. The background task fetches
+    /// the epoch's leaf and recovers the leader counts and stake table itself,
+    /// returning zero rewards for epochs whose header version is < V5
     ///
     /// # Returns
     /// `(total_rewards_applied, changed_accounts)` — the total reward amount
@@ -902,16 +902,28 @@ impl Header {
 
         // Eagerly start the previous epoch's reward calculation if it hasn't
         // been kicked off yet, so the result is ready by the epoch boundary.
-        if epoch > first_epoch + 2 && !reward_calculator.is_calculating(prev_epoch) {
-            tracing::info!(%epoch, %prev_epoch, "triggering catchup reward calculation");
-            reward_calculator.spawn_background_task(
-                prev_epoch,
-                epoch_height,
-                validated_state.reward_merkle_tree_v2.clone(),
-                instance_state.clone(),
-                coordinator.clone(),
-                None,
-            );
+        // A background attempt that already failed is cleared here so it is
+        // retried on the next block instead of staying latched until the
+        // boundary.
+        if epoch > first_epoch + 2 {
+            if let Some(err) = reward_calculator.reap_finished_task(prev_epoch).await {
+                tracing::warn!(
+                    %epoch,
+                    %prev_epoch,
+                    "background epoch rewards calculation failed, respawning: {err:#}"
+                );
+            }
+            if !reward_calculator.is_calculating(prev_epoch) {
+                tracing::info!(%epoch, %prev_epoch, "triggering catchup reward calculation");
+                reward_calculator.spawn_background_task(
+                    prev_epoch,
+                    epoch_height,
+                    validated_state.reward_merkle_tree_v2.clone(),
+                    instance_state.clone(),
+                    coordinator.clone(),
+                    None,
+                );
+            }
         }
 
         if !is_last_block(height, epoch_height) {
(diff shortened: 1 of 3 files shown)


crates/espresso/types/src/v0/impls/reward.rs
```diff
@@ -1120,6 +1134,37 @@ impl EpochRewardsCalculator {
...
+        let finished = matches!(
+            &self.pending,
+            Some((e, CalcTask::Running(handle))) if *e == epoch && handle.is_finished()
+        );
+        if !finished {
+            return None;
+        }
+
+        let Some((pending_epoch, CalcTask::Running(handle))) = self.pending.take() else {
+            unreachable!("pending was just matched as a finished running task");
```
imabdulbasit, line 1154: I think this `unreachable` can be avoided if we take out the pending task at the start of this function and then match on it?

crates/espresso/types/src/v0/impls/header.rs
```diff
@@ -902,16 +902,28 @@ impl Header {
...
-            );
+        // A background attempt that already failed is cleared here so it is
+        // retried on the next block instead of staying latched until the
+        // boundary.
+        if epoch > first_epoch + 2 {
+            if let Some(err) = reward_calculator.reap_finished_task(prev_epoch).await {
+                tracing::warn!(
+                    %epoch,
+                    %prev_epoch,
+                    "background epoch rewards calculation failed, respawning: {err:#}"
```
imabdulbasit, line 913: make the error as a field to easily view in datadog?

crates/espresso/types/src/v0/impls/header.rs
```diff
@@ -920,9 +932,21 @@ impl Header {
...
+        // Take the result. A background task that failed is treated as missing
+        // so the fallback below recalculates fresh, instead of failing
+        // validation on an error latched earlier in the epoch.
+        let result = match reward_calculator.get_result(prev_epoch).await {
+            Some(Ok(result)) => Some(result),
+            Some(Err(err)) => {
+                tracing::warn!(
+                    %epoch,
+                    %prev_epoch,
+                    "background epoch rewards calculation failed, recalculating: {err:#}"
```
imabdulbasit, line 944: same here

crates/espresso/types/src/v0/impls/header.rs
```diff
@@ -902,16 +902,28 @@ impl Header {
...
+        // boundary.
+        if epoch > first_epoch + 2 {
+            if let Some(err) = reward_calculator.reap_finished_task(prev_epoch).await {
+                tracing::warn!(
+                    %epoch,
+                    %prev_epoch,
+                    "background epoch rewards calculation failed, respawning: {err:#}"
+                );
+            }
+            if !reward_calculator.is_calculating(prev_epoch) {
```
imabdulbasit, line 916: this `is_calculating` function now returns true for the finished task as well, so I think it makes more sense to rename this function to has_pending() or something similar?

crates/espresso/types/src/v0/impls/reward.rs
```diff
@@ -1096,13 +1105,18 @@ impl EpochRewardsCalculator {
         &mut self,
         epoch: EpochNumber,
     ) -> Option<anyhow::Result<EpochRewardsResult>> {
-        let (pending_epoch, handle) = self.pending.take()?;
+        let (pending_epoch, task) = self.pending.take()?;
```
imabdulbasit, line 1108: I think one problem here is that we take the task out of pending before awaiting it `(handle.await)`. If this `get_result() `future is dropped at that await, the JoinHandle is dropped too, which detaches the task instead of aborting it, so it keeps retrying until it can fetch the tree. And since pending is already None, the next call spawns the calculation again

PR #4844, opened 2026-08-21
fix(stake-table): store the epoch root header before the early return

sveitser:
`add_epoch_root` wrote the epoch root header after `put_epoch_committee` had already committed the in-memory committee with `header: Some(_)`. A failed write is logged and dropped, and from that point the epoch is stuck: every later `add_epoch_root` short-circuits at the early return, since `block_reward`, `header` and `stake_table_hash` are all `Some`, and catchup never re-enters `fetch_stake_table` because `membership_for_epoch` now succeeds. `epoch_drb_and_root.block_header` stays NULL for that epoch permanently, which is the symptom #4825 set out to fix.

Move the write to just after the `is_epoch_root` guard, in its own lock scope. Any subsequent `add_epoch_root` for the epoch then retries it, and the ordering matches the legacy path in `hotshot-task-impls` `decide_epoch_root`: unconditional, first, keyed off nothing but the header. Nothing between the guard and the old site contributes to the header's validity.

Follow-up to #4825.

### DIFF
--- a/crates/espresso/types/src/v0/impls/committee.rs
+++ b/crates/espresso/types/src/v0/impls/committee.rs
@@ -810,6 +810,24 @@ impl Membership<SeqTypes> for EpochCommittees {
             return Err(Self::Error::NoRootBlock(block_number.into()));
         }
 
+        // Stored ahead of the early return below so a failed write is retried by the next
+        // `add_epoch_root` for this epoch. Downstream of `put_epoch_committee` the committee
+        // is complete, which both arms that early return and stops catchup from re-entering
+        // `fetch_stake_table`, so one dropped error would leave the header unpersisted.
+        //
+        // Keyed by `epoch`, unlike the previous-epoch stake table stored further down, and
+        // not gated behind the stake table consistency check, which doesn't apply here.
+        if let Err(e) = self
+            .fetcher
+            .persistence
+            .lock()
+            .await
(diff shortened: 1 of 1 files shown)


imabdulbasit (approved):

crates/espresso/types/src/v0/impls/committee.rs
```diff
@@ -810,6 +810,24 @@ impl Membership<SeqTypes> for EpochCommittees {
...
+        // `fetch_stake_table`, so one dropped error would leave the header unpersisted.
+        //
+        // Keyed by `epoch`, unlike the previous-epoch stake table stored further down, and
+        // not gated behind the stake table consistency check, which doesn't apply here.
+        if let Err(e) = self
+            .fetcher
+            .persistence
+            .lock()
+            .await
+            .store_epoch_root(epoch, block_header.clone())
```
imabdulbasit, line 825: I think this would overwrite the stake tables at startup because we load the stake tables and then call `add_epoch_root() `for each of them. I don’t think it’s a big issue, but we can probably add a check in store_epoch_root() to return early if the entry/file already exists.

sveitser: You think it's worth it, basically we then have 20 reads on startup instead of writes, except for fs storage where I think it's cleaner. Feels like what we have now should be okay and simpler?

PR #4850, opened 2026-08-24
Fetch a missing payload from peers

twittner:
## Problem

A node that receives a view's proposal and its own VID share, but misses the other voters' share broadcasts, cannot reconstruct that block. It then cannot cast a vote1 on any proposal parented at that view, and nothing re-transmits what it missed: Cliquenet retries a send until it is ACKed, but the coordinator GCs after 2 views, so the last copies of that view's share broadcasts disappear network-wide. All else being equal, widening the GC window would mitigate the issue, but solve it since we have to GC at some point.

The other nodes meanwhile lock the view, so a proposal they will vote for must be parented there, and their locks only move on a new certificate, which needs the votes from the nodes that fell behind, which in turn need the payload they do not receive. Timeout certificates keep forming, so views advance while nothing certifies, and every later view repeats the
previous one.

## Proposed solution

Nodes ask a peer for the missing block.

**Requesting.** On a timeout, or on a timeout certificate, consensus emits `RequestMissingPayload` for the view every proposal it could vote for is parented at: the parent of the latest proposal it holds. Leaders build on their lock, so that is both the view blocking the vote and the only view peers still retain. A node holding the payload of a certified later view would have locked there itself. The certificate comes from that proposal's `justify_qc`, so a node that missed the view's votes and the certificate broadcast can still ask. Views at or earlier than the lock are excluded, since safety pins any votable
proposal's justify_qc there or later, and so are views later than `current_view - GC_MARGIN_VIEWS`, where the transport may still deliver the missed broadcasts on its own.

**Fetching.** The new `Fetcher` unicasts to one peer per round, drawn from the view's
committee, excluding itself and everyone this fetch already asked. Answering spends the peer's slot, so one request buys one verification and one redraw. A peer that cannot help says so and another is drawn at once; only the periodic retry may start the round over.

**Serving.** The new `Server` keeps the payload of the view it is locked at, which is
what a stalled peer asks for, plus a few recent candidates so the one the lock reaches is still there when a certificate arrives late. It answers only within a few views of a timeout, since that is the only situation in which anyone fetches, and stays silent otherwise.

**Verifying.** Nothing from a peer is trusted but the bytes, and those only if they re-commit to the commitment in the node's own proposal under the committee's erasure parameters. The result enters the coordinator through the same path as a share-decoded payload.

## Miscellaneous

- `GC_MARGIN_VIEWS` drives both the fetch delay and the coordinator's network collection bound, so the fetch waits exactly until the transport gives up.
- The `GcScope::Timeout` arm now skips a view holding a certificate. Retiring it tore down the share accumulator at the moment the fetch begins waiting for the transport, leaving two views where neither route could recover the payload. It also aligns the two halves of that gc: `Consensus::gc` already kept its side.
- `BlockReconstructed` retries the latest votable child rather than `view + 1`; on the timeout path the blocked child is further along.
- Requests are unsigned. The transport authenticates the sender, so the signature only re-checked that binding.

## Scope

Whole payloads only. A block plus envelope must fit Cliquenet's message size. A peer whose
payload does not fit answers `TooLarge`. Fetching VID shares instead, for blocks
that cannot be sent whole, could be explored in a subsequent PR.

## Known gaps, not addressed here

- A restart seeds `blocks_reconstructed` from persisted proposals, which are written when a proposal pairs with this node's share rather than when its payload is decoded. A node that restarts mid-stall claims a payload it never had, which also makes the fetch skip that view. Pre-existing, and the one blocked state this mechanism cannot repair; the fix belongs at the seeding, from the DA store.
- A node holding the uncertified sibling of an equivocating leader cannot replace its proposal, so the fetch correctly refuses to target it.
- Outside a stall a lost payload is not recovered: no timeout means no request, and the lock moves past the hole. A data-availability gap, not a liveness one.
- No metrics, and no rate limit on serving beyond the timeout window.


crates/hotshot/new-protocol/src/coordinator.rs
```diff
@@ -592,40 +598,19 @@ where
...
-                                warn!(
-                                    view = %out.view,
-                                    header = %proposal.block_header.payload_commitment(),
-                                    reconstructed = %out.payload_commitment,
-                                    "reconstructed payload commitment does not match proposal header"
-                                );
-                            }
-                        }
-                        return Ok(ConsensusInput::BlockReconstructed(out.view, out.payload_commitment))
+                        return Ok(self.on_payload_obtained(out))
```
bfish713, line 601: On the leader path we never actually reconstruct the block because we created it I think this means the leader can't server it's own block if requested.

bfish713 (commented):
Besides the comment and issue with restarts not seeding the payload (which we discussed) I don't see any issues.  I will spend some more time tomorrow to double check

crates/hotshot/new-protocol/src/coordinator.rs
twittner: Yes, since this is only meant to avoid a general liveness failure I omitted that path, but we can add it.

bfish713 (approved):

PR #4855, opened 2026-08-25
fix(go-sdk): bound every HTTP request and share the caller's deadline across nodes

philippecamacho:
Follows [celo-org/optimism#500](https://github.com/celo-org/optimism/pull/500): the timeout belongs in the SDK. A client timeout alone does not fix it.

## This PR

- Puts a 30s limit on every HTTP request. All five constructors used `http.DefaultClient`. Its `Timeout` is 0, so there was no limit at all. A node could take the connection and never reply. That blocked any caller who set no deadline, forever. See celo-org/optimism#493. The 30s value now lives in one new package, `internal/httpclient`.
- Shares the caller's deadline between endpoints that are tried one after another. This covers fetch and submit in `MultipleNodesClient`, and `BuilderSubmitter`. Before, a dead first node used up the whole deadline. The second node then got none of it. 
- Tidies up code. Behavior does not change. `submitToBuilder` is now its own method. `BuilderSubmitter` holds one HTTP client instead of one per URL. Both submitters now share `decodeSubmitResponse`.

## This PR does not

- Change the public API. Every new name is unexported, or sits under `internal/`.
- Add a setting for users. 30s is fixed. There is no `WithRequestTimeout`.
- Limit a whole walk when the caller sets no deadline. The split does nothing in that case. So N endpoints can still take N × 30s.
- Touch the stream methods: `StreamTransactions`, `StreamTransactionsInNamespace`, `StreamPayloads`. They do not share the deadline. They still fail on the first bad node.
- Cut short an open stream. coder/websocket uses `HTTPClient.Timeout` for the handshake only.

## How to test

```
cd sdks/go && go test ./client/ ./client-dev-node/
```

---
  - https://app.asana.com/0/0/1218179911964595

### DIFF
--- a/sdks/go/client/builder_submitter.go
+++ b/sdks/go/client/builder_submitter.go
@@ -62,37 +62,39 @@ var ErrAllBuildersFailed = errors.New("submission to all builders failed, check 
 func (c *BuilderSubmitter) SubmitTransaction(ctx context.Context, tx types.Transaction) (*types.TaggedBase64, error) {
 	c.previousSubmitErrors = make([]error, 0)
 	for clientIdx, url := range c.builderUrls {
-		response, err := c.tryPostRequest(ctx, url, clientIdx, tx)
-
+		hash, err := c.submitToBuilder(ctx, url, clientIdx, tx)
 		if err != nil {
 			c.previousSubmitErrors = append(c.previousSubmitErrors, err)
 			continue
 		}
-
-		defer response.Body.Close()
-		if response.StatusCode != http.StatusOK {
-			c.previousSubmitErrors = append(c.previousSubmitErrors, fmt.Errorf("%w: %v", ErrEphemeral, response.Status))
-			response.Body.Close()
-			continue
-		}
-
-		body, err := io.ReadAll(response.Body)
-		if err != nil {
-			c.previousSubmitErrors = append(c.previousSubmitErrors, fmt.Errorf("%w: %v", ErrEphemeral, err))
-			response.Body.Close()
-			continue
-		}
-
-		var hash types.TaggedBase64
-		if err := json.Unmarshal(body, &hash); err != nil {
-			c.previousSubmitErrors = append(c.previousSubmitErrors, fmt.Errorf("%w: %v", ErrEphemeral, err))
-			response.Body.Close()
-			continue
-		}
 		// If we receive a successful submission from the builder, we can exit as we don't need to send to other builders.
-		return &hash, nil
+		return hash, nil
 	}
 	return nil, ErrAllBuildersFailed
+}
+
+// Submits the transaction to a single builder and decodes its response.
+func (c *BuilderSubmitter) submitToBuilder(ctx context.Context, baseUrl string, clientIndex int, tx types.Transaction) (*types.TaggedBase64, error) {
+	response, err := c.tryPostRequest(ctx, baseUrl, clientIndex, tx)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%w: %v", ErrEphemeral, response.Status)
+	}
+
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrEphemeral, err)
+	}
+
+	var hash types.TaggedBase64
+	if err := json.Unmarshal(body, &hash); err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrEphemeral, err)
+	}
+	return &hash, nil
 }
 
 // post request handler for the builder submitter.
--- a/sdks/go/client/builder_submitter.go
+++ b/sdks/go/client/builder_submitter.go
@@ -62,7 +62,9 @@ var ErrAllBuildersFailed = errors.New("submission to all builders failed, check 
 func (c *BuilderSubmitter) SubmitTransaction(ctx context.Context, tx types.Transaction) (*types.TaggedBase64, error) {
 	c.previousSubmitErrors = make([]error, 0)
 	for clientIdx, url := range c.builderUrls {
-		hash, err := c.submitToBuilder(ctx, url, clientIdx, tx)
+		builderCtx, cancel := shareRemainingBudget(ctx, len(c.builderUrls)-clientIdx)
+		hash, err := c.submitToBuilder(builderCtx, url, clientIdx, tx)
+		cancel()
 		if err != nil {
 			c.previousSubmitErrors = append(c.previousSubmitErrors, err)
 			continue
--- a/sdks/go/client/builder_submitter.go
+++ b/sdks/go/client/builder_submitter.go
@@ -31,7 +31,7 @@ func NewBuilderSubmitter(builderUrls []string) (*BuilderSubmitter, error) {
 	formattedUrls := make([]string, len(builderUrls))
 	for i, url := range builderUrls {
 		formattedUrls[i] = formatUrl(url)
-		builderClients[i] = http.DefaultClient
+		builderClients[i] = newHTTPClient()
 	}
 
 	return &BuilderSubmitter{
--- a/sdks/go/client/builder_submitter.go
+++ b/sdks/go/client/builder_submitter.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 
 	types "github.com/EspressoSystems/espresso-network/sdks/go/types"
@@ -17,7 +16,7 @@ var _ EspressoBuilderSubmitter = (*BuilderSubmitter)(nil)
 
 type BuilderSubmitter struct {
 	builderUrls          []string
-	builderClients       []*http.Client
+	client               *http.Client
 	previousSubmitErrors []error
 }
 
@@ -27,16 +26,14 @@ func NewBuilderSubmitter(builderUrls []string) (*BuilderSubmitter, error) {
 		return nil, fmt.Errorf("One or more builder url's is required for the builder submitter")
 	}
 
-	builderClients := make([]*http.Client, len(builderUrls))
 	formattedUrls := make([]string, len(builderUrls))
 	for i, url := range builderUrls {
 		formattedUrls[i] = formatUrl(url)
-		builderClients[i] = newHTTPClient()
 	}
 
 	return &BuilderSubmitter{
-		builderUrls:    formattedUrls,
-		builderClients: builderClients,
+		builderUrls: formattedUrls,
+		client:      newHTTPClient(),
 	}, nil
 }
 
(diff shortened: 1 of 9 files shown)


sdks/go/internal/httpclient/httpclient.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"net/http"
+	"time"
+)
+
+// Bounds a request whose caller supplied no deadline: a node that accepts the
+// connection and never answers would otherwise park it forever. coder/websocket
+// applies this to the handshake only, so open streams survive it.
+const Timeout = 30 * time.Second
```
jjeangal: `Client.Timeout` is a whole-request deadline. It caps the body read too, not just the header stall we're fixing. `FetchHeadersByRange` (no range cap) and `FetchNamespaceTransactionsInRange` (count-capped only) could take more than 30s on a slow link and will now fail where they used to succeed, with no escape hatch. `Transport.ResponseHeaderTimeout` targets the black-hole case exactly without cutting a progressing transfer. If we keep `Client.Timeout`, could we add a comment that long-range fetches are capped by it?

sdks/go/client/client.go
```diff
@@ -58,6 +60,17 @@ type Client struct {
...
 
+// Gives one attempt of a sequential walk an even share of what is left of the
+// caller's deadline, so an endpoint that never answers cannot spend the budget
+// of the endpoints after it.
+func shareRemainingBudget(ctx context.Context, remaining int) (context.Context, context.CancelFunc) {
+	deadline, ok := ctx.Deadline()
+	if !ok || remaining <= 1 {
+		return ctx, func() {}
+	}
+	return context.WithTimeout(ctx, time.Until(deadline)/time.Duration(remaining))
```
jjeangal: The even 1/remaining split taxes the healthy case: 5 URLs + a 5s deadline gives the first (usually healthy) endpoint 1s, so a 1.2s round trip time fails everything. We could instead halve (`time.Until(deadline)/2`) which keeps the common case fast and still leaves every later endpoint a share?

Minor: this helper reads more naturally in `utilities.go` next to `decodeSubmitResponse`, and the already-expired-deadline branch isn't covered by `TestShareRemainingBudget`.

sdks/go/client/client.go
```diff
@@ -58,6 +60,17 @@ type Client struct {
...
+}
```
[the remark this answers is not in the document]
jjeangal: Agree with the earlier suggestion to halve instead of splitting evenly. With several URLs the healthy first node gets too little time. Also small: this helper would fit better in `utilities.go`, and there's no test for the already-expired-deadline case.

sdks/go/client/multiple_nodes_client.go
```diff
@@ -37,8 +37,10 @@ func NewMultipleNodesClient(urls []string) (*MultipleNodesClient, error) {
 
 func (c *MultipleNodesClient) FetchLatestBlockHeight(ctx context.Context) (uint64, error) {
 	var errs []error
-	for _, node := range c.nodes {
-		height, err := node.FetchLatestBlockHeight(ctx)
+	for i, node := range c.nodes {
+		nodeCtx, cancel := shareRemainingBudget(ctx, len(c.nodes)-i)
+		height, err := node.FetchLatestBlockHeight(nodeCtx)
+		cancel()
```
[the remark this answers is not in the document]
jjeangal: Could we add the same note that's on `submitToBuilder`? Calling `cancel()` here is only safe because these methods finish reading the response before returning. Worth stating so it isn't accidentally broken later.

sdks/go/client/multiple_nodes_client.go
```diff
@@ -159,8 +161,10 @@ func (c *MultipleNodesClient) SubmitTransaction(ctx context.Context, tx common.T
 
 	// Check if one node is successfully able to submit the transaction
 	var errs []error
-	for _, node := range c.nodes {
-		hash, err := node.SubmitTransaction(ctx, tx)
+	for i, node := range c.nodes {
+		nodeCtx, cancel := shareRemainingBudget(ctx, len(c.nodes)-i)
+		hash, err := node.SubmitTransaction(nodeCtx, tx)
+		cancel()
```
jjeangal, line 167: Same here ^

sdks/go/internal/httpclient/httpclient.go
```diff
@@ -0,0 +1,15 @@
...
+
+func New() *http.Client {
+	return &http.Client{Timeout: Timeout}
+}
```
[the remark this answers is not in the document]
jjeangal: +1 to it being a whole-request limit. My main worry is the two range fetches (`FetchHeadersByRange`, `FetchNamespaceTransactionsInRange`). A slow-but-healthy download that used to work now dies at 30s. Can we switch to `ResponseHeaderTimeout` (which only limits the wait for a reply, not the download itself), or at least note this cap in the comment?

jjeangal (changes requested):
Looks good! This does what was agreed in the #500 thread: put a time limit on the client and share the deadline across nodes. Two things I'd sort out before merging: 
- The 30s whole-request limit will cut off large range fetches that used to work.
- Splitting the deadline evenly starves the healthy first node. Halving would be better. The rest are small!

sdks/go/client/client.go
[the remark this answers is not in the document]
philippecamacho: Yes, done in 7632c037de3.

sdks/go/internal/httpclient/httpclient.go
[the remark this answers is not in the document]
philippecamacho: I went for the first approach: waiting for the reply instead of the whole request.
See 1ed7ff570cc.

sdks/go/client/timeout_test.go
```diff
@@ -0,0 +1,198 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			firstUrl, firstRequests := blackHoleNode(t)
+			secondUrl, secondRequests := blackHoleNode(t)
+			ctx, cancel := context.WithTimeout(context.Background(), callerBudget)
+			defer cancel()
+
+			require.Error(t, tc.call(t, ctx, []string{firstUrl, secondUrl}))
+			require.Equal(t, int64(1), firstRequests())
+			require.Equal(t, int64(1), secondRequests(), "the first endpoint consumed the whole deadline")
+		})
```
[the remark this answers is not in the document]
philippecamacho: Done in aeb3454a7d9.

sdks/go/client/multiple_nodes_client.go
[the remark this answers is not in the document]
philippecamacho: Done 0c3a2972b36.

sdks/go/client/multiple_nodes_client.go
philippecamacho: Done https://github.com/EspressoSystems/espresso-network/commit/0c3a2972b362f0d19f7462b99dacbf18fed8c144..

sdks/go/client/utilities.go
```diff
@@ -39,3 +41,14 @@ func decodeSubmitResponse(response *http.Response) (*types.TaggedBase64, error)
 	}
 	return &hash, nil
 }
+
+// Gives one attempt of a sequential walk half of what is left of the caller's
+// deadline, so an endpoint that never answers cannot spend the budget of the
+// endpoints after it, while a healthy first endpoint still gets most of it.
+func shareRemainingBudget(ctx context.Context, remaining int) (context.Context, context.CancelFunc) {
```
jjeangal: Minor: since the split is now `/2`, `remaining` is only used for the `remaining <= 1 guard`, never in the sizing. So the name reads like it still proportions the share (as it did with the even split), when it's really just asking "is this the last endpoint?". It could take a `bool` instead:
```
func shareRemainingBudget(ctx context.Context, isLast bool) (context.Context, context.CancelFunc) {
      deadline, ok := ctx.Deadline()
      if !ok || isLast {
              return ctx, func() {}
      }
      return context.WithTimeout(ctx, time.Until(deadline)/2)
}
```
Call site:
```
for i, node := range c.nodes {
      isLast := i == len(c.nodes)-1
      nodeCtx, cancel := shareRemainingBudget(ctx, isLast)
      ...
}
```
Non-blocking, the current version is correct, just a readability nit!

jjeangal (approved):
Just a minor comment! Could be merged as is. Tests run all valid on my end.

sdks/go/client/builder_submitter.go
```diff
@@ -61,42 +59,31 @@ var ErrAllBuildersFailed = errors.New("submission to all builders failed, check
 // If all builders fail, this function will return an error. Otherwise, err will be nil.
 func (c *BuilderSubmitter) SubmitTransaction(ctx context.Context, tx types.Transaction) (*types.TaggedBase64, error) {
 	c.previousSubmitErrors = make([]error, 0)
-	for clientIdx, url := range c.builderUrls {
-		response, err := c.tryPostRequest(ctx, url, clientIdx, tx)
-
+	for i, url := range c.builderUrls {
+		builderCtx, cancel := shareRemainingBudget(ctx, len(c.builderUrls)-i)
+		hash, err := c.submitToBuilder(builderCtx, url, tx)
+		cancel()
```
Sneh1999, line 66: when we do cancel here? Will it also cancel the context of the main caller which sent the original context?

Like for example if batch poster is calling this sdk will this cancel the context of the batch poster?

sdks/go/client/builder_submitter.go
```diff
@@ -61,42 +59,31 @@ var ErrAllBuildersFailed = errors.New("submission to all builders failed, check
...
 		if err != nil {
```
Sneh1999, line 67: If we do cancel above will this error we appended?

sdks/go/client/builder_submitter.go
philippecamacho: No, I don't think so. See https://go.dev/blog/context#context

> A Context does not have a Cancel method for the same reason the Done channel is receive-only: the function receiving a cancellation signal is usually not the one that sends the signal. In particular, when a parent operation starts goroutines for sub-operations, those sub-operations should not be able to cancel the parent. Instead, the WithCancel function (described below) provides a way to cancel a new Context value.

sdks/go/client/utilities.go
philippecamacho: Done in ceeb4f9c1cd.

sdks/go/client/builder_submitter.go
philippecamacho: Yes the error is still appended. A small test in 2c0172c9487  has been added to confirm this.

ezdac:
> .. does not: Add a setting for users. 30s is fixed. There is no WithRequestTimeout.

I'm wondering why you decided for instantiating a new client with hard-coded options instead of adding the SDK option to set the whole client  as a user option `WithHTTPClient`? This could always have a default when absent.

I feel like since the caller is able to decide wether the outermost per-submit context has a deadline, and that deadline is made part of the submission loop, they should also be able to tweak the per-request timeouts - as @jjeangal pointed out already, this would lead to somewhat unexpected behavior when the outer context is deliberately chosen very large.

Also, being able to pass in the full http-client would allow the SDK user to:
- share an existing client with the sdk
> as per the go docs:
 // The [Client.Transport] typically has internal state (cached TCP
// connections), so Clients should be reused instead of created as
// needed. Clients are safe for concurrent use by multiple goroutines.
- set the proxy programmatically and not only by env-vars
- tweak other parameters individually
- design a tight retry loop with exponential backoff around the `SubmitTransaction` function.
- choose the overall `http.Client.Timeout` instead of the fine-grained transport timeouts

Sneh1999:
> > .. does not: Add a setting for users. 30s is fixed. There is no WithRequestTimeout.
> 
> I'm wondering why you decided for instantiating a new client with hard-coded options instead of adding the SDK option to set the whole client as a user option `WithHTTPClient`? This could always have a default when absent.
> 
> I feel like since the caller is able to decide wether the outermost per-submit context has a deadline, and that deadline is made part of the submission loop, they should also be able to tweak the per-request timeouts - as @jjeangal pointed out already, this would lead to somewhat unexpected behavior when the outer context is deliberately chosen very large.
> 
> Also, being able to pass in the full http-client would allow the SDK user to:
> 
> * share an existing client with the sdk
> 
> > as per the go docs:
> > // The [Client.Transport] typically has internal state (cached TCP
> > // connections), so Clients should be reused instead of created as
> > // needed. Clients are safe for concurrent use by multiple goroutines.
> 
> * set the proxy programmatically and not only by env-vars
> * tweak other parameters individually
> * design a tight retry loop with exponential backoff around the `SubmitTransaction` function.
> * choose the overall `http.Client.Timeout` instead of the fine-grained transport timeouts

@ezdac Thanks for the suggestion! We did realize that this is not the optimal solution. We were planning to close this PR and just have the client pass the timeout as you suggested. The way to do that will be using the go context. Piers had told us that batch poster already has a network timeout ctx which we can use. So we will pass this ctx to the go sdk. Let us know what you think about that!

PR #4861, opened 2026-08-25
chore(crypto-helper): remove unused dependencies

sveitser:
hotshot-query-service, jf-crhf, jf-rescue and serde are not referenced in sdks/crypto-helper/src. Dropping hotshot-query-service removes the only GPL-3.0-or-later crate from the cdylib dependency tree, which the Go SDK links against.


imabdulbasit (approved):

philippecamacho (approved):
LGTM

PR #4863, opened 2026-08-26
fix: alloy provider responses handling after alloy upgrade

sveitser:
### This PR:

- Fixes `SwitchingTransport` scoring a dead provider as healthy. alloy 2.4.1 returns a non-2xx response with a parseable
  JSON-RPC body as `Ok`, discarding the status
  ([here](https://github.com/alloy-rs/alloy/blob/v2.4.1/crates/transport-http/src/lib.rs#L63), deliberately:
  [test](https://github.com/alloy-rs/alloy/blob/v2.4.1/crates/transport-http/src/lib.rs#L182)). Every error class the
  fleet reports carries such a body, so a dead API key reset the failure counter and never tripped failover. Responses
  are now scored by payload, not by `Result` arm.
- Recognises rate limiting from an HTTP 429 in both `HttpError` and `HttpErrorWithRetryAfter`, honouring `Retry-After`.
- Adds a `test-server` crate for fake HTTP providers: this PR's regression test needs one, staking-cli had the same warp
  helper private to it, and the L1 work ahead needs more.
- `serve_on_random_port` returns a `Url`; all 18 call sites hand-built one from the port.

### This PR does not:

- Add an error taxonomy. `-32005` and `-32007` still take the failure path, as before 2.4.1; infura reuses `-32005` for
  result-count rejections, so the ordering needs its own argument.

### Key places to review:

- `ResponseOutcome` and the tail of `SwitchingTransport::call` in `crates/espresso/types/src/v0/impls/l1.rs`.
- `test_failover_on_dead_provider_with_json_rpc_error_body`: serves alchemy's 403 body ahead of an Anvil, fails on the
  pre-fix code with no failover.
- Fixtures are verbatim bodies from production dumps, except the code-429 one, which is synthetic and marked so: no
  captured response carries that code (alchemy sends `-32600`, infura `-32005`).


crates/espresso/types/src/v0/impls/l1.rs
```diff
@@ -304,6 +304,68 @@ impl SingleTransport {
...
+    RateLimited {
+        retry_after: Option<Duration>,
+    },
+    Failed,
+}
+
+/// HTTP status for "Too Many Requests". Also matched as a JSON-RPC code, since a proxy may echo
+/// the status into the error payload.
+#[cfg(feature = "node")]
+const TOO_MANY_REQUESTS: i64 = 429;
```
sveitser: feels like we don't need to define this ourselves

crates/test-server/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+//! Minimal HTTP test servers built on `warp`.
```
imabdulbasit, line 1: there is already a test-utils crate. I think we should use that instead of creating another crate

imabdulbasit (changes requested):

crates/test-server/src/lib.rs
imabdulbasit: or delete that crate? because that crate only has helper function to reserve the port

crates/espresso/types/src/v0/impls/l1.rs
sveitser: actually based on telemetry the json rpc response is NEVER 429 for any nodes that ship telemetry

sveitser: okay this is dead code since the ethers to alloy migration, will remove the check

crates/test-server/src/lib.rs
sveitser: I was going to do that but test-utils is supposed to be standalone without dependencies other than stdlib. If we put this in there we need to feature gate or pulls http server into everything that uses test-utils. I think both are worse than a new crate. I'm happy to rename the crates or put them into ./crates/test if you think that helps.

crates/espresso/types/src/v0/impls/l1.rs
sveitser: https://github.com/EspressoSystems/espresso-network/pull/4863/commits/c5cfaca10c974ce553ff7933bc3292ee995a19cf

crates/espresso/types/src/v0/impls/l1.rs
```diff
@@ -362,57 +413,52 @@ impl Service<RequestPacket> for SwitchingTransport {
...
-                        .log_failure(&self_clone.opt)
-                    {
-                        // Increment the failovers metric
-                        self_clone.metrics.failovers.add(1);
-                        self_clone.switch_to(current_transport.generation + 1, current_transport);
-                    }
+            if let ResponseOutcome::RateLimited { retry_after } = outcome {
+                // Rate limits should not cause failover, but instead should only cause us to
+                // temporarily back off on making requests to the RPC server.
+                current_transport.status.write().rate_limited_until = Some(
```
imabdulbasit, line 449: from claude:

> retry_after is trusted with no upper bound: a proxy or CDN in front of the provider returning a non-JSON 429 with Retry-After: 7200 sets rate_limited_until two hours out, and since call() returns early during that window and rate limits never count as failures, we never fail over to the healthy backup, the node is just L1-blind until the timer expires. alloy treats this as a footgun too: its retry layer clamps server hints to 5 minutes (RpcErrorExt::backoff_hint(), "a misbehaving server cannot stall retries"), but we bypass that by matching the raw field. Suggestion: use backoff_hint() instead of raw retry_after, or clamp with min(retry_after, cap).

imabdulbasit (approved):

PR #4866, opened 2026-08-26
fix(epoch-membership): recover from wedged epoch stake-table catchup

lukaszrzasik:
### This PR:
* Fixes the 2026-08-14 mainnet wedge: a catchup attempt parked on a stalled persistence query held its `catchup_map` entry forever, answering `Catchup already in progress` for days — no timeout, eviction, or retry.
* Adds a watchdog to `spawn_catchup` (default 5 min): a stuck attempt is abandoned — flagged to stop at its next checkpoint, never aborted — and every entry it claimed is evicted so waiters retry. Abandonment, claims, and completion are all serialized under the `catchup_map` lock (which also closes a pre-existing lost-wakeup for waiters activating between an unlocked broadcast and the entry's removal); a local DRB computation is exempt from the budget.
* Bounds the storage waits: `load_stake_table`'s lock and read phases (60 s each, with won-race checks that answer from the snapshot), and the DRB stored-progress load (60 s), so no stalled query can pin the process-wide locks or the `drb_calculation_map` claim.
* Makes DRB state abort-safe: a computed result is published to the membership before it is persisted, an abandoned attempt stops before entering the hash chain, and a dropped computation's cleanup releases both DRB maps atomically and fires its cancel token, stopping the orphaned hash batch.
* Adds regression tests at both layers — 10 hotshot (hostile membership) and 3 espresso against the real `EpochCommittees`, including one where only watchdog abandonment can recover — each verified to fail on unfixed code.

### This PR does not:
* Abort the stuck attempt — drop-based cancellation could land mid-update (torn membership writes) or not land at all on a synchronously blocked task.
* Make the timeouts operator-configurable; they are compile-time defaults (cross-referenced in their docs) with builder overrides for tests. Mis-sizing degrades to latency, not a wedge: abandoned attempts leave durable progress for the retry.
* Bound other users of the shared `fetcher.persistence` lock (e.g. the stake-table event fetcher's writes).

### Key places to review:
* `crates/hotshot/types/src/epoch_membership.rs`, `spawn_catchup` / `cancel_catchups` — the `abandoned`/`cleaned_up` flag pair, both set and read under the `catchup_map` lock, so the watchdog and the attempt can never both run cleanup or clobber a retry's entries.
* `crates/hotshot/types/src/epoch_membership.rs`, `compute_drb_result_impl` — publish-before-persist ordering, and `clear_drb_state`'s single-critical-section removal + token fire (identity-safe because inserts are gated on the entry it removes).
* `crates/espresso/types/src/v0/impls/committee.rs`, `load_stake_table` — the two bounded phases and the three won-race snapshot checks (entry, between the locks, timeout path).

### Things tested
* `cargo test -p hotshot-testing --test tests_1 epoch_catchup_wedge`: 10/10
* `cargo test -p espresso-types --test epoch_catchup_wedge`: 3/3
* `cargo test -p espresso-types --lib committee`: 12/12; `cargo test -p hotshot-types --lib drb`: 7/7
* `cargo clippy --features testing --all-targets -- -D warnings`: clean
* Full consensus/integration suites left to CI.


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -537,6 +594,11 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
     /// `EpochMembershipCoordinator` by removing the failed epochs from the
     /// `catchup_map` and broadcasting the error to any tasks that are waiting for the
     /// catchup to complete.
+    ///
+    /// It also sweeps any entry whose channel is closed. A closed channel
+    /// means every sender is gone, i.e. the task that claimed the entry died
+    /// without cleaning up (e.g. it panicked while holding entries for
+    /// intermediate epochs), so nothing else can ever complete or remove it.
     fn catchup_cleanup(
```
bfish713, line 634: Don't we need to check `is_abandoned` here?

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -750,13 +834,159 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
...
+        // the middle of a multi-step update (a torn membership write, a
+        // leaked `drb_calculation_map` guard) or not land at all (abort
+        // cannot interrupt a synchronously blocked task). Instead flag the
+        // attempt to stop at its next checkpoint, then evict every
+        // `catchup_map` entry it claimed — the requested epoch's and the
+        // intermediate epochs' — and fail their waiters, so later requests
+        // retry instead of being answered "Catchup already in progress" for
+        // the lifetime of the process.
+        progress.abandon();
+        let stuck_at = progress.last();
```
bfish713: Do we need to also clear the `drb_calculation_map` here

crates/hotshot/types/src/epoch_membership.rs
lukaszrzasik: Yes, before the cleanup we should check whether the catchup attempt has been abandoned. The abandon code takes care of cleaning up so cleaning up again could mess with the next attempt. I've added the check.

crates/hotshot/types/src/epoch_membership.rs
lukaszrzasik: Yes and no. We shouldn't always clear it. The drb calculation can progress normally and so the next attempt would not re-trigger the same calculation. Additionally we now exclude the drb calculation time from the timeout as we expect it to take long time. We should clear in case of panic. I've added a RAII guard for that.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -750,13 +834,159 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
     }
 }
 
+/// Default upper bound on a single catchup attempt. Past this the attempt is
+/// abandoned — flagged to stop at its next checkpoint, never aborted — and its
+/// `catchup_map` entries are evicted so a later request can retry. Generous,
+/// because a legitimate attempt may walk many epochs and compute a DRB, and a
+/// premature abandonment costs duplicate work.
+const DEFAULT_CATCHUP_TIMEOUT: Duration = Duration::from_secs(300);
```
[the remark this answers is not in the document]
lukaszrzasik: Fixed in e4c757cfa10, taking your first option: `CatchupProgress` now carries an `in_drb_compute` flag that `compute_drb_result` raises only while the hash chain itself is running, and the watchdog keeps extending the budget for as long as it is set. Storage awaits inside the computation (including the DRB progress load) stay on the normal budget, so the watchdog still catches a stall in the storage/peer steps. Regression test: `slow_drb_computation_is_not_abandoned`.

The doomed-retry mode also lost its permanent variant from the other side: since 75b3b60e419 the `drb_calculation_map` entry is released via an RAII guard when a computation dies, so "DRB calculation already in progress" can no longer outlive the computation that owns it.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -376,6 +405,9 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
                         let (mut tx, rx) = broadcast(1);
                         tx.set_overflow(true);
                         map_lock.insert(try_epoch, rx.deactivate());
+                        // Recorded under the map lock so the watchdog can
+                        // never observe the entry without the claim.
+                        progress.claim(try_epoch, tx.clone());
                         drop(map_lock);
```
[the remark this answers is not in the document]
lukaszrzasik: Fixed in 1cdfc05ced0, in the shape you suggested: the vacant branch checks abandonment under the `catchup_map` lock before the insert+claim, and the watchdog sets the flag and drains the claims under one hold of the same lock — so an entry is either drained by the watchdog or never inserted. The two later check-then-remove pairs are folded under the same lock as well.

This also made the watchdog's post-abandonment "leftovers" pass unnecessary (a parked attempt can no longer wake up and claim anything), so it is deleted in the same commit. Regression test: `abandoned_attempt_claims_no_new_epochs`, which reproduces exactly the parked-load sequence you describe.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -365,6 +391,9 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
                         // Somebody else is already fetching this epoch, drop
                         // the lock and wait for them to finish
                         drop(map_lock);
+                        progress.checkpoint(format_args!(
+                            "waiting for another task's catchup of epoch {try_epoch}"
```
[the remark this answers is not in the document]
lukaszrzasik: Fixed in 8944bb98624: `Err(RecvError::Closed)` now evicts the orphaned entry — re-checking under the map lock that the mapped entry is still a closed one — so the next loop iteration claims the epoch itself instead of spinning until its own abandonment.

One note: after 1cdfc05ced0 made abandonment atomic with the map claims, a closed orphan should not be reachable at all anymore (the panicking-task case aside), so this lands as defense in depth rather than a routine path — hence the `error!` when it fires.

crates/espresso/types/src/v0/impls/committee.rs
```diff
@@ -709,37 +728,68 @@ impl Membership<SeqTypes> for EpochCommittees {
...
-                None
+        // The whole read phase — lock waits included — is bounded: a stalled
+        // persistence query must not pin `load_from_storage_lock` and the
+        // persistence lock indefinitely, or every epoch's catchup blocks
+        // process-wide. Dropping the in-flight reads on timeout is safe;
+        // they are pure point reads.
+        let read = tokio::time::timeout(self.storage_read_timeout, async {
+            // Ensure there is only one `load_stake_table` at a time:
+            let _guard = self.load_from_storage_lock.lock().await;
+            // Check if someone else won the race:
```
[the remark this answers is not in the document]
lukaszrzasik: Both notes addressed in e33753e1fa3:

1. The lock queueing is now bounded separately from the reads, each phase with its own timeout and its own log line ("timed out queueing for the storage locks" vs "storage reads for the stake table timed out"), so contention is no longer misread as a stalled query and queueing behind healthy loads no longer eats the reads' budget. A caller can still give up after a full budget of pure queueing, but it now says so honestly instead of blaming the DB; the comment also states explicitly that timing out unpins nothing and only lets this caller fall back to peer catchup. I kept a single knob — each phase gets the full `storage_read_timeout` rather than a shorter queue bound.

2. The sqlx caveat is now spelled out in the comment above the read phase: dropping a read mid-result-stream closes the connection instead of returning it to the pool, so under pool exhaustion these timeouts churn connections.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -750,13 +938,190 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
...
     epoch_tx: Sender<Result<EpochMembership<T>>>,
 ) {
     tokio::spawn(async move {
-        coordinator.clone().catchup(epoch, epoch_tx).await;
+        let progress = CatchupProgress::new(epoch);
+        let mut inner = tokio::spawn({
+            let coordinator = coordinator.clone();
+            let epoch_tx = epoch_tx.clone();
+            let progress = progress.clone();
+            async move { coordinator.catchup(epoch, epoch_tx, progress).await }
```
twittner:
```suggestion
            coordinator.catchup(epoch, epoch_tx, progress)
```

sveitser (approved):
I'm not super familiar but I think this is a big improvement and I couldn't spot any actual problems myself or using claude.

PR #4870, opened 2026-08-27
Display IPv6 addresses as socket addresses, and validate addresses

twittner:
`NetAddr` displayed an IPv6 address as `::1:9977`, which nothing else accepts and which separates the port from the last group with the same character that separates the groups. `Display` now writes `[::1]:9977`.

Parsing stays as lax as it was. That means a `NetAddr` can hold a host no peer will reach. `NetAddr::validate` is the check parsing leaves out, applied where an address arrives from someone who can still be told to correct it. It checks neither the port nor the length of the whole name: a missing port and port 0 are one value after parsing, and DNS's 253 characters and the contract's 512 bytes differ, so the length belongs to whoever accepts the address.

Serialization and commitments are unchanged. `main` serialized `to_string()` while `Display` was unbracketed; this branch serializes via `unbracketed_string()`, which gives the same string, and `RegisteredValidator::commit` uses it too.

What does change is the address the staking CLI registers, now `[2001:db8::1]:9977` for IPv6. Existing nodes read it though, since `from_str` has already stripped brackets before.

Also, `Network::create` binds now by destructuring `NetAddr` instead of parsing `Display` output, and `Party` compares peer IPs canonically so an IPv4 address delivered in IPv6 form matches.

For users: the staking CLI now rejects a malformed `--p2p-addr`, port 0, and an address over 512 bytes, and says so when the registered form differs from what was typed. It warns without failing when the address is not routeable or does not resolve. The node validates `--cliquenet-bind-address` and `--cliquenet-advertise-address` at startup, and rejects port 0 for the latter.

### DIFF
--- a/staking-cli/src/p2p_addr.rs
+++ b/staking-cli/src/p2p_addr.rs
@@ -1,0 +1,76 @@
+//! Validation of the p2p address a validator registers in the stake table.
+
+use hotshot_types::addr::NetAddr;
+use tokio::net::lookup_host;
+
+use crate::output::output_warn;
+
+/// The longest address the stake table contract accepts.
+const MAX_ADDR_LEN: usize = 512;
+
+/// Parse a p2p address to register in the stake table.
+///
+/// [`NetAddr::validate`] holds the rules for the address itself; what is added here is
+/// what the stake table asks of one: it fits the contract's length limit, and it names a
+/// port a peer can connect to, which a missing port does not. An address that prints
+/// differently to how it was written is registered in the form it prints in, and the
+/// caller is told so, since that is the string that lands on chain.
+pub fn parse_p2p_addr(s: &str) -> Result<NetAddr, String> {
+    let addr: NetAddr = s.parse().map_err(|e| format!("`{s}`: {e}"))?;
+    addr.validate().map_err(|e| format!("`{s}`: {e}"))?;
+
+    if addr.port() == 0 {
+        return Err(format!("`{s}`: port 0 is not a port a peer can connect to"));
+    }
+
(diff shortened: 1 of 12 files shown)


staking-cli/src/p2p_addr.rs
```diff
@@ -0,0 +1,76 @@
...
+    if registered != s {
+        output_warn(format!("`{s}` will be registered as `{registered}`"));
+    }
+
+    Ok(addr)
+}
+
+/// Warn about an address that is well-formed but that other validators are unlikely to
+/// reach. Not an error: a name may resolve only once the node is deployed.
+pub(crate) async fn warn_if_unreachable(addr: &NetAddr) {
```
sveitser: For the staking CLI I usually return an error by default if anything looks potentially wrong and then provide a flag to to silence the error if it's desired. I think that's better here because the warning can be missed and afterwards the tx goes through as success, which I think provides the stronger signal to the user (e. g. "tx success => I'm done").

So concretely I would error and abort if any of our checks fail with descriptive message on why it failed and what flag to add to force.

This is also easy then to add a few integration tests for.

Happy to make this change myself too.

sveitser (approved):
@twittner LGTM with one caveat.

staking-cli/src/p2p_addr.rs
twittner: Thanks, in https://github.com/EspressoSystems/espresso-network/pull/4870/commits/acb3adb29671861e576cea6ca31df308ad31a7b5 I changed it to error instead of warning and added a `--skip-reachability-check` flag to allow bypassing all of it.

PR #4879, opened 2026-08-28
optimize catchup: Batch scanner fetches for missing blocks and VID

imabdulbasit:
## Problem

- Payloads are stored by hash, so one stored empty payload makes every empty height count as present. That shreds a contiguous missing range into many short runs: measured on mainnet, a 500 height window leaves 343 missing heights in 145 runs, averaging 2.4 heights each.
- The proactive scanner paid a round trip per run. Through the light client, which is what production uses, each run cost two finality proofs on top of the data.

## Change

- Adds `POST /v1/availability/{leaf,block,vid/common}/ranges` and `POST /v1/light-client/payload/ranges`, taking a list of half-open height ranges which must be ascending and disjoint. The availability routes answer from a single SQL statement.
- The endpoints behave like the range endpoints they generalise: heights the node lacks are fetched from its own peers within the usual fetch timeout, they answer in full or not at all, and a 404 means at least one height was not available in time.
- The scanner groups its missing chunks into batches of about one chunk's worth of objects, instead of one request per fragment, merging chunks that are adjacent. A batch of a single range is served by the range endpoint instead, which is a cacheable GET.
- The light client fetches the enclosing span as one range when the gaps are small, so a window costs one finality proof and one chain walk instead of a proof per run, and takes its payload proofs in one request. Verification is unchanged: leaves are pinned by parent-hash chain to a proof-verified anchor, and each payload proof is checked against its own header.
- Both providers fall back to the per-range endpoints when a peer cannot serve a batch, so an upgraded node still catches up from an upstream that predates these routes.

## Failure behaviour

Batching makes one bad height everyone's problem, so most of the review went here.

- A ranges request answers all or nothing, so a height no peer can serve would stall it. The scanner therefore waits a bounded time (`proactive_fetch_timeout`, default 120s, now a node option) and then fetches each chunk on its own, leaving the bad height to hold up only its own chunk. On `main` the equivalent chunk retried forever and the scan never moved past it.
- Ranges fetches no longer retry forever, and a fetch that gives up runs no callbacks, so the scanner asks the fetchers directly rather than waiting out the timeout on a failure that already happened. It gives up once two consecutive one-second polls find nothing fetching, which also covers a task that gives up before the first poll.
- A ranges request resolves only when it covers every height requested, so a peer answering with part of what it was asked for cannot end the fetch.
- The endpoint itself no longer gives the whole request one window: a miss is served range by range, so an unavailable height fails in one fetch timeout instead of holding the connection for the whole set's budget.

## Measured

100k heights of real mainnet history (22,000,000..22,100,000) in a local Postgres, both arms
starting from an identical database and fetching from one server running the node's real handlers.
Payloads are content addressed and one empty payload covers 33,643 of those heights, so the node
is missing 66,357 heights in **29,759 runs averaging 2.23** - the real fragmentation, measured.

| | requests | time |
|---|---|---|
| main | 29,759 range GETs per kind | blocks 3,230s; did not complete |
| this PR | 670 ranges POSTs per kind | **188s**, complete |

17x on main's blocks phase alone against this PR's whole job. Localhost, so the 22x request
reduction is not what dominates; on a real peer at ~1s per round trip it would be.

Serving a fragmented request by walking its ranges pays `chunk_fetch_delay` per range, ~45 sleeps
at that run length: 4.61s one after another, 1.27s a few at a time, 0.008s reading the whole set
at once (a contiguous 100-height GET is 0.42s). The handler therefore reads the whole set first
and only walks the ranges when a height is missing; that took the arm from 1,619s to 188s with the
same request count.

Known, not blocking: on the light-client path the blocks phase already stores the VID that rides
along with it, but the VID phase still works from the range set captured before it ran, so it
re-requests those heights.

The light-client numbers in the previous revision (438 requests / ~433s vs 12 / ~19s) stand as
request counts priced at measured mainnet latencies. They cannot yet be run end to end: no
deployed node serves these routes, so a light client pointed at mainnet gets a 405 and exercises
the fallback instead.

## Notes for reviewers

- Worth the most scrutiny: `fetch_leaves_for_ranges` and `verify_leaf_chain` in `light-client/src/state.rs`, which accept unproven leaves and verify them by hash chain, and the request caps in `validate_ranges`.
- The light client pairs leaves with payload proofs positionally, which is why `validate_ranges` rejects ranges that are not ascending and disjoint. The node's payload-proof handler pairs its two batches by height instead, so neither relies on an unchecked invariant.
- `test_query_service_catchup_from_peer` in `crates/espresso/node/src/api.rs` runs the production wiring end to end: a node leaves consensus, comes back with a fragmented gap, and fills it from a peer over the scanner, `LightClientProvider` and the peer's endpoints. It pins convergence and content, not which route carried it; `test_scanner_backfills_over_ranges_endpoints_only` covers the route.


crates/espresso/api/src/v1/availability.rs
```diff
@@ -105,6 +105,19 @@ pub trait HotShotAvailabilityApi {
...
     ) -> anyhow::Result<Vec<Self::VidCommon>>;
 
+    /// Read the objects stored locally for a set of `(from, until)` half-open height ranges.
+    ///
+    /// These carry a fragmented set of heights in one request, for peers catching up. Like the
+    /// range endpoints they answer in full or not at all, so a height the node does not have is a
+    /// 404 rather than a short response, and no height is ever fetched from another peer to serve
+    /// one.
+    async fn get_leaf_batch(&self, ranges: Vec<(u64, u64)>) -> anyhow::Result<Vec<Self::Leaf>>;
+    async fn get_block_batch(&self, ranges: Vec<(u64, u64)>) -> anyhow::Result<Vec<Self::Block>>;
```
sveitser: Why don't we use `Vec<Range<u64>>` like elsewhere?

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4879/commits/8aec2d7b750abf58c738ff3c2000718009223352

crates/hotshot-query-service/src/data_source/fetching.rs
```diff
@@ -1532,14 +1609,137 @@ where
...
+        if req.clone().might_exist(heights) {
             T::active_fetch(tx, self.clone(), req).await?;
         } else {
             tracing::debug!("not fetching object {req:?} that cannot exist at {heights:?}");
         }
         Ok(())
     }
 
+    /// Store objects fetched at scattered heights, one write per contiguous run.
+    async fn store_runs<T: HeightIndexed>(&self, objs: Vec<T>) -> Vec<NonEmptyRange<T>>
```
sveitser: I think this would silently degrade in performance if obj are not ascending but there's isn't anything documenting this.

crates/light-client/src/state.rs
```diff
@@ -376,6 +390,119 @@ where
...
+            ranges.iter().map(|range| range.start).min(),
+            ranges.iter().map(|range| range.end).max(),
+        ) {
+            let wanted: u64 = ranges.iter().map(|range| range.end - range.start).sum();
+            let span = hi - lo;
+            // hotshot-query-service's default small-object range limit, which the zkVM build
+            // cannot import; test_fetch_leaves_for_ranges pins the two together. Should they
+            // drift anyway, the server rejects the span and the fetch falls back to per-run.
+            let limit = 500;
+            if span <= limit && span <= wanted * 4 {
```
sveitser, line 426: @imabdulbasit why is there a wanted * 4 here?

sveitser (approved):

crates/light-client/src/state.rs
imabdulbasit: We can fetch the whole range if the gaps are small, so that one proof covers everything, or we can fetch the separate ranges, where each range needs its own proof. This check just decides which of the two is better. So let's say you want 0..2 and 4..6. That is 4 heights. Fetching the whole range is 6 leaves and one proof, where fetching the two ranges separately is 4 leaves and two proofs. Now say you want 0..2 and 200..202. Still 4 heights, but the range is 202, so we'd pull 202 leaves and we just fallback to per range. the *4 is basically how many leaves we can waste for each leaf we need.

PR #4908, opened 2026-09-02
Add V2 node module APIs

imabdulbasit:
Completes `NodeService` on v2: every v1 `node` endpoint except `oldest-block` and `oldest-leaf` now has a counterpart, fourteen rpcs covering 34 of the 36 v1 routes. Handlers delegate to the v1 `NodeApi` methods, so both versions share their error classification.

- Scalars first: transaction count, payload size, sync status and block reward. The first v2 endpoints with query parameters and the first enum, so `build/openapi.rs` learns enum schemas.
- Then the rest: block height, limits, the consensus and DA stake tables, validators, paged registered validators, proposal and vote participation, the header window and the VID share.
- Where v1 has a route per epoch plus a `current` route, or a route per way of naming a block, v2 has one route with optional parameters, of which exactly one must be given.
- Two build-time guards, both refusing before any code is generated: `check_bindings` rejects a non-GET binding, since request messages become query parameters and the body mapping is deliberately undecided, and `request_parameters` rejects a repeated, message-typed or enum request field.
- Shapes were taken from v1's own JSON rather than from the types, which caught four encodings a `Display` impl gets wrong: a `U256` stake is a hex quantity, an x25519 key is base58 wherever no serde adapter tags it, a `NetAddr` keeps its unbracketed pre-IPv6 form, and a jellyfish Merkle node has an `Empty` arm written as a bare string.
- `VidShare`'s AvidM arms are the namespaced wrappers, so a share carries one entry per namespace. Those types keep their fields private, so the mapping reads v1's serialized form.
- `/v2/node/block-height` duplicates `/v2/status/block-height` because v1 has both. Its messages and rpc carry a `Node` prefix: the generator uses rpc names as operationIds and keys schemas by short name.
- `test_node_api_v2_agrees_with_v1` compares every endpoint against v1 field by field on a SQL test network, including the recursive ADVZ Merkle proof.
- What a live network cannot reach has unit tests instead: the sync-status range and enum mapping, all six header arms against the reference vectors, all three VID share arms, and the validator mapping, since a short network registers none.

Review focus: `proto/v2/node.proto`, which is frozen surface once released, and the two guards in `build/openapi.rs`. `src/generated/` is regenerated output.


crates/espresso/api/src/axum.rs
```diff
@@ -4854,6 +4854,20 @@ mod tests {
             "/v2/token/circulating-supply-ethereum",
             "/v2/token/total-issued-supply",
             "/v2/token/total-reward-distributed",
+            "/v2/node/transaction-count",
```
sveitser: sort alphabeticallay?

imabdulbasit: https://github.com/EspressoSystems/espresso-network/pull/4908/changes/c9cfc988733bd94b2a056bf14d90c6d6569c614c

API.md
```diff
@@ -26,10 +26,14 @@ descriptor set is exported as `espresso_api::FILE_DESCRIPTOR_SET`).
 
 ### What is served today
 
-`StatusService` and `TokenService`, nine endpoints under `/v2/status/...` and `/v2/token/...`. Everything else a client
-needs is still on v1. Every route in the OpenAPI document is a route `serve_axum` mounts: the tests in
-`crates/espresso/api/src/axum.rs` pin the documented set to a reviewed route list and probe each documented path against
-the mounted v2 router.
+`StatusService`, `TokenService` and `NodeService`: twenty-three endpoints under `/v2/status/...`, `/v2/token/...` and
```
sveitser: It's a small thing but kind of counterproductive to have such counts hardcoded yet claude keeps putting them in.

sveitser (approved):

PR #4909, opened 2026-09-02
feat(new-protocol): upgrade mechanism and V0_7 LargeBlock upgrade

bfish713:
## Summary

Adds a consensus upgrade mechanism to the new protocol (fast finality), and uses it for a first upgrade: **V0_6 → V0_7 (`LARGE_BLOCK_VERSION`)**, a chain-config-only upgrade raising `max_block_size`.

Once a network is on V0_6 the legacy stack (and with it the legacy upgrade task) is torn down, so until now there was no way to upgrade a running new-protocol network.

### Mechanism (`crates/hotshot/new-protocol/src/upgrade.rs`)

Reuses the legacy `UpgradeProposal` / `UpgradeVote` / `UpgradeCertificate` types and the genesis TOML `[[upgrade]]` window configuration, with a flow suited to single-view finality:

1. The leader of a view inside the proposing window broadcasts an `UpgradeProposal`.
2. Voters accept only the exact `expected_upgrade_data` for that view and broadcast `UpgradeVote`s all-to-all. The new protocol has its own offsets, independent of the legacy `UPGRADE_CONSTANTS` (105/130): `decide_by = view + DECIDE_BY_OFFSET` (10) and activation at `view + FINISH_OFFSET` (20), with no empty-block transition window. One-view finality needs no long runway, and mainnet's pending legacy upgrade keeps using the legacy constants.
3. Every node assembles the `UpgradeCertificate` locally at the upgrade threshold (~90%), via a 6th `VoteCollector` (`UpgradeTally` — upgrade data is epoch-less, so the voter's epoch rides on the vote message).
4. The next leader attaches it to the previously always-`None` `Proposal.upgrade_certificate` slot; `ProposalValidator` verifies attached certificates, and nodes that missed the votes adopt them from validated proposals and re-attach on a timeout, until `decide_by`.
5. When a leaf carrying the certificate finalizes, the upgrade is decided. Gap fills can decide carriers out of order, so the **earliest** decided carrier always wins, even after activation: an earlier carrier can only move activation earlier, which aligns a lagging node with the network. `UpgradeLock::set_decided_upgrade_cert` then flips wire encoding and header version at `new_version_first_view` through entirely existing plumbing, and the certificate is persisted (restored on restart, with an anchor-leaf fallback).

Wire compatibility: the two new `ConsensusMessage` variants are appended last; old V0_6 binaries fail to decode only the new messages. A stale-certificate filter in `SystemContext::init` drops a persisted certificate at or below the configured base so a completed upgrade cannot brick a redeployed node.

### V0_7 plumbing

`LARGE_BLOCK_VERSION = 0.7` in `crates/versions`; `Header::V7` reusing the V6 shape (new version tag; `data/v7/` reference vectors); `UpgradeType::LargeBlock { chain_config }` (`[upgrade.large_block]` genesis stanza); windows wired genesis → `HotShotConfig` → `Coordinator::maker` (the coordinator's `UpgradeConfig` is non-optional); light-client verify dispatch extended to 0.7; `data/genesis/demo-large-block-upgrade.toml` (1mb → 10mb, 30-block epochs) with justfile recipes (`demo-native-large-block-upgrade`, `test-demo large-block-upgrade`) and `tests/upgrades.rs::test_native_demo_large_block_upgrade`.

Merged with `main` after #4937 dropped the pre-V6 demo genesis files and their integration tests, so `large-block-upgrade` is the only demo entry this PR adds.

## Where to focus review

- `crates/hotshot/new-protocol/src/upgrade.rs` — `DECIDE_BY_OFFSET` / `FINISH_OFFSET` windows and the exact-expected-data rule
- `Consensus::maybe_decide_upgrade` (`consensus.rs`) — the earliest-carrier decide rule
- `Validator::upgrade_certificate` (`proposal.rs`) — attached-certificate validation
- `crates/hotshot/new-protocol/src/tests/upgrade.rs` — the four multi-node scenarios

## Testing

- Unit: `UpgradeProtocol` windows/expected-data/signing; `UpgradeTally` (certificate at 9/10 votes not 8, duplicates, invalid-signature recovery); `apply_upgrade`; v7 header serialization; `cargo test -p espresso-types reference` (new `test_reference_header_v7`).
- Multi-node (`tests/upgrade.rs`, 10-block epochs so every scenario proposes in one epoch and activates in a later one): happy path through the wire flip; certificate re-attached after dropped carrying proposals; clean expiry without activation when every carrier is dropped past `decide_by`; node restart with persistent storage across activation.
- In-process 5-node `api::test_new_protocol_large_block_upgrade`: base 0.6 from genesis, headers flip to 0.7 carrying the upgraded chain config, every node's decided state resolves to it.
- Full regression: all 248 `hotshot-new-protocol` tests (cutover/safety/liveness/restarts), espresso-types suite (one pre-existing environment flake: `l1::test_failover_update_task_ws`, fails identically on main), light-client quorum tests, clippy clean in both feature variants. `data/v6/new_protocol_messages` vectors regenerated for the appended variants.
- After the merge with `main`: `cargo check -p tests --no-default-features`, clippy on `hotshot-new-protocol` and `espresso-node`, the `reference` tests, and the new-protocol `upgrade` and `legacy_cutover` suites.
- Not run locally: the native-demo e2e (`just test-demo large-block-upgrade`).

## Follow-ups (out of scope here)

- The block builder's hardcoded `BlockBuilderConfig::default()` cap (`max_leader_bytes` = 2 MiB) shadows `ChainConfig::max_block_size`, so bigger blocks are not actually *built* until that is reconciled in a separate PR (cliquenet's 10 MiB `max_message_size` needs checking then too). This PR delivers the mechanism and the validation-side limit.
- All validators must run a 0.7-capable binary before `start_proposing_view`; the 90% upgrade threshold is the on-chain rollout guard.
- dev-node's `DevNodeVersion` still tops out at V0_6.
- `test_native_demo_large_block_upgrade` is not yet in the `test-integration` CI matrix (`.github/workflows/test.yml`).


crates/hotshot/new-protocol/src/upgrade.rs
```diff
@@ -0,0 +1,317 @@
...
+                return None;
+            },
+        };
+        info!(
+            %view,
+            new_version = %data.new_version,
+            decide_by = %data.decide_by,
+            first_view = %data.new_version_first_view,
+            "proposing upgrade"
+        );
```
lukaszrzasik, line 111: Can we make it more visible? error should be fine as upgrade is one off event. alternatively I think there is this new `announce` target which we can filter on to display logs below the generic log level.

crates/hotshot/new-protocol/src/upgrade.rs
```diff
@@ -0,0 +1,317 @@
...
+            &self.private_key,
+            &self.upgrade_lock,
+        ) {
+            Ok(vote) => vote,
+            Err(err) => {
+                warn!(%view, %err, "failed to sign upgrade vote");
+                return None;
+            },
+        };
+        info!(%view, new_version = %data.new_version, "voting for upgrade");
```
lukaszrzasik: Similarly, can we make it target `announce` or even level `error`?

crates/hotshot/new-protocol/src/upgrade.rs
bfish713: Moved to the `announce` target in 963055a2ebfc082a017f44b5c0fca94ae10d9614, which the default OTel filter admits at INFO. Did the same for the "upgrade decided" line in consensus.rs.

crates/hotshot/new-protocol/src/upgrade.rs
bfish713: Moved to the `announce` target in 963055a2ebfc082a017f44b5c0fca94ae10d9614, which the default OTel filter admits at INFO. Did the same for the "upgrade decided" line in consensus.rs.

lukaszrzasik (approved):

PR #4938, opened 2026-09-11
feat(new-protocol): cliquenet observer peers for non-staked query nodes (release)

bfish713:
Release-branch (`release-20260819`, which the `20260910` release is tagged on) version of the cliquenet observer change. The same change rebased onto `main` is #4935; review there first, this one is the backport-shaped counterpart.

Supersedes #4934, which GitHub closed when its head branch was renamed to `release-cliquenet-follower`.

## Why

After 0.6 all consensus traffic runs over cliquenet, whose peer set comes only from the stake table. A non-staked node receives nothing, so its query service is fed purely by HTTP backfill from `--peers`, and the active fetcher can only fill heights below the highest header it already has: without decides the watermark never rises. Our mainnet query nodes are not staked.

## What

Validators can pin "observer" nodes as `Role::Passive` cliquenet peers and forward them validated proposals, cert1 (as `HighQc`), cert2 and epoch changes. An observer decides every block live, without stake, voting, or any wire or protocol change. The consensus state machine already follows without stake (`staked_in_epoch` gates only vote1/vote2, and a parked proposal is adopted when its cert2 arrives); this PR adds the network plumbing and the explicit relay.

Only nodes that set the new options change behaviour. Other operators need no code, config or coordination.

Commits, in review order:

1. `network.rs`: `PeerPolicy` (`StakeTable { observers }` / `StaticOnly { upstreams }`), static peers exempt from `apply_epoch` eviction and `remove_peers`, `send_to_observers`, own-key and duplicate-key validation at startup.
2. `coordinator.rs`: `forward_to_observers` on `SendProposal`, `ProposalValidated` and the `broadcast` helper; `TestRunner` observer support and three runner tests.
3. node: `--cliquenet-observer-peers` / `--cliquenet-upstream-peers` (`BLS_VER_KEY~…@X25519_PK~…@host:port`), `NetworkParams`, observer catchup responders, refuse the `submit` module on observers.
4. Sequencer end-to-end test with a served observer query node.

## Where to focus

- `validate_static_peers` and the two skip points in `apply_epoch` (`network.rs`). Config parties always start `Active`, so `create_with_config` assigns the role right after creation.
- `forward_to_observers`: cert1 goes out as `HighQc` because that is the only certificate intake that advances an observer's view, and the only one not dropped when more than `MAX_VIEWS_AHEAD` ahead. Cert1/Cert2 handlers do not move the view. Like the proposal broadcast, it warns on regular network errors and returns only critical ones.
- `CliquenetPeer::from_str` and the submit-module refusal (`options.rs`).

## Not in this PR

Payloads and VID common still arrive over HTTP from `--peers`. Forwarding them (as `MessageType::External` bytes) is a follow-up, as are deduping duplicate proposal validations on observers with several upstreams and role promotion if a static peer later joins the stake table.

## Testing

- new-protocol: 4 network-layer tests and 3 `TestRunner` observer tests (across epochs, single upstream, restart from persisted anchor); full suite passes.
- node: 4 option-parsing tests; `test_new_protocol_observer_query_node` (5 validators + observer, 3 epochs, payload via HTTP fallback) passes 3/3 with retries disabled.
- Negative control: with `forward_to_observers` stubbed out, the single-upstream runner test does not complete.

### DIFF
--- a/crates/hotshot/new-protocol/src/network.rs
+++ b/crates/hotshot/new-protocol/src/network.rs
@@ -18,7 +18,7 @@ use hotshot_types::{
 };
 use hotshot_utils::anytrace;
 use parking_lot::RwLock;
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 use crate::message::{Message, MessageType, Unchecked, Validated};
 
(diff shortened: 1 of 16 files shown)


crates/hotshot/new-protocol/src/network.rs
```diff
@@ -422,6 +548,45 @@ pub enum NetworkError {
...
         src: x25519::PublicKey,
     },
+
+    #[error("invalid static peer configuration: {0}")]
+    StaticPeers(String),
+}
+
+/// Reject static peers that are this node (that would change our own role) or
+/// share an x25519 key with another peer (indistinguishable on receive).
+fn validate_static_peers<K: Clone + Eq + std::hash::Hash + std::fmt::Display>(
```
twittner, line 558: Maybe this can be placed after the NetworkError's impl block?

twittner (commented):
LGTM. One question: Static peers can also send data to nodes they are connected to. Should `Network::receive` drop messages from static peers or should they be processed like any other?

twittner (approved):

PR #4945, opened 2026-09-14
Anchor catchup at the epochs we hold

twittner:
The broadcast claim map becomes a single-flight registry of shared attempts, each removed from the map when it finishes, fails, or is cancelled. That deletes `catchup_cleanup` and the whole orphaned-entry class of bug: the mainnet wedge was a parked task never releasing its claim. Every external step is bounded by a timeout, and the local DRB computation by progress instead, since it is minutes of hashing by design.

~~**Catchup walks epochs ascending.** It used to walk back from the requested epoch until it found a stake table it held, so an epoch far beyond the tip probed every earlier epoch, each probe a storage load. It now resumes from the latest consecutive pair of stake tables the node holds and walks toward the target, which is where it stops rather than where it starts. An unreachable epoch costs one fetch attempt at the first gap; a gap in what the node holds is bridged rather than fatal. Because the walk supplies the ordering that backward discovery had to recover by recursing, attempts are no longer recursive.~~

**The search for where to start is anchored at what we hold.** Epochs are fetched oldest first. What changed is how catchup decided where "oldest" was. It used to search backward from the *requested* epoch until it found a stake table it held, so the search was as long as the distance from the request to the node's own state. The search now starts from the latest epoch the node holds, so its cost depends on the node's own gaps rather than on what it was asked for, and an unreachable epoch costs one fetch attempt at the first gap.

Also: the root epoch is resolved from memory or storage instead of through `stake_table_for_epoch`, which spawned a nested catchup and could cascade; the
two DRB maps become one; `bootstrap_epoch_window` drops its hand-rolled pair
scan, which catchup now does itself.

**Three tests deleted, because they could not fail**, along with the comments across both suites that still described a watchdog, `catchup_map` and `catchup_cleanup`. Each test waited for an abandoned attempt to resume; a step is abandoned by its `timeout`, which drops the future, so there is nothing to resume.

### Where to focus

- `Attempts` — when a claim is taken and released, and that every await inside an attempt is bounded, since that is what guarantees release.
- `catchup` — a step derives its stake table from the epoch root two epochs earlier and leaves the one in between for the next step, so the walk is seeded from a pair. `set_first_epoch` always seeds one.


crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -824,31 +643,29 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
...
+        let drb = match computed {
             Some(drb) => drb,
             None => {
-                if let Some(progress) = progress {
-                    progress.checkpoint(format_args!(
-                        "fetching externally supplied drb result for epoch {epoch} after \
-                         cancellation"
-                    ));
-                }
                 return self.get_epoch_drb(epoch).await.map_err(|e| {
```
lukaszrzasik: This is now unbounded. Previously it was bounded by the catchup watchdog. We bound the same method call in line `448`. Additionally we've lost the checkpoint functionality here. If we get stuck here, we will have no logs indicating it.

crates/espresso/types/src/v0/impls/committee.rs
lukaszrzasik, line 804: This is not really introduced by this PR but it's worth addressing I think: we should recheck here, in the timeout case,  whether the snapshot already contains the epoch. Similarly to what we do in lines 743-757.

crates/hotshot/types/src/epoch_membership.rs
twittner: Thanks, changed in https://github.com/EspressoSystems/espresso-network/pull/4945/commits/0d23713912b3b94e0cd8b7f007261bf01f6a90b3.

crates/espresso/types/src/v0/impls/committee.rs
twittner: Thanks, changed in https://github.com/EspressoSystems/espresso-network/pull/4945/commits/0d23713912b3b94e0cd8b7f007261bf01f6a90b3.

crates/hotshot/types/src/epoch_membership.rs
```diff
@@ -884,245 +720,129 @@ impl<TYPES: NodeType> EpochMembershipCoordinator<TYPES> {
...
-                progress.last()
-            ),
-        }
-    });
+}
+
+impl Drop for AttemptEntry {
+    fn drop(&mut self) {
+        let removed = self.attempts.map.lock().remove(&self.epoch);
+        drop(removed);
```
lukaszrzasik, line 844: Nit: `drop` is redundant here.

lukaszrzasik (approved):

PR #4947, opened 2026-09-14
fix(node): always record the x25519 key in the validator config

bfish713:
## Problem

`GET /v1/status/keys` (and the `x25519_key` field of `config/hotshot`) reads the node's x25519 public key from `validator_config.x25519_keypair`. The production `init_node` only filled that field on the orchestrator bootstrap path, so a node that loaded its network config from persistence or from `--config-peers` (every mainnet or decaf node after first boot) reported `"x25519_key": null`, even though Cliquenet was running with the configured key.

This is the key an operator needs to paste into `--cliquenet-observer-peers` / `--cliquenet-upstream-peers`.

## Fix

Build the node's own `ValidatorConfig` in a small helper, `local_validator_config`, that always records the x25519 keypair from `NetworkParams::x25519_secret_key`. That is the same key Cliquenet is created with, and `KeySet` has already resolved it from whichever source was configured: `--private-x25519-key`, the key file, or the mnemonic (index-aware). The orchestrator path now only adds `p2p_addr`.

Behaviour elsewhere is unchanged: `ValidatorConfig::public_config()` only produces `connect_info` when both the keypair and `p2p_addr` are set, so the `/ready` payload posted to the orchestrator and the stake-table comparison are the same as before on every path.

## Where to look

- `crates/espresso/node/src/lib.rs`: `local_validator_config` and the two call sites in `init_node`.
- The unit test `local_validator_config_always_records_x25519_key` covers a mnemonic-derived key set and an explicit key, and asserts `connect_info` stays empty without an address.

Applies cleanly to `release-cliquenet-follower` as well.


crates/espresso/node/src/lib.rs
```diff
@@ -442,20 +442,20 @@ where
...
-        private_key: network_params.private_staking_key,
-        stake_value: U256::ONE,
-        state_public_key: state_key_pair.ver_key(),
-        state_private_key: state_key_pair.sign_key(),
+    // Peers reach cliquenet at the advertise address when one is configured. Fall back to the
+    // bind address so `status/keys` always reports where the node listens.
+    let cliquenet_addr = network_params
+        .cliquenet_advertise_addr
+        .clone()
+        .unwrap_or_else(|| network_params.cliquenet_bind_addr.clone());
```
imabdulbasit: why does it default to bind address?

imabdulbasit: isn't this an external address?

bfish713: Good catch, it shouldn't, there is no fallback we should just set it to None if the advertise address is None

imabdulbasit (approved):

PR #4949, opened 2026-09-15
bench: parallelize the transaction-commitment tail of recovery

mrain:
Stacks on `release-mainnet-bench`, which already carries the instrumentation port and the payload shaping.

### This PR:

Traces from `sep-14-8xlarge` showed recovery dominating every view on every node, and the split landed almost entirely on the serial tail:

| block | recover total | parallel decode | **serial tail** |
|---|---|---|---|
| 80 MB | 648 ms | 393 ms | **253 ms** |
| 50 MB | 335 ms | 174 ms | **159 ms** |

That tail is `recover_v_minus_1_decode_end` → `recover_v_minus_1_end`: `from_bytes` plus `transaction_commitments`. Reproduced locally on an 80 MB payload of 81,920 transactions — `from_bytes` 15 ms, the trait default's serial per-transaction Keccak 217 ms, the same work on rayon 27 ms. 232 ms local against the 253 ms measured in the fleet.

Two commits:

- **`TestBlockPayload`** (harness) — fallout from shaping the payload into 1 KiB transactions without taking the other half of that change. One giant transaction hashed once was cheap serially; tens of thousands of small ones are not, and `release-new-protocol-bench` has carried a parallel override for exactly this reason since it made the same payload change.
- **`espresso_types::Payload`** (production) — the same serial default, on the same critical path, with a real consumer: `on_block_reconstructed` evicts reconstructed transactions from the builder's retry buffer, and `retry_pending` is genuinely populated in production via `ClientRequest::SubmitTransaction`. Cost grows linearly with transactions per block, so this is a ceiling on block transaction count rather than a problem at today's volumes.

### Key places to review:

`p3-maybe-rayon` rather than rayon directly, with **no** feature forced: `espresso-types` is built for the SP1 zkVM target and the shim degrades to serial iterators there. It needs no flag because the crate's own non-optional `vid` dependency already enables `p3-maybe-rayon/parallel` in every graph that builds it (verified with `cargo tree`), so this inherits parallelism where it exists and cannot be what breaks a constrained target.

Ordering is the real risk. Callers pair a commitment index with a transaction index (`task-impls/src/helpers.rs:456`, `:652`), so a reorder would misattribute transactions rather than fail loudly. `par_iter` is an indexed parallel iterator and the index sequence is exactly the one `enumerate` walks; both payload types get a test pinning the parallel result against the serial default, the production one across several uneven namespaces.

### Things tested

* `cargo test -p espresso-types` — 257 passed, plus the 32 `reference` vectors per AGENTS.md.
* `cargo test -p hotshot-example-types --lib` — 13 passed.
* `cargo test -p hotshot-new-protocol --lib` — 213 passed.
* `cargo clippy -p espresso-types -p hotshot-example-types --all-targets -- -D warnings` — clean.

The `cargo sort --check` failure on `node-metrics` is pre-existing and identical with and without this change.


crates/hotshot/example-types/src/block_types.rs
```diff
@@ -292,6 +295,23 @@ impl<TYPES: NodeType> BlockPayload<TYPES> for TestBlockPayload {
...
+    ///
+    /// Callers pair a commitment index with a transaction index, so the result
+    /// must stay in [`Self::transactions`] order; `par_iter` is an indexed
+    /// parallel iterator, so `collect` preserves it.
+    fn transaction_commitments(
+        &self,
+        _metadata: &Self::Metadata,
+    ) -> Vec<Commitment<Self::Transaction>> {
+        use p3_maybe_rayon::prelude::*;
+        self.transactions.par_iter().map(|tx| tx.commit()).collect()
```
twittner: Rayon should probably not split the work if there are very few transactions:

```suggestion
        self.transactions.par_iter().with_min_len(8).map(|tx| tx.commit()).collect()
```

crates/espresso/types/src/v0/impls/block/full_payload/payload.rs
```diff
@@ -206,6 +206,36 @@ impl BlockPayload<SeqTypes> for Payload {
...
+    /// serial default gets it through `enumerate`.
+    fn transaction_commitments(
+        &self,
+        metadata: &Self::Metadata,
+    ) -> Vec<Commitment<Self::Transaction>> {
+        use p3_maybe_rayon::prelude::*;
+
+        let indices: Vec<Index> = QueryablePayload::iter(self, metadata).collect();
+        indices
+            .par_iter()
```
twittner, line 228: Perhaps it would make sense to only use rayon if the number of transactions exceeds a certain minimum, e.g. 32. For very small counts the thread management in rayon would outweigh the benefits of running in parallel.

mrain: Done in ff2ac75 and cc1a0e9 — `MIN_PARALLEL_TRANSACTIONS = 32`

bfish713 (approved):

PR #4952, opened 2026-09-15
refactor(node): share node bootstrap between init paths

bfish713:
Pull the consensus-independent parts of `init_node` into helpers that a node not running HotShot can reuse:

- `load_or_fetch_network_config`: the persisted-config and `--config-peers` arms of config resolution; `init_node` still falls through to the orchestrator.
- `apply_genesis_overrides`: the epoch, DRB, stake-table-capacity and DA-committee parameters only genesis knows.
- `init_node_state`: L1 client, state catchup providers, stake table fetcher, membership and coordinator, `NodeState`.

No behaviour change. `init_node` calls the helpers in the same order as before; the cliquenet registration check now reads the membership through the coordinator, which is equivalent.

Part 1 of 5 of the light-client follower stack: a query node without stake that follows the chain through the light client and serves every endpoint a staked query node serves (`doc/light-client-follower.md` lands in the last PR).

**Where to focus:** the moved block is verbatim apart from borrowing `genesis` instead of moving its fields; check that nothing between the removed lines and the new helper call changed order.

## Stack

1. #4952 refactor(node): share node bootstrap between init paths


crates/espresso/node/src/lib.rs
```diff
@@ -895,6 +774,226 @@ where
...
+
+    info!("Spawning update loop");
+
+    fetcher.spawn_update_loop().await;
+    info!("Update loop spawned. Fetching block reward");
+
+    let block_reward = fetcher.fetch_fixed_block_reward().await.ok();
+    info!("Block reward fetched: {:?}", block_reward);
+    // Create the HotShot membership
+    let epoch_height = network_config.config.epoch_height;
```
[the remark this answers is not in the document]
imabdulbasit: @bfish713 I think this should be addressed. We had issues in the past when we used the epoch_height from the config

crates/espresso/node/src/lib.rs
```diff
@@ -235,12 +236,33 @@ pub struct L1Params {
...
+
+pub(crate) enum LoadedNetworkConfig {
+    Persisted(NetworkConfig<SeqTypes>),
+    /// The caller persists it once genesis overrides are applied.
+    Fetched(NetworkConfig<SeqTypes>),
+}
+
+pub(crate) struct NodeStateParts<P> {
+    pub node_state: NodeState,
+    pub coordinator: EpochMembershipCoordinator<SeqTypes>,
```
imabdulbasit: NodeState already has coordinator so we can drop this field I think

crates/espresso/node/src/lib.rs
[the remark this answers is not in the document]
bfish713: Fixed in 6271b0a: `init_node_state` now reads `epoch_height` and `epoch_start_block` from `genesis`, as `init_node` did before the refactor, so the helper has no dependency on `apply_genesis_overrides` having run first.

crates/espresso/node/src/lib.rs
```diff
@@ -235,12 +236,32 @@ pub struct L1Params {
     pub options: L1ClientOptions,
 }
 
+#[derive(Clone, Debug)]
+pub struct CatchupParams {
```
imabdulbasit, line 235: nit: `NetworkParams` now also has all these fields ([here](https://github.com/EspressoSystems/espresso-network/blob/aec9a3f571434ff8d5dd3b3d0c0678c36d135bec/crates/espresso/node/src/lib.rs#L144)) so I think those four fields in `NetworkParams` can be replaced with  `pub catchup: CatchupParams`

imabdulbasit (approved):

crates/espresso/node/src/lib.rs
bfish713: Done in b3087cc: `NetworkParams` now holds `catchup: CatchupParams`, `run.rs` builds it once and `init_node` moves it.

PR #4953, opened 2026-09-15
refactor(api): abstract the API's consensus dependency

bfish713:
`ApiState` reached consensus through `SequencerContext` and `ConsensusHandle` directly, tying every API module to a node that runs HotShot. This puts that dependency behind two traits in `api/context.rs`:

- `ConsensusSource`: the handle methods the API needs (decided leaf and state, per-view state, membership coordinator, upgrade lock, submit, participation, remembering recovered accounts).
- `ApiContext`: the rest of what the API needs from the node (persistence, node state, network config, optional validator keys, state signer and event streamer, a shared light client, VID share requests, task list).

`SequencerContext` and `ConsensusHandle` implement them; `ApiState<C: ApiContext>` and `Options::serve` are generic over the context. Storage-only reads (state certificates, validators, stake table events) go through `ApiContext::persistence` instead of the handle's storage. The `serve` closure receives an `ApiSink`, an `EventConsumer` that can also append decided blocks. The catchup helpers that push recovered accounts into consensus memory take the trait object.

A context without keys makes `status/keys` and `state-signature` answer 404 and `hotshot-events` stream nothing.

## Follow-up commits from review

- `NullEventConsumer`'s `DecideSink::append` fails instead of silently accepting blocks, so a node that hands decided blocks to an API without query storage fails on its first block rather than reporting progress it never persisted.
- `LightClientProvider` opens its SQLite database inside the lazy init, after checking whether the context already runs a light client, so a follower never opens a second pool on the same path.
- The provider holds its light client in a `OnceCell`: a failed open is logged and retried on the next fetch instead of being cached for the life of the node.

For validators the only visible change is that the light-client database is opened on the first peer fetch rather than at startup, and a failure to open it no longer stops the node.

**Where to focus:** `api/context.rs` (the two traits and their impls for the validator), then `ApiState` at the top of `api.rs`, then `LightClientProvider` at the bottom of `api.rs` for the lazy open and retry. The rest of `api.rs` is `<N, P>` to `<C>` substitution; `request_vid_shares` moved to a free function unchanged.

## Stack

1. #4952 refactor(node): share node bootstrap between init paths (merged)


crates/espresso/node/src/api/context.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// [`Options::serve`](super::Options::serve) hands a clone of the context to the API before
+    /// calling this, so the tasks attached here must be shared with that clone: [`TaskList`] is
+    /// that shared handle, and dropping any clone of it aborts every task.
+    fn with_task_list(self, tasks: TaskList) -> Self
+    where
+        Self: Sized;
+}
+
+#[async_trait]
+impl<I> ConsensusSource for ConsensusHandle<SeqTypes, I>
```
imabdulbasit, line 93: Is this trait even needed since all of its methods are called through the `ConsensusHandle`

crates/espresso/node/src/api/update.rs
```diff
@@ -6,28 +6,46 @@ use anyhow::bail;
...
+    async fn append(&self, info: BlockInfo<SeqTypes>) -> anyhow::Result<()>;
+}
+
+pub trait ApiSink: EventConsumer + DecideSink {}
+
+impl<T: EventConsumer + DecideSink> ApiSink for T {}
+
+#[async_trait]
+impl DecideSink for NullEventConsumer {
+    async fn append(&self, _info: BlockInfo<SeqTypes>) -> anyhow::Result<()> {
```
imabdulbasit: from claude

`Issue: NullEventConsumer gets a DecideSink::append that returns Ok(()) without storing anything. It exists only so the no-storage serve arms can produce a Box<dyn ApiSink>. For a validator that's harmless, since nothing calls append. For a follower, append is the only thing that persists the chain, and the loop trusts its return: next advances, decided_leaf updates, logs say "following". A follower wired to that sink discards every block while every health signal reports success.
`

crates/espresso/node/src/api.rs
```diff
@@ -1835,51 +1746,36 @@ where
...
-                    .known_nodes_with_stake
-                    .into_iter()
-                    .map(|peer| peer.stake_table_entry)
-                    .collect(),
-
-                chain_id,
-            };
-            LightClient::from_genesis_with_options(db, client, genesis, opt)
+            let ctx = state.context().await;
+            if let Some(light_client) = ctx.light_client() {
```
imabdulbasit: from claude
`LightClientProvider::new opens a SQLite pool and runs migrations before the lazy future that would use it, and this PR adds an early return that never uses it on a follower. Since the follower's own light client uses the same --light-client-db-path, that's two writer pools and a duplicate migration on one file. Fix: connect inside the future, after the ctx.light_client() check.`

crates/espresso/node/src/api/context.rs
bfish713: `FollowerConsensus` later in the stack implements this trait

imabdulbasit (approved):

PR #4954, opened 2026-09-15
feat(node): light-client follower context

bfish713:
`FollowerContext` is a node that follows the chain through the `light-client` crate instead of running HotShot.

`init_follower_node` resolves the network config from peers, builds the shared node state and stake table coordinator, seeds the first epoch and the stake-table window the way a validator does, and connects a `LightClient` to the upstream query nodes. The follow loop polls the verified chain height and, for each new block, fetches the leaf with its finality proof, the payload and VID common against the header, and the cert2, and appends them to the query database through the API's `DecideSink`. At epoch roots it derives the stake table two epochs out from L1; at transition blocks it supplies the DRB the leaf carries. The merklized state loop every SQL query node runs then rebuilds fee, block and reward state from the stored chain.

The context implements `ApiContext` and `ConsensusSource` with no keys, signer, event streamer or per-view in-memory state, so catchup serves from storage and `submit` forwards to an upstream. The `reward-merkle-tree-v2` catchup route gains a storage fallback.

Part 3 of 5 of the light-client follower stack.

**Where to focus:** `Follower::follow` and `track_epoch` in `follower.rs`, compared with `EpochManager::handle_leaf_decided` in the new protocol; the trust model is in the module doc.

## Stack

1. #4952 refactor(node): share node bootstrap between init paths


crates/espresso/node/src/api.rs
```diff
@@ -991,7 +991,15 @@ impl<C: ApiContext, D: CatchupStorage + Send + Sync> CatchupDataSource for Stora
...
         view: ViewNumber,
     ) -> anyhow::Result<Vec<u8>> {
-        self.as_ref().get_reward_merkle_tree_v2(height, view).await
+        match self.as_ref().get_reward_merkle_tree_v2(height, view).await {
+            Ok(tree) => return Ok(tree),
+            Err(err) => {
+                tracing::info!("reward merkle tree is not in memory, trying storage: {err:#}");
+            },
+        }
+        self.inner()
```
imabdulbasit, line 1000: This is going to load the latest tree in storage which might be an old tree and doesn't match with the height that we are querying

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+
+    async fn membership_coordinator(&self) -> EpochMembershipCoordinator<SeqTypes> {
+        self.coordinator.clone()
+    }
+
+    async fn upgrade_lock(&self) -> UpgradeLock<SeqTypes> {
+        self.upgrade_lock.clone()
+    }
+
+    async fn submit_transaction(&self, tx: Transaction) -> anyhow::Result<()> {
```
imabdulbasit, line 425: Why is the follower submitting the transaction? I think we should only submit to query node

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+        let mut last_err = anyhow!("no upstream query nodes to forward the transaction to");
+        for upstream in &self.upstreams {
+            let client = Client::<ClientErr, SequencerApiVersion>::new(upstream.clone());
+            let result = client
+                .post::<Commitment<Transaction>>("submit/submit")
+                .body_binary(&tx)?
+                .send()
+                .await;
+            match result {
+                Ok(_) => return Ok(()),
```
imabdulbasit, line 435: why are we ignoring the transaction commitment here?

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+    /// The upstreams when unset.
+    pub config_peers: Option<Vec<Url>>,
+    pub catchup: CatchupParams,
+    pub bootstrap_epoch_catchup_timeout: Duration,
+    pub options: FollowerOptions,
+}
+
+#[derive(Derivative, Clone)]
+#[derivative(Debug(bound = ""))]
+pub struct FollowerContext<P: SequencerPersistence> {
```
imabdulbasit, line 113: from Claude:

Every copy of `FollowerContext` can shut down the follower.

- `FollowerContext` is `Clone`, because `ApiContext` requires it.
- All clones share one `TaskList` (an `Arc`), which holds the follow loop.
- `TaskList`'s own `Drop` aborts every task in the shared list, and it runs when any clone is dropped, not only the last one. `detach()` doesn't prevent this, because only `FollowerContext`'s `Drop` checks `detached`.

So if any clone is dropped while the others are still in use, the follow loop dies for all of them. The API keeps serving, but heights stop advancing, with no error.

**Fix:** split the context into two types:
- an owner that isn't `Clone`, holds the `TaskList`, and shuts down on drop
- a cheap `Clone` handle for the API that owns no tasks and implements `ApiContext`

`FollowerConsensus` already takes this approach for the consensus side. With the split, the follower's lifetime is the owner's lifetime, and dropping API clones can't stop it.

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+            vid_common,
+        } = block;
+        let height = leaf.height();
+        let mut info = BlockInfo::new(leaf.clone(), Some(block), Some(vid_common), None);
+        // Upstream stores a cert2 only at the heights it finalized directly, so most answers
+        // are `None`; asking at every height is what keeps the follower serving the same cert2s.
+        if leaf.header().version() >= NEW_PROTOCOL_VERSION
+            && let Some(cert2) = self
+                .light_client
+                .fetch_certificate2(height)
```
imabdulbasit, line 587: fetch_certificate2() function fetches the header as well to verify the cert2 but the header is already available here so I think we can use that header and avoid the extra fetch request.

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+    {
+        Some(LoadedNetworkConfig::Persisted(config)) => (config, false),
+        Some(LoadedNetworkConfig::Fetched(config)) => (config, true),
+        None => bail!("a follower needs peers to fetch the network config from"),
+    };
+    apply_genesis_overrides(&mut network_config, &genesis);
+    if fetched {
+        persistence.save_config(&network_config).await?;
+    }
+    let epoch_height = network_config.config.epoch_height;
```
imabdulbasit, line 175: we should use the epoch height from the genesis file

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+        l1_params,
+        params.catchup.clone(),
+        persistence,
+        &*metrics,
+    )
+    .await?;
+    let coordinator = node_state.coordinator.clone();
+
+    seed_first_epoch(&coordinator, &network_config);
+    if epoch_height > 0 {
```
imabdulbasit, line 194: epoch_height can be never be 0 now so we can panic if it is instead of contining with the wrong value

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+            let client = Client::<ClientErr, SequencerApiVersion>::new(upstream.clone());
+            let result = client
+                .post::<Commitment<Transaction>>("submit/submit")
+                .body_binary(&tx)?
+                .send()
+                .await;
+            match result {
+                Ok(_) => return Ok(()),
+                Err(err) => {
+                    tracing::warn!(%upstream, "forwarding transaction failed: {err}");
```
imabdulbasit, line 437: I think the error here should be made a field to make it easier to view in datadog

crates/espresso/node/src/follower.rs
```diff
@@ -0,0 +1,779 @@
...
+    block: BlockQueryData<SeqTypes>,
+    vid_common: VidCommonQueryData<SeqTypes>,
+}
+
+impl<P: SequencerPersistence> Follower<P> {
+    async fn run(mut self) {
+        tracing::info!(from = self.next, "following the chain");
+        loop {
+            if let Err(err) = self.poll().await {
+                tracing::warn!(next = self.next, "following the chain failed: {err:#}");
```
imabdulbasit, line 499: same here

crates/espresso/node/src/follower.rs
bfish713: it's just forwarding it to the upstream (i.e. the staked query node).  My idea was that this follower should act exactly as a query node, which would include accepting txns.

crates/espresso/node/src/follower.rs
bfish713: I see yeah we should return it to keep the api consistent

crates/espresso/node/src/follower.rs
bfish713: So basically instead of calling fetch_certificate2, we create a new function to fetch the cert with header?  I think that makes sense

imabdulbasit (approved):

PR #4957, opened 2026-09-15
 fix(node): prefetch stake table events before the update loop and bootstrap walk

sveitser:
- Cold-start node scans the whole stake table contract history inside the first `bootstrap_epoch_window` step, blowing its 30s timeout on a slow L1 provider. The walk then treats that as the live tip and the node wedges with a stale stake table window.
- `spawn_update_loop` already ran that scan in a detached task, so a cold node did it twice, concurrently.
- Fix: run it once, synchronously, before `spawn_update_loop`. Every walk step is then one epoch's delta and the 30s default stands.
- The walk's timeout arm now logs at WARN instead of INFO; downgrading to a stale window silently is what made this hard to spot.
- Drive-by: default OTel log filter becomes `warn,announce=info`, and the scan reports progress on the `announce` target every 30s. It was invisible in Datadog before.


sveitser:
Failing tests seem to be flaky on main too.

crates/espresso/node/src/lib.rs
```diff
@@ -734,6 +734,15 @@ where
         genesis.chain_config,
     );
 
+    // Before `spawn_update_loop`, so the cold full-history scan runs once, not twice.
+    prefetch_stake_table_events(
```
imabdulbasit, line 738: This function can return an error if there is a database err so is it okay to crash the node, or should there be retries?

imabdulbasit: or maybe even log the error and continue

sveitser: https://github.com/EspressoSystems/espresso-network/pull/4957/commits/22c37a748a3b0919e67c044c33a50e5e3559604e

imabdulbasit (approved):

PR #4959, opened 2026-09-15
Round down starting point for light client genesis binary search

jbearer:
The light-client-query-service genesis script does a binary search to find the first block where POS was enabled.  The binary search started at `epoch_start_block`, which is an upper bound. This was causing issues for Espresso stack benchmarking, because it mean we couldn't start the stack chains until the `epoch_start_block` was produced. This change rounds `epoch_start_block` down to the first block in its epoch, which makes no semantic difference (epochs always start on an epoch boundary) but allows the binary search to proceed faster, without waiting for an arbitrary block in the middle of an epoch.


sveitser (approved):
I think this is fine but I agree with claude about the comment reading weird since the PoS upgrade happens during an epoch.

PR #4964, opened 2026-09-16
feat(l1): default the finalized safety margin to 100 blocks

sveitser:
The margin was only set in docker/espresso-node.Dockerfile, so operators running the binary directly got no margin and hash-chained every old finalized block back from the head. One operator saw a 2 hour startup on a fast local RPC.

Introduce L1SafetyMargin, whose FromStr takes a block count or `unlimited`, and default the clap arg to 100. `unlimited` restores the previous always-verify behavior. Drop the Dockerfile ENV so the binary default is the single source of truth.


sveitser:
This also changes the value for a node that *intentionally* doesn't set it now. However I don't believe such a case actually exist.

bfish713 (approved):

PR #4966, opened 2026-09-16
Use (view, epoch) as key for vote counting.

twittner:
## Make vote collection consistent with the epoch-bound timeout certificate

Once the epoch of timeouts is signed, votes from two committees are two different statements and cannot be added together. This PR keys the vote collectors by `(view, epoch)` so they are not, and so a node whose epoch has advanced may vote again under the new committee rather than being deduplicated against the vote it cast under the old one. Everything the tally uses — the stake table, the threshold, the membership check on the signer, and the epoch the resulting certificate names — now comes from the key, so a certificate that forms locally verifies for a peer that picks its committee from the field it carries.

The liveness price is a potential timeout round at an epoch boundary. Honest nodes disagree about their epoch until the boundary certificate has spread, so the votes for a view can arrive under two epochs and neither side may reach a threshold. In this case, the view times out again and the votes converge as the nodes do. That price is inherent to a signed epoch, not to this implementation: merging the two tallies would mean counting a signature for a committee its signer did not name.

Before 0.7 there is no such statement to protect, so we do not pay for it: `on_timeout_vote` sets an unbound vote's epoch to the node's own before tallying, and all such votes for a view pool into one tally as they do on `main`. The signature is unaffected, because the commitment never covered the field. What changes against `main` is only that the committee is the receiving node's choice rather than that of whoever sent the first vote.

Widening the collector key lets a peer open a tally per epoch it names, so inbound vote1, vote2 and timeout votes are bounded to one epoch behind and three ahead of the node's own. That bound is placed *after* the catchup evidence on a timeout vote is handled, not before: a node that is behind names an out-of-range epoch by definition, and checking first would deny the catchup handshake to exactly the nodes that need it. For the same reason nothing here bounds a timeout certificate by view. What bounds it instead is the sender. `CertBySenderVerifier` verifies one certificate per peer at a time and parks at most one more while that peer's epoch stake table is unavailable, so a peer cannot buy more than its own slot however many certificates it sends

A node that has fallen behind an epoch still learns it from a certificate: `handle_timeout_certificate` adopts the certificate's epoch when it is later than its own. What it no longer does is adopt one that is earlier. A certificate from the epoch a node has left would pull it back, and because only the first certificate for a view is kept, the certificate from the later epoch is then dropped as a duplicate and the node stays behind, signing its next timeout vote into the committee it had already left.

The same reasoning removes the epoch from `Timer`: `handle_timeout` now signs under `Consensus::current_epoch`, so a node cannot attest, as a member of a committee it has left, that it gave up a view — and its vote lands in the tally its peers are filling.

`CertBySenderVerifier` retires a completed view per `(view, epoch)` where the epoch is signed, since one committee's certificate says nothing about another's. Where it is not signed it retires the view alone: the label records only which node assembled the certificate, so a peer's copy under another epoch is the same aggregate signature relabelled. Without that, the leader of the next view checks one signature per sender for a view it has already certified from the votes.

Relatedly, `run_tally` no longer holds a thread from the blocking pool while a view waits: it receives asynchronously and takes a `spawn_blocking` hop only once enough votes are in hand for a certificate to be possible, sized by `min_votes` and clamped to the stake table.

### Where to focus

`vote.rs` for the key and for `run_tally`; `on_timeout_vote` in `coordinator.rs` for the ordering of the evidence handling against the two gates, and for the pre-0.7 epoch rewrite; `handle_timeout` and `handle_timeout_certificate` in `consensus.rs`.

Cf. https://github.com/EspressoSystems/espresso-network/pull/4962


twittner:
@claude review

bfish713:
Can we also add some multi node tests with some timeouts on 0.7 to test the new code paths?

crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -1752,13 +1754,17 @@ impl<T: NodeType> Consensus<T> {
...
+                debug!(%view, %epoch, "adopting the epoch of a later certificate");
+                self.timeout_certs.insert(view, certificate.into_cert());
+            }
             return Protocol::Continue;
         }
-        let epoch = certificate.epoch();
         self.timeout_certs.insert(view, certificate.cert().clone());
         self.current_view = self.current_view.max(view);
-        self.current_epoch = Some(epoch);
+        let epoch = self.raise_epoch(certificate.epoch());
```
lukaszrzasik, line 1767: Unless I'm missing something, this seems risky. What I have in mind: if we receive a fabricated `TimeoutCertificate2` with a future epoch, we will update `self.current_epoch` to its epoch.
I think we should check `certificate.binds_epoch()` here as well, similarly to line `1758`.

crates/hotshot/new-protocol/src/consensus.rs
```diff
@@ -1752,13 +1754,17 @@ impl<T: NodeType> Consensus<T> {
             );
             return Protocol::Abort;
         }
-        if self.timeout_certs.contains_key(&view) {
+        if let Some(stored) = self.timeout_certs.get(&view).map(HasEpoch::epoch) {
+            if certificate.binds_epoch() && stored < Some(certificate.epoch()) {
+                let epoch = self.raise_epoch(certificate.epoch());
```
lukaszrzasik, line 1759: This silently raises `Consensus`'s epoch. No `ViewChanged` is triggered and so all its handlers in the coordinator stay behind on the old epoch. These are I think: gc, participation and metrics.

lukaszrzasik: Also we don't send `SendTimeoutCertificate` so we won't inform the other nodes stuck in the old epoch.

crates/hotshot/new-protocol/src/consensus.rs
twittner: This is what 0.6 does, the only difference being that `current_epoch` only moves forward. It also requires that the certificate verifier has access to the epoch, so it can not move arbitrarily far ahead. Allowing V2 TCs here can help with catchup, not doing it here requires a lagging node to receive catchup evidence eventually.

crates/hotshot/new-protocol/src/consensus.rs
twittner: There is currently no output for an epoch only change. This code runs when we already emitted a view change for this view, so GC already ran. While something like participation and metrics can lag for another view change the case felt so rare (requiring a timeout vote split at the epoch boundary with two formed certificates), and should normally resolve quickly, that I thought a silent epoch upgrade is preferrable. Finally a TC is not broadcast, but unicast to a leader, which most likely has already received the first TC.

crates/hotshot/new-protocol/src/consensus.rs
lukaszrzasik: I don't think I fully understand the catchup argument. Maybe we can discuss offline.

I do understand that, in order for the TC to be verified, the verifier has to has the stake table for the certificate's epoch. It's true that it limits the fabricated epoch. I think it is entirely possible to fabricate a TC with an epoch 2 greater than the current real epoch though. The nodes which receive such TC and verify it correctly will increment their epoch. This in turn causes `Vote1` and `Vote2` to be rejected via `is_vote_epoch_admissible`. 

Also the `on_timeout_vote` method rewrites the vote's epoch to the internally stored one here:
```
        if let TimeoutVote::V2(vote) = &mut vote {
            vote.data.epoch = Some(self.epoch());
        }
```
so such node will produce TCs with the fabricated epoch.

There's also concern about the GC. The fabricated TC triggers `ViewChanged` with epoch `e+2` (epoch `e` being the correct current one). The GC wipes the state cert for epoch `e` then.

crates/hotshot/new-protocol/src/consensus.rs
lukaszrzasik: Yes, I agree. It's probably not worth addressing. I can't see any real harm.

crates/hotshot/new-protocol/src/consensus.rs
twittner: Thanks, I adopted the guard in https://github.com/EspressoSystems/espresso-network/pull/4966/commits/8c3603d922a624b305714399e5be5f58b25a83ce. The impact for catchup is probably negligible in practice (cf. the commit message for details) and the scenario you described is certainly worse.

lukaszrzasik (approved):

PR #4975, opened 2026-09-18
feat(node): probe storage filesystem and fsync latency at startup

sveitser:
- Resolve the filesystem backing the data directory from `/proc/self/mountinfo` (Linux only)
- Measure fsync latency against a scratch file, bounded to 64 samples or 1s
- Read back the SQLite pragmas actually in effect, including `auto_vacuum`
- Run for the sql and fs backends; the Postgres backend has no local storage to probe
- Log the facts at INFO on the `announce` target, so the healthy baseline ships under the default telemetry filter, not just the problems
- Problem messages name `ESPRESSO_NODE_STORAGE_PATH` and what to set it to
- Export `consensus_disk_info{backend,fs,class,journal_mode,synchronous,auto_vacuum}` and `consensus_disk_fsync_micros{stat}`
- Advisory: never refuses to start, and each part degrades independently
- Implement `fs::Persistence::enable_metrics`, until now a stub, so the fs backend stops discarding its persistence timings


sveitser:
@claude re-check

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -159,6 +161,59 @@ pub fn build_sqlite_path(path: &str) -> anyhow::Result<PathBuf> {
...
+        },
+    }
+}
+
+/// The first failure short-circuits the rest: a pragma query only fails if the pool is unusable.
+#[cfg(feature = "embedded-db")]
+async fn read_pragmas(pool: &sqlx::Pool<Db>) -> Option<SqlitePragmas> {
+    let journal_mode = read_pragma(pool, "PRAGMA journal_mode").await?;
+    // SQLite reports these two back as their numeric settings, not the names used to set them.
+    let synchronous = match read_pragma::<i64>(pool, "PRAGMA synchronous").await? {
```
imabdulbasit: can we include `busy_timeout` as well? Also what is the reason for not printing out all the SqliteOptions and just doing individual queries

crates/espresso/node/src/persistence/sql.rs
```diff
@@ -159,6 +161,59 @@ pub fn build_sqlite_path(path: &str) -> anyhow::Result<PathBuf> {
...
+        Err(err) => {
+            tracing::debug!(pragma, ?err, "storage probe: pragma query failed");
+            None
+        },
+    }
+}
+
+/// The first failure short-circuits the rest: a pragma query only fails if the pool is unusable.
+#[cfg(feature = "embedded-db")]
+async fn read_pragmas(pool: &sqlx::Pool<Db>) -> Option<SqlitePragmas> {
```
imabdulbasit, line 191: can we include `busy_timeout` as well?

imabdulbasit (approved):

PR #4978, opened 2026-09-18
Add tokio runtime metrics

twittner:
Add seven runtime metrics to the existing 5s sampler, so every binary with a metrics registry picks them up without new wiring:

  - tokio_alive_tasks,
  - tokio_blocking_queue_depth,
  - tokio_blocking_threads,
  - tokio_global_queue_depth,
  - tokio_idle_blocking_threads,
  - tokio_worker_busy_seconds_total,
  - tokio_workers

The three blocking-pool gauges are behind tokio's `--cfg tokio_unstable`, now set in `.cargo/config.toml` and declared in check-cfg so the gated code passes clippy. A `RUSTFLAGS` in the environment replaces `build.rustflags` rather than adding to it, so the flag is repeated in the coverage workflow and the coverage dev shell, which also had to start exporting `RUSTFLAGS`: it was assigning a shell variable cargo never saw. The SP1 target check keeps its own flags; it has no runtime to measure. The crate still builds without the flag, warning at init instead of dropping the metrics silently, and `built_with_tokio_unstable` fails if a build path loses it.


sveitser (approved):
LGTM

